PR #694, opened 2013-11-30
Support unicode characters for operators

ahmadsalim:
We already support Unicode for identifiers, this would allow defining operators as Unicode too. There should probably be a flag in the future to disable Unicode support entirely (but that will take some time to implement, where this was a one-liner).


ahmadsalim:
3.5 years later, is this something we should reconsider having again now?
It should be possible to readd support if we wanted to.

david-christiansen:
No, I still think @edwinb's reasons hold.

PR #1652, opened 2014-10-17
Fix incorrectly type-dependence in vectors of VectorSpace type-class

victoredwardocallaghan:
VectorSpace takes two types to construct, a Field and abelian Group
respecively. The second type field was incorrectly typed as the same
as the first.

Signed-off-by: Edward O'Callaghan <eocallaghan@alterapraxis.com>


david-christiansen:
It seems that this PR has been obsoleted, so I'm closing it.

PR #1764, opened 2014-12-05
Agda-style record syntax

ziman:
This patch adds a nice Agda-style way of defining records. A (contrived) example:
```Idris
||| A vector wrapper.
||| @ a An implicit param.
||| @ n An explicit param.
record C : {a : Type} -> (n : Nat) -> Type where
  constructor MkC

  ||| A rather unhelpful implicit because it can't be inferred.
  {unhelpfulField : a}

  ||| The wrapped vector.
  xs : Vect n a
```
This is just syntactic sugar. Behind the scenes, the above generates the following constructor:
```Idris
MkC : {a : Type} -> (n : Nat) -> {unhelpfulField : a} -> (xs : Vect n a) -> C {a = a} n
```
Shortcomings:
* no docstrings for fields and parameters
  * currently, the record pipeline seems to just ignore this info (old-style records suffer from this as well)
  * should be relatively straightforward to fix
* no tactic implicit fields, just plain implicits
  * will probably require decoupling the syntax of implicits from `PPi`
* the `constructor` clause is mandatory because it syntactically disambiguates old-style records from Agda-style records. Maybe we should find a better way to do this and allow automatic `MkT` constructor naming.
  * old-style records with the constructor named `constructor` will be misparsed as agda-style records
* free variables in fields won't get turned into unbound implicits

### DIFF
--- a/src/Idris/ParseData.hs
+++ b/src/Idris/ParseData.hs
@@ -83,9 +83,7 @@ record syn = do (doc, argDocs, acc, opts) <- try (do
         ist <- get
         fc  <- getFC
 
-        -- we discard the second component because
-        -- fields will be described in the following block
-        (ctorDoc, _) <- option noDocs docComment
+        let ctorDoc = parseDocstring . T.pack $ "Constructor of " ++ show recName
         ctorName <- reserved "constructor" *> fnName
 
         fields <- many $ do
@@ -103,7 +101,7 @@ record syn = do (doc, argDocs, acc, opts) <- try (do
                     symbol "}"
                     return (n, t, impl)
 
-            (n, t, plicity) <- impField <$> expField
+            (n, t, plicity) <- impField <|> expField
 
             return (n, t, plicity, doc, argDocs)
 
--- a/src/Idris/ParseData.hs
+++ b/src/Idris/ParseData.hs
@@ -84,9 +84,9 @@ record syn = do (doc, argDocs, acc, opts) <- try (do
         fc  <- getFC
 
         let ctorDoc = parseDocstring . T.pack $ "Constructor of " ++ show recName
-        ctorName <- reserved "constructor" *> fnName
+        ctorName <- indented (reserved "constructor" *> fnName)
 
-        fields <- many $ do
+        fields <- many . indented $ do
             (doc, argDocs) <- option noDocs docComment
 
             let expField = do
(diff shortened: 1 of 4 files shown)


tdidriksen:
@david-christiansen We (@sualitu and myself) like your suggestion as well. The corecord definition we have in mind can quite easily be adapted to this form.

sualitu:
I finally had time to sit down at look at this and have implemented a syntax akin to the one @david-christiansen suggested.

However, during this process I came across (what I think is) a problem. How to express dependencies in the constructor? Between projections is not a problem, but since the constructor type is completely hidden, we can't express all dependencies. Consider the following example in the current record syntax (stolen from the record tests):

```
record Foo : Nat -> Type where
  MkFoo : (param : Nat) -> (num : Int) -> Foo param
```

With a new syntax like:

```
record Foo Nat where
  constructor MkFoo
  param : Nat
  num : Int
```

the dependency on ```param``` is not expressed. Either I am missing something or we need to expand this syntax to have the option to express such dependencies.

We could make it an option to write the type of the constructor (or parts of it). Alternatively we could just keep the old record syntax alongside the new one.

ziman:
@sualitu, what you mean would be written as follows:
```Haskell
record Foo (param : Nat) where
  constructor MkFoo
  num : Int
```
It's basically the same thing as a typeclass declaration.

This is sufficient because records cannot be indexed.

sualitu:
Hmm, it seems we have had a different understanding of what the syntax David suggested meant. I was thinking of still letting Idris infer the parameters from the type constructor, but I understand now why that does not make perfect sense.

What I don't like about your way of doing it is that I cannot have parameters with different plicity in the type constructor and the data constructor. It's definitely a superior solution to my own, but I don't think it's perfect.

I might be biased by the fact that we are currently working with coinduction and a ```Stream``` defined as a record with this syntax could be either of two:

```
record Stream (a : Type) where
  constructor (::)
  head : a
  tail : Stream a
```

resulting in a data constructor with an explicit type parameter (```(::) : (a : Type) -> a -> Stream a```).

```
record Stream {a : Type} where
  constructor (::)
  head : a
  tail : Stream a
```

resulting type constructor with an implicit parameter (```Stream : {Type} -> Type```).

Neither of these I can honestly say I'm happy with.

sualitu:
I like that way of doing things. The only thing missing is a way to make parameters to the constructor extracted from the type declaration explicit.

PR #1803, opened 2014-12-22
Error reflection for Fin

david-christiansen:
Add reflection-based error message improvements for Fin literals. Now, instead of a collection of internal implementation details, a friendly and descriptive message is returned.

Demo:

``` idris
module error005

-- Test the Prelude's error rewrites for Fin

one : Fin 2
one = 1

two : Fin 2
two = 2

hahaha : (n : Nat) -> Fin n
hahaha n = 0

ok : (n : Nat) -> Fin (plus 2 n)
ok n = 1

notOk : (n : Nat) -> Fin (plus 2 n)
notOk n = 2
```

gives

```
error005.idr:11:1:When elaborating right hand side of two:
When elaborating argument prf to function Prelude.Fin.fromInteger:
        When using 2 as a literal for a Fin 2 
                Could not show that 2 is less than 2
error005.idr:14:1:When elaborating right hand side of hahaha:
When elaborating argument prf to function Prelude.Fin.fromInteger:
        When using 0 as a literal for a Fin n 
                Could not show that 0 is less than n
error005.idr:19:1:When elaborating right hand side of notOk:
When elaborating argument prf to function Prelude.Fin.fromInteger:
        When using 2 as a literal for a Fin (S (S n)) 
                Could not show that 2 is less than S (S n)
```

This did require moving the basic parts of `Language.Reflection` into `prelude`, but I think this is a good idea anyway, because it stops the quasiquote syntax from throwing bizzare unknown variable errors when the user hasn't imported them.

I don't want to merge it myself because I want @edwinb's OK to add the reflection datatype to the prelude.

Later I'll find more error messages to pretty up. This is pre-hospital work - don't worry, I'm relaxing and healing :)

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -91,6 +91,8 @@ Extra-source-files:
                        libs/prelude/prelude.ipkg
                        libs/prelude/Prelude/*.idr
                        libs/prelude/Decidable/*.idr
+                       libs/prelude/Language/*.idr
+                       libs/prelude/Language/Reflection/*.idr
                        libs/prelude/Makefile
                        libs/prelude/*.idr
 
@@ -103,7 +105,6 @@ Extra-source-files:
                        libs/base/Data/Vect/*.idr
                        libs/base/Debug/*.idr
                        libs/base/Decidable/*.idr
-                       libs/base/Language/*.idr
                        libs/base/Language/Reflection/*.idr
                        libs/base/Makefile
                        libs/base/Network/*.idr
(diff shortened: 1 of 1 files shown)


LeifW:
A downside of adding Language.Reflection to the Prelude is that it adds some single-char identifiers in scope, e.g. `V` and `P`. The latter gave me an ambiguous identifier error with my `P = NP` proof.

edwinb:
Is this really a problem if disambiguation works properly though? Should Idris not really be able to tell the difference between ProblemClass and TT? Or have you encountered a situation where it can’t? 

Maybe we should avoid single character identifiers in any case, especially for things which won’t be used very often. ’S’ and ‘Z’ might yet prove to be a pain although they’ve seemed okay so far.

Edwin.

Issue #1366, opened 2014-07-07
Failed module load changes the prompt


rizo:
```
Idris>:l xyz
Can't find import xyz
*xyz>
```

As you can see the prompt's namespace changed although there is no such module as `xyz`.

PR #1809, opened 2014-12-26
REPL prompt should only show successfully loaded modules

bmsherman:
Fixes https://github.com/idris-lang/Idris-dev/issues/1366. It was getting on my nerves :).


edwinb:
I've just noticed a minor annoyance with this, which is that if you start Idris with a non-existent file, then try to edit it with :e, it will fail because loading the file failed. It would be good if it still knew which file it was working with so that :e still works, even when the file doesn't exist yet.

I often do this - "idris foo.idr" with the intention of creating foo.idr with :e.

edwinb:
This annoyed me enough that I've just fixed it :).

PR #1820, opened 2015-01-02
add msse2 to config.mk fixes #1819

Heather:
related issue: https://github.com/idris-lang/Idris-dev/issues/1819

Issue #1819, opened 2015-01-02
rts build for windows is broken


Heather:
Here is build output:

related commit: https://github.com/idris-lang/Idris-dev/commit/6dafd1880aa6a6cf934bc8812dd336be26f8bb66
``` shell
"make"[1]: Entering directory `/c/H/P/Idris-dev/rts'
cc -fPIC -O2 -Wall -DHAS_PTHREAD -DIDRIS_ENABLE_STATS  -I/usr/local/include  -DIDRIS_TARGET_OS="\"windows\"" -DIDRIS_TARGET_TRIPLE="\"mingw32\""   -c
-o idris_rts.o idris_rts.c
idris_rts.c:1:0: warning: -fPIC ignored for target (all code is position independent) [enabled by default]
 #include <stdlib.h>
 ^
In file included from idris_rts.h:13:0,
                 from idris_rts.c:10:
c:\mingw\lib\gcc\mingw32\4.8.1\include\emmintrin.h:31:3: error: #error "SSE2 instruction set not enabled"
 # error "SSE2 instruction set not enabled"
   ^
In file included from idris_rts.c:10:0:
idris_rts.h:77:9: error: unknown type name '__m128i'
         __m128i* bits128p;
         ^
idris_rts.c: In function 'MKB8x16const':
idris_rts.c:326:53: error: '__m128i' undeclared (first use in this function)
     Closure* cl = allocate(sizeof(Closure) + sizeof(__m128i), 1);
                                                     ^
idris_rts.c:326:53: note: each undeclared identifier is reported only once for each function it appears in
idris_rts.c:329:34: error: expected expression before ')' token
     cl->info.bits128p = (__m128i*)ALIGN((uintptr_t)cl + sizeof(Closure), 16);
                                  ^
...
```

PR #1822, opened 2015-01-02
compiler (currently) should also use sse2 due rts changes

Heather:
compiler also should use sse2 with the same reason as here (addition rts types) :
 - https://github.com/idris-lang/Idris-dev/pull/1820
 - https://github.com/idris-lang/Idris-dev/issues/1819

I'm agreed that ideally all that should be optional but it's more complicated to implement, now I just want working compiler.


edwinb:
Thanks for this, sorry for the slow merging (busy time lately!)

Heather:
Yeah!

Issue #1126, opened 2014-04-29
Split the .ipkg commands and the HTML doc gen into separate executables


david-christiansen:
This would be a good way to clean up REPL.hs, and it's a good project for a new developer. The idea is to have three (quite thin) top-level executables in the .cabal file, and then have them use the compiler as a library. This would also greatly simplify command-line processing. Suggested names: `idrispkg` and `idrisdoc`.

NobbZ:
Just made my PR for #1071 and knowing cabal hell as well as ruby's bundler, I would like to make some proposals of how package-management should be handled, and hope it is the right place here:

* since @edwinb already said he is agains multiple executables I would propose to have idris with subcommands instead of the current plentora of --flags like the following examples:
 * `idris` or `idris repl` for firing up the REPL
 * `idris package` or `idris pkg` followed by further subcommands for package management
 * and so on
* Save installed packages by using `<pkgname>--<version>` instead of only `<pkgname>` in the libsdirectory.
 * when specifiing some package using `idris -p example` it will choose the latest version of the package example available. When specifing `idris -p example--1.0.0` it will choose exactly that version if available and fail if not.
 * If there is an ipkg-file available in the current directory or its parents, automaticaly choose the latest possible and available version constraints for every invocation of `idris`.
 * `idris package build` (or similar command) should automatically determine latest available versions of specified dependencies (in the ipkg) and install them. Also it creates a ipkg.lock-file in which it tells exactly which versions of dependencies and their dependencies are build against.
 * Whenever this ipkg.lock does exist, it will be used instead of the ipkg for determining the version of packages to load.
* A central catalogue in the style of hackage or rubygems.org would be needed for keeping track of available versions of the packages.
* When executables are added to the packaging system, I would propose to install them in the corresponding folder of its version, and also create a wrapperscript/program in a central idris-bin-dir. This wrapper should then do the following:
 * Check if there is an `ipkg`/`ipkg.lock` in the current working dir or its parrents.
  * if yes: run the executable specified by dependency-constraints
  * if no: run the most recent executable available.

This is quite a lot of writing for someone just trying to enter the project, but I would also be willing to try to get some of this stuff implemented in idris.

jfdm:
@NobbZ cheers for the suggests over what an Idris package manager should look like. Rarely have people put money where their mouth is! However, package management is not small task and there are lots of things to consider as you indicate in your above comment. I too have been working on a feature proposal for package management and see from my notes that we overlap.

[...]

Issue #1071, opened 2014-04-15
A way to list all the packages installed


bvssvni:
Asked about it on IRC, but was told there wasn't any.

NobbZ:
I thought I take this one over, since there was no activity on @steshaw so for since may 14.

But for a first quick and dirty solution I'm wondering if it is okay to introduce a new dependency (directory).

edwinb:
directory is already a dependency, so that'll be fine!

NobbZ:
Oh, ok, it was not a dependency of executable idris, I didn't take a look
into the other targets of the cabal-file.
Am 02.01.2015 20:47 schrieb "Edwin Brady" <notifications@github.com>:

jfdm:
I believe PR #1823 address this issue, so I am closing it. Any objections please reopen.

PR #1823, opened 2015-01-02
Feature/package list

NobbZ:
This is a quick and dirty solution for #1071.

I wish I were able to implement output of metadata, but since there is
no metadata saved (currently) alongside of a package, there is nothing
to display.

Displaying metadata in the output of this subcommand would mean that
there have to happen further improvements in the package-system of
idris.

Please see my proposal in #1126.

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -794,6 +794,7 @@ Executable idris
   Build-depends:  idris
                 , base
                 , filepath
+                , directory
                 , haskeline >= 0.7
                 , transformers
 
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -5,6 +5,7 @@ import System.IO
 import System.Environment
 import System.Exit
 import System.FilePath ((</>), addTrailingPathSeparator)
+import System.Directory
 
 import Data.Maybe
 import Data.Version
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -45,6 +45,7 @@ runIdris opts = do
        when (ShowIncs `elem` opts) $ runIO showIncs
        when (ShowLibs `elem` opts) $ runIO showLibs
        when (ShowLibdir `elem` opts) $ runIO showLibdir
+       when (ShowPkgs `elem` opts) $ runIO showPkgs
        case opt getClient opts of
            []    -> return ()
            (c:_) -> do setVerbose False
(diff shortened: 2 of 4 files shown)


NobbZ:
Sorry for failing tests, when I removed stylish-haskell's interference, it appears that I removed an import too much ;) Will provide an update as soon as possible

david-christiansen:
The libdir can contain subdirectories that aren't packages. For example, on my machine it presently contains the CSS files for idrisdoc in a subdirectory.

I think that your function `showPkgs` should check that the directories contain at least one `.ibc` file.

Additionally, it would be convenient if it were split into two functions: one that found the list of packages, and another that displayed it. That way, the IDE helper can easily be extended to provide a list of available packages for things like autocompletion. But I can do this myself when I get that far.

Thanks for the contribution!

NobbZ:
> The libdir can contain subdirectories that aren't packages. For example, on my machine it presently contains the CSS files for idrisdoc in a subdirectory.

Sorry, I wasn't aware of this possibility.

> I think that your function showPkgs should check that the directories contain at least one .ibc file.

I will go and implement this, but how deep do I need to look? Will there always be an `.ibc` in the root-folder of the package, or do I need to look deeper?

> Additionally, it would be convenient if it were split into two functions: one that found the list of packages, and another that displayed it.

Thought about this, but I wasn't sure where to put the function that creates the list of packages, so I put it into the display for now.

david-christiansen:
I believe they can be arbitrarily deep, though @edwinb may have changed that with the index file that he recently added.

Thanks again!

edwinb:
The index file is nothing more than a list of imports to allow :search and :apropos to know where to look, so it won’t affect anything else.

bmsherman:
I am working on a function `installedPackages :: IO [String]` that lists the installed packages (this will be useful for REPL completion as well). Once I get that, then you can just call that function for the command line function.

david-christiansen:
@edwinb My thought was that it would put a `.ibc` in the root of each package, allowing an easy way to tell packages from non-packages without recursively listing the subdirectories' contents.

LeifW:
What kind of metadata would we want for installed packages? Some kind of
version number, date installed, source, ...?

NobbZ:
@bmsherman 
> I am working on a function installedPackages :: IO [String] that lists the installed packages (this will be useful for REPL completion as well). Once I get that, then you can just call that function for the command line function.

So I leave my function as is for now. and wait until yours is finished, then it should be easily possible to use yours instead of my quickhack.

@LeifW
> What kind of metadata would we want for installed packages? Some kind of
version number, date installed, source, ...?

For displaying the packages, I think, version number is the very least thing. Next most interesting fact is installation source. The source would be most valuable if we get better package-management with different kind of sources, like local, git (branch, tag, commit) (or other VCS), some catalogue (like hackage) etc…

PR #1824, opened 2015-01-02
Default to no REPL completion if not yet implemented

bmsherman:
Before this, an incomplete pattern match would break the REPL. The missing completion that caused this error was for `PkgArgs` (which is specified for the :search and :apropos commands)

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -794,6 +794,7 @@ Executable idris
   Build-depends:  idris
                 , base
                 , filepath
+                , directory
                 , haskeline >= 0.7
                 , transformers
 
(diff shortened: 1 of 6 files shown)


david-christiansen:
Great! Thanks!

david-christiansen:
Can't #1823 be used to add real completion for PkgArgs?

david-christiansen:
Great!

effc954 creates a bit of code duplication with #1823, which is also in the merge queue. Do you think you can integrate this with @NobbZ's package listing command code?

bmsherman:
OK, I just merged in @NobbZ's changes and everything looks good!

david-christiansen:
Great! Thanks!

PR #1827, opened 2015-01-04
Documentation for modules

david-christiansen:
This allows docstrings for modules.

The syntax is the same as for other documentation, right before the "module" declaration. For an example, see the "idrisdoc009" test. Terms in module documentation are elaborated in a context including all of its imports and definitions, so the documentation can provide examples and descriptions.

The :doc command now also shows documentation for modules.


david-christiansen:
Ooops, seems I need to rebase.

david-christiansen:
darn Curses and it's odd character output!

puffnfresh:
Nice stuff :+1:

PR #1828, opened 2015-01-04
Allow building with base-4.8.0.0, deepseq 1.5, lens 4.8

RyanGlScott:
Most of these changes are to fix warnings that arise in GHC 7.10, since the `OverlappingInstances` extension was deprecated and `-fwarn-tabs` is now enabled with `-Wall`. This also bumps the upper version bounds of `deepseq` and `lens`, since none of the changes introduced in `deepseq-1.4` and `lens-4.7` affect `idris`.

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -725,12 +725,12 @@ Library
                 , bytestring < 0.11
                 , cheapskate < 0.2
                 , containers >= 0.5 && < 0.6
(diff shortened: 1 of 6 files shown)


puffnfresh:
:+1:

david-christiansen:
Thanks!

PR #1831, opened 2015-01-04
enable -Werror on travis and encorage to use it locally, but not on hackage!

NobbZ:
Due to an update in the `mtl`-library, a deprecation warning was emitted in many modules, in some of them `-Werror` was set locally for this specific module. This local usage of `-Werror` caused idris to not build anymore, even if it had worked.

There was some discussion going on in IRC, also the PR #1829 was created because of this.

This PR is thought as an addition to already mentioned #1829. If really al `-Werror` are removed from codebase, there might be some cases where we need it! If there are warnings introduced because of dependency updates, or other stuff, they might be overseen easily when compiling locally, but by using this flag on travis there will be some reminder when warnings occur.

**WARNING**: This might fail on travis, because of changes on hackage, I can't foresee, but I did a complete clean and update of all dependencies right before preparing this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ before_script:
   - tar -xf Idris-dev/dist/idris*.tar.gz
   - cd idris*
 script:
-  - cabal configure -f FFI
+  - cabal configure -f FFI -f CI
   - if [[ "$TESTS" != "doc" ]]; then cabal build; fi
   - if [[ "$TESTS" != "doc" ]]; then cabal copy; fi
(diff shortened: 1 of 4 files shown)


david-christiansen:
What's the reasoning behind having a CI flag instead of putting a `--ghc-options=-Werror` on the Cabal command line in the Travis script?

NobbZ:
You can easily switch it on or off during local development if you need.

david-christiansen:
I guess we can also use it to add other Travis-specific hooks in the future.

PR #1834, opened 2015-01-05
Added missing documentation for the CLI.

jfdm:
Although `idris --help` lists all the available options, several of
the more commonly used options were not being listed in *Available
Options*. For instance: `-p <pkg>`.  This was due to a lack of
documentation in `CmdOptions`. This commit fixes this,


david-christiansen:
Thanks! Nice catch!

PR #1838, opened 2015-01-08
Add type class and instances for monad transformers

L8D:
Some of us don't want to use effects...


david-christiansen:
Thanks!

L8D:
Should I turn all the other monads into transformers? MaybeT, EitherT...

david-christiansen:
If you'd like. Be careful to get laziness right - I would need to think a
bit about what should be strict, and I certainly haven't done that.

It would also be nice if you put docstrings on at least the types, if not
their constructors.

2015-01-10 21:42 GMT-08:00 Tenor Biel <notifications@github.com>:

PR #1854, opened 2015-01-15
Handle connection error with --client

elliottt:
When the REPL is started with --client and there's no existing REPL for it to
talk to, catch the exception thrown by `connectTo`, and dump a more useful
message.


david-christiansen:
Thanks!

PR #1857, opened 2015-01-16
Relax lower bound on mtl

david-christiansen:
This lets Idris work with a wider variety of mtl and transformers
versions. In particular, it causes it not to try to replace GHC's own
transformer's lib.

The fix is actually by @glguy, but I couldn't figure out how to
credit him properly in the commit log with the patch file.

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -751,7 +751,7 @@ Library
                 , fingertree >= 0.1 && < 0.2
                 , haskeline >= 0.7 && < 0.8
                 , lens >= 4.1.1 && < 4.8
-                , mtl >= 2.2.1 && < 2.3
+                , mtl >= 2.1 && < 2.3
                 , network < 2.7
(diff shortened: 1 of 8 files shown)


vendethiel:
> The fix is actually by @glguy, but I couldn't figure out how to
credit him properly in the commit log with the patch file.

Doesn't `git commit --author` show up properly on github?

david-christiansen:
Quite possibly. I'll poke around.

/David (from phone)
Den 16 jan 2015 18:30 skrev "ven" <notifications@github.com>:

glguy:
Thanks for the mention, David. Please don't spend any of your time worrying
about putting my name anywhere on my behalf. I'm just glad the change
helped.

david-christiansen:
@glguy : @vendethiel's comment made me realize how to do it without needing to use more than 2 minutes. So I stuck your ID on the commit.

Thanks again! This will hopefully solve conflict issues for some users.

PR #1859, opened 2015-01-17
Haskell Platform compatibility with transformers and mtl

RyanGlScott:
Currently, it is tricky to install `idris` with the latest version of the Haskell Platform, since it is shipped with `mtl-2.1.3.1` but `idris` requires >= 2.2.1. This commit allows older `mtl` and `transformers` versions to be used to build `idris`, using CPP pragmas to import `ExceptT` instead of `ErrorT` if using the latest versions so as to avoid deprecation warnings.

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -751,7 +751,7 @@ Library
                 , fingertree >= 0.1 && < 0.2
                 , haskeline >= 0.7 && < 0.8
                 , lens >= 4.1.1 && < 4.8
-                , mtl >= 2.2.1 && < 2.3
+                , mtl >= 2.0.1 && < 2.3
                 , network < 2.7
                 , optparse-applicative >= 0.11 && < 0.12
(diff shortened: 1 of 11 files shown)


RyanGlScott:
Whoops, I didn't see [this pull request](https://github.com/idris-lang/Idris-dev/pull/1857) that accomplishes the same thing. My bad.

david-christiansen:
Thanks for putting in all that work, and I'm sorry that streams were crossed!

As far as I've heard from people more knowledgeable than myself (namely @cartazio and @glguy), `transformers-compat` is the preferred solution rather than manually doing `cpp` in each program. It certainly seems cleaner to me.

Anyway, thanks again!

RyanGlScott:
You're right, that is an order of magnitude cleaner. In fact, I should probably go back through some of my libraries since I've probably cluttered them up needlessly with `cpp` hacks when I could just use `transformers-compat` instead.

PR #1870, opened 2015-01-23
Support for module import in IDESlave

soimort:
Fixes idris-hackers/idris-mode#295.


david-christiansen:
Great! Travis seems to be doing its thing - hopefully Trifecta has had a hackage release lately.

Perhaps we should add a message to IDE mode that lists the loaded modules, so it can be shown somewhere other than the prompt, such as in a header line akin to the one that shows the help in the prover.

soimort:
I agree - this is just a quick fix for now (on the Idris side), but idris-mode could do more than that (will look further into this definitely!).

PR #1872, opened 2015-01-23
Bump utf8-string upper version bounds

RyanGlScott:
Allow `idris` to build with `utf8-string-1`.

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -765,7 +765,7 @@ Library
                 , trifecta >= 1.1 && < 1.6
                 , uniplate >=1.6 && < 1.7
                 , unordered-containers < 0.3
-                , utf8-string < 0.4
+                , utf8-string < 1.1
                 , vector < 0.11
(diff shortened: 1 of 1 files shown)


edwinb:
I'm a bit scared now to apply changes involving widening the version number bounds, because our experience seems to suggest that a wider range of possible versions leads to more possible build failures, not less.

Is there a particular need for the later version? Generally I'd advise building in a sandbox if it's about other Haskell libraries you have installed.

RyanGlScott:
I'm in no position to make a judgment on keeping dependencies up-to-date, but I can say that the only difference between `utf8-string-0.3.8` and `ut8-string-1` is that `utf8-string-1` removes the `System.Environment.UTF8` and `System.IO.UTF8` modules, since GHC already supports UTF8 IO. `idris` doesn't use the modules anyways, so this doesn't affect the build at all.

PR #1874, opened 2015-01-25
Add IDRIS_LIBRARY_PATH environment variable

puffnfresh:
Allows specifying where to find Idris libraries.

After doing a build, you can even point Idris to the repo's lib path:

    IDRIS_LIBRARY_PATH=libs dist/build/idris/idris


david-christiansen:
This is really a great thing to have!

Is there a reason not to have the path be colon-separated, like with normal $PATH and in things like $PYTHONPATH?

puffnfresh:
@david-christiansen no good reason, I really want to get there.

Issue #1868, opened 2015-01-22
Pattern variables can't start with _


david-christiansen:
The following doesn't work:
```
module Underscore

foo : Nat -> Nat
foo _bar = Z

```

It fails with the error:
```
     When elaborating left hand side of foo:
     Can't unify
             Nat
     with
             argTy -> retTy
```

which makes me think it's not lexing the underscore as part of the name, but rather as a separate pattern.

The following type checks:
```
module Underscore

foo : Nat -> Nat
foo bar = Z
```

edwinb:
Yes, the lexer doesn’t allow this, I’ve noticed this before. I don’t think this was deliberate (I certainly didn’t intend it). It’s probably a quick fix if anyone has a moment to look at it.

david-christiansen:
I agree that it should be easy - I just wanted to document it quickly while working through exercises with some folks here.

PR #1880, opened 2015-02-01
Allow identifiers to start with underscore (_).

ahmadsalim:
Fixes idris-lang/Idris-dev#1868

### DIFF
--- a/src/Idris/AbsSyntaxTree.hs
+++ b/src/Idris/AbsSyntaxTree.hs
@@ -1655,7 +1655,7 @@ prettyName
   -> Doc OutputAnnotation
 prettyName infixParen showNS bnd n
     | (MN _ s) <- n, isPrefixOf "_" $ T.unpack s = text "_"
-    | (UN n') <- n, isPrefixOf "_" $ T.unpack n' = text "_"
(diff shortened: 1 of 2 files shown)


src/Idris/AbsSyntaxTree.hs
```diff
@@ -1655,7 +1655,7 @@ prettyName
   -> Doc OutputAnnotation
 prettyName infixParen showNS bnd n
     | (MN _ s) <- n, isPrefixOf "_" $ T.unpack s = text "_"
-    | (UN n') <- n, isPrefixOf "_" $ T.unpack n' = text "_"
+    | (UN n') <- n, T.unpack n' == "_" = text "_"
```
ahmadsalim: Just FYI for the code reviewer, I am bit unsure if this change breaks something else also everything seemed to work.
If this change was not made, then it would have only displayed "_" for names starting with "_"

david-christiansen:
I think that was my change you asked about - it was due to some internal name getting called something like (UN "__pi_arg") and the change to an MN broke it for some reason. But I'd rather show ugly internal names than hide good names, and the real fix is to use MNs for machine names, so I'd say merge and see what happens (once Travis is happy!).

Thanks!

PR #1884, opened 2015-02-02
Make isTConName and friends exact

david-christiansen:
They should only be used in a disambiguated context anyway to avoid it
picking the wrong one. Before this change, the coloring was wrong on

    :let data A = A.A
    :t A

Now it is fixed.

### DIFF
--- a/src/Idris/AbsSyntax.hs
+++ b/src/Idris/AbsSyntax.hs
@@ -1734,7 +1734,11 @@ stripUnmatchable i (PApp fc fn args) = PApp fc fn (fmap (fmap su) args) where
        | (Bind n (Pi _ t _) sc :_) <- lookupTy f (tt_ctxt i)
           = Placeholder
     su (PApp fc f@(PRef _ fn) args)
-       | isDConName fn ctxt 
+       -- here we use canBeDConName because the impossible pattern
+       -- check will not necessarily fully resolve constructor names,
+       -- and these bare names will otherwise get in the way of
+       -- impossbility checking.
+       | canBeDConName fn ctxt
           = PApp fc f (fmap (fmap su) args)
     su (PApp fc f args)
           = PHidden (PApp fc f args)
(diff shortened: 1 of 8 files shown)


david-christiansen:
Waiting for travis due to slow local tests on Chromebook :-)

david-christiansen:
Thanks to @Melvar for finding the issue.

david-christiansen:
@edwinb: Assuming the tests pass on Travis, I'd appreciate a code review/design description on [one little part of this](https://github.com/david-christiansen/Idris-dev/blob/df574ec148335d7f1ff4efd6dcdb7429fffe6089/src/Idris/Coverage.hs#L31-L33) prior to my merging it.

This fixed the test failure in cases where there isn't overloading for a constructor name, but it seems to me that there still might be issues in totality checking ambiguous constructor names. In particular, the following variant of `totality006.idr` type checks:

```
module CoverageTest

import Data.So

total
blargh : (xs, ys : List a) -> So (length xs > length ys) -> GT (length xs) (length ys)
blargh [] [] Oh impossible
blargh [] (y :: xs) so = absurd so
blargh (y :: xs) [] Oh = LTESucc LTEZero
-- missing case: blargh (y :: xs) (z :: ys) x = ?blargh_rhs_3
```
It seems that this is because of the assumption commented slightly further down in the code, that "matching will ignore the namespace".  That ignoring of namespace was encoded in the old semantics of `isDConName` where it looked if the name _could be_ a constructor. But isn't the property that we really want to look for that it could be a constructor of the family in question? This will also need a lot more pushing things around for sufficiently dependent types, unless there's a better way.

david-christiansen:
The changes have been implemented to fix the coloring bug, but making `isDConName` exact everywhere would have been a rewrite of a big part of the coverage checker. For now, that's allowed to be inexact, and the pretty-print bug is fixed.

The aforementioned request for comment from @edwinb happened on IRC.

PR #1892, opened 2015-02-05
Fix function wrapping bug in JavaScript RTS

### LINKED ISSUE #1531: JavaScript FFI doesn't allow multi-parameter functions
I had a function I wanted to wrap in JavaScript:

```JavaScript
function eventFold(element, eventType, func, initial)
{
    element.addEventListener(eventType, function(eventVal){
        makeFoldable(func, initial, eventVal);
    });
}

function makeFoldable(func, val, eventVal)
{
    if(typeof makeFoldable.accum == "undefined")
        makeFoldable.accum = val;
    makeFoldable.accum = func(val, eventVal);
}
```

With the following Idris to wrap it:

```Idris
Event : Type
Event = Ptr

eventFold : EventType -> (Event -> a -> IO a) -> a -> Element -> IO ()
eventFold {a} eventType f val (MkElem elem) =
    let eventStr = show eventType
    in mkForeign (FFun "eventFold(%0, %1, %2, %3)"
                          [FPtr, FString,
                           FFunction FPtr (FFunction (FAny a) (FAny (IO ()))),
                           FAny a]
                          FUnit)
                          element eventStr f val
```

And when I try to use this function like `eventFold Click update 0 slideshow`, it passes `update` the correct value for the first argument, but on the second, polymorphic argument, it seems to pass a JavaScript value with the type "MouseEvent", but cast as whatever the polymorphic type `a` is, in this case `Int`.

I think this is a known problem, as I saw it referenced very generally in a project, but I couldn't find an issue for it.

fiendfan1:
`i$ffiWrap` was basically apply the first argument over
and over instead of for looping through each argument.
ie.

```JavaScript
foreach(arg in arguments) {
  while(function is not fully applied) {
    func.apply(arg);
  }
}
```

This has made it impossible to use native JavaScript
higher-order functions through the FFI.

`make test_js` passes, but I'm not familiar with the
js codegen/rts, so someone should check and make sure this
is correct.

This fixes #1531

### DIFF
--- a/jsrts/Runtime-common.js
+++ b/jsrts/Runtime-common.js
@@ -68,20 +68,18 @@ var i$ffiWrap = function(fid,oldbase,myoldbase) {
     var res = fid;
 
     for(var i = 0; i < (arguments.length ? arguments.length : 1); ++i) {
-      while (res instanceof i$CON) {
-        i$valstack_top += 1;
-        i$valstack[i$valstack_top] = res;
-        i$valstack[i$valstack_top + 1] = arguments[i];
-        i$SLIDE(2);
-        i$valstack_top = i$valstack_base + 2;
-        i$CALL(_idris__123_APPLY0_125_,[oldbase])
-        while (i$callstack.length) {
-          var func = i$callstack.pop();
-          var args = i$callstack.pop();
-          func.apply(this,args);
-        }
-        res = i$ret;
+      i$valstack_top += 1;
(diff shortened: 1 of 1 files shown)


david-christiansen:
@raichoo I believe reviewing this one is in your side of things :-)

edwinb:
I'm not especially familiar, but it looks okay to me (and if it fixes things and the tests still pass, that's usually fine :))

PR #1902, opened 2015-02-09
Added the Zero Constructor to Data.Sign

JakobBruenker:
Added the Zero Constructor to Sign and adapted the ZZ instance
this addition.

### DIFF
--- a/libs/base/Data/Sign.idr
+++ b/libs/base/Data/Sign.idr
@@ -1,7 +1,7 @@
 module Data.Sign
 
 ||| A representation of signs for signed datatypes such as `ZZ`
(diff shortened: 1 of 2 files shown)


JakobBruenker:
Oops, messed up the commit message. 'This addition' shouldn't be there.

JakobBruenker:
To elaborate on why I think this makes sense: The mathematical signum function maps numbers to a set of three possible values: -1, 0, and 1. But Data.Sign only allowed for Minus and Plus.

david-christiansen:
Seems reasonable enough. I didn't really have any use in mind when I made Sign, so it was probably wrong :-)

PR #1905, opened 2015-02-10
update contributors to include myself

JakobBruenker:
since I have submitted two pull requests which have been accepted,
D. Christiansen suggested I should be in this file.


david-christiansen:
I'll merge this when I get the chance to realphabetize the list.  Right now my local tree is a bit of a mess, but it should come soon.

PR #1906, opened 2015-02-10
Parse lambdas with impossible

Melvar:
An attempt at making things like `the (True = False -> Void) (\(Refl) impossible)` work. This would reduce the need for small lemmas that only consist of a single clause with `impossible`.

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -609,6 +609,9 @@ Extra-source-files:
                        test/totality008/run
                        test/totality008/*.idr
                        test/totality008/expected
+                       test/totality009/run
+                       test/totality009/*.idr
+                       test/totality009/expected
 
                        test/tutorial001/run
                        test/tutorial001/*.idr
--- a/src/Idris/ParseExpr.hs
+++ b/src/Idris/ParseExpr.hs
@@ -234,6 +234,14 @@ internalExpr syn =
      <|> app syn
      <?> "expression"
 
+{- | Parses the "impossible" keyword
+@
+Impossible ::= 'impossible'
+@
+-}
+impossible :: IdrisParser PTerm
+impossible = PImpossible <$ reserved "impossible"
+
 {- | Parses a case expression
 @
 CaseExpr ::=
@@ -250,13 +258,13 @@ caseExpr syn = do reserved "case"; fc <- getFC
 {- | Parses a case in a case expression
 @
 CaseOption ::=
-  Expr '=>' Expr Terminator
+  Expr (Impossible | '=>' Expr) Terminator
   ;
 @
 -}
 caseOption :: SyntaxInfo -> IdrisParser (PTerm, PTerm)
 caseOption syn = do lhs <- expr (syn { inPattern = True })
-                    r <- PImpossible <$ reserved "impossible" <|> do symbol "=>"; expr syn
+                    r <- impossible <|> symbol "=>" *> expr syn
                     return (lhs, r)
                  <?> "case option"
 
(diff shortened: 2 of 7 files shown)


david-christiansen:
I like this a lot, but as it's a bigger language change, I'd like @edwinb to weigh in before I press the "merge" button.

A test of how this interacts with the totality checker would also be nice to have.

edwinb:
Seems good to me - perhaps we should also have "impossible" on case expressions, to have it all consistent. Also I agree a test would be useful (probably in the "totality" section).

Melvar:
I’ll come up with a test. Do you want me to immediately also adjust case expressions this way, or was that merely a hypothetical?

david-christiansen:
If you can do it easily, please do! I'd be in favor of merging this once there's a test, though, and making an issue to track the similar feature for `case`.

david-christiansen:
@Melvar how is this coming? Do you need some help with it or is it just a matter of finding time? Let me know if I can be useful!

Melvar:
The current holdup is because I’m handing in my Bachelor’s thesis in two days, i.e. the latter.

david-christiansen:
I'm glad that your priorities are in order :-)

david-christiansen:
@Melvar I think we had an IRC conversation about what needs to happen with this but I can't for the life of me remember what we figured out. What is needed for this? Is it just a test or was there something else?

Melvar:
I need to get back to this. Needs rebasing, cleaning up, and the test. I believe what we figured out is that making this work causes an INTERNAL ERROR to leak out, and I’m not sure where to fix it, so you should look at it once I have it that far properly.

david-christiansen:
All right. I'll wait for a ping from you to do something about this one.

Melvar:
Now, something like `\x impossible`, where `x` is *not* a constructor, fails with `INTERNAL ERROR: Unelaboratable syntactic form impossible`, while `\False impossible` correctly complains `(input):1:8:case block in {val0} pargTy pretTy False lamp is a valid case`. The problem occurs when one has a lambda that has only bare identifiers on the left side (gets parsed in the first branch of `lambda`) and is *not* later identified as “pattern-matching after all” from some of the identifiers referring to known constructors.

I don’t know how this problem should be fixed.

david-christiansen:
I suppose the way to fix this is to update ElabTerm.hs to have an explicit case for "impossible" that gives a better error message - even just having it send a message like:
    "impossible" doesn't make sense here - have you used it with a non-pattern-matching lambda?
That would be an extra case on `elab'` in `ElabTerm.hs` - something like
   elab' ina fc PImpossible = lift . tfail . Msg $ "\"impossible\" doesn't make sense here - have you used it with a non-pattern-matching lambda?"
should do the trick for that error message.

david-christiansen:
Here's something that doesn't make sense to me using this branch.
```
foo : Dec (2 = 3)
foo = No $ \(Refl) impossible
```
type checks fine, but

```
foo : Dec (2 = 3)
foo = No $ \Refl impossible
```
gives the error:
```
case block in foo Refl lamp is a valid case
```

Is this expected behavior?

david-christiansen:
It seems to be a difference between pattern-matching lambdas created by the implicit-adder and pattern-matching lambdas created by the parser.

Melvar:
I had noticed this, which is why the very first comment of this PR includes the parens. I had then forgotten it would be an obvious usability problem, thanks for reminding me. It seems to be restricted to `Refl`, which makes me suspect it’s something about `Refl` being special-cased in the parser or elsewhere.

Possibly the *`Name`* `Refl` doesn’t exist or doesn’t refer to a constructor in the right way at the right time, because `Refl` is normally parsed specially?

david-christiansen:
(for future readers: we worked out on IRC that it's the special handling of Refl in various places that is the problem here)

PR #1907, opened 2015-02-10
Added static analysis to C RTS

ahmadsalim:
Added `cppcheck` to run on the C Idris RTS folder.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ before_install:
   - sudo apt-get install hscolour
   # test dependency
   - sudo apt-get install -qq expect
+  - sudo apt-get install -qq cppcheck
   - cabal install alex-3.1.3
 install:
   - cabal install -f FFI --enable-tests --dependencies-only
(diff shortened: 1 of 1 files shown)


ahmadsalim:
One could also investigate adding http://www.coverity.com, which is free for open-source projects.
This however requires someone with commit rights.

puffnfresh:
cppcheck looks like it gave some useful analysis already. Nice! :+1:

jfdm:
Ah yes, use of Coverity Scan would be nice as well. I have been meaning to look at integrating some static analysis of the rtses.  I will look into it during my hackthon time.

mattias-lundell:
I tried coverity for Idris rts a month ago https://scan.coverity.com/projects/4009. However it seems that only members of the project can view results.

jfdm:
The reason being is [responsible disclosure](https://scan.coverity.com/faq#who-can-have-access), which I am all in favour for. I am all in favour of using coverity scan but the final decision should lie with those who have commit rights, as they will see the increase in mail.

PR #1912, opened 2015-02-12
Rename ideslave to ide-mode

david-christiansen:
This patch changes the name of the machine-readable REPL syntax from
ideslave to ide-mode. The old command line options still work, but are
announced as deprecated for a while.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,8 @@
+New in 0.9.17:
+--------------
+* The --ideslave command line option has been replaced with a --ide-mode
+  command line option with the same semantics.
+
 New in 0.9.16:
 --------------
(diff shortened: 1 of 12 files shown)


peterohanley:
There is one surviving instance of ideslave, which is only used by Prover.hs
https://github.com/idris-lang/Idris-dev/blob/master/src/Idris/Output.hs#L156
https://github.com/idris-lang/Idris-dev/blob/master/src/Idris/Prover.hs#L70

david-christiansen:
Darn those inconsistent cases breaking my find-replace!

I'll fix it soon. Thanks!

2015-02-12 22:33 GMT+01:00 peterohanley <notifications@github.com>:

PR #1918, opened 2015-02-13
Added optional argument to intros to tell idris how many assumptions should be introduced

JakobBruenker:
Writing "intros" does the same thing it always did, so there are no problems regarding backwards compatibility. Writing "intros n", where n is a natural number (including 0), tries to introduce n assumptions. If there aren't that many assumptions left, it will fail. (Which means "intros 0" will always be successful and will do nothing.)

I'm not entirely sure if I did everything correctly, though. Specifically:
 - In AbsSyntaxTree.hs, the size of Intros was 1, and I left it at 1.
 - In ElabTerm.hs, I moved Intros from reify to reifyApp, and let it convert the one term it accepts into a constant. I'm not sure how I can test whether I did the reflected representation stuff correctly.
 - I would have preferred to use Maybe Int instead of Int (where -1 represents Nothing), but I don't think that's possible because of the reflected representation. I can convert a Term to a Const, but Const doesn't have a Maybe Constructor

But it typechecks, works, and passes the tests.

### DIFF
--- a/src/Idris/AbsSyntaxTree.hs
+++ b/src/Idris/AbsSyntaxTree.hs
@@ -824,7 +824,8 @@ mapPT f t = f (mpt t) where
   mpt x = x
 
 
-data PTactic' t = Intro [Name] | Intros | Focus Name
+data PTactic' t = Intro [Name] | Intros Int | Focus Name
+                  -- Intros (-1) corresponds to the tactic w/o arguments
                 | Refine Name [Bool] | Rewrite t | DoUnify
                 | Induction t
                 | CaseTac t
@@ -859,7 +860,7 @@ deriving instance NFData PTactic'
 !-}
 instance Sized a => Sized (PTactic' a) where
   size (Intro nms) = 1 + size nms
-  size Intros = 1
+  size (Intros _) = 1
   size (Focus nm) = 1 + size nm
   size (Refine nm bs) = 1 + size nm + length bs
   size (Rewrite t) = 1 + size t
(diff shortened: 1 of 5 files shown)


david-christiansen:
I think a Maybe would be much cleaner. It does make the reflection work slightly more complicated, but I think it's worth it. Grab me on IRC, and I'll walk you through it.

Also, shouldn't the argument be `Nat` rather than `Int`? Does it ever make sense to introduce a negative number of assumptions?

JakobBruenker:
It's probably going to take 5 days or so before I get the time to take a deeper look at how reification works and finish this

JakobBruenker:
5 days seems to have been a little too optimistic... I might get back to this at some point, but it probably makes sense to close this for now, as I don't seem to be able to find the time at the moment.

david-christiansen:
OK.

In the meantime, I might be able to get the new tactics language working. Then you can define your own `intros`:

```idris
introN : Nat -> Tactical ()
introN Z = return ()
introN (S n) = do intro
                  introN n
```

david-christiansen:
Or perhaps more realistically:
```idris
introN : Nat -> Tactical ()
introN Z = return ()
introN (S n) = do (_, g) <- goal
                  case g of
                    (Bind n (Pi _ _ _) _) => do intro (Just n) ; introN n
                    _ => do intro Nothing ; introN n
```

JakobBruenker:
Oh, that looks cool, I'm looking forward to it.

david-christiansen:
Yeah, that's what I get for writing my code outside the type checker!

PR #1919, opened 2015-02-13
Added error reporting on eliminator generation failure

ahmadsalim:
This PR enables showing errors in log when failing to generate eliminators.

### DIFF
--- a/src/Idris/Elab/Data.hs
+++ b/src/Idris/Elab/Data.hs
@@ -317,12 +317,12 @@ elabCaseFun ind paramPos n ty cons info = do
   let eliminatorDef = PClauses emptyFC [TotalFn] elimDeclName eliminatorClauses
   elimLog $ "-- case function definition: " ++ (show . showDeclImp verbosePPOption) eliminatorDef
   State.lift $ idrisCatch (rec_elabDecl info EAll info eliminatorTyDecl)
-                    (\err -> do { e <- showErr err ; logLvl 2 ("Error while elaborating eliminator type declaration: " ++ e)})
+                    (ierror . Elaborating "type declaration of " elimDeclName)
   -- Do not elaborate clauses if there aren't any
   case eliminatorClauses of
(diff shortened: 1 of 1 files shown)


david-christiansen:
Shouldn't this cause elaboration to abort and say something like:

```
Test.idr:9:1-12:18:
When elaborating eliminator for datatype Foo:
  ERROR HERE
```
?  Should be easy enough with the current `Err` constructors.

ahmadsalim:
@david-christiansen For some reason I believe that there was a reason why eliminators have silent error messages. Perhaps, it may not be valid anymore.

david-christiansen:
I think that if a user asks for an eliminator and doesn't get one, then it's better to fail at that time. I don't remember any reason we would have had to hide them.

edwinb:
I can't remember what the reason was if there was one... I think it's better to fail immediately too.

My preferred behaviour would still be to generate eliminators on demand when "induction" is first used on something (and report an error if that fails) though it will be a bit fiddly to implement this. This is partly because the '%elim' annotation is a bit ugly even if it's useful for now, partly because a lot of the time eliminators simply aren't needed for a type so it's a waste of time to generate them, and partly because you might want eliminators even when a library author hasn't though to mark a type as %elim.

david-christiansen:
Agree on both points.

2015-02-14 23:39 GMT+01:00 Edwin Brady <notifications@github.com>:

ahmadsalim:
@edwinb @david-christiansen 
On the first point, I will try to change it to fail immediately soon.

On the second point, I think that perhaps the eliminator infrastructure might need to be rewritten again (preferably directly on TT datatypes) and in that version I agree that the eliminator should be generated on demand (when used by elim_for or the induction tactic). 
Time is a bit pressing so it may take some time, but if someone else is interested to rewrite soon (in a smaller project) I can give some pointers.

david-christiansen:
The new tactic system that I'm working on may be enough to move eliminator generation into Idris code - perhaps it might be good to hold off a bit on the big rewrite!

PR #1929, opened 2015-02-18
add missing HAS_PTHREAD wrapper

Heather:
``` diff
 void init_threadkeys() {
+#ifdef HAS_PTHREAD
     pthread_key_create(&vm_key, (void*)free_key);		     pthread_key_create(&vm_key, (void*)free_key);
+#endif
 }
```


edwinb:
Ah yes, thanks.

PR #1931, opened 2015-02-18
Operator Standardization Proposal

L8D:
`(<.>)` from Data.Matrix is now `(<:>)`.
`(<*>)` from Prelude.Algebra is now `(<.>)`.
`(<$>)` from Prelude.Applicative is now `(<*>)` as in Haskell.
`map` from Prelude.Functor is now `(<$>)` with `map` as an alias.

Let me know what you guys think ;)


david-christiansen:
I'm generally in favor of not having pointless incompatibilities with other languages' APIs.

As @Melvar pointed out on IRC, you should probably also rename `$>` and `<$` to their `*` equivalents. If that happens, the proposal has my +1.  I would want to look the code over once as well, but based on our earlier conversations I don't expect to find any monsters in  there :-)

L8D:
`(<$)` and `($>)` from Prelude.Applicative are now `(<*)` and `(*>)` as of ac17903

test/basic001/reg005.dSYM/Contents/Info.plist
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
david-christiansen: Please remove this file from the PR and, if possible, add a pattern for it to `.gitignore`. No need for us to have this kind of thing in the repo!

david-christiansen:
There seems to have been some Mac filesystem noise committed by accident in `test/basic001/reg005.dSYM`. Would you mind updating the PR with a commit to remove it again?

Otherwise, the code looks good to me.

edwinb:
This looks great, thanks.

I think I slightly prefer having `map` in the type class and `<$>` as the alias though. I think because `map` is a fairly conventional name for this, and I'd rather minimise the number of operators that people need to learn at first, especially people who don't come from a Haskell background. (Also, I suppose, because I generally prefer using names to symbols.)

It's probably not that important because of the alias, and because interactive editing gives prototype type class bodies automatically, but I'm just thinking about minimising difficulty when trying to explain Functors for the first time.

(Why don't I think the same about `<*>` in Applicative? Not sure... just a feeling. Maybe just because `map` is the name I've been using for this for ages. Or because `<*>` feels like it's standing for something that ought to be a space...)

david-christiansen:
RE `map` vs `<$>`: this was actually my first reaction as well. @L8D explained it as being motivated by consistency with `<*>` and `>>=`, which makes sense to me, but I do actually still like the name being `map`. In any case, though, I wouldn't really mind either way.

jonsterling:
Please don't take what I have to say too seriously since I don't have much Idris code except a few experiments lying around, but I would like to say that I also prefer to keep full names in type classes, and then give operators as aliases separately.

I'd even prefer to have `<*>` and `>>=` be aliases for a known name, since in mathematical practice, choice of operator symbol is often local...

But anyway, I don't feel so strongly about it that I'd have more to say than this tidbit. Cheers!

LeifW:
I'd tend to concur with jonsterling. I'm already used to having `map`, `ap`, and `bind` be the names of things defined on the typeclasses from scalaz, with the operators being add-on syntax.

L8D:
Then perhaps we flip this around and do:

`(<$>)` from Prelude.Applicative becomes `ap` with `(<*>)` as an alias (and `($>)`, `(<$)` get changed accordingly).
`(>>=)` from Prelude.Monad becomes `bind : (a -> m b) -> m a -> m b` with `(>>=) : m a -> (a -> m b) -> m b` being the flipped alias
Prelude.Functor gets `(<$>)` as an alias for `map`

Does that sound more reasonable?

Melvar:
I can sorta see that `ap` and `bind` could be useful in some compositions, and it’s probably better to have more written-out named functions as more basic. I think I like the proposal of having them as the methods better than the one changing `map` to `(<$>)` as the method.

david-christiansen:
Sorry for the lack of action on this - I think that we just need a decision now. I'm happy with anything that increases Haskell API compatibility. 

I'm actually not a big fan of calling `<*>` `ap` and calling `>>=` `bind`, for only one reason: while working with Idris, we often encounter normalized terms that are reduced as far as possible. This occurs especially in error messages, but it can also just happen while playing at the REPL. Usually, normalization is able to resolve type classes, and replace `+` with `plus` in the case of `Nat`, and it will probably do the same for `map` and `bind` and `ap`. To keep terms in error messages as close to source terms as possible, I'd like to suggest that we use (or keep) the most-used names as type class method names.

The use of `map` vs `<$>` seems to be fairly evenly split. Thus, I'd say that this criterion doesn't tell us which to use, so we might as well stick with the `map` if we otherwise have a preference for non-operator names. But `>>=` is everywhere, as is `<*>`. In these cases, I think we should stick to the operator names in @L8D's original PR.

I can update the patch to this standard in the next couple of days.

edwinb:
For the record, my preference is 'map' as a name, and the others as operators.

In general, my preference is for names where possible, but the operators are not only everywhere but also very often nested (and have syntactic sugar on top of them in the form of do notation and idiom brackets).

My preference is a very weak one anyway :).

L8D:
Based on @david-christiansen's suggestion, we should pull at `e18c469`

david-christiansen:
OK, it seems like we've gotten a rough consensus together here. I'll do some git cherry-picking and get the operators from e18c469 in place along with the test updates, and merge.

david-christiansen:
I believe that #1970 collects the consensus decision from this PR - is that right? If so, mind if I close this one?

PR #1932, opened 2015-02-20
Start of new tactics system

david-christiansen:
This also fixes bit rot in reflection and IBC.hs as well as letting quasiquotes support `Raw` in expression contexts.


david-christiansen:
This is ongoing work - merging now because it incidentally fixed some things that others need.

PR #1934, opened 2015-02-20
update sqlite3 bindings

Heather:


david-christiansen:
Thanks for updating these!

Is there a reason why you don't use https://github.com/david-christiansen/IdrisSqlite? The reason I'm interested is that I think the one in `contrib/`should probably either be deleted or spun out into a separate repo and I'd like to figure out which option.

Heather:
@david-christiansen it's just recently I was searching for `mkForeign` and discovered that there are something inside repository which is not updated yet so I desired to update it, just in this case.

jfdm:
I have added the IdrisSqlite project to the third-party pages. This should help with discoverability of projects.

david-christiansen:
Thanks, @jfdm. In the meantime, it doesn't hurt to have things in the repo actually work :)

PR #1937, opened 2015-02-22
Set encoding to UTF8 on Windows and fix :exec

melted:
For some inexplicable reason Haskell hardcodes latin1 as the encoding on
Windows. This stops the output when an unicode character is reached. It's better
 all over to set the encoding to UTF8, if the user has a UTF8 terminal like
mintty it looks good. If the user has a terminal with latin1, there will
be some bad chars but they will see the entire text otherwise.

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -3,7 +3,6 @@ module Main where
 import System.Console.Haskeline
 import System.IO
 import System.Environment
-import System.Info
 import System.Exit
 import System.FilePath ((</>), addTrailingPathSeparator)
 import System.Directory
@@ -30,6 +29,7 @@ import Idris.CmdOptions
 import IRTS.System ( getLibFlags, getIdrisLibDir, getIncFlags )
 
 import Util.DynamicLinker
+import Util.System
 
 import Pkg.Package
 
@@ -40,7 +40,7 @@ import Paths_idris
 
 main :: IO ()
 main = do
-          when (os == "mingw32") $  do hSetEncoding stdout utf8
+          when isWindows $ do hSetEncoding stdout utf8
           opts <- runArgParser
           runMain (runIdris opts)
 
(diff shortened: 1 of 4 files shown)


melted:
The :exec fix was made on the utf8 fix, so they have to go in the same PR.

main/Main.hs
```diff
@@ -38,7 +39,9 @@ import Paths_idris
 -- on with the REPL.
 
 main :: IO ()
-main = do opts <- runArgParser
+main = do
```
Heather: I don't like new line here

Heather:
@melted doesn't encoding came from system defaults?
What if there will be linux with non-utf8 locale as default one?

LeifW:
We've talked about just ignoring system specified encoding and just assuming .idr files are in utf8 (gotten build errors in the past when the stdlib files have unicode in them and people try builing them in a C locale or something).
Do we want to "if windows" check before `hSetEncoding stdout utf8`? Or should we just set it that way on all platforms?

melted:
This fixes the output at the repl and the terminal that will choke with Unicode characters if the encoding is set to latin1. The specific case that bugged me was that the help text was truncated thanks to a Unicode apostrophe.

Från: "Leif Warner" <notifications@github.com>
Skickat: ‎2015-‎02-‎23 07:33
Ämne: Re: [Idris-dev] Set encoding to UTF8 on Windows and fix :exec (#1937)

We've talked about just ignoring system specified encoding and just assuming .idr files are in utf8 (gotten build errors in the past when the stdlib files have unicode in them and people try builing them in a C locale or something).
Do we want to "if windows" check before hSetEncoding stdout utf8? Or should we just set it that way on all platforms?

Heather:
Error msg:DeleteFile "test.exe": does not exist (╨Э╨╡ ╤Г
╨┤╨░╨╡╤В╤Б╤П ╨╜╨░╨╣╤В╨╕ ╤Г╨║╨░╨╖╨░╨╜╨╜╤Л╨╣ ╤Д╨░╨╣╨╗.)

I'm so happy with UTF8 :dancer:

PR #1938, opened 2015-02-22
Better error messages for parsing operators

ahmadsalim:
The idea is to make better errors for parsing expressions with operators without fixity, so instead of

```idris
Idris> 2 *** 2
(input):1:3: error: expected: "$",
"$>", "&&", "+", "++", "-",
"->", ".", "/", "/=", "::", "<",
"<#>", "<$", "<$>", "<*>",
"<+>", "<->", "<<", "<=", "<|>",
"=", "==", ">", ">=", ">>",
">>=", "\\\\", "`", "||", "~=~",
ambiguous use of a left-associative operator,
ambiguous use of a non-associative operator,
ambiguous use of a right-associative operator,
end of input, function argument,
matching application expression
2 *** 2<EOF> 
```

It will write:

```idris
Idris> 2 *** 2
(input):1:7: error: unexpected
    Operator without known fixity:
    ***, expected: space
2 *** 2<EOF>
```

### DIFF
--- a/src/Idris/ParseHelpers.hs
+++ b/src/Idris/ParseHelpers.hs
@@ -326,7 +326,7 @@ operatorLetter = oneOf opChars
 -- | Parses an operator
 operator :: MonadicParsing m => m String
 operator = do op <- token . some $ operatorLetter
-              when (op `elem` [":", "=>", "->", "<-", "=", "?=", "|"]) $
+              when (op `elem` [":", "=>", "->", "<-", "=", "?=", "|", "**", "==>", "\\", ",", "--", "|||"]) $
                    fail $ op ++ " is not a valid operator"
               return op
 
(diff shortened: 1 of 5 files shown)


ahmadsalim:
However, I have an issue now with external syntax so it states e.g.

```idris
(input):1:7: error: unexpected
    Operator without known fixity:
    .., expected: space
```
I am unsure how to fix this issue, so I thought I will create the PR if anyone else has a good idea.

ahmadsalim:
OK, I have fixed the issues by inferring `simpleExpr` when an expression in a syntax definition is followed by operator-like characters. This can probably be relaxed to `expr'` but since it wasn't there I was uncertain whether I should have added support for it (which may confuse a bit more).

edwinb:
There's a failing test here, but is it just due to having a proper error message now?

ahmadsalim:
@edwinb Yeah, essentially. I will fix it an resubmit.

PR #1941, opened 2015-02-24
Make the Makefile cabal-sandbox aware.

jfdm:
The Makefile has been augmented to construct a relative location to a
pre-installed Idris executable based upon whether a sandbox was
involved. This fix resolves problems when using `make relib` and
running testing commands that rely upon a preinstalled idris binary.

The Makefile checks for the existence of the directory: `.cabal-sandox`.
If the directory exists then
    `IDRIS := .cabal-sandbox/bin/idris`
else
    `IDRIS := dist/build/idris`

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -38,3 +38,10 @@ else
 	SHLIB_SUFFIX    :=.so
 endif
 endif
+
+SBOX=.cabal-sandbox
+ifneq "$(wildcard $(SBOX))" ""
+	IDRIS :=.cabal-sandbox/bin/idris
+else
+	IDRIS :=dist/build/idris/idris
+endif
--- a/Makefile
+++ b/Makefile
@@ -20,16 +20,16 @@ build: dist/setup-config
 test: doc test_c
 
 test_c:
-	$(MAKE) -C test IDRIS=../dist/build/idris
+	$(MAKE) -C test IDRIS=../$(IDRIS)
 
 test_java:
-	$(MAKE) -C test IDRIS=../dist/build/idris test_java
+	$(MAKE) -C test IDRIS=../${IDRIS} test_java
 
 test_llvm:
-	$(MAKE) -C test IDRIS=../dist/build/idris test_llvm
+	$(MAKE) -C test IDRIS=../${IDRIS} test_llvm
 
 test_js:
-	$(MAKE) -C test IDRIS=../dist/build/idris test_js
+	$(MAKE) -C test IDRIS=../${IDRIS} test_js
 
 test_all:
 	$(MAKE) test
@@ -37,7 +37,7 @@ test_all:
 	$(MAKE) test_java
 
 lib_clean:
-	$(MAKE) -C libs IDRIS=../../dist/build/idris/idris RTS=../../dist/build/rts/libidris_rts clean
+	$(MAKE) -C libs IDRIS=../../${IDRIS} RTS=../../dist/build/rts/libidris_rts clean
 
 relib: lib_clean
 	$(CABAL) install $(CABALFLAGS)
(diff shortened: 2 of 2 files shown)


Melvar:
Was the “fixup!” commit meant to be part of the previous?

jfdm:
Ah I forgot to autorebase...

jfdm:
there rebase successful.

david-christiansen:
There's a permissions error on `make lib_doc` - is it possible that an execute bit isn't getting set?

david-christiansen:
Oh, and thanks for doing this!

jfdm:
argh, the path for building outwith a sandbox is incorrect. I have a fix and am running more exhaustive tests before I push it.

Melvar:
Huh, actually looking at this, does installing into a sandbox not build into `dist/build/` first? Since the idea with that is using the path of the `idris` built from this repo rather than any installed one.

jfdm:
@Melvar You are partially correct. Cabal sandboxes do build in `dist` first but not under `dist/build`, but rather it builds under a more unique directory name of the form `dist/dist-sandbox-<hash>`. The binaries are then installed into the cabal sandbox. 

When people build using cabal sandboxes the sandbox directory `.cabal-sandbox` is created first and appears to be a more reliable test for whether we are using a sandbox or not, hence why I chose it.

This PR really affects the Makefile in being able to call targets that use idris from the local build process. Cabal handles that build process. 

Also I fixed the error yesterday, and all the tests seem to be passing as before.

config.mk
```diff
@@ -38,3 +38,10 @@ else
 	SHLIB_SUFFIX    :=.so
 endif
 endif
+
+SBOX=.cabal-sandbox
+ifneq "$(wildcard $(SBOX))" ""
+	IDRIS :=.cabal-sandbox/bin/idris
```
david-christiansen: Does this cause the libraries to be built with the just-built Idris, or with the one that was last installed in the sandbox?

jfdm: The `$(IDRIS)` variable is only used to define the Idris version that will be used to: a) run the tests; b) clean the libraries; and (c) build/clean the documentation. This version will be the one that was last installed in the sandbox.

The libraries are still built as normal using the CABAL machinery.

jfdm:
Sorry to be that guy, but any chance this can get merged. I think it is an innocuous PR.

david-christiansen:
I haven't done it because I don't yet understand it. I'll try to dig in a bit more after lunch. Thanks for the bump!

david-christiansen:
What I'm confused about still is whether it's running the last installed version of Idris from the sandbox, or the one that was just built.  In other words, if I say `cabal build` then `make test`, won't this test mean that it uses the last-installed Idris instead of the one I just built to run the tests?

jfdm:
Ah okay, I see what you are getting at now. I think if we can isolate where the latest build version of Idris is we can hopefully amend the value to that. Hopefully it will be the value I have specified---fingers crossed---and if not we can easily amend it.

david-christiansen:
What about a hack like `find dist/ -name idris` or something?

jfdm:
That could work, but I prefer I 'native' solution. Having just cloned a fresh copy of Idris and built it using `cabal build` I get an executable `idris` placed in `dist/build/Idris/`. Prior to a sandbox install. If this is the same value on others clean install for both sandboxed and none sandboxed builds. I suggest we just make the `IDRIS` variable be: `list/build/Idris/idris` with the correct relative locations for passing onto sub directory makefiles

jfdm:
I am going to close this PR as it turns out that I believe more of the build process is incorrect wrt to testing under a sandbox i.e. it used the most recently installed rather most recently built. See PR #2108 for instance.

PR #1945, opened 2015-02-24
Revert "Set encoding to UTF8 on Windows and fix :exec"

Heather:
Discussion could be here.

Because I don't want to install some hacky utf8 terminal now I don't like to see messages like that: `╨Э╨╡ ╤Г
╨┤╨░╨╡╤В╤Б╤П ╨╜╨░╨╣╤В╨╕ ╤Г╨║╨░╨╖╨░╨╜╨╜╤Л╨╣ ╤Д╨░╨╣╨╗.`

It's `Не удается найти указанный файл.` encoded in utf8 and that's how default windows user will see it (maybe English will be fine but not my language) so I strongly suggest to revert it but it's my personal opinion.

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -29,7 +29,6 @@ import Idris.CmdOptions
 import IRTS.System ( getLibFlags, getIdrisLibDir, getIncFlags )
 
 import Util.DynamicLinker
-import Util.System
 
 import Pkg.Package
 
@@ -39,9 +38,7 @@ import Paths_idris
 -- on with the REPL.
 
 main :: IO ()
-main = do
-          when isWindows $ do hSetEncoding stdout utf8
-          opts <- runArgParser
+main = do opts <- runArgParser
           runMain (runIdris opts)
 
 runIdris :: [Opt] -> Idris ()
--- a/src/Idris/REPL.hs
+++ b/src/Idris/REPL.hs
@@ -1104,12 +1104,10 @@ process fn Execute
                            (tmpn, tmph) <- runIO tempfile
                            runIO $ hClose tmph
                            t <- codegen
-                           -- gcc adds .exe when it builds windows programs
-                           progName <- return $ if isWindows then tmpn ++ ".exe" else tmpn
                            ir <- compile t tmpn m
                            runIO $ generate t (fst (head (idris_imported ist))) ir
                            case idris_outputmode ist of
-                             RawOutput h -> do runIO $ rawSystem progName []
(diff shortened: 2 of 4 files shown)


melted:
Then everything that isn't latin1 will get an error message. Can you try substituting utf16 instead of utf8 in the line?

Heather:
@melted it's idris compiler generated message WARNING: Cannot remove file Error msg:DeleteFile, I did nothing with it.

Also this revert is not fully correct because it also reverts "fix :exec", however "fix exec" commit also contains changes related to first commit so I can't revert just one commit, maybe it should be some kind of manual revert...

melted:
The broken thing I was trying to solve was that "idris --help" spews an error on windows. And not outputting unicode ever seems like a hard thing to avoid. Maybe it will work better with utf16, which is what windows uses when it is on a unicode codepage, can you just try to substitute utf8 with utf16?

Heather:
Also what is `latin1` chcp returns [866](http://en.wikipedia.org/wiki/Code_page_866)

melted:
@Heather just remove the added line to the main function, the rest is useful in any case.

melted:
The encodings that goes into hSetEncoding are here: http://hackage.haskell.org/package/base-4.7.0.2/docs/System-IO.html#v:latin1

Heather:
@melted with PR reverted I get no error on `idris --help`

melted:
Hmmm... weird, I get that, Can you make a new PR that just removes the hSetEncoding line from main?

Heather:
@melted with UTF16 it goes wide... :dancer: 

```shell
   T y p e   c h e c k i n g   . \ B u i l t i n s . i d r
 T y p e   c h e c k i n g   . \ P r e l u d e \ A l g e b r a . i d r
 T y p e   c h e c k i n g   . \ P r e l u d e \ U n i n h a b i t e d . i d r
 T y p e   c h e c k i n g   . \ P r e l u d e \ B o o l . i d r
 T y p e   c h e c k i n g   . \ P r e l u d e \ B a s i c s . i d r
 T y p e   c h e c k i n g   . \ P r e l u d e \ C l a s s e s . i d r
 T y p e   c h e c k i n g   . \ P r e l u d e \ F o l d a b l e . i d r
 T y p e   c h e c k i n g   . \ P r e l u d e \ C a s t . i d r
 T y p e   c h e c k i n g   . \ P r e l u d e \ M a y b e . i d r
```

PR #1946, opened 2015-02-24
disable windows specific UTF8 encoding

Heather:
discussion here: https://github.com/idris-lang/Idris-dev/pull/1945

Because I don't want to install some hacky utf8 terminal now I don't like to see messages like that: ╨Э╨╡ ╤Г ╨┤╨░╨╡╤В╤Б╤П ╨╜╨░╨╣╤В╨╕ ╤Г╨║╨░╨╖╨░╨╜╨╜╤Л╨╣ ╤Д╨░╨╣╨╗.

It's Не удается найти указанный файл. encoded in utf8 and that's how default windows user will see it (maybe English will be fine but not my language).

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -29,7 +29,6 @@ import Idris.CmdOptions
 import IRTS.System ( getLibFlags, getIdrisLibDir, getIncFlags )
 
 import Util.DynamicLinker
-import Util.System
 
 import Pkg.Package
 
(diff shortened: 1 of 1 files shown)


melted:
Apparently this is more complicated than I expected, so I think removing it is the right thing to do. I will try to come up with a better solution.

david-christiansen:
If both @Heather and @melted think this is a good idea, then I'll go for it. We really need to get suitable Appveyor testing set up!

jfdm:
@david-christiansen As you keep on dropping hints at the need for a Windows CI setup, I have added it as an issue on the issue tracker---Issue #1948.

david-christiansen:
@jfdm Thanks!

PR #1947, opened 2015-02-24
Replace a fancy apostrophe with a basic ASCII one

melted:
A Unicode apostrophe will cause problems with locales with
non-Unicode text encodings, like CP850 that is default in
Windows terminals.


david-christiansen:
Best branch name ever :-)

PR #1950, opened 2015-02-25
Correct format strings for idris_gcInfo.

Chobbes:
Just a small change to make sure we're using size_t's where we should be and matching the other integer sizes in our format strings!

See pages 277 and 199 of the C99 standard if you don't trust me :).


david-christiansen:

PR #1951, opened 2015-02-26
SETHEAP macro fix.

Chobbes:
SETHEAP, which is never used, had a copy/paste bug. SETTY uses the argument 't', and SETHEAP copied this, even though it's called 'y' for SETHEAP.


david-christiansen:

edwinb:
This will bite when I finish the thing that was intended to use it (which may be never, who knows...). Not entirely sure why this macro is in the repo in the first place, but since it's there better it's right than wrong :). Thanks.

PR #1952, opened 2015-02-26
Changed closure type to a uint8_t.

Chobbes:
Drop memory usage down by changing the closure type to a uint8_t. This was using a uint32_t before, so there's potential to shave 3 bytes off of every value in an Idris program (depending on alignment and whatnot).

This could be a significant savings on memory constrained devices (such as AVR chips).

### DIFF
--- a/rts/idris_rts.h
+++ b/rts/idris_rts.h
@@ -57,12 +57,12 @@ typedef struct {
 } ManagedPtr;
 
 typedef struct Closure {
-// Use top 16 bits of ty for saying which heap value is in
-// Bottom 16 bits for closure type
+// Use top bit of ty for saying which heap value is in
+// Bottom 7 bits for closure type
 //
 // NOTE: ty can not have type ClosureType because ty must be a
-// uint32_t but enum is platform dependent
-    uint32_t ty;
+// uint8_t but enum is platform dependent
+    uint8_t ty;
     union {
         con c;
         int i;
@@ -155,14 +155,14 @@ typedef void(*func)(VM*, VAL*);
 #define CTAG(x) (((x)->info.c.tag_arity) >> 8)
 #define CARITY(x) ((x)->info.c.tag_arity & 0x000000ff)
 
-// Use top 16 bits for saying which heap value is in
-// Bottom 16 bits for closure type
+// Use top bit for saying which heap value is in
+// Bottom 7 bits for closure type
 
-#define GETTY(x) ((x)->ty & 0x0000ffff)
-#define SETTY(x,t) (x)->ty = (((x)->ty & 0xffff0000) | (t))
+#define GETTY(x) ((x)->ty & 0x1f)
+#define SETTY(x,t) (x)->ty = (((x)->ty & 0xe0) | (t))
 
-#define GETHEAP(x) ((x)->ty >> 16)
-#define SETHEAP(x,y) (x)->ty = (((x)->ty & 0x0000ffff) | ((y) << 16))
+#define GETHEAP(x) ((x)->ty >> 7)
+#define SETHEAP(x,y) (x)->ty = (((x)->ty & 0x7F) | ((y) << 7))
 
 // Integers, floats and operators
 


Chobbes:
In case anybody is wondering, this enum represents the values that the closure type can be

https://github.com/idris-lang/Idris-dev/blob/master/rts/idris_rts.h#L26

This could fit in 5 bits, so the 7 given in this pull request should be plenty :).

Melvar:
… Isn’t this going to fail horribly as soon as you have more heaps? (IIRC there’s one heap per thread.)

markuspf:
The C compiler will pad the structure, so I don't see what problem is solved here? Sure you could make the struct packed, but still I don't see how this should be solving any problems. 

If you want to run on an "embedded device" and you really actually assume you want to do that, *and* you assume that it's not got 32bit register size and address space (and the air is getting really thin here), then you want to make a different rts anyway (and this is being done...)

Chobbes:
@Melvar I don't think the heap value is ever actually used. It doesn't seem like it should be necessary to me with the Cheney garbage collector. As far as I can tell there's only one active heap, even if you have a bunch of threads. I'm happy to be wrong - I just don't see it right now! :)

@markuspf It will pad the structure by default, and in that case it won't matter (and this will be preferred in most cases). Changing it to a uint8_t would give the compiler the option to pad the structure more tightly depending on the architecture, and if I decided this mattered to me I would have the option to tell the compiler to pack my structs. I don't see explicitly telling the compiler that we need more space than we actually do as being a win. Hinting that we can use less seems preferable, even if in the common case it won't make a difference. I could be missing something, though :). I'll gladly be proven horribly wrong!

In either case can somebody explain the advantage of using a uint32_t like this? If we need 16 bits for the type of the closure, and 16 bits for the size of the heap, then perhaps it would be preferable to have a separate uint16_t for each? What is being gained from the bit hacking here?

Thanks!

Chobbes:
@markuspf is the work on the new RTS you have mentioned public? I would love to take a gander if you have a link. I have not been able to find anything :(

david-christiansen:

melted:
The natural thing to avoid the padding (which is standard-decreed, a compiler can't optimize a struct without a pragma or an attribute no matter how much it wants to) would be to move the small value to the last spot. Seeing that the union makes room for a 128-bit value that will be wasted for nearly all uses of this struct, it doesn't seem space efficiency is a high priority. So I don't think this is the best optimization that can be made. For general comprehensibility the biggest win would probably be to avoid the bitops, I've a hard time seeing the need for them here. I'd vote for splitting the uint32 into two uint16. Going to a byte and retaining the bitops  just looks like so much premature optimization.

Chobbes:
@melted I think the union is only 64-bits because this is a pointer:

__m128i* bits128p;

This is perhaps beside the point, but it does make the potential savings more significant. For the most part it just makes no sense to me to have a uint32_t for this value. It won't matter because of padding most of the time, but when I bumped into it I was very confused about why this had to have so much space reserved for it. It's not a high priority change, but if it has the potential to be slightly better and doesn't make a difference otherwise... why not?

I completely agree with going for comprehensibility and avoiding the bit operations all together (I just left them in because that's how it was before). I would be happy to go through and make yet another pull request for this change if we can figure out what we actually want :). I'm not entirely convinced that we should have two uint16_t's, though.

I have been completely unable to find where the heap value is actually used. It's not used in the RTS, nor the code generation as far as I can tell. It looks like we can just toss it.

melted:
Right, it's a pointer. 

But the union must be 64bit aligned so there will be 4 bytes of padding in the original and 7 bytes in your version. And both will be 16 bytes (you can verify that with sizeof). 

Actually moving the field to the back wouldn't work as it still needs the 64-bit alignment for the union in an array of the struct so it would pad it anyway.

I don't think packing is an option as bad alignment can be a drag on performance.



Från: "Calvin Beck" <notifications@github.com>
Skickat: ‎2015-‎03-‎01 23:19
Ämne: Re: [Idris-dev] Changed closure type to a uint8_t. (#1952)

@melted I think the union is only 64-bits because this is a pointer:
__m128i* bits128p;
This is perhaps beside the point, but it does make the potential savings more significant. For the most part it just makes no sense to me to have a uint32_t for this value. It won't matter because of padding most of the time, but when I bumped into it I was very confused about why this had to have so much space reserved for it. It's not a high priority change, but if it has the potential to be slightly better and doesn't make a difference otherwise... why not?
I completely agree with going for comprehensibility and avoiding the bit operations all together (I just left them in because that's how it was before). I would be happy to go through and make yet another pull request for this change if we can figure out what we actually want :). I'm not entirely convinced that we should have two uint16_t's, though.
I have been completely unable to find where the heap value is actually used. It's not used in the RTS, nor the code generation as far as I can tell. It looks like we can just toss it.

melted:
Also, I tried sizeof on both structs and got 16.

Från: "Calvin Beck" <notifications@github.com>
Skickat: ‎2015-‎03-‎01 23:19
Ämne: Re: [Idris-dev] Changed closure type to a uint8_t. (#1952)

@melted I think the union is only 64-bits because this is a pointer:
__m128i* bits128p;
This is perhaps beside the point, but it does make the potential savings more significant. For the most part it just makes no sense to me to have a uint32_t for this value. It won't matter because of padding most of the time, but when I bumped into it I was very confused about why this had to have so much space reserved for it. It's not a high priority change, but if it has the potential to be slightly better and doesn't make a difference otherwise... why not?
I completely agree with going for comprehensibility and avoiding the bit operations all together (I just left them in because that's how it was before). I would be happy to go through and make yet another pull request for this change if we can figure out what we actually want :). I'm not entirely convinced that we should have two uint16_t's, though.
I have been completely unable to find where the heap value is actually used. It's not used in the RTS, nor the code generation as far as I can tell. It looks like we can just toss it.

Chobbes:
@melted On my Arduino mega I get 12 bytes for the original structure, and 9 bytes for the updated structure. Structure padding is platform dependent, and while on common platforms this makes no difference, I don't see why we should intentionally make it worse on other (admittedly less suitable) platforms.

We have some existing code for using the runtime with Arduinos:

https://github.com/idris-lang/Idris-dev/tree/master/rts/arduino

Even if it isn't a major focus, I really don't see why we shouldn't change the data type. This change seems like a no brainer to me.

Plus, I think we can all agree that types have meanings, and we should use the more appropriate type for clarity, even if it doesn't make a difference performance wise.

edwinb:
Does this actually save anything in practice on current Idris programs? (i.e. have you measured it?)

I don't think this RTS is the right one to use for systems with small heaps in the end - it is pretty easy to retarget Idris using a different back end and I would prefer to go that way rather than tinker with the main RTS.

(There's a merge conflict, in any case, so I can't merge quite yet.)

Chobbes:
@edwinb Should merge now (I changed SETHEAP to match here, but there was a fix in another patch too).

I'll run a test in a couple days. I don't have my Arduino with me, and I have some other work to focus on right now.

melted:
It saves nothing. It would be better to fork the rts for 8-bit systems, then it could shave off far more stuff. 

Having the same codebase for 8-bit and 64-bit will just break the former all the time.

Didn't the last person who built it on Arduino fork the rts?

Från: "Edwin Brady" <notifications@github.com>
Skickat: ‎2015-‎03-‎05 23:19
Ämne: Re: [Idris-dev] Changed closure type to a uint8_t. (#1952)

Does this actually save anything in practice on current Idris programs? (i.e. have you measured it?)
I don't think this RTS is the right one to use for systems with small heaps in the end - it is pretty easy to retarget Idris using a different back end and I would prefer to go that way rather than tinker with the main RTS.
(There's a merge conflict, in any case, so I can't merge quite yet.)

david-christiansen:
It seems that the consensus is that this PR is not something we want to incorporate at this time.  Thanks for sending it though!

markuspf:
I think what one should do (and what I have done) is produce a new code generator with it's own RTS. As soon as I am back in St Andrews, I'll upload this to github.

Chobbes:
@markuspf fair enough! Let me know when it's up. I'd love to take a gander at it :).

PR #1953, opened 2015-02-26
uint32_t type needed even if stats are not enabled.

Chobbes:
Includes were guarded by IDRIS_ENABLE_STATS, but we need the uint32_t type regardless.


david-christiansen:

PR #1955, opened 2015-02-26
Added a citation file.

jfdm:
Given the large number of academic users, I have pulled the citation from the Unofficial FAQs and placed it in a file in the project root. Save some people for searching for the canonical reference for `Idris`.


david-christiansen:
Seems reasonable. Should we add other Idris papers as well?

jfdm:
For papers that talk about idris internals I think so.

On a similar note maybe we should start to think of updating the list of papers on http://www.idris-lang.org/documentation/ if there are other papers that use idris.

PR #1962, opened 2015-02-27
Simplify comment parser

### LINKED ISSUE #1958: Empty comments extend to the next line
This issue was first reported at https://github.com/idris-hackers/idris-mode/issues/300 by @jasonhemann.

Idris sucessfully parses and type checks the following module:

```
module Foo
-- 
malformed stuff here
Alias : Type
Alias = Nat
```

It seems that the comment handling code is a bit off.

Pinging @ahmadsalim in case he has time to diagnose it before me.

david-christiansen:
Since we no longer have special documentation comments, the comment
parser can be much simpler. The simpler version also works better - it
fixes #1958.

### DIFF
--- a/src/Idris/ParseHelpers.hs
+++ b/src/Idris/ParseHelpers.hs
@@ -105,30 +105,28 @@ isEol :: Char -> Bool
 isEol '\n' = True
 isEol  _   = False
 
+-- | A parser that succeeds at the end of the line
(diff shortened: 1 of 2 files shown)


david-christiansen:
This is also related to the other half of https://github.com/idris-hackers/idris-mode/issues/300.

PR #1967, opened 2015-02-28
GC check allocation of new heap.

Chobbes:
When allocating the new heap for the Cheney garbage collector, check the status of malloc. Could potentially try doing something smarter, but failing gracefully would be nice.

### DIFF
--- a/rts/idris_gc.c
+++ b/rts/idris_gc.c
@@ -119,6 +119,14 @@ void idris_gc(VM* vm) {
     char* newheap = malloc(vm->heap.size);
     char* oldheap = vm->heap.heap;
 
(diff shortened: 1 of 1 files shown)


Chobbes:
So, I have realized that this part of the garbage collector is essentially a copy of what alloc_heap does. I'm working on a small generalization that will allow alloc_heap to be used for this here:

https://github.com/Chobbes/Idris-dev/tree/heap_fixes

Right now I'm just running the tests. I might have to make a few fixes, but it should be good to go.

The advantage of this is that if the heap structure ever changes we won't have this functionality scattered throughout the code and have to adjust it in multiple places.

Cheers! :)

PR #1968, opened 2015-02-28
Generalized alloc_heap so it could be reused for garbage collection.

Chobbes:
The garbage collector had duplicated code / functionality from alloc_heap. This pull request generalizes alloc_heap in order to keep this functionality in one place, so if the heap structure ever changes we won't have this functionality scattered throughout the code and have to adjust it in multiple places.

A previous pull request addresses the return value of malloc not being checked in the garbage collector:

https://github.com/idris-lang/Idris-dev/pull/1967

But this is actually covered by alloc_heap in these changes.

### DIFF
--- a/rts/idris_gc.c
+++ b/rts/idris_gc.c
@@ -116,19 +116,8 @@ void idris_gc(VM* vm) {
     if (vm->heap.old != NULL)
         free(vm->heap.old);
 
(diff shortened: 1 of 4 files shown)


edwinb:
Hmm, should I not have merged #1967? There seems to be a merge conflict here now...

Issue #1869, opened 2015-01-22
Keep original names of DSL-block variables


david-christiansen:
It would be nice if the binder constructors under DSL notation would receive the original variable name as a string in addition to the body. This would help in generating good runtime errors.

david-christiansen:
@edwinb Any objections to this change?

david-christiansen:
This was implemented in #1969.

PR #1969, opened 2015-03-01
DSL extension: Named lambda and let

chetant:
This adds two new entries to the dsl specification:
* lambda_named
* let_named
which override lambda and let, when present.

The use case is when we want to use the AST to translate to another language and keep the variable names similar to the ones used in the AST.

For example:

```haskell
data Test : Type where
  Var : Nat -> Test
  Lam : Test -> Test
  LamNm : String -> Test -> Test
  Let : Test -> Test -> Test
  LetNm : String -> Test -> Test -> Test
  Lit : Integer -> Test
  Add : Test -> Test -> Test

plus : Test -> Test -> Test
plus = Add

dsl test_named {
  let_named = LetNm
  variable = Var
  index_first = Z
  index_next = S
  lambda_named = LamNm
}

dsl test {
  let = Let
  variable = Var
  index_first = Z
  index_next = S
  lambda = Lam
}

t1 : Test
t1  = test       (let ztest = (Lit 12) in (let y = (Lit 24) in (ztest `plus` y )))

t1n : Test
t1n = test_named (let ztest = (Lit 12) in (let y = (Lit 24) in (ztest `plus` y )))

t2 : Test
t2  = test       (\x,y => (x `plus` y))

t2n : Test
t2n = test_named (\x,y => (x `plus` y))
```

Gives the following:

```haskell
*test> t1
Let (Lit 12) (Let (Lit 24) (Add (Var 1) (Var 0))) : Test
*test> t1n
LetNm "ztest"
      (Lit 12)
      (LetNm "y" (Lit 24) (Add (Var 1) (Var 0))) : Test
*test> t2
Lam (Lam (Add (Var 1) (Var 0))) : Test
*test> t2n
LamNm "x" (LamNm "y" (Add (Var 1) (Var 0))) : Test
```

### DIFF
--- a/src/Idris/DSL.hs
+++ b/src/Idris/DSL.hs
@@ -17,21 +17,43 @@ desugar :: SyntaxInfo -> IState -> PTerm -> PTerm
 desugar syn i t = let t' = expandDo (dsl_info syn) t in
                       t' -- addImpl i t'
 
+mkTTName :: FC -> Name -> PTerm
+mkTTName fc n =
+    let mkList fc []     = PRef fc (sNS (sUN "Nil") ["List", "Prelude"])
+        mkList fc (x:xs) = PApp fc (PRef fc (sNS (sUN "::") ["List", "Prelude"]))
+                                   [ pexp (stringC x)
+                                   , pexp (mkList fc xs)]
+        stringC = PConstant . Str . str
+        intC = PConstant . I
+        reflm n = sNS (sUN n) ["Reflection", "Language"]
+    in case n of
+         UN nm     -> PApp fc (PRef fc (reflm "UN")) [ pexp (stringC nm)]
+         NS nm ns  -> PApp fc (PRef fc (reflm "NS")) [ pexp (mkTTName fc nm)
+                                                     , pexp (mkList fc ns)]
+         MN i nm   -> PApp fc (PRef fc (reflm "MN")) [ pexp (intC i)
+                                                     , pexp (stringC nm)]
+         otherwise -> PRef fc $ reflm "NErased"
+
 expandDo :: DSL -> PTerm -> PTerm
 expandDo dsl (PLam fc n ty tm)
     | Just lam <- dsl_lambda dsl
-        = let sc = PApp fc lam [pexp (var dsl n tm 0)] in
-              expandDo dsl sc
+        = let sc = PApp fc lam [ pexp (mkTTName fc n)
+                               , pexp (var dsl n tm 0)]
+          in expandDo dsl sc
 expandDo dsl (PLam fc n ty tm) = PLam fc n (expandDo dsl ty) (expandDo dsl tm)
 expandDo dsl (PLet fc n ty v tm)
     | Just letb <- dsl_let dsl
-        = let sc = PApp (fileFC "(dsl)") letb [pexp v, pexp (var dsl n tm 0)] in
-              expandDo dsl sc
+        = let sc = PApp (fileFC "(dsl)") letb [ pexp (mkTTName fc n)
+                                              , pexp v
+                                              , pexp (var dsl n tm 0)]
+          in expandDo dsl sc
 expandDo dsl (PLet fc n ty v tm) = PLet fc n (expandDo dsl ty) (expandDo dsl v) (expandDo dsl tm)
 expandDo dsl (PPi p n ty tm)
     | Just pi <- dsl_pi dsl
-        = let sc = PApp (fileFC "(dsl)") pi [pexp ty, pexp (var dsl n tm 0)] in
-              expandDo dsl sc
+        = let sc = PApp (fileFC "(dsl)") pi [ pexp (mkTTName (fileFC "(dsl)") n)
+                                            , pexp ty
+                                            , pexp (var dsl n tm 0)]
+          in expandDo dsl sc
 expandDo dsl (PPi p n ty tm) = PPi p n (expandDo dsl ty) (expandDo dsl tm)
 expandDo dsl (PApp fc t args) = PApp fc (expandDo dsl t)
                                         (map (fmap (expandDo dsl)) args)
--- a/test/dsl001/test001.idr
+++ b/test/dsl001/test001.idr
@@ -35,8 +35,11 @@ using (G : Vect n Ty)
       If  : Expr G TyBool -> Expr G a -> Expr G a -> Expr G a
       Bind : Expr G a -> (interpTy a -> Expr G b) -> Expr G b
 
+  lam_ : TTName -> Expr (a :: G) t -> Expr G (TyFun a t)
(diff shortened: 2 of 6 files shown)


david-christiansen:
Great! Someone else did #1869 so I didn't have to! Thanks!

I do have a couple of concerns about the particulars of this design/implementation:

1. Why do we need to have separate named and unnamed rules, instead of making all binders provide variable names?  It seems that `lambda_named` is strictly a generalization of `lambda`, and that the unnamed version can be supported by eta-expanding a constructor and adding a dummy parameter to throw away the argument. I don't think that backwards compatibility here is a huge concern - I'd rather have a clean design than be backwards compatible with all 12 or so users of DSL notation :-)

2. Is there a reason that only `lambda` and `let` can preserve their names? What about `pi`?

3. What do we do for binders that don't have a name that is easily available? For instance, I might have a `\x, _ => x`, where the `_` should become some internal name behind the scenes. This is even more of an issue for `pi`, as it's quite common not to explicitly name the variable bound in a function type. It seems that the easy way out here is to pack the original source names in a `Maybe` instead, providing them if they are available. The code in the patch calls all such variables `x_unknown`, which might conflict with the name of a variable chosen by the user.

Otherwise, the code seems clean, and this is a nice extension. Thanks again!

src/Idris/DSL.hs
```diff
@@ -17,13 +17,30 @@ desugar :: SyntaxInfo -> IState -> PTerm -> PTerm
 desugar syn i t = let t' = expandDo (dsl_info syn) t in
                       t' -- addImpl i t'
 
+getNameStr :: Name -> PTerm
+getNameStr (UN nm)   = PConstant $ Str $ str nm
+getNameStr (NS nm _) = getNameStr nm
+getNameStr (MN _ nm) = PConstant $ Str $ str nm
+getNameStr _         = PConstant $ Str "x_unknown"
```
david-christiansen: It seems to me that this would be much cleaner as a `Maybe String`, rather than picking a special sentinel value here.

src/Idris/DSL.hs
```diff
@@ -17,13 +17,30 @@ desugar :: SyntaxInfo -> IState -> PTerm -> PTerm
 desugar syn i t = let t' = expandDo (dsl_info syn) t in
                       t' -- addImpl i t'
 
+getNameStr :: Name -> PTerm
+getNameStr (UN nm)   = PConstant $ Str $ str nm
+getNameStr (NS nm _) = getNameStr nm
```
david-christiansen: If a user explicitly wrote a namespace, shouldn't we keep it?

src/Idris/DSL.hs
```diff
@@ -17,13 +17,30 @@ desugar :: SyntaxInfo -> IState -> PTerm -> PTerm
...
+getNameStr (MN _ nm) = PConstant $ Str $ str nm
```
david-christiansen: `MN` denotes a name chosen by the machine rather than by the user. It might be a name for a `_` pattern or for the binder in a non-dependent function type. It seems to be like a bad idea to lift these.

david-christiansen:
One more question: is there a reason not to use `Language.Reflection.TTName` instead of `String` for the names of binders? I see this as an interesting design question rather than me arguing for a particular choice.

chetant:
I agree with all your concerns/suggestions.
Also thanks for point me to TTName, I agree that its a better fit for this than String (esp considering #1869)
Updated pull request.

test/dsl001/test001.idr
```diff
@@ -35,8 +35,11 @@ using (G : Vect n Ty)
       If  : Expr G TyBool -> Expr G a -> Expr G a -> Expr G a
       Bind : Expr G a -> (interpTy a -> Expr G b) -> Expr G b
 
+  lam_ : TTName -> Expr (a :: G) t -> Expr G (TyFun a t)
+  lam_ _ = Lam
+
   dsl expr
-      lambda = Lam
```
chetant: One thing I noticed, if I replace this with:
```
    lambda = (\_ => Lam)
```
or other lambda variants, idris ate up all available memory on my system while typechecking.
Not sure if this is known issue.

jfdm:
Now here is the big question:

    As one of the approx. twelve users of DSL notation, how will this change affect my work?

david-christiansen:
@jfdm: you'll need to write some smart constructors to ignore the names, most likely. Or add them to your AST.

david-christiansen:
@chetant: Would you mind adding a note to the CHANGELOG file as well? Something like:
```
* DSL notation now provides binding forms with a representation of the name that occurred in the source code, to assist in better error reporting for DSLs. This means that binders in DSL expansions take an extra argument. This change is backwards-incompatible.
```

jfdm:
Is it possible that the tutorial/documentation be updated for this feature update? I am still not sure how precisely these changes will affect my work or what I will need to change. Sorry for being 'stupid' about it but I am still learning how to use DSL notation, and it is still not heavily intuitive for me still.

david-christiansen:
Yes, documentation will need to be updated for the next release.

I don't consider this stupid at all - there's a reason I reached out for comments before just pressing the "merge" button. Here's an example of how it would affect your work. If you previously had:
```
data Term : Nat -> Type where
  Var : Fin n -> Term n
  Lam : Term (S n) -> Term n

dsl terminator
  index_first = FZ
  index_next = FS
  variable = Var
  lambda = Lam
```

you could fix it in two ways: by ignoring the name provided at the binding site, or by saving it. To ignore it, do the following:
```
data Term : Nat -> Type where
  Var : Fin n -> Term n
  Lam : Term (S n) -> Term n

mkLam : TTName -> Term (S n) -> Term n
mkLam _ tm = Lam tm

dsl terminator
  index_first = FZ
  index_next = FS
  variable = Var
  lambda = mkLam
```

To save it  (perhaps for use in error messages or code generation), do this:
```
data Term : Nat -> Type where
  Var : Fin n -> Term n
  Lam : TTName -> Term (S n) -> Term n

dsl terminator
  index_first = FZ
  index_next = FS
  variable = Var
  lambda = Lam
```

jfdm:
Ah.......gotcha now. Cheers.

david-christiansen:
@jfdm So you're on board with this?

jfdm:
I would have been onboard with it regardless, just wanted to know what I would need to do to adjust my work.

Merge away.

david-christiansen:
All right. I'll do it after #1970 is merged, because I expect a couple minor conflicts, and if @chetant is sleeping I'll write the CHANGELOG entry.

david-christiansen:
I just created #1971 with the CHANGELOG update. Once Travis is done, I'll merge.

chetant:
Thanks for fixing the change log on my behalf!

david-christiansen:
No prob - I figured I'd bugged you enough :-)

PR #1970, opened 2015-03-02
The Great Operator Rename

david-christiansen:
This PR supersedes #1931. It contains the commits that I believe there was a consensus for, plus a bit of documentation.

The main change that it implements is that `Applicative.(<$>)` is now `Applicative.(<*>)`  and `(<$>)` is now an alias for `Functor.map`. Correspondingly, `($>)` and `(<$)` have been renamed to `(<*)` and `(*>)`. The cascading effects of this rename are that `Algebra.(<*>)` has been renamed to `Algebra.(<.>)` and `Matrix.(<.>)` is now `Matrix.(<:>)`.


david-christiansen:
I forgot to update the tests again after  ff7214e - argh!

PR #1972, opened 2015-03-02
Update error message in reg050 from PR #1938

david-christiansen:
To get this ready for merging, a conflict had to be resolved.

### DIFF
--- a/src/Idris/ParseHelpers.hs
+++ b/src/Idris/ParseHelpers.hs
@@ -119,7 +119,7 @@ singleLineComment :: MonadicParsing m => m ()
 singleLineComment = (string "--" *>
                      many (satisfy (not . isEol)) *>
                      eol *> pure ())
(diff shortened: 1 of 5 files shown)


Melvar:
Should the prefix-like sigils (`!`, `?`, `@`, `~`) not be in that list?

david-christiansen:
Possibly!

david-christiansen:
@ahmadsalim I merged this to prevent nasty conflicts from coming with the next merge from the new tactics branch, but what do you think about @Melvar's question?

ahmadsalim:
@Melvar @david-christiansen I think it is reasonable, to want to have these as invalid operators since the way we parse tokens now (if it was like Agda then it would have been different).

However, this also brings to light how to better handle invalid operators and such, but that can be handled in a different commit.
I will take a look and provide a new PR when I have time.

PR #1977, opened 2015-03-05
Added 'Print' functions to STDIO effect.

jfdm:
The STDIO effect was missing a `print` function, that calls `show` on the passed in argument.
This missing function was added with in-place print and print line varients.

### DIFF
--- a/libs/effects/Effect/StdIO.idr
+++ b/libs/effects/Effect/StdIO.idr
@@ -9,8 +9,8 @@ import Control.IOExcept
 
 ||| The internal representation of StdIO effects
 data StdIO : Effect where
-     PutStr : String -> { () } StdIO () 
-     GetStr : { () } StdIO String 
+     PutStr : String -> { () } StdIO ()
+     GetStr : { () } StdIO String
      PutCh : Char -> { () } StdIO ()
      GetCh : { () } StdIO Char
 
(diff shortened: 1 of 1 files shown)


david-christiansen:
Great!

Melvar:
This mismatches with `print` elsewhere, since that adds a newline.

jfdm:
mhm it does doesn't it. However, thinking about it, I think that `print` is a mismatch concerning the passing of strings to `STDOUT`. We already have:

+ `putStr` which puts a string to `STDOUT`.
+ `putStrLn` which puts a string with a trailing new line to `STDOUT`.

Why not change `print` in the prelude to have the same conventions, together with adding a `putCharLn` for completeness?

david-christiansen:
I agree that changing `print` to mean `putStr . show` and adding `printLn` to mean `putStrLn . show` would be a nice change.

david-christiansen:
As long as we're breaking all the code anyway :-)

LeifW:
:thumbsup:

jfdm:
i will do a PR and light all flares after I have done some chores.

Melvar:
I question if one ever wants `print` without a newline. Or at least if one wants it often enough to have a short name for that.

jfdm:
Well I think the addition of two extra characters for that functionality is really a moot point. I think by having the two functions we not only add some interesting functionality to the Idris prelude, but bring the 'put to `STDOUT`' set of functions in line semantically. speaking.

Melvar:
I’m not sure what you mean by “add some interesting functionality”. It’s a small shortcut to something I’ve never heard of a need for.

edwinb:
Just because you haven't heard of a need for it doesn't make it not useful by anyone else. This exists in IO and I've often wanted it but never quite enough to add it myself. Thanks @jfdm!

jfdm:
the interesting functionality, is the ability to put multiple showable characters on the same line without the need of `++` or `unwords`. For example compare:

```idris
doStuff : IO ()
doStuff = do
  print x
  putStr ", "
  printLn y
``` 

```idris
doStuff : IO ()
doStuff = do
  putStrLn $ show x ++ ", " ++ show y
  putStrLn $ unwords [show x, ",", show y]
```

PR #1981, opened 2015-03-05
Breaking Changes: Align semantics of 'print' functions.

jfdm:
Address semantic differences in putting things to STDOUT. The changes are as follows:

+ `print` is for putting showable things to STDOUT.
+ `printLn` is for putting showable things to STDOUT with a new line
+ `putCharLn` for putting a single character to STDOUT, with a new line.

Effects has been updated accordingly.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -20,6 +20,11 @@ New in 0.9.17:
   an extra argument of type TTName. This allows more understandable
   dynamic errors in DSL code and more readable code generation results.
 * DSL notation can now be applied with $
+* Effectful and IO function names for sending data to STDOUT have been
+  aligned, semantically.
+    + `print` is now for putting showable things to STDOUT.
+    + `printLn` is for putting showable things to STDOUT with a new line
+    + `putCharLn` for putting a single character to STDOUT, with a new line.
 
 New in 0.9.16:
 --------------
@@ -57,7 +62,7 @@ New in 0.9.15:
   solved automatically.
 * More efficient representation of proof state, leading to faster elaboration
   of large expressions.
-* EXPERIMENTAL Implementation of uniqueness types 
+* EXPERIMENTAL Implementation of uniqueness types
 * Unary negation now desugars to "negate", which is a method of the Neg type class.
   This allows instances of Num that can't be negative, like Nat, and it makes correct
   IEEE Float operations easier to encode. Additionally, unary negation is now available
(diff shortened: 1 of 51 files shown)


L8D:
I personally don't prefer this, it creates unnecessary differences with Haskell. But there's probably a good reason I'm missing to do this.

Also, be careful about committing whitespace changes.

jfdm:
Well I believe that Haskell has it wrong in the provision of convenience functions for showable things. We are not Haskell, and can thus provide improvements where necessary. I think this is one such case. The prelude is inconsistent in that it has already set precedence that functions that 'put' things to STDOUT must have 'Ln' as a postfix. The `print` function does not, and this PR brings that inline. Secondly, having a `print` that does not include a newline character facilitates the ability to put multiple showable characters on the same line without the need of `++` or `unwords`. For example compare:

```idris
doStuff : (Show a, Show b) -> a -> b -> IO ()
doStuff x y = do
  print x
  putStr ", "
  printLn y
``` 
with
```idris
doStuff : (Show a, Show b) -> a -> b -> IO ()
doStuff x y = do
  putStrLn $ show x ++ ", " ++ show y
  putStrLn $ unwords [show x, ",", show y]
```

I understand that printing can be a contentious issue, but I think this can be beneficial in the longer term.

david-christiansen:
This change seems to me to increase the consistency of the API while decreasing Haskell compatibility, though in a way that is not a big deal IMHO. Would you mind pinging the mailing list before we merge this? It could be that someone there has some insight that we're missing.

jfdm:
@david-christiansen yes I am planning to ping the mailing list, I just wanted to make sure that all the tests were passing.

david-christiansen:
I'm in favor of this. I say merge if nobody else comments soon - I think internal consistency trumps Haskell API compatibility.

It might be a good idea to make a `.dir-locals.el` that turns off the automatic whitespace squishing inside the Idris repo. Perhaps something like
```elisp
((nil . (eval (remove-hook 'before-save-hook 'delete-trailing-whitespace t))))
```
would do the trick.

L8D:
I'm not against the merge but I think we should be moving towards (if we're breaking Haskell compatibility in this case) generalizing print-like functions into some common logging interface so they become composable and reusable and what-not. Instead of `Show a => a -> IO ()` we might have `(Loggable a, Logger a m) => a -> m ()` or something along those lines. Perhaps separate the Effect.StdIO module into a logging module and an console input module.

L8D:
Also I think `putCharLn` should be re-formatted to only perform one write instead of two. Other than that, :+1:

david-christiansen:
What do you think `Loggable` and `Logger` should contain? I've long seen `Show` as a kind of a place to put some hacky string debug thing so you don't have to remember all the names, rather than a place to really do lots of thoughtful work.

In any case, it could be that a much bigger change is a good idea, and I'd encourage anyone to make such an improved library so we can try it out. I don't think it's potential existence should block for small improvements along the way though.

david-christiansen:
As far as two writes go, you're suggestion is that it be something like:

```idris
putCharLn : Char -> IO ()
putCharLn c = putStrLn (singleton c)
```
right?

L8D:
Yep, that works.

jfdm:
I agree that anything other than the changes I suggest should be left for another day. Not that I do not agree or disagree, but as @david-christiansen suggested it is a more longer term project. I shall make the update to `putCharLn`; I was nto aware of singleton.

Issue #1943, opened 2015-02-24
`|||` breaks instance declarations


nicolabotta:
The code
````
> ||| The singletons
> class Lala a where
>   lala : (x : a) -> (y : a) -> x = y

> -- ||| () is a singleton
> instance Lala () where
>   lala () () = Refl
```
type checks but uncommenting the line `> -- ||| () is a singleton` yields the error
````
- + Errors (1)
 `-- ./instance.lidr line 6 col 0:
      error: expected: "class",
         "|||", accessibility modifier
     instance Lala () where 
     ^
```

david-christiansen:
This is because individual instances can't have documentation strings. I considered it to not be a good idea, because there's not really a syntax to look up those docs, but on further reflection I think that this assumption was bad.

Perhaps a sensible interface might be something like:
```
Idris> :doc Uninhabited
Type class Uninhabited
    A canonical proof that some type is empty

Parameters:
    t

Methods:
    uninhabited : Uninhabited t => t -> Void
        If I have a t, I've had a contradiction
       
Instances:
    Uninhabited Void:
      Void is the canonical empty type, thus it is trivially uninhabited.
    Uninhabited (0 = S n)
      The constructors of Nat are disjoint.
```

where the instance docstrings would be shown in the documentation for the class itself. This would also be nice for named instances, come to think of it.

Thanks for a good idea!

Marking this as "low-hanging fruit" because it should be a matter of following the code for the existing docstring handling and adding in a bit more.

nicolabotta:
Thanks a lot David! Best

david-christiansen:
This was fixed by #1983. Thanks @adamse!

PR #1983, opened 2015-03-06
Instance docs

adamse:
Implementation of instance docs as proposed in #1943.

This is not quite finished, here are some items to do:

- [x] Update types & plumb types
- [x] Parse instance docs
    - [x] Refactor parser to remove added `try`s?
- [x] Add `Docstring` to compiler state
- [x] Tests
- [x] Show instance docs in `:doc`
- [x] <del>Show instance docs in html</del>
- [x] <del>Update documentation where applicable</del>
- [x] Add instance docs to standard libraries

Example
~~~
> echo ":doc Functor" | ./dist/build/Idris/idris ./libs/prelude/Prelude/Functor.idr
[...]
Instances:
    Functor id
      The id-entity functor
    Functor List
[...]
~~~
with
~~~idris
||| The id-entity functor
instance Functor id where
    map f a = f a
~~~
in `Prelude.Functor`.


david-christiansen:
This is really great! Thanks!

Are you proposing that we merge this as-is and create issues for the non-checked boxes in your todo list above, or are you creating this PR to help us track ongoing work, with the intention that it be merged when the boxes are all checked? Either way is fine by me, though I'd want some instances in the library to have docs so that it doesn't bitrot - then a simple test can just `:doc` them and make sure we get output.

Is there are reason why the `try`s need to be at the outer level, rather than around the parsing of the docstring itself down in the specific parsers for parsers and instances?

adamse:
I'd prefer to wait until all checkboxes are checked before merging!

Regarding the `try`s:

Without them I get
~~~
./Prelude/Functor.idr:11:5: error: expected: ")",
    ";", "instance", "|||"
    map : (m : a -> b) -> f a -> f b
    ^
~~~
when type checking `Prelude.Functor`. And the obvious way for me to fix this was to add the outer `try`s, but perhaps there is a way to move the `try`s inside?

adamse:
Also, what are some reasonable instances from the standard libraries to have docs? And what should they be?

david-christiansen:
RE parsing, my understanding is that it can't parse anymore without the `try`s because the parser has consumed the docstring, but then fails. `try` allow backtracking after all. So I would think that the `try` can be down inside the parser for instances / classes, wrapped around the docstring bit (and potentially the keyword `class` or `instance`). Or, perhaps better, there could be a "documentable declaration" parser that potentially parses docs and then sends the parsed docs onward to an inner parser, which would be classes or instances for now.

@ahmadsalim: any thoughts?

david-christiansen:
As far as instances in the library, I think the `Monad` instance for `Vect` surprises people who are used to the `Monad` instance for `List` - it might be good to point out that `>>=` on `Vect` is doing the diagonal. Some of the `Cast` instances might also document limitations, for instance that `cast {from=Int} {to=Nat}` maps negative numbers to `Z`.

adamse:
I like the idea of a "documentable declaration" parser as it would reduce code duplication. But wouldn't it include type declarations, data declarations, postulates, records and functions as well?

david-christiansen:
It is possible that the documentation comments in the parser are out of date - please update them if you notice that they've become wrong! Unfortunately we don't have a good way to machine-check those in Travis.

Right now, it should be possible to document modules (though that's parsed and stored specially), classes, functions, datatypes, and constructors. With your code, we'll be able to add instances to that. Eventually, we should probably also allow docstrings for records and their constructors/projections as well as namespace declarations.

I have an idea that the record thing might happen at the developer meeting in St Andrews that's coming up.

david-christiansen:
One other thing to do: please add yourself to CONTRIBUTORS, and make a note about the new feature in the CHANGELOG file.

ahmadsalim:
@adamse @david-christiansen I think I was a bit late, but the latest commit seems to do the right thing. 

Using `try` too much hurts performance and error messages, so the scope must not be larger than necessary.
See http://blog.ezyang.com/2014/05/parsec-try-a-or-b-considered-harmful/.

david-christiansen:
This looks like a definite improvement to the parser code generally. Thanks!

david-christiansen:
This is great progress!

adamse:
@david-christiansen there is now a test case, and some docs in the standard library, is there anything else that need doing?

I figure that https://github.com/idris-lang/Idris-dev/wiki/Documenting-Idris-Code should be updated once this PR is landed.

david-christiansen:
There's one thing that would be nice in the test / implementation: docstrings for parameters to instances. For example, could I write the following?

```
||| Showable things become strings.
||| @ a the showable type
instance Show a => Cast a String where
  ...
```

david-christiansen:
This is looking nice! I restarted the Travis job that failed.

Thanks so much!

david-christiansen:
One thing that's new in your code that's a big difficult to reason out based on diffs and commit messages is the binding of the name "subclasses" to a list. Is this because class docs now will show subclasses as well as superclasses?

adamse:
In the original code subclasses and instances were split and pretty printed in `pprintDocs`, I moved the splitting into `docClass`. This list of subclasses is nothing new, I simply moved code around.

david-christiansen:
Ah, that sounds good.

You have a "update documentation" point in your todo list, but that seems like something that's not part of a merge, right?

adamse:
Indeed, that was in case there was any documentation to be updated in the source repo, no need for it to stop a merge!

david-christiansen:
So you're ready to go?

I see one more thing to be done: the new test files (for `interactive008`) need to be added to the Cabal file, so that `cabal sdist` will include them. Otherwise the test doesn't get run! When you look in the Cabal file, then it will be pretty obvious how to do it.

After that, I think this is ready.

adamse:
Ah, I noticed that the test case was not run by travis.

david-christiansen:
Travis runs the result of a `cabal sdist` so that we don't accidentally break the release process. Perhaps we need to find a way to be more explicit about this.

david-christiansen:
Well, it's rubbing your test now. I'll merge it when Travis has finished, next time I'm on the net. Thanks again!

david-christiansen:
Thanks!

PR #1996, opened 2015-03-14
Type provider docs

adamse:
No docs for provided postulates, but there are already some:

~~~
(fmap (const (Left $ Msg "")) .
 parseDocstring $ T.pack "Provided postulate")
~~~

perhaps this should be used as a default if no explicit docs are given?


david-christiansen:
I actually think the Right Thing here is to get rid of the default for provided postulates, and parse it just like one would for provided values. I don't know why I put that in there :-)

So if you want to do that too, please go ahead! Merging this now because it's already a real improvement.

Thanks!

PR #2005, opened 2015-03-16
Update Ahmad's cppcheck branch to work

david-christiansen:
This pull request, assuming Travis passes, supersedes #1907. It does the following:
 1. The `realloc` leak identified by `cppcheck` is fixed in the RTS
 2. Other errors were found in mini-GMP (which we simply ship) and the Windows RTS (which I can't effectively test changes to). Thus, those files are now excluded from cppcheck. It would be nice if the Windows RTS files were improved and added to cppcheck!


david-christiansen:
Trying again with a different way of giving cppcheck the file list

david-christiansen:
@ahmadsalim Do you know why it's not claiming errors, yet still returning exit code 1?

jfdm:
@david-christiansen The error I see in the build logs:

```
src/Idris/Core/TT.hs:159:25: parse error on input ‘t’
make: *** [doc] Error 1
```

look eerily similar to a problem I had with PR #2008, even though I hadn't touched the code base. Try rebasing against master and see if that fixes it.

ahmadsalim:
Perhaps it is due to the warnings and style errors it provides,
maybe we could tone down the `--enable=all` to `--enable=error`.

ahmadsalim:
For example:
`[rts/idris_opts.c:26]: (warning) scanf without field width limits can crash with huge input data`

and
`[rts/idris_net.h:8]: (style) struct or union member 'idrnet_recv_result::result' is never used
[rts/idris_net.h:15]: (style) struct or union member 'idrnet_recvfrom_result::result' is never used`

david-christiansen:
Thanks!

Issue #1993, opened 2015-03-11
Type class unexpected behavior


nicolabotta:
The following type checks
````
> import Control.Monad.Identity
> %default total 

> class Monad m => ContainerMonad (m : Type -> Type) where
>   Elem    : a -> m a -> Type
>   ret     : a -> m a
>   ret     = pure
>   spec1   : {x : a} -> Elem {a = a} x (pure {a = a} x)

> instance ContainerMonad Identity where
>   Elem a1 (Id a2) = a1 = a2
>   spec1 = Refl
```
But replacing 
````
>   spec1   : {x : a} -> Elem {a = a} x (pure {a = a} x)
```
with
````
>   spec1   : {x : a} -> Elem {a = a} x (ret {a = a} x)
```
yields
````
- + Errors (1)
 `-- typeClasses2.lidr line 12 col 8:
     When elaborating right hand side of Main.Main.Identity instance of Main.ContainerMonad, method spec1:
     Can't unify
             x1 = x1
     with
             Main.Identity instance of Main.ContainerMonad, method Elem a x (Main.Identity instance of Main.ContainerMonad, method ret a x)
```
Is this the expected behavior ? How can I say that `Identity` is an instance of `ContainerMonad` when `spec1` is defined in terms of `ret` ? Thanks, Nicola

edwinb:
Curiously, if you give an explicit definition of 'ret' rather than using the default, this works fine (at least it does after I fixed a thing which arose from the recent type class fixes). This suggests default instances are not quite working as they should in the presence of dependent type classes, in that reduction is being blocked.

Issue #1814, opened 2014-12-31
Spinning out `base` into separate repository.


jfdm:
Taken from a recent mailing list discussion. The `base` package contains useful libraries that can aid, but are not crucial, to development of Idris programs. However, the `base` package should ideally be spun out into its own repository e.g. `idris-contribs`, that is under the `idris-lang` organisation control. Secondly, this new repository does not necessarily have to be monolithic and could be broken down further into other packages. 

If spinning out `base` the relevant tests from `test` need to be extracted as well, and linked into the travis builds.

edwinb:
I think it would be nice at least to organise base into some meaningful packages rather than treating it as a dumping ground for things that are likely to be useful. This is as much to help organise documentation as anything. It would also be good, while we're at it, to think a bit more about package management - .ipkg files are just a quick hack.

It probably isn't so useful to have separate repositories, at least not at the moment.

david-christiansen:
I definitely agree that meaningful packages is useful, especially now that tools are improving to deal with them.

I wrote code this morning to add IdrisDoc to modules, but I'm not satisfied with it and I'm going to redo it before submitting. Package docs as separate Markdown files that Idris can search and/or display will also make this nicer.

jfdm:
Regarding the separate repository idea. I think that in future we might want to consider the use of git submodules so that the `libs/` can be spun out *and* kept integrated within the project. This way for a minimal overhead we get the best of both worlds. Moreover, I suggest that *if* base and other libs are to be spun out we do such that [the version history is kept](http://stackoverflow.com/questions/359424/detach-subdirectory-into-separate-git-repository/17864475#17864475).

As for splitting `base` into separate packages, taking inspiration from [Apache Commons](http://commons.apache.org) I think the following split might be useful:

+ `data` for everything in `Data/`.
+ `lang` for code that affects the language e.g. `Syntax.idr`, `Providers.idr`, `Decidable/`
+ `comms` for code that deals with system IO and network IO i.e. communication.
+ `utils` for utilities that do not fit into the above packages. i.e. Debug and System.

Finally thinking about it if we were to spin out the new 'base' a better name for the package/repository might be either:

+ `commons` as in code that is commonly used but not necessary.
+ `contribs` the (AFAIK) standard FP name to detail contributed but not required code.

jfdm:
[...]

PR #2007, opened 2015-03-16
Initial reorganization into contrib

david-christiansen:
This PR is intended to be the beginning of a discussion. Please don't merge it yet.

Right now, the "base" library in Idris contains a wide variety of things. Some of them are widely-used, maintained, and supported, such as `Vect`. Others are experiments that haven't ever been used in actual code, as far as I can tell, such as `Data.Rel`. This is perhaps confusing, because it's hard to tell what things are fully supported and just not in the default namespace vs what things are brief experiments.

My proposal is to have another package that needs explicit use with `-p` called `contrib` that contains code that may be useful, that we want to avoid bitrot in, but that we don't really consider part of the Idris system _per se_. This package can contain things that are experimental, dubious, or just plain unused. Then, in the future, we'll be able to be quite free about PRs of little code bits for `contrib` while being much more selective about `base` - I suggest that, in the future, `base` only be open to modules that have _already_ been used in multiple programs, rather than speculative experiments. `contrib` provides a place for these to grow up.

So there's two parts to this proposal:
 1. Should we do this?
 2. Have I split things correctly in my initial pass through the library, as is here in this PR?

Your feedback is welcome.


markuspf:
I am all for this approach. In particluar weeding out all the half-baked experiments that don't really work or have plain wrong axioms in them (Algebra.idr, I am looking at you). I think these things should not look official, as "the user" seems to just assume they are correct, or asks question along the lines of "Is there a reason why my favourite obscure algebraic structure is not there", whereas them being in "some other place" makes them hopefully subject to the scrutiny they deserve (and the specialisation that makes them usable for that matter).

Of course there'll be the problem of what to call consesus or what's in "multiple programs" as soon as there are about 400 versions of Algebra.idr which are all the be-all-and-end-all of Algebra.idr, but neither of them really works, so this will call for a (separate) standardisation project. Later on.

suhr:
It's definitely a good idea. Also, in the future we can have more libraries with useful things.

xoltar:
Not only is this a great idea, but our next mission should be to come up with a mechanism for managing and distributing packages - Nix, go, npm, hackage all have interesting aspects, and this seems crucial to me for adoption and growth of the user community.

jfdm:
> Should we do this?
yes
> Have I split things correctly in my initial pass through the library, as is here in this PR?
i believe so.

I think we should also not forget to update CONTRIBUTING.md (and other documentation) with the new ideas concerning the existence and relationship between: `prelude`, `base,` and `contrib`. Here it should be made explicit (i.e. copy, paste, and tidy the idea above) what the rules and progress are for submits to the respective packages.

One thing that might be useful is to say what the criteria is for a good contribution to `prelude`, `base`, and `contrib`. For example:

+ Contributions for `prelude` & `base`:
  + *must have* associated tests
  + *must have* documentation
  + For `base` it *must be* shown to have potential use in multiple `Idris programs`
  + For `prelude` it *must be* shown to be potentially essential for all `idris` programs.
  + *must contain* a rationale for inclusion in the PR
+ Contributions to `contrib`:
  + *it would be nice but not essential* to have tests.
  + *should be documented*
  + *must contain* a rationale for inclusion in the PR.

Also I think that this PR can address Issue #1814, if we think about organising base into better packages. For example, taking inspiration from [Apache Commons](http://commons.apache.org) I think the following split might be useful:

+ `data` for everything in `Data/`.
+ `lang` for code that affects the language e.g. `Syntax.idr`, `Providers.idr`, `Decidable/`
+ `comms` for code that deals with system IO and network IO i.e. communication.
+ `utils` for utilities that do not fit into the above packages. i.e. Debug and System.

Although, the changes in splitting `base` into `base` & `contrib` might be sufficient to address #1814, and the resulting packages from splitting base might be too small to necessitate an individual package.
I do not think it is necessary **yet** to think about spinning the libraries out as separate git repos and included as git submodules.

david-christiansen:
We decided not to do #1814, so I've closed it.

With regards to rules, I really don't think we need that degree of formality - if we think a contribution needs things like tests, then we should certainly ask for them, but I don't think that a hard-and-fast approach is useful in a project this size.

WRT the alternative organization that you propose, dividing libraries after theme or content rather than expected stability: I see the primary benefit of making a "contrib" package as being a way to communicate what degree of stability, coherency, and general maturity things have. This will make it easier for us to accept initial experiments, and also easier for users to know how to complain about certain topics! But requiring "-p data" to use the functions in `Data.List` seems a bit odd to me - what would the concrete benefit be?

Also, nobody has yet commented on the specific split that's in this PR. For example, should monad transformers go into base or contrib? What about sockets? I think Vect should clearly stay in base, but what about BoundedList? I'd like to hear what others think about this, keeping in mind the criteria of stability and support.

david-christiansen:
@xoltar Yes, it would be nice to have a package manager, but that's a bit beyond the scope of this particular suggestion :-)

jfdm:
@david-christiansen I wasn't entirely clear with my suggestion concerning splitting `base`. My suggestion would be to divide it into packages based on theme, but still keep the implicit import that base has thereby maintaining functionality but making the packages organisation better.

I also do not disagree about wanting to project notions of: degree of stability, coherency, and general maturity. I welcome the contrabass proposal, I think that the division of base into thematic packages aides that.

david-christiansen:
In what specific way does it aid that?

jfdm:
The way I see it is that `contribs` presents us with an incubator to guide development of nascent code before it is to be include. if at all, into the 'core idris system'. `base` whether we like it or not does infact represent part of the 'standard library' for idris, and therefore why shouldn't we consider making it more coherent in its organisation. Thus, once experimental code that has reached maturity and is intended to be included into the 'core' package set, can be included into a more suitable home than `base`, `effects`, or `prelude`.

Also my suggestion does not argue against having a `contribs` package, nor does it argue against organising things wrt to expected stability. All it is arguing for is reorganising base into a better set of packages.

david-christiansen:
In what way are packages sorted by theme "more suitable"? (Sorry if I'm being dumb - I just don't understand the argument yet)

david-christiansen:
Also, how is that particular split more coherent? I don't think it will make it easier to find libraries for users, and I don't think it will improve compile times, and the lack of a story about mutual package dependencies will likely cause difficulty. Which is fine if there's a concrete, tangible benefit.

jfdm:
First, I do not know where you got the idea that my suggestion will improve compile times; AFAIK i do not make mention to that at all!

In fact now that you mention it, I do think that a more thematic organisation of base would help developers find things. The current mapping of `base` is a catch all things. Thus, if I have to find something say related to Networking, I have to trawl through everything in `base` and hope that it is all in the folder named `Network` and if not then double check everything. By having a more thematic split of `base`, as well as having the `contribs` package, we can start to say things like: All code related to networking that is mature will be in the `comms` package, and there is some interesting stuff in development in `contribs`. All mature code related to collections/data structures will be in `data`, and some interesting stuff in development in `contribs`.

david-christiansen:
I was just trying to think of any reason why it would be useful to have 3-4 packages with 3-4 modules each instead of one package with 12 modules, and that was all that I could think of :)

The point of something like `base` is really to have a basic set of handy stuff that can be relied upon, not to have large networking libraries that are difficult to navigate. Such a large library should perhaps be moved into a new package to make it easier to choose something else, but we don't have anything like that right now.

jfdm:
I think that you are right in that: Given the size of the modules in `base` this 'issue' of splitting `base` up doesn't present itself well. Maybe, this discussion should be left until base has grown considerably larger. At which point the role of base to have a 'basic set of handy stuff' gets more contentious, and there might be a need to spinout, on an adhoc basis, a 'comms.ipkg' or a 'lang.ipkg', or as well as the default  'base.ipkg'

nicolabotta:
Let me play the devil's advocate:

`Should we do this?`

Not yet. At this point, it is not very clear (to me) how Idris is going to support modular components. I would postpone the identification of a `base` set of components to the time this question has been answered. 

The rationale. We currently have a number of language constructs that can be used to modularize code: modules, type classes and perhaps records and parameter blocks. But we do not have clear guidelines (again, modulo my ignorance which is broad and deep) for using these constructs for library design. Different guidelines will lead to different hierarchies of components and, possibly, to  different notions of a `base` set.

The possible roles of modules and type classes in component design. Modules support partial definitions. These can be used to introduce a module context (specification, interface) in the form of a set of metavariables. Clients are required to implement these metavariables in order to use the module. This approach is viable but, as far as I know, has not been very much practiced and might have shortcomings related with name resolution rules. Type classes can also be used to introduce a context. In a library design based on type classes, client will access component functionalities by type class instantiation. But type classes also have not been systematically used for component design and there are issues (see for instance #1993) which might limit or even prevent such usage.

Another reason why I am not sure that a separation between `base` and `contrib` libraries should be done at this point is that it is not clear (again, to me) that there are enought Idris applications to meaningfully discriminate between what is widely used (and shall be considered proper part of the Idris system) and what can be considered to be contributions.

david-christiansen:
@nicolabotta:
Thanks for the feedback and for playing devil's advocate! That's very important!

Please note that this proposal is silent with regards to modules and structuring large codebases. There's things I'd really like to do in that space if I had the time, but  I don't. This will affect _packages_ - that is, a collection of things for which "import" works. Some of the things for which "import" works will no longer be available by default, requiring a command-line flag the way that the effects library does.

It doesn't mean that these things are set in stone forever, and if I ever get that module system together, then I hope that we will update everything to match! But that's got a bit of a long horizon, and in the meantime, we have a problem: the quality and completeness of the things in the bit called `base` varies quite a lot. Some things (like `Vect`) are quite mature, and other things (like `SortedSet`) are actually wrong, relying on false assumptions about type class coherence. From time to time, we get pull requests for small bits of code, and it would be nice to have a place for them to grow and develop that isn't part of a globally available namespace.

In essence, we have two areas today: `contrib` and `prelude` (though `contrib` is called `base` and available by default to imports). We already found the bits of the old Prelude that are either a bit tricky (like `So`) or specialized (like `Erased`) and moved them into what's now called `base`, and all the things that are less solid also came along. The idea of this is to get the things that really shouldn't be in `prelude` but are still fairly solid onto the default import path, while a flag is needed for things that are more experimental or new in character.

jfdm:
@david-christiansen FYI, I was just reading the last paragraph in your last comment, and it doesn't fit with what I thought was planned. `So` and `Erased` have been in base for a while now, and what you suggest implies that we are doing away with base.....

I thought the plan was to create a 'third' package called `contribs` that contains things that are more experimental/new in character. `prelude` would still be the things that are necessary for `Idris` programs, and `base` would be for things that are really useful and have solid foundations but not necessary for an `Idris` program. With the plan be to redistribute the packages already in `base` and `prelude` across the three packages.

david-christiansen:
I think I was a bit unclear in my writing there - sorry. You've successfully understood me, with one exception: I'm only proposing to move things from current-`base` to `contrib`, not from `prelude`, which I think we got fairly well sorted out at the last meeting in St Andrews. With the possible exception of some of the more obscure things in `Algebra.idr`.

david-christiansen:
And `So` and `Erased` were _already_ removed from the prelude - that's what I was trying to say.

jfdm:
no worries.

david-christiansen:
I blame hackathon brain and distraction by giant state monads :)

david-christiansen:
So, if we assume that `contrib` will happen, does anyone else have thoughts about which things stay in `base` and which things move?

markuspf:
Here's a list of random thoughts I had while looking at libs/prelude and libs/base


* I am going to go out and vote for moving anything even mildly controversial about Algebra out of
  prelude or base, move them into contrib, and then lets paint those bikesheds in all colours of the 
  rainbow. (That particularlly includes any form of arithmetic towers and such) 
* I'd move `Network` out of base, because it's neither portable nor does it work very well. 
* There's also Bits which I don't know how much of it is actually necessary (and did I mention that the explosion of primitives is some concern there as well?)
* While we're there, I'd remove DecEq for Float.
* Category is not necessary for idris or a runnable (average) idris program. I have seen this thing being implemented half-heartedly in about 100 different github repositories. While it's fun, by the guidelines that one actually *needs* it for many programs, it should go.
* I don't know about Isomorphism, this might become quite handy in many places

(I could probably rant on about this a long time, but I am also sure there are dependencies that I completely ignored and that I'll be slammed for in a minute)

BlackBrane:
I have a general question: What is the motivation for requiring `-p contribs`, or `-p Effects` for that matter? Shouldn't these things just be handled as imports? I imagine that if we're working on some project that uses these packages over a long period of time the (admittedly minor) inconvenience of this extra typing might seem to add up to more of an annoyance, and its not clear what the benefit is to me. Maybe just because it hasn't been explained to me yet.

As for the specific suggestions, I'm fine with all the specific moves David has done, but I have my own ideas on what should be done with Algebra.idr. Its been brought up before that Field isn't defined correctly (inverse should be defined only for non-zero elements), but this to me proves only that the definition should be corrected. My own tentative proposal, which I haven't PRed myself yet but maybe I should, is to keep everything up to Field together in prelude (of course we could be more aggressive with this too, but that seems like a sensible enough conceptual unit to preserve), and factor out the rest to two different base/contrib libraries. One for all the Lattice stuff, which I have no experience/confidence with and so could certainly go intro contrib as far as I'm concerned. And one for VectorSpace related stuff, in particular Modules, VectorSpaces and InnerProductSpaces. The latter I'd personally prefer stay in base, because these things are pretty ubiquitous in mathy computing, the correctness of the definitions is easy to verify, and at least I've used them a fair amount in Idris code. But since I'm probably the only real user so far, I'd of course understand if the community wants to override my preference on that account. In my branch where I did these changes I categorized these libraries under Control.Algebra.Lattice and Control.Algebra.VectorSpace. If we want to remove from Prelude even more aggressively, perhaps in addition to all this we could factor out all the "Verified" classes up to VerifiedField into a base library as well.

I certainly like the general idea, in particular having a library that we could contributed code without worrying that its fully up to the standards of what should be expected from Prelude and Base, and where people could then comment on and experiment with provisional proposals.

jfdm:
@BlackBrane Technically, they are handled as imports, but we need to tell `Idris` about them first. By default installed libraries are not added to the 'build path'. The `-p <package>` flag is adding these libraries to the build path for that compilation step.

For long term libraries you can always define an '.ipkg' file in which you can specify package dependencies in the `opts` option.  For example:

```
package foobar 

opts = "-p effects -p contrib"

modules = ...
```

BlackBrane:
@jfdm Ah, thanks for reminding me and clearing that up.

puffnfresh:
:+1:

edwinb:
I think we should just do this. Putting something into base implies we have some confidence in it, and there's a lot of stuff in base I have no confidence in at all... 

I think the split is roughly consistent with that - I would have kept System.Concurrency.Raw just because that is a than layer over the VM, but the rest looks fine.

david-christiansen:
@BlackBrane : I think that something else needs to be done for the Verified* classes - namely, they need to be parametrized over the instance as well. That  makes it reasonable to implement them.  More on that later!

david-christiansen:
@markuspf, @edwinb Thanks for the comments! I'll update this.

markuspf:
I completely agree with @edwinb that this should just be done.

@blackbrane shall we make an idris-algebra on idris-hackers and make an attempt at the algebra bikeshed? I saw someone suggesting starting at Magma instead of Semigroup, and we'd be able to work out how the (Verified) things are parametrised by instances as well (and @david-christiansen seems to have something up his sleeve there?)

david-christiansen:
WRT the Verified* stuff: I noticed that it's not actually implemented anywhere. When doing it, one quickly runs into problems with equalities over type class dictionaries. This is because we don't have a coherence principle.

The key is that you're not verifying the fact that (eg) List is a Functor, you're verifying that _the particular instance_ implements Functor for List correctly. So the VerifiedFunctor should be a relation between a type and its instance, not a function from instances to verification.

Here's a quick example of definition, instantiation, and use of what I had in mind:
```
module VeryFun

class VeryFun (f : Type -> Type) (dict : Functor f) | f where
  funIdentity : {a : Type} -> (x : f a) ->
                map @{dict} id x = id x
  funComposition : {a : Type} -> {b : Type} ->
                   (x : f a) -> (g1 : a -> b) -> (g2 : b -> c) ->
                   map @{dict} (g2 . g1) x = (map @{dict} g2 . map @{dict} g1) x

instance VeryFun List %instance where
  funIdentity []        = Refl
  funIdentity (x :: xs) = cong $ funIdentity xs
  funComposition []        g1 g2 = Refl
  funComposition (x :: xs) g1 g2 = cong $ funComposition xs g1 g2

test : (xs : List Nat) -> map ((+1) . (+2)) xs = map (+1) (map (+2) xs)
test xs = funComposition xs (+2) (+1)
```
In the meantime, Verified* as it exists today is getting moved to `contrib` until we actually have instances and uses. Then we can feel more free to tweak it than we can the things in the prelude.

david-christiansen:
If Travis passes, then 81b6b29 will get merged, unless any serious arguments come up in the intervening period.

Thanks for the feedback and discussion!

markuspf:
@david-christiansen I'll take a note of this, it makes a lot of sense to me right now!

PR #2011, opened 2015-03-19
Proposed revision of Algebra.idr: Classes beyond Field moved to two base libraries

BlackBrane:
There has been a lot of interest in cleaning up the Prelude, and Algebra.idr is rightfully considered a prime target: A lot of stuff in there is clearly not going to be used in typical idris programs. On the other hand, the most basic algebraic classes may frequently be useful, and are in fact used in the Prelude and probably elsewhere.

Therefore I propose this modest revision of Algebra.idr that cuts it almost in half, and moves the rest into base while preserving sensible conceptual units. In particular, the classes beyond Field are moved to `Control.Algebra.Lattice` and `Control.Algebra.VectorSpace`, both of which are self-contained.

This PR also fixes the definition of Field to be correct (inverse is defined only for non-zero elements). And also a minor, unrelated docstring formatting blemish (Sigma).

A couple comments and notes:

* While Field is now correct, there may be a way to do it better. the signature of `inverseM` is now `(x : a) -> (notId : x == neutral = False) -> a`, which means that Fields must be an instance of Eq, and that Classes and Bool must be imported.
* A few of the Lattice classes are instantiated in Nat.idr. For now I simply put these in the Lattice.idr file. They could instead go into their own file Algebra/Lattice/Nat.idr if we like, but it didn't seem worth it.
* The moved classes are also referenced by a couple base libraries, namely Heap now imports Lattice while Matrix imports VectorSpace.

### DIFF
--- a/libs/prelude/Prelude/Algebra.idr
+++ b/libs/prelude/Prelude/Algebra.idr
@@ -6,7 +6,6 @@ import Builtins
 infixl 6 <->
 infixl 6 <+>
 infixl 6 <.>
-infixl 5 <#>
 
 %access public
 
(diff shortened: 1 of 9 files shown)


libs/prelude/Prelude/Algebra.idr
```diff
@@ -286,44 +160,19 @@ class (VerifiedBoundedJoinSemilattice a, VerifiedBoundedMeetSemilattice a, Verif
...
+||| + InverseM of `<.>`, except for neutral
+|||     forall a /= neutral,  a <.> inverseM a == unity
+|||     forall a /= neutral,  inverseM a <.> a == unity
 ||| + Distributivity of `<.>` and `<->`:
 |||     forall a b c, a <.> (b <+> c) == (a <.> b) <+> (a <.> c)
 |||     forall a b c, (a <+> b) <.> c == (a <.> c) <+> (b <.> c)
-class RingWithUnity a => Field a where
-  inverseM : a -> a
+class (RingWithUnity a, Eq a) => Field a where
+  inverseM : (x : a) -> (notId : x == neutral = False) -> a
```
zraffer: 1. We can simplify `(notId : x == neutral = False)` into `(notZero: not (x == neutral))` or even `(x /= neutral)`.
2. ...or rather we can use propositional equality `(Not (x = neutral))` to avoid dependency on `Eq a`. What is better?

BlackBrane: As for the `So` option, wouldn't that, by nature of being a run-time check, make it inherently difficult to use `inverseM` in functions that pass the type checker? I never use `So`, so maybe I'm mistaken about that.

I like the `(Not (x = neutral))` option much better, and I tried it myself. I just encountered a [type class resolution error](https://gist.github.com/BlackBrane/6c277f97cc565d1bf684) when trying to define `VerifiedField` in the straightforward way, and so I proceeded with this definition instead. But maybe you or someone else can resolve it; it certainly seems like it shouldn't really be an issue. 

Also, along the lines of the previous question: Would this be more difficult to use in practice? What does it look like in general to supply a proof of `Not (x = neutral)`?

zraffer: I agree that two equalities `==` and `=` can be of different complexity to proof them or their negation. The `Eq` class seems more being a run-time check (even it is not so). 

Perhaps we should try to write a pair functions to establish the equivalence of `x == something = False` and `Not (x = something)`. Or rather of  `x == something = True` and `x = something`. Perhaps these functions should be members of a new `VerifiedEq` class. Then we will be able to provide a propositional inequality `Not (x = neutral)` via a call of such function (in instance declaration) while the `Field` class still being free of dependency on `Eq`.

david-christiansen:
Would you mind reorganizing this inside of the `contrib` package? For now, `prelude` and `base` are going to go up to `Monoid`, and the other bits will wait for a broader consensus.

BlackBrane:
Certainly @david-christiansen. I've started this PR over at #2140, so I'll close this one now.

PR #2013, opened 2015-03-19
Add "ship it" tactic, aliased to qed

puffnfresh:
I feel like an academic when saying "quod erat demonstrandum" - while
Idris is a research language, I think we need to be more practical.

I propose we start following Edwin's philosophy:

https://twitter.com/edwinbrady/status/431415892233428992

Which gives a better interactive proof session:

    Type checking ./Main.idr
    Metavariables: Main.onePlusOneIsTwo
    *Main> :p onePlusOneIsTwo

    ----------                 Goal:                  ----------
    {hole0} : 2 = 2
    -Main.onePlusOneIsTwo> trivial
    onePlusOneIsTwo: No more goals.
    -Main.onePlusOneIsTwo> shipIt
    Proof completed!
    Main.onePlusOneIsTwo = proof
      trivial

We should also consider using "move_fast_and_break_things" for
"believe_me" next.

### DIFF
--- a/src/Idris/ParseExpr.hs
+++ b/src/Idris/ParseExpr.hs
@@ -1234,6 +1234,7 @@ Tactic ::= 'intro' NameList?
        |   'term'
        |   'undo'
        |   'qed'
+       |   'shipIt'
        |   'abandon'
        |   ':' 'q'
        ;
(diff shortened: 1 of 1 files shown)


vendethiel:
> "move_fast_and_break_things" for
> "believe_me"

For April fool's? :P

david-christiansen:
This is funny, but I don't think it has a place in the language. :0)

fponticelli:
I wish it happened :)

L8D:
love it. I think someone should start a poll and we'll see how the community feels about these kinds of changes.

jfdm:
I agree with @david-christiansen, as funny as this might be it certainly has no place in the language. Even though you might feel like an 'academic', and I do not know what bad connotations you are throwing around with that sentiment, and that Idris has its origins as a research language, the use of `qed` in theorem proving and `believe_me` are innately practical function names. These names describe accurately their purpose.

Issue #287, opened 2013-04-22
Universe consistency check should be done at each new constraint


bitonic:
As discussed on IRC, constraints for the type hierarchy are not checked eagerly enough, and it's probably possible to send the compiler into a loop exploiting that.

jfdm:
Just a bump saying that this is still a problem. Below is the code snippet from above updated to reflect changes in syntax.

```idris
module TypeInType;

data Tree : Type where
  sup : (a : Type) -> (f : a -> Tree) -> Tree

a : Tree -> Type
a (sup a _) = a

f : (t : Tree) -> a t -> Tree
f (sup a f) = f

normal : Tree -> Type
normal t = (y : a t ** (f t y = sup (a t) (f t))) -> Void

nt : Type
nt = (t : Tree ** normal t)

p : nt -> Tree
p (x ** _) = x

q : (y : nt) -> normal (p y)
q (x ** y) = y

r : Tree
r = sup nt p

lemma : normal r
lemma ((y1 ** y2) ** z) = y2 (replace {P = (\ y3 => (y : a y3 ** f y3 y = sup (a y3) (f y3)))} (sym z) ((y1 ** y2) ** z))

russel : Void
russel = lemma ((r ** lemma) ** Refl)
```

ahmadsalim:
Still seems to be an issue in 0.11:

```idris
module TypeInType

data Tree : Type where
  Sup : (a : Type) -> (f : a -> Tree) -> Tree

a : Tree -> Type
a (Sup a _) = a

f : (t : Tree) -> a t -> Tree
f (Sup a f) = f

normal : Tree -> Type
normal t = (y : a t ** (f t y = Sup (a t) (f t))) -> Void

nt : Type
nt = (t : Tree ** normal t)

p : TypeInType.nt -> Tree
p (x ** _) = x

q : (y : TypeInType.nt) -> normal (p y)
q (x ** y) = y

r : Tree
r = Sup nt p

lemma : normal TypeInType.r
lemma ((y1 ** y2) ** z) = y2 (replace {P = (\ y3 => (y : a y3 ** f y3 y = Sup (a y3) (f y3)))} (sym z) ((y1 ** y2) ** z))

russel : Void
russel = lemma ((r ** lemma) ** Refl)
```

edwinb:
I've fixed the underlying issue with the inconsistency here, though the snippet in question will still type check for a different reason. I've opened a new issue, #3194, about that, so I think it's time to close this one.

[...]

PR #2021, opened 2015-03-20
Universe solver

ozgurakgun:
I think the new universe solver is ready to be merged.

I am fairly certain that it is correct now. Well, until someone finds a problem with it. :)
All the tests pass.

There were a few simple things I could do to improve the performance a bit. It isn't as fast as I wanted, I think it is a bit slower than the old one. However, it has the added benefit of being correct.

It should also be easy to extend it with new kinds of constraints in the future.

NOTES
- There are two interlinked TODOs left in the code. Edwin knows about them.
- I updated the error message for `error001`. It still fails the universe check, but the error message is different & more verbose. The error message for universe inconsistencies should be improved and expected file for this test should be updated accordingly.
- I updated the error message for `error006`. This seems to be a regression that I have introduced. This should be fixed before a merge.
- I have validated that #287 generates a consistent set of constraints. So it is not fixed yet, but it shouldn't be the solvers fault.

Cheers!

### DIFF
--- a/src/Idris/Core/Binary.hs
+++ b/src/Idris/Core/Binary.hs
@@ -594,8 +594,8 @@ instance {- (Binary n) => -} Binary (TT Name) where
                            x2 <- getWord8
                            return (Proj x1 ((fromEnum x2)-1))
                    6 -> return Erased
-                   7 -> do x1 <- get
-                           return (TType x1)
+                   7 -> do _x1 <- get :: Get UExp
+                           return (TType (UVar (-1)))             -- TODO: temp hack, make Edwin fix this!
                    8 -> return Impossible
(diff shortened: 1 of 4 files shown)


david-christiansen:
I am really looking forward to putting some nice user interface on top of actually having universe levels available! Thanks for implementing this, @ozgurakgun!

edwinb:
This all seems to work fine, thanks! I've put a location on top of the error and updated the tests.

It does seem to be a bit slower than the old solver, but has the advantage of actually adding concrete levels, so that's much better. I'll take a closer look and see if I can speed it up on the pathological example I have (that's the echo protocol code in ConcIO).

PR #2028, opened 2015-03-20
Prevent variable capture when desugaring custom syntax

### LINKED ISSUE #1172: Variable capture in syntax declarations with lambdas
You can have syntax declarations with lambdas on the right hand side:
```
syntax delay  [x] = \z => case z of { () => x }
syntax delay' [z] = delay z

broken : a -> () -> a
broken x = delay' x
```
However, there are some issues with variable capture:
```
*SyntaxBug> :r
Type checking ./SyntaxBug.idr
SyntaxBug.idr:4:32:
When elaborating right hand side of SyntaxBug.case block in broken:
Can't unify
        ()
with
        a

Specifically:
        Can't unify
                ()
        with
                a
Metavariables: SyntaxBug.broken
```
The error goes away if you rename the bound variable from z to zz.

david-christiansen:
Bound variables are replaced with globally unique symbols now.

Fixes #1172.


david-christiansen:
Please don't merge this until a consensus about the capturing change has been reached (see mailing list post).

david-christiansen:
OK, this has been discussed on the mailing list and it seems to be just fine.

PR #2029, opened 2015-03-21
Replace div/mod functions with divNZ/modNZ

canndrew:
This commit replaces the `div` and `mod` functions in the Integral class:

    class Integral a where
      div : a -> a -> a
      mod : a -> a -> a

With functions that take an extra parameter, a proof that the second
argument is non-zero.

    class Integral a where
       zero : a
       divNZ : a -> (y: a) -> Not (y = zero) -> a
       modNZ : a -> (y: a) -> Not (y = zero) -> a

It also add a `Math` module containing functions `div` and `mod` that return
`Either (y = zero) a` instead of `a`.

### DIFF
--- a/libs/base/Data/Bits.idr
+++ b/libs/base/Data/Bits.idr
@@ -1,24 +1,30 @@
 module Data.Bits
 
+import Math
 import Data.Fin
 
 %default total
 
-divCeil : Nat -> Nat -> Nat
-divCeil x y = case x `mod` y of
-                Z   => x `div` y
-                S _ => S (x `div` y)
+divCeilNZ : Nat -> (y: Nat) -> Not (y = Z) -> Nat
+divCeilNZ x y p = case (modNZ x y p) of
+  Z   => divNZ x y p
+  S _ => S (divNZ x y p)
+
+divCeil : Nat -> (y: Nat) -> Either (y = Z) Nat
+divCeil x y = case decEq y Z of
+  Yes p => Left p
+  No  p => Right (divCeilNZ x y p)
 
 nextPow2 : Nat -> Nat
-nextPow2 Z = Z
-nextPow2 x = if x == (2 `power` l2x)
-             then l2x
-             else S l2x
-    where
-      l2x = log2 x
+nextPow2 x = case log2 x of
+  Left _    => Z
+  Right l2x => if x == (2 `power` l2x)
+               then l2x
+               else S l2x
 
 nextBytes : Nat -> Nat
-nextBytes bits = (nextPow2 (bits `divCeil` 8))
+nextBytes bits = (nextPow2 (divCeilNZ bits 8 nz))
+  where nz Refl impossible
 
 machineTy : Nat -> Type
 machineTy Z = Bits8
--- a/libs/contrib/Data/Hash.idr
+++ b/libs/contrib/Data/Hash.idr
@@ -38,7 +38,8 @@ byte n w = prim__lshrB64 (prim__andB64 mask w) offset
 ||| Modulo of an integer, downsized to a Bits64
 private
 mod64 : Integer -> Bits64
-mod64 i = assert_total $ prim__truncBigInt_B64 (abs i `mod` 0xffffffffffffffff)
+mod64 i = assert_total $ prim__truncBigInt_B64 (modNZ (abs i) 0xffffffffffffffff nz)
+  where nz Refl impossible
 
 instance Hashable Bits64 where
   saltedHash64 w salt = foldr (\b,acc => (acc `prim__shlB64` 10) + acc + b)
--- a/libs/prelude/Math.idr
+++ b/libs/prelude/Math.idr
@@ -1,0 +1,49 @@
+module Math
+
+import Builtins
+import Prelude.Basics
+import Prelude.Classes
+import Prelude.Nat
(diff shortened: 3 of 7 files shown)


canndrew:
Prior to this patch the way `div` and `mod` (mis)behave depends on the type. For primitive types like `Integer` it crashes the process. For `Nat`, `x / 0 => x + 1`. Needless to say this is pretty terrible. In some languages division by zero raises an exception. In Idris a good analogy for this behaviour is for a function to return an `Either` so that is what the new `div` and `mod` functions do.

edwinb:
Thanks for the contribution. I'm not sure this is exactly the right solution to this problem though (at least, not yet), for the same reason we have separate Verified* type classes rather than requiring proofs of laws. I don't think the Idris Prelude should force you to write in a way which *requires* totality checking, proofs, etc, even if it does encourage it. There are lots of possible solutions which do require totality checking, and I don't think we're in a position to say which is best yet.

I think this should be discussed on the mailing list (or even in the comments here) before we decide exactly what to do (indeed this is what we suggest for Prelude changes in CONTRIBUTING.md).

My first thought would be to add a `SafeIntegral` type class of the form in the patch, to the `contrib/` package, so that anyone who wants to write total programs with `div` and `mod` can use that rather than the `Integral` versions. Indeed, this is something I've wanted for some things in the past but never got around to doing.

The reason for the Travis fail, by the way, is that if you add a module to a library it needs to be added to the corresponding .ipkg file too or it won't be installed.

canndrew:
When I was writing this my first thought was to replace `div` and `mod` with versions that return `Maybe` or `Either`:

    class Integral a where
      div : a -> a -> Maybe a
      mod : a -> a -> Maybe a

In most languages divide by zero would raise an exception. We could do that here aswell but I personally prefer the `Maybe`/`Either` solution. They're monads so it's easy to push the error condition to outside of whatever calculation the user is doing. And if someone does want an exception they can just use `fromJust` or equivalent.

Writing totality proofs can be very difficult and so it's not a good idea to force people to write them just for such basic operations as divide and modulo. However if we were going to have the sort of `div` and `mod` functions I describe above the more flexible solution would be to implement them in terms of `divNZ` and `modNZ` functions. ie:

    div: a -> a -> Maybe a
    div x y = case decEq y 0 of
      Yes _ => Nothing
      No  p => divNZ x y p

Which is basically what this patch implements.

> My first thought would be to add a `SafeIntegral` type class of the form in the patch, to the `contrib/` package, so that anyone who wants to write total programs with `div` and `mod` can use that rather than the `Integral` versions.

Okay, suppose we had a `SafeIntegral` instance of `Nat`. That still leaves the question of what the `Integral` instance should look like. Currently it looks like this:

    total
    divNat : Nat -> Nat -> Nat
    divNat left Z         = S left               -- div by zero
    divNat left (S right) = ...

Which is madness. Or take `Integer` division:

    divBigInt : Integer -> Integer -> Integer
    divBigInt = prim__sdivBigInt

Presumably C functions can't raise exceptions or return complex types like `Maybe`. So whatever wraps `prim__sdivBigInt` is going to have to check for zero divisors by the time Idris has a mature standard library (unless we're okay with having standard library functions that generate a `SIGFPE`). Again, the most flexible way to do this is to force this restriction directly in the type of the function that wraps `prim__divBigInt`. ie.

    divBigInt : Integer -> (y: Integer) -> Not (y = 0) -> Integer`
    divBigInt x y _ = prim__sdivBigInt x y

Then `divBigInt` can be wrapped by more convenient functions that raise an exception or return a `Maybe`/`Either`.

canndrew:
> The reason for the Travis fail, by the way, is that if you add a module to a library it needs to be added to the corresponding .ipkg file too or it won't be installed.

Fixed.

edwinb:
> In most languages divide by zero would raise an exception.

In Idris, trying to execute a missing case in a partial function should crash with an error message stating what the missing case was.

> [On the Nat instance of Integral] Which is madness. 

Yes, this should be a partial function too. I would prefer to correct this.

> unless we're okay with having standard library functions that generate a SIGFPE

One point of the totality checker is to be able to identify where there are programs which could possibly fail at run-time. 

I think ultimately we would like to have a completely safe standard library, with proofs in all sensible places, but we're a long way from the point where people will be willing to program like this. That's why at the moment we prefer having more "conventional" code (and type classes) in the library, with the Safe/Verified versions available if desired.

This patch is a breaking change for the Prelude, and as such I'm not going to apply it directly without some wider discussion and consensus (see CONTRIBUTING.md for more on this).

If there is a "Safe" version of the class, we can see how well it works in practice for those who are aiming to write provably total programs, and encourage its use. I think this *should* be in the library, but not as the *only* way to do division.

david-christiansen:
I'd like to see a safe version in `contrib`, and if it gets multiple users then we can graduate it to `base` or `prelude`.

Melvar:
Do these functions actually need a new typeclass? Is `Num` plus the existing `Integral` not sufficiently powerful? (I was vaguely surprised to find `Integral` does not require `Num`.)

david-christiansen:
I'm going to close this PR for now, as there's been quite some time without new activity. Please feel free to re-open this one or submit a new one with a safe version of Integral for `contrib`, so the community can get a chance to work with it for a while and iron out the issues.

PR #2038, opened 2015-03-21
A RWST bug fixed.

TorosFanny:
A Writer Monad should sum up all the log messages


Melvar:
Huh, that looks like it has a few more bugs than that …

PR #2044, opened 2015-03-22
Improve universe solver

ozgurakgun:
I have done a few obvious things to hopefully improve the performance a bit. Happy to look at this again later if I can get my hand on a tricky problem to do with universe checking.

The error message for error001 is changed due to the order of constraints. This is expected.

The error message should probably be updated not to print that much info anyway.

### DIFF
--- a/src/Idris/Core/Constraints.hs
+++ b/src/Idris/Core/Constraints.hs
@@ -62,14 +62,34 @@ solve maxUniverseLevel inpConstraints =
                         ]
                     }
 
-        -- | a map from variables to the list of constraints the variable occurs in.
-        constraints :: M.Map Var (S.Set (UConstraint, FC))
-        constraints = M.fromListWith S.union
+        lhs (ULT (UVar x) _) = Just (Var x)
+        lhs (ULE (UVar x) _) = Just (Var x)
+        lhs _ = Nothing
+
+        rhs (ULT _ (UVar x)) = Just (Var x)
+        rhs (ULE _ (UVar x)) = Just (Var x)
+        rhs _ = Nothing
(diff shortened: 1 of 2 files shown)


edwinb:
Thanks for this! I was tinkering with the solver most of yesterday (mostly trying to reduce the number of constraints it gets) and this certainly speeds things up.

There's still the odd pathological case - Control.Arrow in base is currently taking about 33 seconds to check, for me. But when I try to profile it, its suddenly fast, so I'm wondering if that's something odd at my end instead.

edwinb:
One question - if we made a better effort to put concrete levels in (a lot are obviously just 0 for example) and allowed users to make levels explicit, would this have a significant effect on the speed of the solver?

Important as it is, it does seem to be a bottleneck so if we can give it help and have it run much faster that'd be great!

ozgurakgun:
I will take a look at what happens with `Control.Arrow`, probably tomorrow.

Setting variables to 0 if they are known to be 0 should help. `x <= 0` will only be propagated once, so that's good.

There are also other things we can do to improve the performance. I started with a simple implementation hoping that it would be fast enough already. We can keep adding tricks gradually until it is fast enough. :)

edwinb:
It turns out that the problem is simply that constraintsLHS and constraintsRHS aren't cached like you'd expect, since they're in a where clause and built every time. Possibly a different set of optimisations would do this, but I've just cached them explicitly in the state and suddenly it's running more or less instantly. Yay!

ozgurakgun:
Oh *that* is great news! :)

Haskell's performance characteristics still manage to puzzle me.

PR #2052, opened 2015-03-23
Partial div mod

canndrew:
Make `divNat` and `modNat` functions partial and change various functions that this effects. Make the other integer division/modulo functions fail on a missing case.

### DIFF
--- a/libs/base/Data/Bits.idr
+++ b/libs/base/Data/Bits.idr
@@ -4,11 +4,13 @@ import Data.Fin
 
 %default total
 
+partial
 divCeil : Nat -> Nat -> Nat
 divCeil x y = case x `mod` y of
                 Z   => x `div` y
                 S _ => S (x `div` y)
 
+%assert_total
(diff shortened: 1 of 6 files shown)


Melvar:
I believe `%assert_total` is deprecated.

edwinb:
Thanks for this. I think the test fails here are okay (it's just spotting a different reason for totality failure) but they should still be cleaned up before merge. You can do this with 'runtest.pl testname00x update' in the tests directory, or simply with './run > expected' in the specific test directory.

One thing I notice is that natEnumFromThenTo has become partial - it shouldn't need to be, and indeed it was written this way so that it could be total. If 'inc' is 0 it should return an empty list (we follow the behaviour of e.g. [2,1..10] in Haskell here). This is just an effect of the old behaviour of 'div' so will need to be written slightly differently.

edwinb:
Thanks for fixing this up. Sorry to leave the PR lurking so long.

PR #2056, opened 2015-03-24
Fixes build errors relating to use of Sphinx 1.2.2.

jfdm:
ReadTheDocs works with Sphinx v1.2.2. Most platforms will install a more recent version of Sphinx (v1.3). This PR fixes the last remaining errors caused my version mismatch in Sphinx directive support.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -10,7 +10,13 @@ To build the manual the following dependencies must be met. We assume that you h
 ### Sphinx-Doc
 
 Python should be installed by default on most systems.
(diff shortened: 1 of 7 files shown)


jfdm:
Apparently Travis is failing to report that this build failed on the Javascript test. I shall resubmit this PR to see if that makes a difference.

*note* I have no idea why this PR would affect the javascript test output.

edwinb:
Sometimes they stall rather than fail, in which case just ask travis to restart it and see what happens...

jfdm:
ah okay, I just resubmitted the PR.

PR #2067, opened 2015-03-26
Re-export Effect.Monad from effects package

L8D:
For some reason when `Effect.System` was added in the commit `4ba5460` the export of `Effect.Monad` got replaced.


L8D:
I guess I was mistaken or something, since this code is way out of date. Someone needs to remove this file and replace the definitions in `Effects.idr` with class instances.

PR #2068, opened 2015-03-27
Clarify Data.Fin meaning for people who haven't seen this usage before

xoltar:
I'm still worried that Data.Fin seems an awfully abstract way of talking about indexing a vector, but hopefully this helps for some folks.

### DIFF
--- a/docs/tutorial/typesfuns.rst
+++ b/docs/tutorial/typesfuns.rst
@@ -334,7 +334,7 @@ importing ``Data.Fin``, or can be declared as follows:
        FZ : Fin (S k)
        FS : Fin k -> Fin (S k)
 
-From the signature,  we can see that this is a data type that takes a ``Nat``, and produces a type.
+From the signature,  we can see that this is a type constructor that takes a ``Nat``, and produces a type.
 So this is not a set in the sense of a collection that is a container of objects,
 rather it is the canonical set of unnamed elements, as in "the set of 5 elements," for example.
 Effectively, it is a type that captures integers that fall into the range of zero to ``(n - 1)`` where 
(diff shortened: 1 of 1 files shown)


xoltar:
I can't see any way the build errors apply to this change.

jfdm:
Sometimes travis gets confused for some reason. I have restarted the build, will see what happens.

edwinb:
It seems to be upset today. I think it's safe to ignore for this patch.

edwinb:
This is a useful addition, but one comment I'd make is that 'Fin' itself isn't a type, as you write in "data type that takes a ``Nat``, and produces a type". Rather, it's a family of types indexed by a Nat. We need to be careful with wording here, since we use 'Type' in a more precise way than other languages do.

xoltar:
Thanks, @edwinb I'll make that change. I agree I was using the term "type" a bit loosely. It looks like we do introduce the term "indexed" in the section above, so using that should be OK. Shouldn't I refer to Fin (in this context) as a "type constructor," though?

I've still not understood why sometimes a type constructor is indexed and sometimes it's parameterized (as in "We say that ``Vect`` is *indexed* over ``Nat`` and
*parameterised* by ``Type``."). Is it "indexed" when we provide a value, and "parameterized" when we provide a type? What if we provide a type constructor instead of a type, what would that be called?

If someone can clarify that for me, I'll happily update that section in the tutorial as well.

edwinb:
We might be better not distinguishing indices and parameters at that point, actually - it's important in various contexts, but not yet, and most of the time you can get by without know.

A parameter doesn't vary across a structure (e.g. the type of elements) and an index does (e.g. the length of the sub lists). Parameters are automatically in scope in the body of a declaration, but indices aren't.

xoltar:
@edwinb OK, that almost makes sense, let me think about that for a little bit. In the meantime I've changed type to type constructor and made a couple of small changes (like removing "Obviously,..." because nothing in the Data.Fin section was obvious the first time I read it and I suspect there are others in my position...)

edwinb:
Yes, "obviously" is in the set of words that have no place in tutorial materials along with things like "straightforward" and "just" and "clearly" and "trivial" and so on. Still, they do creep in from time to time. Thanks!

PR #2079, opened 2015-03-30
Implement bundled toolchain feature

melted:
This commit makes it possible to build a freestanding Idris that uses a bundled toolchain (or a toolchain in a specific absolute location). The path (relative or absolute) to the toolchain is specified in the env variable IDRIS_TOOLCHAIN_DIR when running a build with the freestanding flag.

I have used something similar but slightly more hardcoded to build the windows binaries.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 New in 0.9.18:
 --------------
+* Add support for bundled toolchains.
 * Strings are now UTF8 encoded in the default back end
 * Idris source files are now assumed to be in UTF8, regardless of locale
   settings.
--- a/Setup.hs
+++ b/Setup.hs
@@ -137,17 +137,33 @@ generateTargetModule verbosity dir targetDir = do
                                     ++ "   dir <- getDataDir\n"
                                     ++ "   return (dir ++ \"/\" ++ name)"
 
+-- a module that has info about existence and location of a bundled toolchain
+generateToolchainModule verbosity srcDir toolDir = do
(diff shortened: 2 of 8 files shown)


david-christiansen:
Would you please add a note about this to the CHANGELOG as well as to the ReST docs?

melted:
I have added docs (and made ghc 7.6 happy).

david-christiansen:
Thanks!

What would it take to get a test for this into Travis? Or will this enable AppVeyor testing?

melted:
It should work just fine to test on Linux, it's not Windows-specific. I suppose the biggest hurdle to test it properly is that it would have to get a toolchain from somewhere.

melted:
I'd would be hesitant to make the Travis builds even slower to test such a niche feature. That it doesn't interfere with normal builds is what's important.

david-christiansen:
@edwinb I haven't done so much with the backend guts and C - would you mind reviewing this before I merge it?

edwinb:
Looks okay to me. I think I understand what it's doing! If it can ultimately make the windows build easier that's a good thing too. Seems to be a merge conflict somewhere though.

melted:
Yeah, I'll rebase it, probably just the changelog.

melted:
Turns out I changed Utils/System.hs a couple of hours ago. Of course :-/ Should be easy enough to merge.

melted:
Updated, let's hope Travis is in a good mood.

PR #2082, opened 2015-03-31
Heading normalisation, and reST usage fixes.

jfdm:
The main contribution of this PR is the normalisation of the reST syntax used to denote document sections. There are inconsistent uses of notation within the documentation. 

By design reST & Sphinx detect sectioning by noting 'runs' of notation used, as long as the notation used is consistent within a file this should not matter. Thus inconsistent usage across files should not affect the build result.

However, to ensure consistency in the syntax used the headers have been changed to use the [convention used for Python Documentation](http://sphinx.readthedocs.org/en/latest/rest.html#sections). Such that:

+ As each folder in `docs` represents a 'part', the headers will be denoted using the part syntax.
+ As each file in a part represents a chapter, the headers will begin with the chapter syntax, and 'subsections' will be bumped up a level in syntax used.

A minor secondary contribution is fixes for some minor reST inconsistencies.


david-christiansen:
You have a merge bit, right? If so, I'll wait for you to merge your own PRs unless you say otherwise in a comment.

PR #2084, opened 2015-03-31
Division and modulus for Nat, plus associated utilities

david-christiansen:
This adds some trivial lemmas to `Prelude.Nat`, as well as the following `contrib` modules:
 * `Data.Nat`: contains division and modulus for `Nat`, using well-founded recursion
 * `Control.WellFounded`: contains low-level operations for well-founded recursion


Melvar:
I find the naming of `lteSucc` a bit odd, after all, it is importantly different from `LTESucc`. If you think this will not lead to confusion, carry on.

david-christiansen:
You're right. I'll rename it.

PR #2080, opened 2015-03-31
Idris does not build with GHC 7.10.1

jstolarek:
Idris depends on `filepath` < 1.4 but GHC 7.10.1 ships with version 1.4.0.0. This causes havoc when cabal tries to resolve dependencies.

Idris also depends on `lens` version 4.7 and that does not build with 7.10.1 either. Perhaps bump dependency to 4.8?

`annotated-wl-pprint` dependency does not build either but I've already created a pull request that fixes the build problem.

Finally, Idris itself fails to build because some modules define `<$>` operator and that clashes with GHC 7.10.1 `Prelude`.

After bumping the dependencies (filepath to 1.4, lens to 4.8) and fixing the build errors everything seems to work fine.

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -817,10 +817,10 @@ Library
                 , containers >= 0.5 && < 0.6
                 , deepseq < 1.5
                 , directory >= 1.2 && < 1.3
-                , filepath < 1.4
+                , filepath < 1.5
                 , fingertree >= 0.1 && < 0.2
                 , haskeline >= 0.7 && < 0.8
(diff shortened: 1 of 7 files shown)


jstolarek:
I am surprised to see build failure. It looks like one of the test jobs (`TESTS=lib_doc` for GHC 7.6) was killed due to timeout. Is it possible to re-run that job?

jfdm:
@jstolarek The build appears to have stalled, I have restarted the build process. It looked like everything else went through so fingers crossed...

PR #2088, opened 2015-04-01
Replace the Foldable List instance

treeowl:
Implementing `foldr` in the obvious way should be just as fast as
the tail recursive way, and much more useful for proofs.

Defining `foldl` directly may improve performance, and will
probably make some proofs easier.

### DIFF
--- a/libs/prelude/Prelude/List.idr
+++ b/libs/prelude/Prelude/List.idr
@@ -309,13 +309,12 @@ mapMaybe f (x::xs) =
 -- Folds
 --------------------------------------------------------------------------------
 
-||| A tail recursive right fold on Lists.
-total foldrImpl : (t -> acc -> acc) -> acc -> (acc -> acc) -> List t -> acc
-foldrImpl f e go [] = go e
-foldrImpl f e go (x::xs) = foldrImpl f e (go . (f x)) xs
-
 instance Foldable List where
-  foldr f e xs = foldrImpl f e id xs
+  foldr c n [] = n
+  foldr c n (x::xs) = c x (foldr c n xs)
+  
+  foldl f q [] = q
+  foldl f q (x::xs) = foldl f (f q x) xs
 
 --------------------------------------------------------------------------------
 -- Special folds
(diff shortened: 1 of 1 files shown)


edwinb:
I think we've talked about this offline in the past for this function 
but never did anything about it - my preference is for the more 
fundamental library functions like this to be in the form that's easy to 
prove things about, and if there are ways to make it more efficient they 
can be set up as transformation rules (which are mostly implemented but 
not documented yet)

One question though: you say "should" be just as fast, and "may" improve 
performance, but that sort of phrase rings alarm bells. Have you 
measured this? In general, I'm wary of things which claim results about 
efficiency without any evidence, although in this specific case there 
are reasons other than efficiency for taking the simple version.

treeowl:
No, I have not made any measurements. I'm just running, in this case, on what I know:

1. A right fold of a right-strict function is inherently inefficient, because no matter how it's implemented, it needs to traverse the whole list, remembering something about each cons it passes, before it can start calculating anything. Whether the path is remembered "on the heap", as in the current approach (storing composition closures) or "on the stack" as in my approach, is not likely to make enough of a difference to worry about.
2. The current version of `foldl` is more complicated, and certainly would need more compiler transformations than the proposed version to be efficient. Unlike `foldr`, `foldl` is *very* useful in practice in a strict environment, so its efficiency actually does matter.

JanBessai:
I'm far to absent from the project to vote/decide on anything, but here are a few (historic?) notes, why the current implementation is there:

Implementing `foldr` without tail-recursion causes crashing programs if the stack is exhausted, which - even on modern PCs - happens quickly and independently of the total amount of ram present. Controling how much stack the task of the program will get is not easy and users probably have to run the program multiple times before they are able to set the right value. Depending on the back-end (e.g. JavaScript) it might be impossible to reliably do so.
Usually there is more heap available and for use cases where the tail recursive solution would require e.g. 40MB of heap space, no time or space problem would be caused.

Considering proofs,  these should be easier when foldl is implemented in terms of foldr, since the algebraic properties are inherited. Code outside of Prelude doesn't have access to the implementation internals anyway, so its proofs are unaffected. The Coq standard library has a solid set of lemmas that might be of interest to expose:
https://coq.inria.fr/distrib/current/stdlib/Coq.Lists.List.html

Compiler optimizations will probably aim at fusion and whoever implements that will come up with the best solution along the way.

treeowl:
@JanBessai, I didn't realize Idris had a fixed limited stack. GHC had such a thing for a long time, but it caused so many headaches they eventually scrapped that design. If `foldr` is to be defined tail-recursively, then we will indeed need laws and will need to get them ASAP.

edwinb:
The limited stack is more a problem with compiling via C than Idris itself. At some point we'll need to address this.

Personally I prefer the simple definition in this patch to the tail recursive version in any case, because it's easier to prove things about and Lists are the sorts of things we'd expect to see in types. We do have transformation rules which would allow more efficient implementations without affecting proofs, but I haven't documented these yet and the syntax for introducing them might change in lots of ways - they are used internally. So there is more to say on this, but not right now...

In general, library changes which claim benefits in efficiency ought to be backed up with experimental evidence, though - there are lots of considerations which might apply in one setting (e.g. Haskell) that don't apply here, so surprising things sometimes happen.

PR #2089, opened 2015-04-01
Remove assert_total from Prelude.List.transpose

treeowl:
This pragma was apparently needed once, but does not seem to be
any more. The totality check must have improved.


david-christiansen:
It seems that this is still needed, based on the test output on Travis where the modified function fails the totality checker.

PR #2095, opened 2015-04-01
Don't let tests depend on colouring.

melted:


david-christiansen:
This is great! Thanks!

PR #2097, opened 2015-04-01
Don't hardcode 'cc' as the compiler

melted:
Fixes ffi006 on windows


melted:
Grumble.

david-christiansen:
Darn that travis!

melted:
I got a good feeling about parameter expansion.

PR #2100, opened 2015-04-02
Replace the Monoid instance on Maybe

david-christiansen:
The current default Monoid instance for Maybe is like that in Haskell, where a semigroup operator on the underlying type is used to collect successes. Conor [argues convincingly](https://www.reddit.com/r/haskell/comments/30s1t2/proposal_make_semigroup_as_a_superclass_of_monoid/cpvdco1) that this is not the right notion of computation for Maybe, so I figured I'd fix it while we're young.

The old behavior is available in a named instance, and both are documented.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -21,6 +21,9 @@ New in 0.9.18:
   requiring a separate lemma.
 * Case alternatives are allowed to be impossible:
   case Vect.Nil {a=Nat} of { (x::xs) impossible ; [] => True }
+* The default Semigroup and Monoid instances for Maybe are now prioritised
+  choice, keeping the first success as Alternative does. The version that
+  collects successes is now a named instance.
 
 New in 0.9.17:
 --------------
--- a/libs/prelude/Prelude/Maybe.idr
+++ b/libs/prelude/Prelude/Maybe.idr
@@ -78,17 +78,25 @@ instance (Eq a) => Eq (Maybe a) where
   (Just _) == Nothing  = False
   (Just a) == (Just b) = a == b
 
--- Lift a semigroup into 'Maybe' forming a 'Monoid' according to
--- <http://en.wikipedia.org/wiki/Monoid>: "Any semigroup S may be
--- turned into a monoid simply by adjoining an element i not in S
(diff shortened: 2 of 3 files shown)


LeifW:
I can see having an Alternative instance on Maybe that does this. Personally I find the current behaviour quite useful; I've used it a lot in code at work (in Scala), and documented that behaviour to coworkers https://gist.github.com/LeifW/5537089
I can't think of as many uses for the proposed instance.

david-christiansen:
Nevertheless, it does fit better with the `Applicative`, `Monad`, `Alternative`, etc instances, which I think is reason enough. The other one is still available, after all - perhaps we should make a type `Option` representing optional success rather than potential failure?

milessabin:
Discussing this in https://gitter.im/non/cats the following came up:

> in Scala, an Option really just represents a result -- you'd need to talk about Lazy[_] or something to
> get at a computation.
> doesn't the whole use case fall apart with strictness?
> well -- maybe not totally, but it just seems way less compelling.
> i guess my point is that almost every time i find myself doing x orElse y i find myself realizing i wish i
> had not calculated y until i evaluated x and found it to be None.

Does the same consideration apply here?

LeifW:
The named instance doesn't seem usable for `Monoid`. If I try invoking something with a `Monoid` constraint with `@{collectJust}`, I get
```
When elaborating argument constrarg to function Main.combineThings:
        Can't unify
                Semigroup (Maybe a) (Type of collectJust)
        with
                Monoid iType (Expected type)
```
I can define one locally:
```idris
instance  [collectJust] Monoid (Maybe a) where
  neutral = Nothing
```
and then pass it to that method, but the instance that gets used for the `<+>` `Semigroup` op is this `Alternative` one.

If try to define it next to the `[collectJust]` `Semigroup` instance in Maybe.idr, it fails to compile with `Prelude.Maybe.collectJust already defined`. Should it have a different name for the subclass?

Should there be different types / typeclasses for the different behaviours? Do we really want a `Monoid (Maybe a)` instance in Prelude? Or have it as something separate for people to import or define locally?

PR #2101, opened 2015-04-02
Added informal definition of dependent types and additional material abo...

xoltar:
...ut vectors and finite sets

### DIFF
--- a/docs/tutorial/typesfuns.rst
+++ b/docs/tutorial/typesfuns.rst
@@ -243,12 +243,24 @@ So, for example, the following definitions are legal:
 Dependent Types
 ===============
 
+Dependent types are types that depend on values. It's quite possible you have never encountered such types before,
+so let's take a momement to answer the question: What does it mean for a type to depend on a value?
+First consider a type that depends on a type. ``List a`` is one such example. ``List`` is a type constructor
+that takes a type ``a`` (``a`` is a metavariable here meaning any type), and produces the type of lists that 
+contain only elements of the type provided. For example, ``List Nat`` is the type of lists that contain only ``Nat`` s. 
+
+So if we understand ``List`` as an example of a type that depends on a type,
+a type that depends on a value has a type constructor that takes a value as a parameter. Suppose we could parameterise 
+``List`` not only by the type of elements it contains, but also by the length of the list? Then we could 
+write functions that only apply to lists of three floats, for example. These functions would only accept lists that 
+had been verified (either statically, by the compiler, or dynamically, with runtime tests) to have exactly three elements. 
+That might be a good way to build a three dimensional graphics library, for example. 
+  
 Vectors
 -------
 
-A standard example of a dependent type is the type of “lists with
-length”, conventionally called vectors in the dependent type
-literature.  They are available as part of the Idris library, by
+The accepted name for "lists with length" in the dependent types community is "vectors".
+They are available as part of the Idris library, by
 importing ``Data.Vect``, or we can declare them as follows:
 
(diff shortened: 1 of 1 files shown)


jfdm:
Some of this material is already covered in the introduction. Do we *really* to repeat the informal definitions again?

edwinb:
I think we need to be a bit careful about the scope of the tutorial too - it really is just a crash course and while lots of things could use more explaining, I'm not sure the tutorial is the right place to do it. It might be better to lift this out into a more in-depth introduction to some basic dependent types covering lists, vects and fins.

xoltar:
Hmm, OK. To me, dependent types are the main thing we have to cover in the tutorial, since this is a dependently typed language. Everything else is secondary. When I first read the tutorial, this is the section where I went from saying, "yes, yes, that's fine" to "what??"

Vectors are a good example, but frankly they're the only example anyone gives of what a dependently typed language can do. Data.Fin isn't a great choice for a second example because it seems as if we're just picked a deeply strange way to implement vector indexing, and that is so distracting that it's hard to care about it being an example of dependent types. 

What I'd really like to see as a next example is Edwin's gist on safe division, though to build up to that will take some additional material. 

If we don't want that level of detail in the tutorial, I think we risk producing the Idris version of "A Gentle Introduction to Haskell," which long ago was the only way to learn Haskell, and was gentle only for someone already accustomed to Gofer, Miranda or at most SML. These days we have "Learn You a Haskell for Great Good" and similar, which everyone agrees is a better situation. 

If we want "Learn You an Idris for Great Good" (working title :-) ) to be the place where we actually talk about what dependent types are good for, and not the tutorial, then we should start writing that book immediately, because that's what people actually need. I'd love to help.

jfdm:
@xoltar In case you were not aware, and to re-enforce the comments of others: The 'Idris Tutorial' was designed as a crash course on using *Idris the programming language*, and not solely about using dependent types. Prior knowledge of programming in functional languages (and to some extent programming with dependently types languages) was assumed and the essentials were explained. Sometimes you *just* have to make some assumptions to bootstrap your project.

We have always known about the deficiencies with the tutorial i.e. its content and assumptions over prior knowledge made. We have always wanted to, but not necessarily have had the time, to construct better tutorials on using: (a) Idris, and (b) Programming with Dependent Types.

I think the ideas and suggest content on how to introduce dependently typed programming through the medium of Idris does look good, and as suggested by @edwinb should go in a separate tutorial. Given the 'new' user documentation system we have, then the 'guides' section might be the more appropriate place to put such an introduction.

xoltar:
@jfdm I wasn't suggesting the tutorial should only be about dependent types, only that if you bill your language as "a dependently typed language," you should provide a new user with some sense of your your language handles dependent types *in their first encounter with the language*, which is this tutorial. For someone with Haskell background but not previous experience with dependent types, this tutorial doesn't do that. I think that's a problem.

But I hear you when you say you don't, message understood. I'll look into making a guide.

We could also change the title of this tutorial so it's not the first thing most people click on. Something like "Idris Tutorial for Dependent Types Researchers". 

Thanks,
Bryn

LeifW:
Idris is certainly being approached by a lot of people new to dependent types. It's positioned (somewhat uniquely) as "a practical programming language with dependent types". I assume many people are coming to from languages like Haskell and Scala. FWIW, the tutorial was enough to to bootstrap me from Haskell. Reading the intro to [Ulf's Agda tutorial](http://www.cse.chalmers.se/~ulfn/papers/afp08/tutorial.pdf) previously may have helped.

PR #2102, opened 2015-04-02
Set encoding to read and write source files.

melted:
Not all systems use utf-8 by default, Windows has latin-1. This enforces utf-8 on 
source files. It's needed to make Unicode in source work on Windows.

It may seem strange to open a file and not close it, but that is how readFile works, 
this just adds the set encoding step before hGetContents. hGetContents is lazy 
and will close the file when everything has been read.

readFile source:
    readFile name   =  openFile name ReadMode >>= hGetContents

### DIFF
--- a/src/Idris/Chaser.hs
+++ b/src/Idris/Chaser.hs
@@ -15,6 +15,7 @@ import Control.Monad.State
 import Data.List
 
 import Debug.Trace
+import Util.System (readSource, writeSource)
 
 data ModuleTree = MTree { mod_path :: IFileType,
                           mod_needsRecheck :: Bool,
@@ -133,7 +134,7 @@ buildTree built fp = btree [] fp
   children lit f done = -- idrisCatch
      do exist <- runIO $ doesFileExist f
         if exist then do
-            file_in <- runIO $ readFile f
+            file_in <- runIO $ readSource f
             file <- if lit then tclift $ unlit f file_in else return file_in
             (_, _, modules, _) <- parseImports f file
             -- The chaser should never report warnings from sub-modules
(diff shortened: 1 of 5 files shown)


david-christiansen:
Great, we've been meaning to do this for a while! Thanks!

A couple comments:

 1. Shouldn't `readSource` and `writeSource` be in `Util.System` instead of `Idris.Parser`? The parser module already has all kinds of crazy extra stuff in it due to the need to read dependencies to get the syntax information, but it seems unfortunate to increase the amount of non-parsing code there.

 2. This should really be listed in the CHANGELOG, and probably also in the ReST documentation.

I don't think there's any reason to add options for other encodings, at least not until someone has a need for it. UTF8 is a pretty sane encoding to mandate.

melted:
Moving it makes sense, I'll do that.

Från: "David Christiansen" <notifications@github.com>
Skickat: ‎2015-‎04-‎02 10:29
Ämne: Re: [Idris-dev] Set encoding to read and write source files. (#2102)

Great, we've been meaning to do this for a while! Thanks!
A couple comments:
Shouldn't readSource and writeSource be in Util.System instead of Idris.Parser? The parser module already has all kinds of crazy extra stuff in it due to the need to read dependencies to get the syntax information, but it seems unfortunate to increase the amount of non-parsing code there.
This should really be listed in the CHANGELOG, and probably also in the ReST documentation.
I don't think there's any reason to add options for other encodings, at least not until someone has a need for it. UTF8 is a pretty sane encoding to mandate.

david-christiansen:
I took care of CHANGELOG. Thanks for the contribution!

Issue #1909, opened 2015-02-11
Don't sugar listy constructors if there's an incomplete term


david-christiansen:
Right now, incomplete term holes are not shown for list-like constructors: 

```
> with List []
(input):0:0:Incomplete term []
```

It should say:
```
> with List []
(input):0:0:Incomplete term Nil {a = ([__])}
```

The fix is just a matter of adding the same check for holes to the sugaring helper function in the pretty printer that is used to determine whether or not to print implicit arguments.

PR #2104, opened 2015-04-02
Clarify failing searches

### LINKED ISSUE #2091: :search, :apropos should give feedback if nothing is found
Currently, :search doesn't return anything if there is nothing that matches the required type. It would be better if the user got some kind of feedback, like a message saying "No results" or something similar.

The same applies to :apropos, although :apropos returns two empty lines if nothing is found (which, obviously, is not much better than nothing.)

japesinator:
When `:apropos` or `:search` can't find anything, an error is now shown. Fixes #2091

### DIFF
--- a/src/Idris/TypeSearch.hs
+++ b/src/Idris/TypeSearch.hs
@@ -30,7 +30,7 @@ import Idris.Core.Unify (match_unify)
 import Idris.Delaborate (delabTy)
 import Idris.Docstrings (noDocs, overview)
 import Idris.Elab.Type (elabType)
-import Idris.Output (iputStrLn, iRenderOutput, iPrintResult, iRenderResult, prettyDocumentedIst)
+import Idris.Output (iputStrLn, iRenderOutput, iPrintResult, iRenderError, iRenderResult, prettyDocumentedIst)
 import Idris.IBC
 
 import Prelude hiding (pred)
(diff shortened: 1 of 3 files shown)


src/Idris/TypeSearch.hs
```diff
@@ -55,9 +55,13 @@ searchByType pkgs pterm = do
          displayScore theScore <> char ' ' <> prettyDocumentedIst i docInfo
                 | (n, theScore) <- names']
   case idris_outputmode i of
-    RawOutput _  -> do mapM_ iRenderOutput docs
+    RawOutput _  -> do if (not (null docs))
```
david-christiansen: I think the repeated conditional here is a bit inelegant, because the False case is duplicated. Why not do:
```
if (not (null docs))
  then case idris_outputmode i of ...
  else iRenderError $ text "No results found"
```

japesinator:
That makes sense, thanks for pointing it out

david-christiansen:
That looks nice and clean - thanks a ton! Ping me if you want a good second project :-)

japesinator:
I'm absolutely interested! What would you suggest?

david-christiansen:
Sorry for the slow answer.

I'd suggest #1909: it uses some code that you've seen (the pretty-printer), it's fairly straightforward once you've understood the context, and it familiarizes you with another bit of the system.

The place to update is the pretty-printer's helper function [`slist'`](https://github.com/idris-lang/Idris-dev/blob/5d269e39d1a2894881763f755ec73b7ff03117b3/src/Idris/AbsSyntaxTree.hs#L1692-L1694). Use [`containsHole`](https://github.com/idris-lang/Idris-dev/blob/5d269e39d1a2894881763f755ec73b7ff03117b3/src/Idris/AbsSyntaxTree.hs#L1692-L1694) to determine if a subterm is a reference to an unsolved goal, in which case it should just return `Nothing`.

You'll know it works when
```
Idris> List.Nil
(input):0:0:Incomplete term []
```

says
```
Idris> List.Nil
(input):0:0:Incomplete term Nil {a = ([__])}
```
instead.

japesinator:
Thanks for the suggestion! I can totally do this

PR #2106, opened 2015-04-02
Don't require expect for tests with timeouts

melted:
Turns out there's a timeout command in core utils that is
convenient.

Fixes basic010 and reg039 on Windows.

Reason for these is that expect is a troubling dependency on Windows, the alternatives are to shell out money to ActiveState or download five year old janky versions. Neither is appealing.

### DIFF
--- a/test/basic010/run
+++ b/test/basic010/run
@@ -1,28 +1,5 @@
 #!/usr/bin/env bash
 
-# From http://unix.stackexchange.com/questions/43340/how-to-introduce-timeout-for-shell-scripting
-# Here copied from reg039.
-#
-# Executes command with a timeout
-# Params:
-#   $1 timeout in seconds
-#   $2 command
-# Returns 1 if timed out 0 otherwise
-timeout() {
-
-    time=$1
-
-    # start the command in a subshell to avoid problem with pipes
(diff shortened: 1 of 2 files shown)


Melvar:
Wait, if you don’t need to turn the whole thing into one string, is there any reason for `extraargs`? It should work right to just use `"$@"` there; won’t this introduce spurious quotes now if `timeout` is available?

david-christiansen:
I don't think that BSD has timeout - that was why I went with expect back
in the day.

melted:
@Melvar fixed that

@david-christiansen expected there was a reason, that's why I kept expect and checked for timeout

david-christiansen:
Ah, ok, I'd just read the comment on my phone when I got up for a glass of water during the night :-)

Is there a reason why the body of the else isn't indented like the `if` is?

melted:
I guess the reason is to keep the diff small for the poor reviewer. On the other hand, indenting would boost my stats. Hmmm.... :-)

david-christiansen:
I care more about clean code than clean diffs!

melted:
Of course I indented it at another computer that didn't have the other fix...

PR #2108, opened 2015-04-03
runtest.pl : fix error when there is no sandbox

Heather:
If directory doesn't exist `abs_path` just fails. This patch resolves issue for me

``` shell
basic001/../../.cabal-sandbox/bin: No such file or directory at ./runtest.pl line 13
make: *** [basic001.test] Error 2
```

### DIFF
--- a/test/runtest.pl
+++ b/test/runtest.pl
@@ -10,10 +10,11 @@ my @idrOpts    = ();
 
 sub sandbox_path {
     my ($test_dir,) = @_;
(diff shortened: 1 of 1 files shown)


jfdm:
Thanks.

Although, I think this also points out a larger issue with the testing....

1. For sandboxes, we are running the tests against the most recently installed version of idris and not the most recently built.
2. For non-sandboxed builds, does this also result in the same behaviour as above that is build against most recently installed and not most recently built...

Heather:
@jfdm I think that should be posted as new issue also.

PR #2109, opened 2015-04-03
HTML path slashes are always '/'.

melted:
Using the file system separator makes it "\" on Windows, which works, but is odd. (not to mention, makes idrisdoc006 fail)


david-christiansen:
Thanks!

PR #2110, opened 2015-04-03
Make a mechanism for os-dependent output of tests

melted:
Fixes interactive007 on Windows.

This might be slightly controversial, but there are good reasons here for the output being different. I feel it's better to make a mechanism that accommodates the (hopefully rare) cases of that rather than make spurious changes in Idris just to shut up the tests.

I don't know what perl uses for OS strings for the BSDs, but they can be added to the comment by people in the know.


david-christiansen:
Seems perfectly reasonable.

PR #2116, opened 2015-04-04
Type Signatures use IO' l where they can instead of IO

lenary:
This is just a minor cleanup so that these functions no longer rely on the C FFI, and can be used with an IO monad parameterised by whichever FFI they want.

In this case, `fork` is entirely implemented with primitives, and `while` is entirely implemented with the monad instance of `IO' l`, which doesn't use the C FFI either. `run__provider` seems just to be a reimplementation of `call__IO`.


lenary:
:smiley:

PR #2118, opened 2015-04-04
Show incomplete term holes for listy constructors

### LINKED ISSUE #1909: Don't sugar listy constructors if there's an incomplete term

japesinator:
Fixes #1909


david-christiansen:
Great!

If you're looking for more intro projects, check out the "Low-Hanging Fruit" tag on the issue tracker. If you have something bigger in mind,I'll try to find some pointers for you.

PR #2131, opened 2015-04-09
Fix default for execonly and Nix build

puffnfresh:


puffnfresh:
This should fix https://github.com/NixOS/nixpkgs/issues/7014

puffnfresh:
Confirmed that Nix is fixed by this patch. It would be great to have a Hackage release with this change so we can unbreak Nix.

PR #2138, opened 2015-04-09
Fix crashes when garbage collecting bigints

### LINKED ISSUE #2122: Bigints cause segfaults in gc
folding002 segfaults on Windows. At first I thought it was that the tail call optimization was broken. But the call stack and disassembly looked just fine (not to mention that it took a few hundred thousand recursions to appear), so I started to check for memory corruption. Because the problem occurs at the tenth collection it seemed to be something subtle. 

After some false starts I saw that it flakes out in cheney when it encounters a suspiciously small CON, so I intstrumented it to look for small allocations. Turns out that it is the bigints' fault (I'm using mini-gmp), they are allocated on the idris heap and  if a limb is zero, cheney will happily assume that it's a CON and call copy with bad consequences. And the reason it took a while to appear is of course that the big ints weren't needed until then.

I "fixed" the problem by checking if the chunk was at least sizeof(Closure) + sizeof(size_t) before switching on the type in cheney, but that seems a bit unprincipled. @edwinb Can you think of a better solution?

melted:
Invent a new Closure type for raw memory that is used by idris_alloc, add it to copy in the gc,

Takes a bit more memory, but hey, it works!

Fixes #2122, and loads of more potential problems.


melted:
The idea here is basically that the GC expects the heap to contain Closure structs, so the path of least resistance is to make a Closure struct for a blob of data, where the info field contains the size, and use that for idris_alloc. It falls out very nicely, not even idris_realloc has to be changed.

PR #2140, opened 2015-04-11
Fix Field definition and separate Lattice and VectorSpace modules from Algebra

BlackBrane:
This is a revision of my previous PR #2011 following the reorganization of Algebra.idr into the contrib package. The motivation is largely the same: Both the Lattice and VectorSpace class hierarchies are independent offshoots of Algebra that do not need to be in the same module, and I think we could benefit from this sensible organization into conceptual units. As before I've included a fix to the currently-incorrect definition of `Field`.

Notes:

* As @zraffer suggested I've opted to use propositional equality in the definition of Field, instead of `So` or the `Eq` class, so the definition now reads: `inverseM : (x : a) -> Not (x = neutral) -> a`. I think this may be the best approach, although this class has hardly been used as yet, so that could change. Views on this question are again welcomed.
* Another small refinement I've included is to move the long list of instance declarations of algebraic classes for numeric types that were previously thrown into `Data.Matrix` into their own module at `Control.Algebra.NumericInstances`. (I've found I need to use these instances a lot: They provide the only reasonable way to abstract over operations on complex number types since, for example, something like `Complex ZZ` can't be made an instance of `Num` because it doesn't support `abs`.)

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -114,6 +114,7 @@ Extra-source-files:
                        libs/contrib/Classes/*.idr
                        libs/contrib/Control/*.idr
                        libs/contrib/Control/Isomorphism/*.idr
+		       libs/contrib/Control/Algebra/*.idr
                        libs/contrib/Data/*.idr
                        libs/contrib/Decidable/*.idr
                        libs/contrib/Network/*.idr
--- a/libs/contrib/Classes/Verified.idr
+++ b/libs/contrib/Classes/Verified.idr
@@ -81,8 +81,8 @@ class (VerifiedRing a, RingWithUnity a) => VerifiedRingWithUnity a where
   total ringWithUnityIsUnityR : (r : a) -> unity <.> r = r
 
 class (VerifiedRing a, Field a) => VerifiedField a where
-  total fieldInverseIsInverseL : (l : a) -> l <.> inverseM l = unity
-  total fieldInverseIsInverseR : (r : a) -> inverseM r <.> r = unity
+  total fieldInverseIsInverseL : (l : a) -> (notId : Not (l = neutral)) -> l <.> (inverseM l notId) = unity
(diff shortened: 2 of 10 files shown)


edwinb:
Thanks for this. I don't feel qualified to check this properly before merging though - does anyone have any comments?

david-christiansen:
Sounds like we should remove `abs` from `Num` :-)

I don't know enough about the structures in question to evaluate whether these definitions fit them, but it doesn't seem to be a step backwards at all.

@markuspf, do you have any comments?

markuspf:
Well, this is a complex topic, and everyone has an opinion, so here are a couple of mine:

* I wonder whether we actually expect people to tell idris that the element they're inverting is not zero?
* I find the notion of `(<+>)` and `sum` for monoids weird, it evokes the sense of commutativity here (which I assume to not be true for Monoids in general over monoid this is a fold, over a (semi-)ring one might call that fold a sum, if one wants to)
* Things like `Complex ZZ` do my head in. Is this extension so important? I see how people would want Complex numbers in some incarnations (approximations by doubles or floats, fractions, whatever else). How often will Gaussian Integers be relevant? Even worse, I wouldn't want to have Complex (GF 2), because that is just not even an integral domain anymore
* Of course one can define a magnitude on Complex ZZ, so maybe introduce a class that has a (useful!) abs? We did this with Neg already
* Looking at the comments, I think we have to be *very* careful about overengineering what should essentially be what Num always wanted to be.

I could not spot anything inherently wrong with the axioms yet, and I think since this is an improvement it should be merged, but I also maintain that (some of) this bikeshedding should take place outside of the idris code base in a separate library. That might not be as visible as being in the main repository, but there things can be broken as we please.

david-christiansen:
I'll go ahead and merge this. But, @markuspf, I encourage you to extract the code in question to idris-algebra on idris-hackers, and iterate through a few designs there.

david-christiansen:
Oh, @markuspf, thanks for checking this out for us!

PR #2141, opened 2015-04-12
Add fibs to Stream.idr in the prelude.

markfarrell:
Found myself needing an infinite stream of Fibonacci numbers for solving a Project Euler problem. Also noticed that an infinite stream of Fibonacci numbers is an example of how to use infinite streams in the Scala standard library documentation. So, thought it would be useful to include in Stream.idr, especially since `fib n` is included in Nat.idr. 

```Idris
||| An infinite stream of Fibonacci numbers
fibs : Stream Nat
fibs = 0 :: 1 :: zipWith (+) fibs (tail fibs)
```

Signed-off-by: Mark Farrell <m4farrel@uwaterloo.ca>


markfarrell:
> fibs : Stream Nat fibs = 0 :: 1 :: zipWith (+) fibs (tail fibs)

Issue #2073, opened 2015-03-30
getLine inconsistent behaviour


JakobBruenker:
The output of the following program

```idris
main : IO ()
main = map show getLine >>= putStrLn
```
results in the following output, if the user enters `foo`:
```idris
"foo\n"
```
Which means getLine appends a '\n' to its result.

However, in the REPL, writing
```idris
:x the (IO ()) $ map show getLine >>= putStrLn
```
results in this, if the user enters `foo`:
```idris
"foo"
```
So here, getLine doesn't append a '\n'.

edwinb:
I'm not bothered which way this goes, as long as it's consistent. Any volunteers?

PR #2143, opened 2015-04-13
Fix getLine inconsistency

soimort:
This change affects the behavior of `getLine`, as it fixes the inconsistency problem in #2073.

The string buffer returned by getline() contains the trailing newline character, now it will be removed from the return value of `idris_readStr`. (analogous to Haskell `getLine`)


edwinb:
Great, thanks.

PR #2153, opened 2015-04-16
Delete obsolete tailcall test

melted:
Also update the comment.


edwinb:
To be honest, I think we should remove this test rather than make it more lenient... it's not actually testing anything that isn't covered elsewhere.

melted:
Ok, I'll remove it

melted:
Timeout in travis, seven of eight is enough...

jfdm:
In future, if Travis has timed out can you restart the affected build rather than merge the PR; a commit bit allows you to do that on Travis if you log in. Given that the timeout was on the C backend it would be better if the PRs built cleanly prior to a merge.

PR #2155, opened 2015-04-17
fix divNat segfaulting

### LINKED ISSUE #2125: divNat segfaults
Compiles and works:

```idris
main : IO ()
main = putStrLn $ show $ divNat 261971 1
```

Segfaults:

```idris
main : IO ()
main = putStrLn $ show $ divNat 261972 1
```

If it helps, I'm running on Linux 64-bit, compiling with whatever the default is for "idris t.idr -o t"

identicalsnowflake:
fixes #2125


david-christiansen:
Would you mind adding a regression test, so we make sure that the issue stays fixed?

identicalsnowflake:
I made one, but I couldn't figure out how to get it to integrate into the testing framework. I tried to just follow the pattern used in all the other tests, but I must be missing something. I'll try again this weekend and see if I can get it to work. Is there anywhere I can read about the testing framework used?

melted:
The testing framework is very simple. It's in runtest.pl if you want to have a look. What it needs is an executable file called `run`, typically a shellscript, and a file with the expected output called `expected`. `run` needs to take idris options as an argument and do the right thing with them (so different backends can run the same tests). The runtest script executes `run` and pipes the output into a file and comparses it to `expected`

In the makefile in the test directory, there is a makefile target to run just a single test: `make <testname>.test`.

david-christiansen:
If you'd like, you can post the test code here and I'll add it tomorrow
when I get a moment.

identicalsnowflake:
Ok, I think it's working now.

@melted Yeah, that's what I thought, but for some reason I was getting different results when I ran the test from when I just compiled it myself manually, even for just `main = putStrLn "hello"`. I have no idea what the problem was. It must have been a character encoding issue or something, because I just deleted the directory and recreated everything again, and suddenly it all works now.

PR #2157, opened 2015-04-18
Rename test to mark it as a regression test, and note the issue

melted:


melted:
I have no clue what's wrong here. And it's not that important anyway.

PR #2159, opened 2015-04-18
Convert Text to ByteStrings directly

melted:
Heap allocation for a trivial compilation goes from 1.9 to 1.1 GB, a decent win.


david-christiansen:
That's a great improvement! Thanks!

PR #2163, opened 2015-04-20
changed Vect.findIndex to return Fins

JakobBruenker:
`Vect.findIndex` had the type `(a -> Bool) -> Vect n a -> Maybe Nat`, but it makes a lot more sense if it return `Maybe (Fin n)`, for two reasons:
 - `Vect` uses `Fin` for indices everywhere else
 - It's a lot more convenient to convert a `Fin` to a `Nat` than the other way around. The easiest way to convert the `Nat` that `findIndex` returns to a `Fin` was to completely rewrite `findIndex`, which, of course, this Pullrequest does.

This was brought up by DanC on IRC (although in the end it turned out that they needed a different function)

Note: This will break existing code that uses this function.

`elemIndexBy` and `elemIndex` were also changed, for the same reasons and because their implementation uses `findIndex`.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -28,6 +28,8 @@ New in 0.9.18:
   choice, keeping the first success as Alternative does. The version that
   collects successes is now a named instance.
 * :exec REPL command now takes an optional expression to compile and run/show
+* The return types of `Vect.findIndex`, `Vect.elemIndex` and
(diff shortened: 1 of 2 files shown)


david-christiansen:
This seems perfectly reasonable. Thanks!

Would you please add a note to the CHANGELOG, because this is a breaking change to code in `base`?

JakobBruenker:
Done.

david-christiansen:
Thanks!

PR #2165, opened 2015-04-20
Add common functions for running State monad

icasperzen:
The `Control.Monad.State` module exports a type alias `State s a` for `StateT s Identity a`, but requires the user to manually unwrap the transformer stack. This requires the user to also import `Control.Monad.Identity`. So I've re-exported the Identity module from and added `runState`, `evalState`, and `execState` to the State module.

### DIFF
--- a/libs/base/Control/Monad/State.idr
+++ b/libs/base/Control/Monad/State.idr
@@ -1,6 +1,6 @@
 module Control.Monad.State
 
-import Control.Monad.Identity
+import public Control.Monad.Identity
 import Control.Monad.Trans
 
 %access public
@@ -55,3 +55,15 @@ gets f = do s <- get
 ||| The State monad. See the MonadState class
 State : Type -> Type -> Type
 State s a = StateT s Identity a
+
+||| Unwrap a State monad computation.
+runState : State s a -> s -> (a,s)
+runState m = runIdentity . runStateT m
+
+||| Unwrap a State monad computation, but discard the final state.
+evalState : State s a -> s -> a
+evalState m = fst . runState m
+
+||| Unwrap a State monad computation, but discard the resulting value.
+execState : State s a -> s -> s
+execState m = snd . runState m


david-christiansen:
Great, thanks! When Travis is done, I'll merge it.

Melvar:
Historical note: the `Identity` there used to be `id` and thus the `StateT`-related functions did the right thing already. Now however the necessary instances for `id` can no longer be written because of restrictions on what can have instances.

david-christiansen:
I imagine that we'll have a few of these things as fallout from the injectivity restriction.

eldargab:
Since idris supports overloaded names, probably `run`, `eval`, `exec` are better than `runState`, `evalState`, `execState`.

icasperzen:
That didn't even occur to me, but it seems like a good idea. Should I go ahead and make the changes?

david-christiansen:
I don't think there's any real win from overloading here. Overloading is nice when you're wanting to have similar APIs but can't instantiate a type class (e.g. Functor on Set, or using do-notation or idiom brackets for a DSL with its own closed type universe). It also has downsides, like making type checking slower and error messages worse. Here, I don't see that `run` instead of `runState` will make any code easier to read, and I do think we should maintain Haskell API compatibility when there's not a good reason to change it.

I'm open to the change in principle, but I'd like to a reason other than "we can overload, thus we should overload".

jfdm:
I agree with @david-christiansen we should overload just because we can. I have run into trouble in the past with the elaborator. I experienced slow compile times with large memory overheads trying to work out with version of a list like data structure I was using. While you can give the elaborator hints, I think one should avoid that.

Melvar:
Well, I see it this way: If you include the type or module name in the function name like `runState`, it’s always going to be there, no way around it. If instead you rely on the namespace to make it unique, it’s perfectly possible to write `State.run` at one extra character, but also to use `run` if it’s the only one in scope, or to alias the namespace as desired and write perhaps `S.run`. Putting such a thing right into the function name is *inflexible*, from this point of view, because one doesn’t have these choices.

PR #2048, opened 2015-03-31
Append Char to String

cchantep:
It would be nice to have simpler way to append a `Char` to a `String`.

```idris
||| Adds a character to the end of the specified string.
|||
||| ```idris example
||| strAppend 'A' 'B'
||| ```
||| ```idris example
||| strAppend "" 'A'
||| ```
strAppend : String -> Char -> String
strAppend s c = s ++ (strCons 'c' "")

||| Alias of strAppend
(+) : String -> Char -> String
(+) = strAppend
```

PR #2166, opened 2015-04-20
Don't append a literal 'c'

melted:
Fix my note in #2048


david-christiansen:
This is clearly a proof that two heads are better than one. Thanks for catching my review fail, @melted, and I'm sorry I misunderstood your comment on the last PR!

PR #2168, opened 2015-04-21
Experimental support for more GHC versions in Travis

david-christiansen:
This is to see how Travis works with multi-ghc-haskell and Idris. Please don't merge until I get a chance to look at it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,31 +1,68 @@
-language: haskell
-ghc:
-  # Idris won't build on 7.4 and earlier due to dependency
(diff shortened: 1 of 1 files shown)


david-christiansen:
If anyone out there has experience with multi-ghc-travis, I'd love hints on how to make the 7.10 builds fast enough to actually complete.

david-christiansen:
I hear that 7.10.2 will be faster, so perhaps that will prevent Travis timeouts.

Issue #1826, opened 2015-01-03
There's too much stuff in IBC files


edwinb:
Currently, IBC files contain everything that is generated by elaboration - that's all the typechecked data types and functions, as well as recording the original definitions, metavariables, what's on each line for interactive editing, documentation, etc.

All of this stuff is useful to have available, but the problem is that when it's not used (which is most of the time) it really hurts both in memory usage and loading time. Currently, even when elaborating from scratch, rebuilding an IBC is the biggest bottleneck in elaboration (that will happen to some extent of course since importing will load the IBC).

I'd suggest splitting it into two files - one containing the things which are absolutely necessary when importing (that's types, definitions and other bits and pieces to help with elaboration) which is always loaded on 'import', and one containing the things that are useful at the REPL, which will only be loaded on demand.

My recent adventures in profiling suggest this will be a big win. We really are eating quite a lot of memory at the minute and I think a lot of it is unnecessary!

(While we're at it, it might be good to put generated files in a different directory rather than dumping them with the source. That's probably a separate issue though.)

edwinb:
I should add that I also suspect there to be huge loss of sharing when writing out IBC files. Terms by their dependently typed nature have lots of shared subterms, and no attempt is made to deal with this when writing out, so when read back in, sharing has gone.

I haven't measured either of these effects precisely, by the way, so take with a pinch of salt. All I know is that 'get' in Binary/IBC is allocating all the memory. That is not at all surprising, of course, but it's still worth looking for places to reduce it.

edwinb:
Also, prodding with different profiling options, I see most memory is eaten by Text and Bytestrings (most likely in variable names but I haven't poked that deeply). I suspect a massive loss of sharing here so we need to work harder to make sure strings are shared.

By the way, I'm just leaving notes here - I don't have time to poke further at the minute but I don't want to forget what I've found so far... if anyone finds anything from some profiling that's eating too much please also make a note here.

melted:
It should be possible to improve the performance without discarding data from IBC files. The expanded IBC files in libs are 24,2 MB (I shut off the compression to get a sense of how much is in there), compiling a trivial Idris program churns through 1,9 GB, nearly all of it in loadModule, A heap profile shows heap objects from Core/Binary.hs and IBC.hs that are hundreds of kb, some over a meg.

It's a bit hard to pin down exactly what is churning memory with names like these in the profile:
```
[...]

PR #2170, opened 2015-04-21
Use zip as IBC container

melted:
This starts off what I wrote about in #1826.

This commit in itself doesn't lead to better performance, rather the IBC files grow slightly as they contain the same stuff, now just split up on entries in a zip container. However, this is needed to begin work on improving the module load times. The two ways to attack the loading times is to either load less, much of the stuff can be loaded lazily on need, or use a smarter file format, so we for example don't blast many copies of the same string to disk. A random access format is necessary to do the first, and it will be helpful with the second.

Other advantages with using a container format: data don't have to have a Binary instance to go down into the container, things that are better as text don't have to be shoehorned; using folders we could in the future amalgate several modules into one ibc-file; and it is easily inspectable with standard apps.

### DIFF
--- a/src/Idris/IBC.hs
+++ b/src/Idris/IBC.hs
@@ -35,10 +35,11 @@ import qualified Control.Monad.State.Strict as ST
 import System.FilePath
 import System.Directory
 import Codec.Compression.Zlib (compress)
+import Codec.Archive.Zip
 import Util.Zlib (decompressEither)
 
 ibcVersion :: Word8
-ibcVersion = 104
+ibcVersion = 105
 
 data IBCFile = IBCFile { ver :: Word8,
                          sourcefile :: FilePath,
@@ -103,9 +104,12 @@ loadIBC reexport fp
                                 Just p -> not p && reexport
                 when redo $
                   do iLOG $ "Loading ibc " ++ fp ++ " " ++ show reexport
-                     ibcf <- runIO $ (bdecode fp :: IO IBCFile)
-                     process reexport ibcf fp
-                     addImported reexport fp
+                     archiveFile <- runIO $ B.readFile fp
+                     case toArchiveOrFail archiveFile of
+                        Left _ -> ifail $ fp ++ " isn't loadable, it may have an old ibc format.\n"
+                                          ++ "Please clean and rebuild it."
+                        Right archive -> do process reexport archive fp
+                                            addImported reexport fp
 
 -- | Load an entire package from its index file
 loadPkgIndex :: String -> Idris ()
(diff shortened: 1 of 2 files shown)


src/Idris/IBC.hs
```diff
@@ -240,65 +289,68 @@ ibc i (IBCUsage n) f = return f { ibc_usage = n : ibc_usage f }
 ibc i (IBCExport n) f = return f { ibc_exports = n : ibc_exports f }
 ibc i (IBCAutoHint n h) f = return f { ibc_autohints = (n, h) : ibc_autohints f }
 
+snarf :: (Binary b, NFData b) => FilePath -> Archive -> Idris b
```
david-christiansen: This function name doesn't say much to me... is there a reason why it's not called something like `extractIBCPart`?

david-christiansen:
This seems like a generally good idea to me. Hopefully, it will also make it more feasible to have things like `:search` and `:browse` performantly search non-loaded IBC files.

jfdm:
I agree with @david-christiansen in thinking that this is a good idea. However, it would be nice if there were more code level documentation, and better names than snarf used. Also I think it might be a prime opportunity to think about reference docs to describe the general idea behind think file format.

Also I see a new dependencies have been been added. How are the bounds wrt LTS Haskell....can they be satisfied by them, or do they need to be adjusted....

david-christiansen:
Well, the docs are no worse than they were before the PR. It's nice to improve them, but it's still a valuable contribution if it doesn't. Also, I'd prefer not to document the file format, because I think that the right way to process a .ibc file is to use Idris as a library, not to roll one's own.

WRT LTS Haskell: if we want to make that a requirement, then we really need to get it into Travis, don't you think?

jfdm:
> Well, the docs are no worse than they were before the PR. It's nice to improve them, but it's still a valuable contribution if it doesn't. 

I agree.

> Also, I'd prefer not to document the file format, because I think that the right way to process a .ibc file is to use Idris as a library, not to roll one's own.

I agree, but I was thinking more in terms of maintainability of the Idris library, and making sure that design decisions are accessible from outside the code and the people who wrote it.

> WRT LTS Haskell: if we want to make that a requirement, then we really need to get it into Travis, don't you think?

Obviously the LTS Haskell thing shouldn't detract from this PR. I am just trying to ensure that when we do eventually switch over to LTS Haskell it should be painless as possible.

melted:
Awww, snarf is in the dictionary and all, http://dictionary.reference.com/browse/snarf

I suppose I should change it to getEntry or something.

It's too early for documentation, this is just the start and it will likely change a lot as I try to squeeze out performance. Not even the exports are safe if I introduce load on demand. 

As for LTS, if they don't have what we need, what to do? It would be possible to just snarf zip-archive into idris as it's one smallish file. Syb wouldn't be a problem, it's a classic (actually I don't use it more than the import here, I should probably remove it until I start to use it)

Skickat: ‎2015-‎04-‎22 08:38
Ämne: Re: [Idris-dev] Use zip as IBC container (#2170)

I agree with @david-christiansen in thinking that this is a good idea. However, it would be nice if there were more code level documentation, and better names than snarf used. Also I think it might be a prime opportunity to think about reference docs to describe the general idea behind think file format.
Also I see a new dependencies have been been added. How are the bounds wrt LTS Haskell....can they be satisfied by them, or do they need to be adjusted....

jfdm:
> Awww, snarf is in the dictionary and all, http://dictionary.reference.com/browse/snarf
> I suppose I should change it to getEntry or something.

I know snarf is in the dictionary but it would be nice to use more self explanatory words, and snarf is not necessarily that widely known.

> It's too early for documentation, this is just the start and it will likely change a lot as I try to squeeze out performance. Not even the exports are safe if I introduce load on demand.

Very true, but it doesn't mean we shouldn't think about it.

> As for LTS, if they don't have what we need, what to do? It would be possible to just snarf zip-archive into idris as it's one smallish file. Syb wouldn't be a problem, it's a classic (actually I don't use it more than the import here, I should probably remove it until I start to use it)

@melted TBH forget about the LTS Haskell needs. LTS Haskell has all of hackage but with a subset pinned to versions that are known to work together. The question really is have the bounds you set for the new packages conform with the ones that are in Stackage..

david-christiansen:
@jfdm until we get an automated check of package bounds against Stackage, I think you should be the one who checks that rather than having everyone check it themselves - I believe it's more efficient that way, because you'll know where to look at all.

@melted Is there a reason you prefer SYB to Uniplate? Lots of Idris already uses Uniplate, and I think the codebase is more understandable if we stick to one generic traversals library.

src/Idris/IBC.hs
```diff
@@ -113,15 +115,60 @@ loadPkgIndex pkg = do ddir <- runIO $ getIdrisLibDir
...
-bencode f d = B.writeFile f (compress (encode d))
 
-bdecode :: Binary b => FilePath -> IO b
-bdecode f = do d' <- B.readFile f
-               either
-                 (\(_, e) -> error $ "Invalid / corrupted zip format on " ++ show f ++ ": " ++ e)
-                 (return . decode)
-                 (decompressEither d')
+entries :: IBCFile -> Idris [Entry]
+entries i = do
```
david-christiansen: Why does this function need to be in the Idris monad? It looks pure to me.

Also, I think it would be cleaner to drop the "do" and just have the expression in the body, eg

```
entries :: IBCFile -> [Entry]
entries = [ toEntry....
```
or
```
entries :: IBCFile -> Idris [Entry]
entries = return [ toEntry....
```
for the monadic version.

david-christiansen:
@melted Sorry to be such a style stickler here, but I do think that keeping things clean and readable is important.

melted:
I used to set the time to something there, but I decided that nobody cares what the entries inside got as timestamp.  I can remove the impurity, any dirty transforms I will do in ibc anyway.

As for syb. The point of using Data.Generics is that one can do data foo ... derive (Generic) and instance Binary foo and don't have to have that ugly Binary boilerplate. I intend to use it for any new sharing-friendly data structures that will go in there.

david-christiansen:
Sounds good to me! Getting rid of the ugly old code seems to be a good idea too, when it can be replaced by generics and deriving.

david-christiansen:
Looks good now!

PR #2173, opened 2015-04-22
add intersect / intersectBy to Data.List

Heather:
```idris
syntax [a] "∩" [b] = intersect a b

lst1 : List Int
lst1 = [1, 2, 3]

lst2 : List Int
lst2 = [2]

putStrLn $ "intersect: " ++ (show (lst1 ∩ lst2))
```

``` shell
intersect: [2]
```

### DIFF
--- a/libs/base/Data/List.idr
+++ b/libs/base/Data/List.idr
@@ -37,15 +37,14 @@ isElem x (y :: xs) with (decEq x y)
         mkNo f g Here = f Refl
         mkNo f g (There x) = g x
 
+||| The intersectBy function returns the intersect of two lists by user-supplied equality predicate.
 intersectBy : (a -> a -> Bool) -> List a -> List a -> List a
-intersectBy _  [] _     =  []
-intersectBy _  _  []    =  []
 intersectBy eq xs ys    =  [x | x <- xs, any (eq x) ys]
 
-||| The intersect ∩ function returns the list intersect of the two lists. For example,
+||| Compute the intersection of two lists according to their `Eq` instance.
 |||
 ||| ```idris example
-||| intersect [1,2,3,4] [2,4,6,8]
+||| intersect [1, 2, 3, 4] [2, 4, 6, 8]
 ||| ```
 |||
 intersect : (Eq a) => List a -> List a -> List a
--- a/libs/base/Data/List.idr
+++ b/libs/base/Data/List.idr
@@ -39,7 +39,7 @@ isElem x (y :: xs) with (decEq x y)
 
 ||| The intersectBy function returns the intersect of two lists by user-supplied equality predicate.
 intersectBy : (a -> a -> Bool) -> List a -> List a -> List a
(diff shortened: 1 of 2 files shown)


libs/prelude/Prelude/List.idr
```diff
@@ -537,6 +537,8 @@ delete = deleteBy (==)
 (\\) =  foldl (flip delete)
 
 unionBy : (a -> a -> Bool) -> List a -> List a -> List a
+unionBy _  [] _  =  []
```
david-christiansen: What is the purpose of these extra cases?

libs/base/Data/List.idr
```diff
@@ -36,3 +36,17 @@ isElem x (y :: xs) with (decEq x y)
                Elem x' (y' :: xs') -> Void
         mkNo f g Here = f Refl
         mkNo f g (There x) = g x
+
+intersectBy : (a -> a -> Bool) -> List a -> List a -> List a
```
david-christiansen: Why doesn't intersectBy get documentation? And why are there specific cases for []?

libs/prelude/Prelude/List.idr
Heather: @david-christiansen faster solution for empty list? is it Haskell only way?

libs/base/Data/List.idr
```diff
@@ -36,3 +36,17 @@ isElem x (y :: xs) with (decEq x y)
...
+intersectBy _  [] _     =  []
+intersectBy _  _  []    =  []
+intersectBy eq xs ys    =  [x | x <- xs, any (eq x) ys]
+
+||| The intersect ∩ function returns the list intersect of the two lists. For example,
```
david-christiansen: I'd phrase the docstring as more like:
````
||| Compute the intersection of two lists according to their `Eq` instance.
|||
||| ```idris example
||| intersect [1, 2, 3, 4] [2, 4, 6, 8]
||| ```
````

This is because the first paragraph of the docstring is shown as an overview in various contexts (type search, IDE tooltips, etc), so it should make sense on its own. Also, this docstring explains the difference from `intersectBy`.

libs/base/Data/List.idr
Heather: @david-christiansen I followed `unionBy` which has no documentation also, but I can add it to both, `[]` cases reason is same as for `unionBy`, it's more interesting for me why I should not add it.

david-christiansen:
Thanks! I put in a couple minor questions and docstring style requests - would you mind taking a look?

Heather:
@david-christiansen sure, I will 
 - add documentation to intersectBy and unionBY
 - change documentation for intersect to suggested

But I don't understand the cases with empty List `[]`

Heather:
@david-christiansen done

Heather:
Failed to install tagsoup-0.13.3

Something wrong with Travis?

david-christiansen:
The failing builds have been restarted. Should be OK soon.

PR #2175, opened 2015-04-22
Latest metaprogramming updates

david-christiansen:
This branch contains my latest metaprogramming updates.

Summary:

 * Lookup of definitions of datatypes from global context
 * Reflection for special names
 * Quotation syntax for names
 * Various bug fixes along the way


Melvar:
I was wondering why the reflected `TTName` didn’t have `SN`.

david-christiansen:
I think only because whoever first implemented it never needed it :-)

This will, BTW, make it trivial to get ahold of instance constructors. Just sayin'.

PR #2180, opened 2015-04-23
Corrected the definition of Control.Algebra.Field.

morganthomas:
If a is a Field, inverseM is not a function a -> a, because it's not defined for neutral. (The additive identity doesn't have a multiplicative inverse in a field.) I made the type of inverseM a dependent type which needs to know that its argument is nonzero.

I also added functions for subtraction and division in fields, and (formally incorrect) functions for mutiplicative inverse and field division which return Maybe values.

This change caused a type error which I don't understand in compiling VerifiedField, so for now I've commented that out. Help with correcting this error would certainly be appreciated!

### DIFF
--- a/libs/contrib/Classes/Verified.idr
+++ b/libs/contrib/Classes/Verified.idr
@@ -8,6 +8,12 @@ import Control.Algebra
 -- implementations, then some of them can move back to base (or even
 -- prelude, in the cases of Functor, Applicative, Monad, Semigroup,
 -- and Monoid).
+--
+-- N.B.: One thing to note here is that most numeric classes aren't
+-- really rings, fields, etc. E.g., Int is not a ring because of integer
+-- overflow. Float and Double are not fields, for essentially the same reason.
+-- One can't hope to prove something that isn't true! On the other hand,
+-- Integer probably really is a ring.
 
 class Functor f => VerifiedFunctor (f : Type -> Type) where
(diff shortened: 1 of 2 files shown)


markuspf:
This is related to #2140, in which @BlackBrane did something similar. I *still* think one should spin these "advanced" structures out into idris-algebra, try out there how to fix them, and then point people towards that. Be it as it may, I will have a look at this shortly.

morganthomas:
@markuspf, thanks for pointing out this other pull request. @BlackBrane's modification to Field is more or less identical to mine. The principal difference is that mine includes in the Field class a proof that one is not zero. There's an obvious argument that this should belong to VerifiedField. My reason for putting it in Field is that it's required to make -1 well-typed. The other thing that's different (with respect to Field) between our PRs is the additional operations I defined on fields. 

What puzzles me is that @BlackBrane apparently got their (essentially identical) modification to VerifiedField to type-check, whereas I got an error I didn't understand. I'd like to know what's going on with that.

morganthomas:
Ah, so here's an error in what I just said, which also points out an error in my PR. -1 is, obviously, the additive inverse of 1, not the multiplicative inverse. This means that my definition of negativeOne is incorrect, and that there's no need to include a proof that one is not zero in the Field class. I'm going to close this PR for that reason, and my apologies for the incorrect code.

PR #2189, opened 2015-04-24
IBC tweaks

melted:
Already at version 106 means that it's a good idea to give the version field some more bits.

### DIFF
--- a/src/Idris/IBC.hs
+++ b/src/Idris/IBC.hs
@@ -22,6 +22,7 @@ import Paths_idris
 import qualified Cheapskate.Types as CT
 
 import Data.Binary
+import Data.Functor
 import Data.Vector.Binary
 import Data.List as L
 import Data.Maybe (catMaybes)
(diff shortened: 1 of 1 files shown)


src/Idris/IBC.hs
```diff
@@ -168,9 +168,8 @@ entries i = catMaybes [Just $ toEntry "ver" 0 (encode $ ver i),
                        makeEntry "ibc_patdefs"  (ibc_patdefs i),
                        makeEntry "ibc_postulates"  (ibc_postulates i),
                        makeEntry "ibc_externs"  (ibc_externs i),
-                       case ibc_parsedSpan i of
-                            Nothing -> Nothing
-                            Just s -> Just $ toEntry "ibc_parsedSpan" 0 (encode s),
+                       ibc_parsedSpan i
+                            >>= \s -> Just $ toEntry "ibc_parsedSpan" 0 (encode s),
```
david-christiansen: Wouldn't this be simpler as `fmap (toEntry "ibc_parsedSpan" 0 . encode) (ibc_parsedSpan i)` or  `toEntry "ibc_parsedSpan" 0 . encode <$> ibc_parsedSpan i`?

melted:
Sounds reasonable

PR #2190, opened 2015-04-24
Idris anticlassical

treeowl:
Add @david-christiansen's translation of Chung-Kil Hur's Agda proof that type constructor injectivity is incompatible with LEM.

### DIFF
--- a/libs/contrib/Decidable/ExcludedMiddleInconsistent.idr
+++ b/libs/contrib/Decidable/ExcludedMiddleInconsistent.idr
@@ -1,0 +1,72 @@
+||| A port of Chung-Kil Hur's Agda proof that type constructor
+||| injectivity is incompatible with LEM.
+|||
+||| See https://lists.chalmers.se/pipermail/agda/2010/001526.html
+||| and its follow-ups for a good discussion.
+module AntiClassical
+
+%default total
+
+||| Law of the excluded middle, with an appropriately classical name
+postulate TertiumNonDatur : (a : Type) -> Dec a
+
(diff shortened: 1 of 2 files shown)


david-christiansen:
I don't really think that this is appropriate for the libraries shipped with Idris. It is fun as a curiosity, and if it can derive a general inconsistency then it should perhaps be a regression test at that time, but its utility for programming is nil as far as I can see.

treeowl:
That's fair. However, I wonder if there is a place where such things can be put that will be readily accessible; it seems like useful documentation.

jfdm:
One option might be to add it as a guide on the user docs. There is a potential [tutorial series there](http://idris.readthedocs.org/en/latest/guides/index.html#guides-index).

david-christiansen:
Honestly, I don't think it fits in the user docs either. This is blog post worthy, perhaps.

david-christiansen:
I'll go ahead and close this. I don't really think a postulate of LEM makes any sense in the library.

treeowl:
@david-christiansen, huh? The postulate was in your original code, not the code in the pull request I certainly wouldn't want a bad postulate in the library! You may be right that the ultimate theorem is insufficiently useful, but on the other hand, there may be enough other things that can prove LEM that it could be good to have its refutation available. I don't really know.

david-christiansen:
Ah, yeah, the second commit changed it into a `parameters` block. Sorry!

In any case, I still don't think we should have this kind of "meta"-principle here. I still can't think of any actual program that would use it, so until one shows up, I think we should keep it out.

PR #2191, opened 2015-04-25
Permit use of spaces in paths when entered from REPL

### LINKED ISSUE #1320: :l and :lto doesn't like paths with spaces
Wrapping the path with quotes doesn't help
Neither does escaping the space with a \

I noticed this when trying to use idris-vim

melted:
Fixes #1320 :l and :lto doesn't like paths with spaces.

Works well with the paths given by Haskeline tab completion.

### DIFF
--- a/src/Idris/Chaser.hs
+++ b/src/Idris/Chaser.hs
@@ -65,6 +65,16 @@ getModuleFiles ts = nub $ execState (modList ts) [] where
                                   then getSrc x : updateToSrc path xs
                                   else x : updateToSrc path xs
 
+-- Strip quotes and the backslash escapes that Haskeline adds
+extractFileName :: String -> String
+extractFileName ('"':xs) = takeWhile (/= '"') xs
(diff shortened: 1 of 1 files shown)


david-christiansen:
Great! That one's been sitting around for a while!

melted:
Needs another tweak though, files given as args are handled correctly with a backslash, but not with quotes without a backslash, guess the quotes are stripped before getting into the program. It should be easy to just add a backslash when processing the args.

I guess there is no reason not to handle single quotes also.

melted:
Now I wonder why it is terminating on space at all when everything that is fed into there should be a file path, if some command takes a filepath and something after, shouldn't it need to parse that before anyway? And then why would it feed the unprocessed line to buildtree? Stripping the backslah is sound as haskeline adds it.  But rather than adding backslashes to command line args, I think I'd rather remove the space termination and fix any breakage that occurs in the callers.

melted:
Seems to work well without breaking on space.

PR #2193, opened 2015-04-26
Check for x86 before using SSE2.

melted:
It's not a valid flag on other archs like ARM.

Also allow MACHINE to be imported from the environment.

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -2,7 +2,7 @@ CC              ?=cc
 CABAL           :=cabal
 # IDRIS_ENABLE_STATS should not be set in final release
 # Any flags defined here which alter the RTS API must also be added to src/IRTS/CodegenC.hs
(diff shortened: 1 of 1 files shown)


edwinb:
I don't think we need this flag at all any more do we? It was there for the bit vectors but I simplified the primitives - bit vectors should be added as external primitives to specific back ends if they are needed now.

markuspf:
-msse and -msse2 is automatically active for x86_64, i.e. the *vast* majority of current users, there are some weird configurations on MacOS that compile to i386 for incomprehensible reasons, but I don't know whether it's worth setting -msse2 for those people (it'll do some good for the minigmp code).

melted:
I built a 32-bit binary for windows last release, there it would make a difference to what is built. Though I doubt the old FPU instructions would be much of a drag compared to other inefficiencies.

LeifW:
People often use i386 on server VMs to save memory.

melted:
This patch is insufficient anyhow, -msse2 is hardcoded in CodegenC.hs too.

PR #2200, opened 2015-04-28
add Data.CoList with unfoldr to contrib

Heather:
``` idris
putStrLn $ "unfoldr: " ++ (show
      ( unfoldr
          (λ b → if b ≡ 0 then Nothing else Just (b, b-1)) 10
      )
)

putStrLn $ "unfoldr: " ++ (show
      ( unfoldr
           (λ b → if b ≡ 0 then Nothing else Just (b, b-1)) 30
      )
)
```

``` idris
unfoldr: [10, 9, 8, 7, 6, 5, 4, 3, 2, 1]
unfoldr: [30, 29, 28, 27, 26, 25, 24, 23, 22, 21, 20, 19, 18, 17, 16, 15, 14, 13, 12, 11, ...]
```

### DIFF
--- a/libs/contrib/Data/CoList.idr
+++ b/libs/contrib/Data/CoList.idr
@@ -1,7 +1,7 @@
 module Data.CoList
 
 %access public
--- %default total
+%default total
 
 ||| Idris will know that it always can produce a new element in finite time
 codata CoList : Type -> Type where
--- a/libs/contrib/Data/CoList.idr
+++ b/libs/contrib/Data/CoList.idr
@@ -7,8 +7,6 @@ codata CoList : Type -> Type where
   Nil : CoList a
   (::) : a -> CoList a -> CoList a
 
-infixr 7 ++
-
 ||| Append two CoLists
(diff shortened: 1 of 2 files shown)


libs/contrib/Data/CoList.idr
```diff
@@ -0,0 +1,32 @@
+module Data.CoList
+
+%access public
+
+||| Idris will know that it always can produce a new element in finite time
+codata CoList : Type -> Type
+  where {
```
david-christiansen: This doesn't match the style used in the rest of the library. Why not

```
codata CoList : Type -> Type where
  Nil : CoList a
  (::) : a -> CoList a -> CoList a
```
?

Heather: good suggestion

david-christiansen:
It might also be nice to have things like Functor and Show, but that can wait for future contributions as well.

Heather:
@david-christiansen agreed, not sure if that is possible to make total show

david-christiansen:
Lookin' good now!

PR #2202, opened 2015-04-28
Add Data.List.ContainsKey proof for association lists

icasperzen:
I'm not sure if this could have been implemented more elegantly (using `map fst` and `Elem` somehow?), but this is a proof construct that I've been using a lot lately.

Should Data.Vect have something similar?

### DIFF
--- a/libs/base/Data/List.idr
+++ b/libs/base/Data/List.idr
@@ -37,6 +37,37 @@ isElem x (y :: xs) with (decEq x y)
         mkNo f g Here = f Refl
         mkNo f g (There x) = g x
 
+namespace ContainsKey
+
+  ||| A proof that an associative list contains a pair with given key.
+  data ContainsKey : List (k,v) -> k -> Type where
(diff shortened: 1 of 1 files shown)


david-christiansen:
Why not something like `Data.Vect.Quantifiers.Any`, but on lists?
```
using (a : Type, p : a -> Type)
  data Any : (a -> Type) -> List a -> Type where
    Here : p x -> Any p (x :: xs)
    There : Any p xs -> Any p (x::xs)

ContainsKey : List (k, v) -> k -> Type
ContainsKey xs key = Any match xs
  where match : (k, v) -> Type
             match (key', x) = key' = key
```

Is type checking or implicit resolution easier with the version in this PR?

icasperzen:
Thanks! I didn't know about `Data.Vect.Quantifiers.Any`. I'll write it up a version for `List`s and rewrite `Elem` in terms of it, too. Probably not before tonight, though.

david-christiansen:
No worries! I'll close this PR then, in anticipation of the next one.

PR #2205, opened 2015-04-28
Relations

treeowl:
Add basic  definitions and theorems relating to binary relations to `contrib`.  The names, module locations, and finer points of API are open to discussion. Attempts to use more implicits, which would make things less verbose in some cases, led me into horrible fights with the compiler; perhaps someone else will have an idea how to make it work. I did not look at too much of the Agda library for this (because I wanted to see how far I could get on my own), but it would be worth checking that I don't trample *too* hard on its terminology, in case we want to translate some of it.


melted:
The "rc.." and "tc..." names looks very Hungarian. I'm not keen on stuffing namespacing into the function names like that in such a cryptic way,

treeowl:
@melted, I'm currently fixing some things up to match Agda's approach (because things like antisymmetry and, to a lesser extent, irreflexivity in the classical sense, are of relatively limited value). When I finish that, if I do, I will attempt to learn how to use namespaces.

david-christiansen:
I agree with @melted about the names. Also, Travis is failing.

melted:
The Travis failures is because the files haven't been added to idris.cabal, so they aren't in the source package.

libs/contrib/Control/Relations/Basics.idr
```diff
@@ -0,0 +1,298 @@
+||| General definitions and theorems about relations
+module Relations.Basics
```
melted: Should be Control.Relation.Basics

libs/contrib/Control/Relations/Orders.idr
```diff
@@ -0,0 +1,128 @@
+||| General definitions and theorems about preorders, orders, etc.
+module Relations.Orders
+import Basics
```
melted: Should import Control.Relation Basics

libs/contrib/Control/Relations/ReflexiveClosure.idr
```diff
@@ -0,0 +1,63 @@
+||| General definitions and theorems about
+||| reflexive closure.
+module Relations.ReflexiveClosure
+import Basics
```
melted: Same here

libs/contrib/Control/Relations/TransitiveClosure.idr
```diff
@@ -0,0 +1,76 @@
+||| General definitions and theorems about relations
+module Relations.TransitiveClosure
+import Basics
```
melted: and here

dckc:
Yeah; my fix to Control.Relations.Basics was just one example of things that need to be fully qualified. I guess I should have done the rest.

edwinb:
I've restarted travis on this because the failure didn't look related to the contents of the patch and I wonder if recent changes have been interfering somewhat.

treeowl:
I think there are still a few things to fix on my side, and also some
restructuring of the modules is needed before this can go in.

edwinb:
I'm going to close this since there hasn't been movement for a while and much has changed since this was originally submitted. Please feel free to reopen if/when the issues are resolved though, since these do look useful.

PR #2208, opened 2015-04-29
Set colour off by default on Windows

melted:
The ANSI lib doesn't emit the commands on Windows anyway. And the coloured prompt
printing adds a couple of ugly chars that gets printed on a windows console.


david-christiansen:
Good idea. One of these days I'll get annotated-wl-pprint updated with `displayM ::  Monad m => SimpleDoc a -> (a -> m ()) -> (String -> m ()) -> (a -> m ()) -> m ()` that can be used to display colors on Windows.

PR #2216, opened 2015-04-30
Fix formatting in docs/tutorial/classes.rst

nathan:


jfdm:
@nathan thanks for spotting the many typo's in the tutorial. It can be hard spotting them if you work close to the source. One thing that might constitute a better workflow than submitting lots of individual PRs, is to hold off for a bit and submit a bigger PR that contains more fixes.

nathan:
Oh, sorry about that—sometimes I hear multiple PRs are better in case the owners don't like one of the changes, but I'll submit fixes in a single PR from now on.

jfdm:
Do not worry, about. We are generally quite flexible about these things, and do not mind most. Personally, if it is a whole host of typo's then just a single PR is more appropriate. If it is a whole suite of features, on the other hand, then it would be better to do them as separate PRs. Generally, do what is suitable for the thing you are submitting a PR on.

PR #2218, opened 2015-04-30
icon for easy donations with primecoins

Heather:
People donate primecoins ( https://github.com/primecoin/primecoin ) to projects. When someone's commit is accepted to the project repository, those coins automatically transferred to the author.

(same as tip4commit but I can't automatically register project there anymore)

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 [![Build Status](https://travis-ci.org/idris-lang/Idris-dev.svg?branch=master)](https://travis-ci.org/idris-lang/Idris-dev)
 [![Documentation Status](https://readthedocs.org/projects/idris/badge/?version=latest)](https://readthedocs.org/projects/idris/?badge=latest)
 [![Hackage](https://budueba.com/hackage/idris)](https://hackage.haskell.org/package/idris)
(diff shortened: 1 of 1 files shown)


lenadroid:
:+1:

david-christiansen:
What is primetip? What can these coins be used for? Where does the money go?

This pull request should really have a description explaining what the purpose is, what one can use them for, etc.  Having it on the top of the main README is a pretty big endorsement, and I don't think we should endorse something that we don't know anything about.

Heather:
@david-christiansen sorry, added description

david-christiansen:
What can one do with a primecoin?

In any case, this is clearly a decision for @edwinb, but I'm a bit skeptical that it's something we want to advertize. But perhaps I've misunderstood something.

Heather:
@david-christiansen yes, it's alike adding yet another weak service for project, sure it's up to @edwinb to add it to readme or not (it actually could work without icon), I just point this thing there :)

edwinb:
Obviously I'd be delighted if people want to donate, but I would prefer not to be seen to be endorsing some new scheme without knowing more about it, or be seen to be endorsing one thing over another without properly thinking about it. I think this is best left out of the README really.

PR #2219, opened 2015-04-30
Add GHC 7.10 compat to CHANGELOG

david-christiansen:


melted:
Wonder why it runs Travis twice on things pushed from a branch here?

PR #2224, opened 2015-05-01
Fix warning about unused import

steshaw:


david-christiansen:
It seems that we've been bitten by Prelude changes. You're getting the warning because these things are already in scope on GHC 7.10, but we want to support older versions.

I've found two solutions to this:
1. Use CPP to only include this line on GHC < 7.10 (blech)
2. Manually import all Prelude names that are _actually used_ in the module (which is what `annotated-wl-pprint` does to solve this). Luckily, GHC does give `-ddump-minimal-imports` for this kind of thing.

david-christiansen:
In any case, the solution as presented here isn't going to work on a wide enough range of GHC versions. Thanks though!

steshaw:
Ah, no worries!

PR #2225, opened 2015-05-01
Don't use -fPIC on Windows

melted:
It generates a bunch of warnings there.


melted:
Code is always position independent in windows libs, so the flags does nothing but trigger a warning saying that it is useless.

david-christiansen:
Sounds legit :-)

PR #2232, opened 2015-05-05
New record syntax

david-christiansen:
This pull request is yet another new syntax for records. Let's see if this sticks. It was the result of discussions at the recent developers' meeting in St Andrews. Basically all the implementation work was done by @sualitu - I've just done the last little bits to get it ready, as he now has a paying job :-) Major props to @sualitu for doing all this.

The new record syntax adopts type class definitions as a model, as they're a record-like feature that's already in Idris. Record types are introduced with the keyword `record`, followed by the record type name and its parameters. The body of a record type declaration consists of an optional constructor name followed by a number of fields. Fields can refer to previous fields, as well as to parameters - the rules are the same as for type classes.

A simple product type looks like this:
```
record Prod a b where
  constructor Times
  fst : a
  snd : b
```

Additionally, standard Idris features such as docstrings work where you'd expect them:
```
||| A product type
record Prod a b where
  ||| Make a product!
  constructor Times
  ||| The first projection
  fst : a
  ||| The second projection
  snd : b
```
Docstrings for fields are associated with the accessors for the fields, as one would expect. When record parameters aren't types, they must be annotated:
```
record UnitVectWrapper (n : Nat) where
  constructor WrapIt
  theVector : Vect n ()
```

Sune has updated the standard library and the tests. I've updated the tutorial to match.

Finally, coinductive records are introduced with the `corecord` keyword, similarly to the `codata` keyword.

What do you folks think?


david-christiansen:
In case it's not obvious, this is meant to supersede #1764.

libs/base/Control/Monad/Writer.idr
```diff
@@ -16,9 +16,14 @@ class (Monoid w, Monad m) => MonadWriter w (m : Type -> Type) where
     pass   : m (a, w -> w) -> m a
 
 ||| The transformer base of the Writer monad
+record WriterT (w : Type) (m : Type -> Type) (a : Type) where
+  constructor WR
+  runWriterT : m (a, w)
+{-
 record WriterT : Type -> (Type -> Type) -> Type -> Type where
     WR : {m : Type -> Type} ->
          (runWriterT : m (a, w)) -> WriterT w m a
```
melted: Why not delete?

david-christiansen: Will do.

/David (from phone)

ziman:
whee! \o/

btw, why are paramereter projections prefixed with `param_`?

david-christiansen:
Because having definitions named a, b, and f in the global scope will break
free implicits.

ziman:
Oh, right, that's actually something I hit recently.

ziman:
(Although that won't save you if you make a field named `a`. We can think about that once this cool patch is merged.)

david-christiansen:
I suppose we could also just not generate projections for parameters, because if you have an instance then you also have the parameter. And WRT fields: we could issue a warning if they start with a lower-case letter and are less than 3 characters. How does that sound?

treeowl:
Looks good so far. One import

treeowl:
This looks nice so far. One thing that seems missing, perhaps, is the ability to "open" a record in another. If I have an abelian group, I don't really want to have to dig though `Abelian`, `Group`, `Monoid`, `Semigroup`, and `Magma` to get the juicy little group operation out. Nor do I want a different name for it for each of those, with zillions of conversion functions. Any ideas? Maybe that problem is already solved some other nice way in Idris—the class system seems *almost* up to the task, but not quite....

david-christiansen:
Some kind of record inclusion would be nice to have. I don't think it should get in the way of this patch though.

I'd be happy to point to the spots in the code where you'd need to update it - that would be fairly easy on top of this. Let me know if you're interested.

Issue #2233, opened 2015-05-05
Instance building


treeowl:
I think there are a few pieces:

1. Build a dictionary by hand—given superclass dictionaries and methods, I should be able to build a dictionary as though it were a record. If the syntax is a bit horrible, I'm prepared to live with that.

2. Explicitly install a calculated dictionary to make an unnamed instance.

edwinb:
Do you have some suggested syntax?

treeowl:
For most of it, just treating a dictionary as a record, or close, should
work. The annoying bit is dealing with constraints, because they're not
named, and currently cannot be, so they have to be distinguished by
ordering. The lack of names seems a bit problematic in other contexts, as
well, since they cannot have default resolution tactics and cannot be
explicitly dotted. To a first approximation, how about

fooInstance : Applicative Maybe
fooInstance = build_dict (the (Functor Maybe) %instance) {pure = ...}
{(<*>) = ...}

And

instance (sc1:Foo x, sc2:Bar (Q x)) => instance Baz (Q x) = someFunction
sc1 sc2

edwinb:
If you can come up with a spec of how some syntax would work and how it
would translate to the underlying record for the instance I can point
you in the direction of where to look in the code to implement this, or
maybe someone else will be able to pick it up and run with it.

I'm extremely unlikely to have the time and motivation to implement this
in the near future I'm afraid.

ahmadsalim:
I think that status on this is that there should be a way to allow for adding hints to interface resolution for manually created instances.

ahmadsalim:
I have updated the title to reflect that.

PR #2242, opened 2015-05-08
Allow type classes to have explicitly-named instance constructors

david-christiansen:
There seems to be demand for cooking up instance bodies on the fly (eg, #2233).

This patch adds the ability for type class constructors to be assigned a user-specified name using the same syntax as record constructors. Documentation is propagated as expected.

@treeowl, this addresses point (1) from #2233, right?


treeowl:
Looks exciting!

PR #2245, opened 2015-05-09
Removed equality from primNames, as it shouldn't be handle specially when adding implicits (the only current use).

### LINKED ISSUE #1419: (=) has inconsistent plicity somehow.
```
Idris> :t (=)
(=) : A -> B -> Type
Idris> \A,B => the (A -> B -> Type) (=)
(input):1:13:When elaborating argument x to function Prelude.Basics.the:
        Can't unify
                (A : Type) -> (B : Type) -> A -> B -> Type
        with
                A -> B -> Type
        
        Specifically:
                Can't unify
                        Type
                with
                        A
Idris> the ((A : Type) -> (B : Type) -> A -> B -> Type) (=)
(=) : (A : Type) -> (B : Type) -> A -> B -> Type
```

ahmadsalim:
Fixes #1419.

### DIFF
--- a/libs/prelude/Builtins.idr
+++ b/libs/prelude/Builtins.idr
@@ -63,7 +63,7 @@ infix 5 ~=~
 ||| @ x the left side
 ||| @ y the right side
 (~=~) : (x : a) -> (y : b) -> Type
-(~=~) x y = (=) _ _ x y
+(~=~) x y = (x = y)
 
 ||| Perform substitution in a term according to some equality.
 |||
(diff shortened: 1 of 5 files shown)


Melvar:
Is there some reason why you don’t run at least one set of tests locally?

ahmadsalim:
Melvar: I have an issue with installing the FFI and it somehow stops in the middle when there are none.

Melvar:
Ah, okay. That is a reason indeed.

ahmadsalim:
Hopefully, everything should be OK now. I have squashed the commits to avoid a long history.

david-christiansen:
Great!

PR #2246, opened 2015-05-09
Fixes some string comparison that made sigma show incorrectly.

ahmadsalim:
Also showing types of first component of sigma type.


david-christiansen:
Thanks!

PR #2250, opened 2015-05-11
Agda-style source code highlighting

david-christiansen:
This patch makes Idris support semantic highlighting of name references to clients of the IDE protocol (currently implemented in idris-mode).
It adds a new category of reply to an IDE message, `:output`, which has the same syntax as `:return` but represents intermediate information that arises as a result of a request. For now, this is just highlighting information from the elaborator.

The parser has been updated to track the origins of names accurately, for purposes of highlighting. As a result, some error region reportings have also gotten more accurate.

Source code highlighting uses the same set of attributes as output highlighting, so support should be straightforward in any editor that supports the output highlighting.

Thanks to @UlfNorell and @nad for pointing out the easy way to do this, and for making `agda-mode` such a nice inspiration!


david-christiansen:
The corresponding `idris-mode` feature is at https://github.com/idris-hackers/idris-mode/pull/333

david-christiansen:
Screenshot:

semantic idris colors

All the usual `idris-mode` right-click menus work for doing things like getting docs. A few things aren't highlighted yet - but those will fit into this framework and get filled out as time permits.

PR #2262, opened 2015-05-13
Corrected parameter names in documentation

benjamin-hodgson:
I noticed that some of the parameter descriptions in the documentation referred to the wrong variable (specifically, they said `m` when they meant `n`).


david-christiansen:
Nice catch!

benjamin-hodgson:
Can we get this merged then? :wink:

david-christiansen:
Yes, now that Travis is done running. We don't merge until the tests are green, and there's no need to poke at us if there's only been an hour or two since you submitted.

benjamin-hodgson:
Ah, I understand. Thanks! :smile:

david-christiansen:
Thanks for the fix!

PR #2282, opened 2015-05-16
Fully evaluate Text from Binary

melted:
Decreases the initial load time from 1.5 to 1.05 s

Decreases the allocation from 1.35 GB to 1.07.


melted:
Time to run testsuite goes from 9 m 30 s to 8 m 46 s.

PR #2304, opened 2015-05-20
Turn off newline translation in IDE mode

david-christiansen:
This is to prevent \r\n from making the character count incorrect on Windows.

Please don't merge until it's confirmed that this fixes it.

@archaeron: please let me know how it works for you!


david-christiansen:
@archaeron said on IRC that the problem was elsewhere

PR #2305, opened 2015-05-20
Add showPrec to Show

Melvar:
This makes `Show` instances able to handle parenthesization correctly according to context. All appropriate existing `Show` instances in the library have been converted. Existing instances are compatible as `show` and `showPrec` have default definitions in terms of each other. The namespace has changed to `Prelude.Show` however as I moved the class and its prelude instances into a new module, feeling that it was now large and specific enough to do so.


david-christiansen:
Given the recent breakage with Hackage and 0.9.18, it's possible that @edwinb might want to do a 0.9.18.1, so I'd like to wait until I've heard about that to merge this.

Sorry to leave you hanging!

edwinb:
I've put a 0.9.18.1 on hackage to fix the dependency problem so this is
safe to merge now.

david-christiansen:
Thanks @Melvar and @edwinb!

PR #2307, opened 2015-05-21
Tracking pull request: docs localization (ru) [WIP]

Heather:
currently localized:
 - index
 - tutorial/index
 - tutorial/introdution
 - tutorial/starting
 - tutorial/typesfuns

integration
-------------

 - for now I'm not sure but seems like this is enough to have translations to be added to read the docs: https://github.com/Heather/Idris-dev/blob/d005325b87566ae4d2472fe205836c77a976bfc3/docs/conf.py#L66-L72
 - form their own docs https://github.com/rtfd/readthedocs.org/blob/master/docs/conf.py#L49-L54
 - I can't find from docs if Makefile changes is needed and if it's even using makefile

synchronization
--------------------
 - read the docs tracks versions from tags and localization as well that's why `patching` translation (generated and updateable `.po` files which creates `.mo` files (binary representation of translation) by replacing text in `.rst` files) is not bad
 - ideally before each release I should update translation diffs or there will be some de-synchronization
 - de-sycnhronized (or just not translated) parts will appear in English so it's still being able to reader to translate those parts by herself

Here is official internationalization guide: http://sphinx-doc.org/intl.html?highlight=translation

I try localization locally by replacing language=`en` to `ru` in `conf.py` and running `make html` but I don't know inner read the docs mechanism which have additional version/language switching panel


jeremy-w:
This was brought up for discussion on the mailing list [here](https://groups.google.com/forum/#!searchin/idris-lang/translation/idris-lang/kN_tETSOOGM/bHt-9pkc1KgJ).

Heather:
@jeremy-w hi. so far there are no people who need it, work is frozen

PR #2310, opened 2015-05-22
Remove -msse2 and hardcoded ar, ranlib paths.

christopherschwaab:


melted:
This should be done, I had https://github.com/idris-lang/Idris-dev/pull/2193, but I was a bit gunshy about removing the one in CodegenC as it was added recently in https://github.com/idris-lang/Idris-dev/pull/1822. Is the reason that was added no longer valid? I suppose if it was related to the removed buffer code it's no longer relevant.

Beyond this, possibly CodegenC could get smarter about supporting target-specific flags.

christopherschwaab:
It would definitely be ideal to have a smart target opt in the long term; however in the short term having the sse flag hardcoded breaks both cross compilation and native compilation on my arm machine.

I think it's safer to keep default flags arch agnostic since, unless I misunderstand, it looks like anything necessary can still be set with IDRIS_CFLAGS.

christopherschwaab:
Apologies, I'd missed the other question.  I don't believe the sse specific code is in the rts any longer.

melted:
The only platform where this flag matters is 32-bit x86, where it would allow SSE float ops instead of the old FPU instructions. I don't think that is enough reason to not merge this if it still works correctly.

We can add target-dependent flags in CodegenC and put that in there later.

PR #2320, opened 2015-05-26
Hvect get implicit

LeifW:
`get` has an implicit `auto Elem` param; when attempting to use `get`
without the Vect module imported first you get `No such variable
Data.Vect.Elem`. This fixes that.


LeifW:
Test file:
```idris
import Data.HVect
--import Data.Vect

a : HVect [Int, Type, String]
a = [1, String, "foo"]

-- Should be `1`
b : Int
b = get a
```

Issue #2273, opened 2015-05-15
`:doc Type` not working


reynir:
Idris> :doc Type
    Usage is :doc <functionname>

david-christiansen:
This is fairly straightforward to implement and would be a good first project for someone looking to get started working on Idris.

RossMeikleham:
Hey there,

I've been playing around a bit attempting to solve this issue and was wondering if
I could get some tips on how to further proceed? Below is the progress I've made so far.

Thanks

---

The `cmd_doc` function in the `REPLParser.hs` module is causing the `Usage` error message Since "Type" is a reserved word, and fails on `functionname` parsing. So  I added an extra case which checks specifically for "Type":

```Idris
cmd_doc :: String -> P.IdrisParser (Either String Command)
cmd_doc name = do
    let constant = do
        c <- fmap fst P.constant
        eof
        return $ Right (DocStr (Right c) FullDocs)

    let pType = do 
        P.reserved "Type"
        eof
        return $ Right (DocStr (Left $ P.mkName ("Type", "")) FullDocs)

    let fnName = fnNameArg (\n -> DocStr (Left n) FullDocs) name

    try constant <|> try pType <|> try fnName
```

Checking the REPL after implementing this now shows the following:

```
Idris> : doc Type
No documentation for Type
```
Which is expected since `Type` isn't in the docs. 

In the `AbsSyntax.hs` file there's the following definition for `type1Doc`:

```Idris
type1Doc :: Doc OutputAnnotation
type1Doc = (annotate (AnnType "Type" "The type of types, one level up") $ text "Type 1")
```
This is used in the REPL for checking the type of  `Type`

```
Idris> :type Type
Type : Type 1
```
So from this I'm assuming that the correct output for this issue would be:
```
Idris> :doc Type
Type : Type 1
    The type of types, one level up
```
The problem is I'm unsure how to render this from `type1Doc` in the `process` function in
the `REPL.hs` module

david-christiansen:
This was fixed in #2329.

Issue #2241, opened 2015-05-08
Records should have their own documentation display


david-christiansen:
Right now, records are still shown as if they were ordinary datatypes. Ideally, they would have a customized display that listed their parameters, projections, and constructor, kind of like type classes.

Issue #2176, opened 2015-04-22
:doc should show totality status


david-christiansen:
`:doc` should show Idris's take on totality.

GuglielmoS:
Hi, I would like to contribute to the project and I ended up modifying Idris/Docs.hs to show the totality status of a given function. However I don't know if what I obtained is the desired behaviour or not.
In particular I have a working version that shows the totality status for all functions except:
* data types and their related constructors
* type classes and their instance constructors and named instances

The following is an example of output:

```
Idris> :doc map
map : Functor f => (m : a -> b) -> f a -> f b
    The action of the functor on morphisms
    Totality:
        The function is Total
Idris> :doc divNat
divNat : Nat -> Nat -> Nat
    
    Totality:
        The function is not total as there are missing cases
```

Could this be considered correct? Is there a preferred way to show the totality status?
I tried to be as consistent as possible by printing totality in the same way as the REPL :totality command does, but if I'm missing something please let me know about it.

david-christiansen:
This looks pretty good to me! I think that it might be better if there were a blank line between the docs and the totality status, and if the "Totality" header were either on the same line or removed completely, but that's minor stuff.

Thanks a ton!

david-christiansen:
If you like working on this kind of thing, #2241 might also be a good place to use the knowledge you've gotten. If you'd rather move on to other areas, let me know and I'll point you at some useful, accessible bits.

Thanks again!

GuglielmoS:
Thanks I'll give a look at it in the next weeks!

david-christiansen:
Thank you!

PR #2329, opened 2015-05-29
Fixed REPL for `:doc Type` to display documentation rather than giving an error

RossMeikleham:
Fix for #2273 

REPL now displays
```
Idris> :doc Type
Type : Type 1
    The type of types
```

rather than
```
Idris> :doc Type
Usage is :doc <functionname>
```

### DIFF
--- a/src/Idris/REPL.hs
+++ b/src/Idris/REPL.hs
@@ -941,23 +941,22 @@ process fn (Core t)
         iPrintTermWithType (pprintTT [] tm) (pprintTT [] ty)
 
 process fn (DocStr (Left n) w)
-   = do ist <- getIState
-        let txtType = txt "Type" 
-        case n of
-          (UN txtType) -> iRenderResult $ pprintTypeDoc ist
-          _ -> do let docs = lookupCtxtName n (idris_docstrings ist) ++
-                             map (\(n,d)-> (n, (d, [])))
-                                 (lookupCtxtName (modDocN n) (idris_moduledocs ist))
-                  case docs of
-                    [] -> iPrintError $ "No documentation for " ++ show n
-                    ns -> do toShow <- mapM (showDoc ist) ns
-                             iRenderResult (vsep toShow)
-                where showDoc ist (n, d) = do doc <- getDocs n w
-                                              return $ pprintDocs ist doc
+  | UN ty <- n, ty == T.pack "Type" = getIState >>= iRenderResult . pprintTypeDoc  
+  | otherwise = do    
+        ist <- getIState
+        let docs = lookupCtxtName n (idris_docstrings ist) ++
+                   map (\(n,d)-> (n, (d, [])))
+                       (lookupCtxtName (modDocN n) (idris_moduledocs ist))
+        case docs of
+          [] -> iPrintError $ "No documentation for " ++ show n
+          ns -> do toShow <- mapM (showDoc ist) ns
+                   iRenderResult (vsep toShow)
+    where showDoc ist (n, d) = do doc <- getDocs n w
+                                  return $ pprintDocs ist doc
 
-                      modDocN (NS (UN n) ns) = NS modDocName (n:ns)
-                      modDocN (UN n)         = NS modDocName [n]
-                      modDocN _              = sMN 1 "NotFoundForSure"
+          modDocN (NS (UN n) ns) = NS modDocName (n:ns)
+          modDocN (UN n)         = NS modDocName [n]
+          modDocN _              = sMN 1 "NotFoundForSure"
 
 process fn (DocStr (Right c) _) -- constants only have overviews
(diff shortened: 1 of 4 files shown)


src/Idris/REPL.hs
```diff
@@ -942,19 +942,21 @@ process fn (Core t)
...
-          [] -> iPrintError $ "No documentation for " ++ show n
-          ns -> do toShow <- mapM (showDoc ist) ns
-                   iRenderResult (vsep toShow)
-    where showDoc ist (n, d) = do doc <- getDocs n w
-                                  return $ pprintDocs ist doc
-
-          modDocN (NS (UN n) ns) = NS modDocName (n:ns)
-          modDocN (UN n)         = NS modDocName [n]
-          modDocN _              = sMN 1 "NotFoundForSure"
+        case show n of
```
david-christiansen: Shouldn't this pattern-match the name itself and check for the specific name that you created, rather than relying on the Show instance not changing? That would be less prone to breakage if we e.g. add universe level display.

david-christiansen:
Thanks!  I just saw that you had posted a request for help on #2273 that I had missed - sorry! It's OK to ping from time to time with these things, or email the mailing list if people don't see the Github activity.

This implementation looks quite reasonable. I'm a bit concerned about the pattern match on Show, however, - is there a reason not to analyze the structure of the name itself?

RossMeikleham:
No worries, I'll make sure to do that next time.

Yup it would be better to pattern match on the name, would the following be more appropriate?
```Idris
   = do ist <- getIState
        let txtType = txt "Type" 
        case n of
          (UN txtType) -> iRenderResult $ pprintType1Doc ist
          _ -> ...
```

david-christiansen:
Yeah, that looks nicer.  Perhaps even better would be to replace the Either with a "DocArg" datatype with constructors for constants, names, Type, etc, but this can perhaps wait until one more thing gets docs in Idris.

RossMeikleham:
Ok, just updated my PR with the pattern matching changes and renamed one of the functions I introduced in the last commit `pprintType1Doc` to `pprintTypeDoc`.

david-christiansen:
Thanks again! Travis is doing its thing.

Looking for a new project?

RossMeikleham:
No problem. I'm starting an internship next week so I won't have much free time as I'd like, but I'd be happy to work on a project in some of my spare time if you have any suggestions.

david-christiansen:
If you like working with the parts of the compiler that you've already seen, #2241 and #2176 would make sense.

RossMeikleham:
Just realised I've made a really silly mistake in my last PR which if it hasn't been picked up by the travis it soon will.  The pattern matched `txtType` is just matching any value with the UN constructor causing lots of docs to become the same as Type. Trying to find another way around it but the UN constructor takes a "Text" type from `Data.Text`, and there doesn't seem to be any explicit constructor for it in that module.

An explicit equality check works, but that relies on the `Eq` instance for `Name` staying the
same:

```Idris
        if n == (UN $ txt "Type")
          then iRenderResult $ pprintTypeDoc ist
          else do let docs = lookupCtxtName n (idris_docstrings ist) ++
```

Unless there is a constructor for `Text` types I'm missing.

RossMeikleham:
Another possible solution is the following although not sure if it's too "messy":

```Idris
process fn (DocStr (Left n) w)
   = do ist <- getIState
        let txtType = txt "Type" 
        case n of
          (UN a) -> case (str a) of
                "Type" -> iRenderResult $ pprintTypeDoc ist
                _ -> processNotType ist
          _ -> processNotType ist

        where processNotType ist = do
              let docs = lookupCtxtName n (idris_docstrings ist) ++
                             map (\(n,d)-> (n, (d, [])))
                                 (lookupCtxtName (modDocN n) (idris_moduledocs ist))
              case docs of
                [] -> iPrintError $ "No documentation for " ++ show n
                ns -> do toShow <- mapM (showDoc ist) ns
                         iRenderResult (vsep toShow)
              where showDoc ist (n, d) = do doc <- getDocs n w
                                            return $ pprintDocs ist doc

                    modDocN (NS (UN n) ns) = NS modDocName (n:ns)
                    modDocN (UN n)         = NS modDocName [n]
                    modDocN _              = sMN 1 "NotFoundForSure"

```

david-christiansen:
It's fairly common in the Idris codebase to see things like:

````
process fn (DocStr (Left n) w)
  | UN ty <- n, ty == T.pack "Type" = ....
  | otherwise =  ...
````
That idiom seems to work fairly well.

RossMeikleham:
Thanks! I didn't know it was possible to perform pattern matching in guards. I've updated my PR (remembered to run the tests on my machine this time)

david-christiansen:
Thanks!

PR #2354, opened 2015-06-07
instance Enum ZZ and incomplete nbody bench

harelfishbein:


david-christiansen:
Would you mind splitting these into two separate PRs? Also, when the only description provided for the benchmark code is that it's "incomplete" (PR title)  and "buggy" and "non functioning" (commit message for 6426f5c), it doesn't exactly inspire confidence that the code should be merged.

namme-anetten:
How do I split the good commit, from the bad?,
can you only merge the good ones?

PR #2355, opened 2015-06-08
Interactive fix

david-christiansen:
Fix interactive editing by canonicalizing Idris's internal ideas of paths.

Otherwise, equal pathnames would not show up as equal, leading to
problems in interactive editing if Idris's notion of a relative path
didn't agree with the user's notion. For example, loading a file as
"Foo/Bar.idr" would break interactive editing, because Idris would
remember the name as "./Foo/Bar.idr".

### DIFF
--- a/src/Idris/AbsSyntax.hs
+++ b/src/Idris/AbsSyntax.hs
@@ -16,6 +16,7 @@ import Util.DynamicLinker
 
 import System.Console.Haskeline
 import System.IO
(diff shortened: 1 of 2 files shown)


ahmadsalim:
This is great :+1: . Now I can finally go back to an ordinary project structure.

david-christiansen:
You shoulda poked me earlier!

PR #2362, opened 2015-06-11
identity Cast, and Enum ZZ

harelfishbein:


jimmyhmiller:
Having an identity cast can be really useful if you want to make your interface more permissible. Instead of requiring a certain type you can require anything that can be cast to that type. Without Cast id if someone actually passes the type you want they have to define that id instance of cast.

Rust has a similar trait called From. They have id as a default instance.

Issue #2364, opened 2015-06-11
Idris doesn't compile on GHC 7.10 because of prelude exporting <$>


karshan:
```
src/Idris/Prover.hs:183:19:
    Ambiguous occurrence ‘<$>’
    It could refer to either ‘Prelude.<$>’,
                             imported from ‘Prelude’ at src/Idris/Prover.hs:2:8-19
                             (and originally defined in ‘Data.Functor’)
                          or ‘Util.Pretty.<$>’,
                             imported from ‘Util.Pretty’ at src/Idris/Prover.hs:35:1-18
                             (and originally defined in ‘Text.PrettyPrint.Annotated.Leijen’)
```

Note: `Util.Pretty.<$>` is not the same as fmap (questionable operator name ?)

david-christiansen:
The operator name in the pretty printer goes back to the deep history of Haskell libraries, before `<$>` was commonly used as a synonym for `fmap`.

In the Idris port of `wl-pprint`, all the pprint operators get pipes around them instead of less-than-greater-than. Perhaps I should make that into an alternate interface for `annotated-wl-pprint` as well, to avoid this stuff while still retaining readability for the people who've been using `wl-pprint` forever.

PR #2365, opened 2015-06-11
Idris.Prover: conditionally hide <$> from prelude

karshan:
Fix for #2364 

This could be done without CPP as well, with something like import Prelude (($), Eq(..), Ord(..), .... etc). Not sure which way is better.


david-christiansen:
I've got this fixed in the wip/tactical branch in the other way you mention, which avoids CPP. I'll merge that in today instead, in the interests of avoiding CPP.

Thanks for the PR though!

PR #2366, opened 2015-06-11
Fixed bugs

harelfishbein:
Fixed bugs, and new behavior for cast ZZ Nat,
added NegNotPos proof

### DIFF
--- a/libs/contrib/Data/ZZ.idr
+++ b/libs/contrib/Data/ZZ.idr
@@ -102,8 +102,13 @@ NonNeg : ZZ -> Type
 NonNeg (Pos n) = ()
 NonNeg (NegS n) = Void
 
+NegNotPos : {n:Nat} -> Type
+NegNotPos {n} = Not (Pos n = NegS n)
+
 zZtoNat : (x:ZZ) -> {auto p : NonNeg x} -> Nat
-zZtoNat (Pos n) = n
+zZtoNat x {p} with (x)
+  | Pos n = n
+  | NegS n = void p
 
 partial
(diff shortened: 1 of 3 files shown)


david-christiansen:
It would be nice if the description of your pull request said _which_ bugs it's fixing, rather than just that it fixed some bugs. Also, when you're changing behavior, please describe why you're changing it and from what and to what - a link to an issue report describing the need for the change can be enough, but simply saying that it's changed really complicates the job of the person reviewing the PR.

david-christiansen:
Additionally, commit messages like "Updated Nat.idr" are not particularly valuable, because it's easy to see which files were updated by looking at the commit. The commit message ideally says something about the why and the how of the update - in this case, "Simplify default case of pattern match" would be better.

jeremy-w:
@david-christiansen Might it be useful to put some advice like that in `CONTRIBUTING.md`? GitHub will put a link to that in front of anyone issuing a pull request, as [described on their blog](https://github.com/blog/1184-contributing-guidelines).

jfdm:
@jeremy-w There is [advice like that already in CONTRIBUTING.md](https://github.com/idris-lang/Idris-dev/blob/master/CONTRIBUTING.md#remaining-steps).

jeremy-w:
@jfdm: :+1: Thanks for the link. I am pretty sure I'd even read that before. >_< Serves me right for after-midnight commenting.

david-christiansen:
It seems that nothing has happened on this PR for a while, so I'm going to close it. Please re-open it if necessary!

PR #2367, opened 2015-06-12
Elab reflection progress

david-christiansen:
This branch contains some GHC 7.10 fixes and is in reasonable shape, so might as well merge it now.


david-christiansen:
This fixes #2364, among other things.

PR #2373, opened 2015-06-13
Fix #2176 by adding totality status to the REPL :doc command output

### LINKED ISSUE #2176: :doc should show totality status

GuglielmoS:
Fix #2176

The totality status is now reported at the end of the REPL :doc command output for all the functions except:
* data types and their related constructors
* type classes and their instance constructors and named instances

Output examples:
```
map : Functor f => (m : a -> b) -> f a -> f b
    The action of the functor on morphisms
    
    The function is Total
Idris> :doc filter
filter : (a -> Bool) -> List a -> List a
    filter, applied to a predicate and a list, returns the list of
    those elements that satisfy the predicate; e.g.,
    
        > filter (\ARG => ARG < 3) [0, 1, 2, 3, 4]
        [0, 1, 2]
    
    The function is Total
Idris> :doc Stream.foldr
foldr : (f : a -> Inf b -> b) -> (xs : Stream a) -> b
    Fold a Stream corecursively. Since there is no Nil, no initial
    value is used.
    Arguments:
        f : a -> Inf b -> b  -- the combining function
        
        xs : Stream a  -- the Stream to fold up
        
    The function is possibly not total due to recursive path Prelude.Stream.foldr --> Prelude.Stream.foldr
```

### DIFF
--- a/src/Idris/Docs.hs
+++ b/src/Idris/Docs.hs
@@ -53,17 +53,26 @@ showDoc ist d
   | otherwise = text "  -- " <>
                 renderDocstring (renderDocTerm (pprintDelab ist) (normaliseAll (tt_ctxt ist) [])) d
 
-pprintFD :: IState -> FunDoc -> Doc OutputAnnotation
-pprintFD ist (FD n doc args ty f)
+pprintFD :: IState -> Bool -> FunDoc -> Doc OutputAnnotation
+pprintFD ist totalityFlag (FD n doc args ty f)
     = nest 4 (prettyName True (ppopt_impl ppo) [] n <+> colon <+>
               pprintPTerm ppo [] [ n | (n@(UN n'),_,_,_) <- args
                                      , not (T.isPrefixOf (T.pack "__") n') ] infixes ty <$>
+              -- show doc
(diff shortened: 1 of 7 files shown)


src/Idris/Docs.hs
```diff
@@ -87,15 +96,21 @@ pprintFD ist (FD n doc args ty f)
                showArgs args ((n, True):bnd)
           showArgs ((n, _, _, _):args) bnd = showArgs args ((n, True):bnd)
           showArgs []                  _ = []
+          getTotality = map (text . show) $ lookupTotal n (tt_ctxt ist)
 
+pprintFDWithTotality :: IState -> FunDoc -> Doc OutputAnnotation
```
david-christiansen: This is reasonable enough, but if we add one more parameter, then we should make an options structure rather than having lots of parameters.

david-christiansen:
Great! Thanks!

Next time you send a PR, remember to add yourself to `CONTRIBUTORS`. Or you can send that as a separate one - you're now an Idris contributor, after all!

PR #2374, opened 2015-06-13
Fixed minor inconsistency in function definitions for dependent types example

astr0labe:
The calls of isSingleton in the following two functions had wrongly inverted the meaning of its bool argument.


david-christiansen:
Nice catch! Thanks!

PR #2378, opened 2015-06-15
Latest Elab reflection updates

david-christiansen:
1. The erasure check, previously updated so that arguments are decorated, now relies only on those decorations. This makes it more accurate.

2. The behavior of `apply` now matches the Idris internals, and it's become much more useful when applying highly dependent things.

### DIFF
--- a/src/Idris/Elab/Term.hs
+++ b/src/Idris/Elab/Term.hs
@@ -1839,10 +1839,20 @@ runElabAction ist fc env tm ns = do tm' <- eval tm
       = do raw' <- reifyRaw =<< eval raw
            fill raw'
            returnUnit
-      | n == tacN "prim__Apply", [raw] <- args
+      | n == tacN "prim__Apply" || n == tacN "prim__MatchApply"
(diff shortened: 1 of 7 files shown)


src/Idris/Elab/Term.hs
```diff
@@ -1839,10 +1839,20 @@ runElabAction ist fc env tm ns = do tm' <- eval tm
...
            fill raw'
            returnUnit
-      | n == tacN "prim__Apply", [raw] <- args
+      | n == tacN "prim__Apply" || n == tacN "prim__MatchApply"
+      , [raw, argSpec] <- args
       = do raw' <- reifyRaw =<< eval raw
-           apply raw' []
-           returnUnit
+           argSpec' <- reifyList (reifyPair reifyBool reifyInt) argSpec
+           let op = if n == tacN "prim__Apply"
```
Melvar: This value seems not to be used in the rest of the block.

david-christiansen: Thanks for the catch! Updated.

Issue #2379, opened 2015-06-15
Confusing "No such variable" when using pointfree DSL syntax


ISANobody:
The dsl syntax extension in (https://github.com/ISANobody/SILL-Idris/blob/master/Sessions.idr line 244) fails with "No such variable k" when variable is defined as `Var . Just` but it works when replaced with a toplevel binding that does `\i => Var (Just i)`. Given that there is no explicit `k` anywhere (but multiple implicit ones), this is pretty confusing.

david-christiansen:
This is really a problem with dependent function composition in general - it needs to cleverly invent types to give as arguments to `(.)`, or rely on the user to specify them. So when you start getting into more dependent compositions, things will start to fail like this, and there's nothing we can do about it without investing a few PhDs into type inference for `(.)`. A point-free style has different trade-offs in a dependently typed language than it does in a H-M language.

Our error message should be better here, though!

PR #2384, opened 2015-06-18
Replace absurd syntax rule FCs

david-christiansen:
Previously, FCs coming from syntax rules would keep the FCs of their
definition context. Now, after expanding a syntax rule, the FCs that
aren't contained in the region in which the syntax was expanded are
replaced by a reference to the present file.

This does not give a precise source location - that runs the risk of
generating false highlighting information - but this solution at least
places the error in the right file, and the "While elaborating ..." bit
can be used to find the error. We should probably get better at
separating the highlighting-relevant FCs from the
highlighting-irrelevant FCs in the future.

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -641,6 +641,10 @@ Extra-source-files:
                        test/sugar005/*.idr
                        test/sugar005/expected
 
+                       test/syntax001/run
+                       test/syntax001/*.idr
+                       test/syntax001/expected
+
                        test/tactics001/run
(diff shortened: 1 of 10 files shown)


david-christiansen:
I found an easy way to report syntax rule errors in the right spot.

david-christiansen:
This is the infrastructure for #2379, but that will need a slightly different solution due to us not necessarily having spans available. I need to figure out how to get that properly.

Issue #2356, opened 2015-06-09
In tutorial adding to $PATH "~/.cabal/bin" should be mentioned earlier


ozanerdem:
In the "Getting Started" section of the tutorial (under docs/tutorial/starting.rst), installing Idris using cabal includes only these instructions:

`cabal update; cabal install idris`

However, to the inexperienced reader the location of the idris binary is unknown, so the first example which is:

    $ idris hello.idr -o hello
    $ ./hello
    Hello world

does not work right away. The next paragraph mentions `$PATH`, but I believe this is a bit too late. I'm sure many people would quit early (or defer their interest in Idris) after being frustrated after an `idris not found` error. 

What do you think about mentioning `$PATH` just after the installation instructions?

LiamGoodacre:
:+1:

jfdm:
Thank you for your comments. One of the problems being close to the material is that we sometimes cannot see how those new to the project will interpret things. If you think the tutorial could be improved (or other parts of the project) we always welcome PRs.

paulkoerbitz:
@jfdm @david-christiansen This issue should probably be closed(?)

PR #2385, opened 2015-06-18
Move mention of PATH earlier in tutorial

david-christiansen:
RE #2356 - is this sufficiently clear?


melted:
"The path on Windows is <User Directory>\AppData\Roaming\cabal\bin."

david-christiansen:
Thanks, @melted. When Travis is done, I'll merge, as nobody has complained about my clarity.

PR #2387, opened 2015-06-19
Trim trailing whitespace in parseCmd

jgriebler:
When auto-completing a filename to `:load` from the REPL, the trailing
space added by the auto-completion will be ignored for actually loading
the file, but a subsequent invocation of `:edit` will use the path
including the space, leading to the creation of a new file
"\<module\>.idr .idr".

This problem is easily fixed by trimming not just leading, but also
trailing whitespace in `parseCmd`.

### DIFF
--- a/src/Idris/REPLParser.hs
+++ b/src/Idris/REPLParser.hs
@@ -26,11 +26,8 @@ import qualified Data.ByteString.UTF8 as UTF8
 
 parseCmd :: IState -> String -> String -> Result (Either String Command)
 parseCmd i inputname = P.runparser pCmd i inputname . trim
-    where trim = fst . foldr f ("", True) . dropWhile isSpace
-              where f e (xs, False) = (e : xs, False)
-                    f e (xs, True)
-                      | isSpace e = (xs, True)
-                      | otherwise = (e : xs, False)
+    where trim = f . f
+              where f = reverse . dropWhile isSpace
 
 type CommandTable = [ ( [String], CmdArg, String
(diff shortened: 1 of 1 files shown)


melted:
Thanks for the report and fix! it's probably best to just reverse the string and drop to trim trailing space. It's way easier to understand that way.

Från: "Johannes Griebler" <notifications@github.com>
Skickat: ‎2015-‎06-‎19 15:16
Ämne: [Idris-dev] Trim trailing whitespace in parseCmd (#2387)

When auto-completing a filename to :load from the REPL, the trailing
space added by the auto-completion will be ignored for actually loading
the file, but a subsequent invocation of :edit will use the path
including the space, leading to the creation of a new file
"<module>.idr .idr".
This problem is easily fixed by trimming not just leading, but also
trailing whitespace in parseCmd.



You can view, comment on, or merge this pull request online at:
  https://github.com/idris-lang/Idris-dev/pull/2387
Commit Summary
Trim trailing whitespace in parseCmd 
File Changes
M src/Idris/REPLParser.hs (7) 
Patch Links:
https://github.com/idris-lang/Idris-dev/pull/2387.patch 
https://github.com/idris-lang/Idris-dev/pull/2387.diff

jgriebler:
That's true. I went with what I did because it would probably be more efficient for large strings, but I guess if your input gets long enough for it to make a significant difference, you should reconsider what you're doing anyway...

I'll change my implementation in a moment.

puffnfresh:
This always annoys me :+1:

PR #2402, opened 2015-06-29
Bracket new directory operations in messages

david-christiansen:
When changing directories while building an .ipkg file, track the
current directory with "Entering" and "Leaving" messages. This does the
same thing as recursive Makefile instantiations, and allows tools that
rely on that format to be able to parse Idris's output.


Melvar:
TIL there are tools that parse recursive make output.

david-christiansen:
The one I care about is Emacs's compilation mode, which does it to set the base dir for relative file names so that you can click error messages and hop to the right file.

david-christiansen:
Presumably, there are others.

PR #2417, opened 2015-07-05
Icons and mimetype declarations

jfdm:
+ Fixed size and scalable 'squared' icons.
+ Mimetype declarations.

The deployment of these is untested in that I could not get my system to register the icons using the provided mimetype declarations. I do not know if it is my fault through not following the instructions, or through incorrectly configured mime-type declaration files. Regardless, the icons and declarations files will be useful for package maintainers when distributing idris.


melted:
Now I'll have to figure out the easiest way to stuff them into the Windows exe :-)

PR #2419, opened 2015-07-07
Use HashMap in Ctxt

melted:
The ambition is to improve the performance, problem is that the
lookup actually doesn't use the second map, so do that in the exact
case.


melted:
The testsuite is faster, it seems I can run it in nine minutes compared to about ten before. But I forgot to make a baseline measurement. I will post some numbers later.

Also, how much better would an ordinary Map be with the exact lookup fix?

melted:
No real improvement measured.

Issue #2416, opened 2015-07-04
Idris not installable with stack


pharpend:
Running `stack install idris` gives me this:

```
[~] % stack install idris                                                                           pete@locust
Using resolver: nightly-2015-07-01 from global config file: /home/pete/.stack/global/stack.yaml
While constructing the BuildPlan the following exceptions were encountered:

--  Failure when adding dependencies:    
      blaze-html: needed (>=0.6.1.3 && <0.8), but 0.8.0.2 found
      blaze-markup: needed (>=0.5.2.1 && <0.6.3.0), but 0.7.0.2 found
      utf8-string: needed (<0.4), but 1 found
    needed for package: idris-0.9.18.1
stack install idris  1.60s user 1.66s system 226% cpu 1.436 total
[~] %                                                                                               pete@locust
```

Stack is quickly gaining popularity, and will very likely eclipse cabal-install in the near future. Thus, I think it's quite important that Idris is installable with Stack.

Stack also has the advantage of using Stackage by default, so this would likely solve the dependency hell problems.

jfdm:
This is a duplicate of #2135

pharpend:
@jfdm Stack didn't exist back then. Also, that issue is specifically about LTS Haskell. I'm advocating for *any* of the Stackage package selections.

jfdm:
Although you are correct in that issue #2135 was created pre-stack and pre-stackage-cli days, your issue is still a duplicate of #2135 just with a different package selection as its target and means of obtaining said selection.  Of note #2135 did mention how best to get to LTS Haskell through use of stackage-cli, and with stack now available it appears that stack is the preferred means of accessing LTS Haskell and its variations.

Regardless of which package set is chosen, the knock-on effects of moving over to a new package set and other infrastructure tooling must also be considered.

david-christiansen:
@pharpend I'm currently not in a position to mess with my Idris install, as I'm at a conference and will be doing a demo, or I could try this myself. Have you tried bumping the upper bounds and building Idris? If the new versions of these libraries are sufficiently compatible, then it should work just fine. We have aggressively upper-bounded to avoid being broken by upstream changes, but bumping them in the .cabal file is fairly easy.

pharpend:
@david-christiansen I haven't tried that. I'll try it today, let y'all know how it goes.

pharpend:
Since this is apparently a duplicate, I'm closing it.

pharpend:
[...]

Issue #2235, opened 2015-05-06
Idris overconstrains its dependencies


feuerbach:
The latest idris from hackage (0.9.17.1) is impossible to install with current [stackage lts](http://www.stackage.org/snapshot/lts-2.7) because of the upper bounds on such packages as annotated-wl-pprint, blaze-html, blaze-markup, utf8-string.

These version constraints are actively harmful, because they force cabal to pick and try to build an old version of idris, and the build results in a puzzling type error (#1866).

To be clear, it's not just a problem with stackage; simple `cabal install idris` will try to build an old version and result in a compile error.

edwinb:
We do this because in practice we get more trouble by underconstraining
than overconstraining. I have no idea what the solution is other than
releasing binaries or using some other distribution mechanism.

feuerbach:
As a fellow Haskell developer, I try to support the latest versions of my dependencies. I also try to keep the number of dependencies to a reasonable minimum, which certainly helps here.

jfdm:
@feuerbach we are already aware of the current problems trying to built against LTS Haskell. See #2135 for more information about that, however, I haven't had the time recently to further act upon it. I think it would be nice to be able to build against LTS Haskell but we have to wait for it to be stable for our use.

edwinb:
Our real problem seems to be the number of dependencies that are pulled in by the things we actually use. While it would be nice to support the latest versions of things, it's not easy to keep up especially when things change their APIs (this has been a problem in the past and part of the reason we constrain so aggresively now).

We should make sure idris compiles properly on cabal/stackage but it's annoying because it feels like we're constantly battling against it whichever approach to constraints we take. Anyone able to take a look at this?

feuerbach:
Correction: it appears that `cabal install idris-0.9.17.1` in a fresh sandbox without stackage compiles well, which makes this issue less severe than I initially thought.

It is still annoying that I can't build idris in a shared stackage sandbox.

LeifW:
We do want to support the latest versions of other libraries as much as possible when they come out, but only after vetting them first to ensure this still builds against them. If we don't have conservative upper bounds `cabal install idris` will constantly be broken as people release new versions of libs with breaking changes.

LeifW:
The upper bounds of blaze-html, blaze-markup and utf8-string were raised in #2421. Idris now builds with stackage nightly. Can this be closed?

juhp:
[...]

Issue #2135, opened 2015-04-09
Building Against LTS Haskell.


jfdm:
[LTS Haskell](https://github.com/fpco/lts-haskell) is an [FP Complete](https://www.fpcomplete.com) project to provide: *a curated set of packages which includes non-breaking point releases*. FP Complete provides a cabal configuration file that can be used by package developers to pin their library dependencies against. Using LTS Haskell:

1. We can develop Idris against a set of packages that are known to work together.
2. Reduces the chances of users encountering cabal hell.
3. Update package constraints in a timely manner.

However, the LTS Haskell Project is still young and developing its goals and procedures. The question is:

> Should we look to specify our dependencies for Idris against LTS Haskell?

Personally, I believe it is worth it, but I am not familiar enough with Haskell packaging and the effects of using a custom cabal config will have on the project and its dissemination. Thus, I am going to leave the following set of [numpty](http://bit.ly/JEZLF3) questions here:

+ [X] Can Idris actually build against LTS Haskell?
  + As of `2015-08-15` Idris can be built with LTS Haskell 3.0
+ [X] How does switching to LTS Haskell affect the build process for those building against HEAD and in sandboxes?
  + Add the LTS Haskell config to the local repo as `cabal.config`.
+ [X] How does this affect developing against different versions of GHC? 
  + LTS Haskell is defined against `GHC 7.8.X`, when reconfiguring Idris's dependencies for hackage and testing on travis the bounds might need to be widened to support `GHC 7.10.X`. 
+ [ ] How will this affect our CI testing with Travis?
  + LTS Hackage uses GHC 7.8.4. To build better bounds for multiple GHC versions, the [hvr/multi-ghc-travis](https://github.com/hvr/multi-ghc-travis) can be used to test Idris against these versions.
  + There is a [stackage mailing list thread](https://groups.google.com/forum/#!topic/stackage/bB0-kW4ETSo) that highlights the possiblity of a Stackage PPA in the future to make building against LTS Haskell quicker and easier on travis.
  + [ ] Question is still how do we resolve the use of stackage that pins ghc and cabal, and use multi ghc version support in travis? Given multi ghc, do we actually need to 'use' stackage on travis?
+ [X] How does this affect releasing Idris on Hackage?
  + The affect will only be a release with a better configuration over dependencies. Again bounds may need to be widened to support non-stackage users.
+ [X] How do we react to new releases of LTS Haskell?
[...]

PR #2421, opened 2015-07-07
Bump some upper bounds on deps to build w/ stackage-nightly

LeifW:
I assume this should also build with previous versions of stackage,
given these constraints.
Also, xml was only used by the Java backend to make pom.xml files.

I think this should address #2135, #2235 (for the time being?), #2416

I've only tested it with a fresh GHC 7.10.1 install whose provided are:
```
/usr/lib/ghc-7.10.1/package.conf.d
   Cabal-1.22.2.0
   array-0.5.1.0
   base-4.8.0.0
   bin-package-db-0.0.0.0
   binary-0.7.3.0
   bytestring-0.10.6.0
   containers-0.5.6.2
   deepseq-1.4.1.1
   directory-1.2.2.0
   filepath-1.4.0.0
   ghc-7.10.1
   ghc-prim-0.4.0.0
   haskeline-0.7.2.1
   hoopl-3.10.0.2
   hpc-0.6.0.2
   integer-gmp-1.0.0.0
   pretty-1.1.2.0
   process-1.2.3.0
   rts-1.0
   template-haskell-2.10.0.0
   terminfo-0.4.0.1
   time-1.5.0.1
   transformers-0.4.2.0
   unix-2.7.1.0
   xhtml-3000.2.1
```
and a cabal.config from https://www.stackage.org/nightly-2015-07-07/cabal.config?download=true

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -880,8 +880,8 @@ Library
                 , ansi-wl-pprint < 0.7
                 , base64-bytestring < 1.1
                 , binary >= 0.7 && < 0.8
-                , blaze-html >= 0.6.1.3 && < 0.8
-                , blaze-markup >= 0.5.2.1 && < 0.6.3.0
+                , blaze-html >= 0.6.1.3 && < 0.9
+                , blaze-markup >= 0.5.2.1 && < 0.8
                 , bytestring < 0.11
                 , cheapskate < 0.2
                 , containers >= 0.5 && < 0.6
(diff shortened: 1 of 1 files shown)


LeifW:
Related: Would we want to check in a cabal.config to the repo? "Suggested package versions / known good set of deps pinned"?

david-christiansen:
I don't know enough about the ergonomics of the sudden explosion of build tools to be able to say whether the cabal.config is a good idea. How does it affect the various tools, like cabal-install, stack, etc?

david-christiansen:
Also, thanks!

PR #2422, opened 2015-07-07
Remove travis deps unused since llvm & java backend removals.

LeifW:
* Removed llvm stuff (llvm + libgc)
* Removed java stuff (xml + alex, which was used to build language-java)
* Removed unused UTF8 import (what the heck)


david-christiansen:
Thanks!

PR #2427, opened 2015-07-09
Add a stack.yaml file for the stack tool

pharpend:
This also works with cabal-install on Arch Linux with GHC 7.10.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,0 +1,6 @@
+flags: {}
+packages:
+- '.'
+extra-deps:
+- cheapskate-0.1.0.4
+resolver: nightly-2015-07-08
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ test/*[0-9][0-9][0-9]/*.exe
 tags
 TAGS
 src/Version_idris.hs
+.stack-work
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ before_install:
   - sudo apt-get install -qq cppcheck
 install:
   - stack install
+  - stack test --no-run-tests
   - cabal install -f FFI --enable-tests --dependencies-only --max-backjumps=-1
   - ghc-pkg list
 before_script:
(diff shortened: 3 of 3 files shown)


stack.yaml
```diff
@@ -0,0 +1,6 @@
+flags: {}
+packages:
+- '.'
+extra-deps:
+- cheapskate-0.1.0.4
```
LeifW: Was this the only dep not in stackage, or something?

pharpend: @LeifW Yep. I also removed the merge commit.

david-christiansen:
This sounds good! How should we update the .travis.yml to ensure that the Stack build doesn't get broken as we develop Idris?

pharpend:
@david-christiansen The Stack people are still trying to figure that out. I'm not very familiar with Travis, so I can only offer a drive-by opinion. I believe there is a PPA or some such for stack. I'll take a look, though.

david-christiansen:
It looks like Stack is wanting GHC 7.10, but that doesn't work right for Idris on Travis yet, because the slower compilation times cause it to time out while building the dependencies.

david-christiansen:
In any case, the right way to add a Stack test would be to add to the Travis build matrix rather than having it be part of the main script. That way, we ensure that it still builds with ordinary cabal-install as well, and we get some parallelism.

jfdm:
I'm tempted to close this PR due to PR #2469 adding a stack configuration file. Unless there are any objections...

pharpend:
@jfdm That has the same problem that this originally had, where there weren't Travis tests with Stack. I don't use Travis on my own projects, so I don't know how to fix the issue.

LeifW:
To test "does Idris build with the deps specified by that stackage nightly", one could just drop the provided cabal.config into the dir, then cabal-install will use those constraints.

suhr:
Maybe use something else beside Travis? For example, drone.io supports containers from Docker index: http://readme.drone.io/usage/build/images/

david-christiansen:
Travis also has a new infrastructure available since we last poked at this that supports cached directories, so it should be reasonable to get it to install GHC 7.10.2 once, build all the dependencies into a single sandbox, then cache it for future builds.

jfdm:
So PR #2527 provides a stack file that allows Idris to be compiled using stack. If no one objects, as PR #2527 supersedes this PR, I shall close this at the end of day.

pharpend:
No objections.

PR #2431, opened 2015-07-11
Remove zombie dependency to zlib

melted:


LeifW:
Are .ibc's no longer zipped, or is that achieved some other way now?

melted:
It uses zip-archive so the various parts can be accessed independently and isn't one big Binary tangle.

melted:
Grumble, Travis is only happy if I wave a dead chicken over the commit and keep the zlib line in cabal. Well, let's merge this and remove that in a separate commit.

PR #2432, opened 2015-07-12
Fix copy-paste bug when reifying reflected tactics

bkoropoff:
Map `Case` to `CaseTac` (not `Induction`)


david-christiansen:
Nice catch! Thanks!

PR #2433, opened 2015-07-12
Remove zlib dependency

melted:


enolan:
I can't reproduce the test failures, and the segfault is *extremely* weird for what the change is. Seems likely there was a problem at Travis' end.

melted:
Yeah, it works fine on my end too. I can see one possible failure scenario, and that is that zlib > 0.6 is bugged (it still is pulled in by zip-archive) and that the constraint helps against that, the errors seem awfully odd in any case (sometimes it even gets Illegal instruction).

LeifW:
I can intermitently reproduce a segfault w/  GHC 7.10.2, hackage zlib-0.6.1.1, and c library zlib 1.2.8.
It occured during the `../../dist/build/idris/idris --build prelude.ipkg` command in `libs/prelude`.
One time gdb said it was a segmentation fault in ghczmprim_GHCziTypes_Izh_con_inf; the second time it was in ??.
The stacktrace for the second time is:
```
in ?? ()
#1  0x00007ffff7bceaea in inflateEnd () from /usr/lib/libz.so.1
#2  0x0000000001da23f6 in runCFinalizers ()
#3  0x0000000001da24d8 in scheduleFinalizers ()
#4  0x0000000001dae5fb in GarbageCollect ()
#5  0x0000000001da32e6 in scheduleDoGC ()
#6  0x0000000001da3ea9 in schedule ()
#7  0x0000000001da4ee7 in scheduleWaitThread ()
#8  0x0000000001da9b70 in hs_main ()
#9  0x000000000041163f in main ()
```
Is some zlib functionality being called from a GC finalizer or something? I'm confused.

Sounds like Agda also gets segfaults w/ zlib 0.6: https://github.com/agda/agda/issues/1518

LeifW:
Sounds like this'll be fixed in GHC 7.10.3: https://ghc.haskell.org/trac/ghc/ticket/10904
Or maybe there'll be another zlib release which works around this?

david-christiansen:
Thanks for following up on this!

LeifW:
Using a ghc build of the 7.10 branch in git, I can compile idris and have it compile its libs using zlib-0.6.1.1.

melted:
Let's open this so it's not out of sight.

It's good that it builds with 7.10, but there really should be an updated zlib, so 7.6 and 7.8 will work too.

But if we can't get that, can we fix up idris.cabal so the zlib dependency is pinned at the older version just for ghc 7.8 and older?

enolan:
I'm pretty sure Leif meant it builds with the 7.10 branch of GHC in git. i.e. we'd have to wait for 7.10.3 if they decide to make one. From the GHC wiki: [We have not yet decided when, or even whether, to release GHC 7.10.3.](https://ghc.haskell.org/trac/ghc/wiki/Status/GHC-7.10.3) Although from the ghc-devs mailing list discussion it seems we are not alone in having major problems with 7.10.2.

We should hold off on the zlib version bump until the next GHC release and then either require latest GHC or use some cabal magic to switch between zlib versions based on compiler version.

LeifW:
The proposal on the corresponding Agda ticket was to make the constraint conditional on the GHC version used. Something like "if GHC < 7.10.3, zlib < 0.6.1"? So people on newer GHC can use the latest zlib version.

PR #2434, opened 2015-07-13
Add an export list to Idris.REPL

melted:
Also remove some commented out imports.


melted:
Added some more export lists

PR #2436, opened 2015-07-13
Add a make target to build an Idris without optimization

melted:
The use case is having a fast compile when developing.


melted:
Build time with this was 2x faster when doing a clean build. It is a bit faster than that when ibcs have already been build as that build uses a slower Idris.

PR #2437, opened 2015-07-14
Fix missing case in `casetac`

bkoropoff:
This looks like a refactoring snafu.


Melvar:
It looks like that’s going to come out as, for example, `Can't doinductionhere`, missing a couple of spaces.

bkoropoff:
Fixed

david-christiansen:
Oh the joys of type inference...

david-christiansen:
Thanks!

PR #2440, opened 2015-07-15
Detect outdated IBC files in 'buildTree' function in Chaser.hs. Fixes #1163.

### LINKED ISSUE #1163: idris complains about outdated ibc files instead of replacing them
The error message is

    Incorrect ibc version --- please rebuild

The resolution is to delete ibc files and rebuild them. The compiler should just do this automatically.

paulkoerbitz:
In addition to testing whether an IBC function is outdated we now also check if it is for an old version and set 'needsRecheck' to true in both cases.
    
IBC files with old versions are still found sometimes, my guess is that this happens because the dependency tracking of the IBC files is not 100% correct (in this case the respective code in 'buildTree' is not triggered). In most cases the IBC files with old versions are found and replaced.

### DIFF
--- a/src/Idris/Chaser.hs
+++ b/src/Idris/Chaser.hs
@@ -116,29 +116,26 @@ buildTree built fp = btree [] fp
                              else return []
                    mt <- idrisCatch (runIO $ getModificationTime fn)
                                     (\c -> runIO $ getIModTime src)
-                   ok <- checkIBCUpToDate fn src
-                   return [MTree (IBC fn src) ok mt ms]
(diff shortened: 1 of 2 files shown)


david-christiansen:
@paulkoerbitz Great to see you around here again! :-) And thanks for the patch!

paulkoerbitz:
@david-christiansen yeah, I wanted to get back for a while ... hopefully a bit longer this time :)

paulkoerbitz:
Abandoning this PR to fix the (<$>) import (that's what I get for building with ghc 7.10 I guess :) )

PR #2441, opened 2015-07-16
Detect outdated IBC files in 'buildTree' function in Chaser.hs. Fixes #1163.

### LINKED ISSUE #1163: idris complains about outdated ibc files instead of replacing them

paulkoerbitz:
In addition to testing whether an IBC function is outdated we now also check if it is for an old version and set 'needsRecheck' to true in both cases.

IBC files with old versions are still found sometimes, my guess is that this happens because the dependency tracking of the IBC files is not 100% correct (in this case the respective code in 'buildTree' is not triggered). In most cases the IBC files with old versions are found and replaced.

### DIFF
--- a/src/Idris/Chaser.hs
+++ b/src/Idris/Chaser.hs
@@ -116,29 +116,26 @@ buildTree built fp = btree [] fp
                              else return []
                    mt <- idrisCatch (runIO $ getModificationTime fn)
                                     (\c -> runIO $ getIModTime src)
-                   ok <- checkIBCUpToDate fn src
-                   return [MTree (IBC fn src) ok mt ms]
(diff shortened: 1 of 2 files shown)


paulkoerbitz:
Second attempt :)

melted:
Merged. You don't have to open another PR, you could just have modified the last commit and force-pushed it to the same branch with -f, then the PR would have been updated.

paulkoerbitz:
OK, thanks for the heads up on pushing to a PR.

PR #2442, opened 2015-07-17
More parser-driven highlighting

david-christiansen:
Now, expression keywords like "case", "let", and "of" are highlighted by the parser. Additionally, names embedded in directives like "%name" are highlighted post-disambiguation, to make it easier to see how they were disambiguated.


david-christiansen:
Reopening for Travis

PR #2444, opened 2015-07-17
Add inductive divMod to contrib

Melvar:
Also move the well-founded recursion one, having it all in Data.Nat didn’t make much sense to me.


david-christiansen:
Looks good! (pinging @nicolabotta - this is what you needed, right?)

david-christiansen:
Great! Thanks for fixing the module name!

nicolabotta:
David Christiansen <notifications@github.com> wrote:

> Looks good! (pinging @nicolabotta - this is what you needed, right?)

Thanks David, Melvar! Do you have an example of how to use the module to
implement

> divNatNZ : (m : Nat) -> (n : Nat) -> Not (n = Z) -> Nat

? Thanks, Nicola

david-christiansen:
It should be sufficient to project out the first argument from `MkDivMod` after running @Melvar's code.

nicolabotta:
David Christiansen <notifications@github.com> wrote:

> It should be sufficient to project out the first argument from MkDivMod after running @Melvar's code.

Right, but it is not obvious to me how this should be done. "Projecting
after running Melvar's code" means pattern matching on a value computed
by |divMod|: this is the only function in Melvar's module that returns a
value of type |DivMod m n| apart from the constructor. We could do
something like:

> divmodNatNZ : (m : Nat) -> (n : Nat) -> Not (n = Z) -> (Nat, Nat)
> divmodNatNZ m  Z    p = void (p Refl)
> divmodNatNZ m (S n) p with (divMod m n)
>   | MkDivMod k r rLTm = (k, r)

The problem here is that |MkDivMod k r rLTm| is a value of type |DivMod
(r + k * (S n)) (S n)| but the type of |divMod m n| is |DivMod m (S
n)|. Thus I am not sure this is the intended usage of |divMod|.

I am probably missing something obvious but I think it would be good if
there was a simple example illustrating how the module is meant to be
used for computing |divNatNZ|.

Best,
Nicola

Best,
Nicola

Melvar:
Something like this:

```
divNZ : (m, n : Nat) -> Not (n = Z) -> Nat
divNZ m Z notZ = void $ notZ Refl
divNZ m (S n') notZ with (divMod m n')
  divNZ (r + q * (S n')) (S n') notZ | (MkDivMod q r rLtn) = q
```

For better understanding, I don’t mix `n` and `n'` here, so `n = S n'` always.

nicolabotta:
Thanks Melvar! Best, Nicola

PR #2447, opened 2015-07-19
need to import `Data.Vect` and `Data.Fin` for `interp.idr` to compile

cantora:
I had to add these lines to the `interp.idr` sample in order to compile it. It would also be good to mention that these imports are needed in the relevant section of the tutorial.

### DIFF
--- a/samples/tutorial/interp.idr
+++ b/samples/tutorial/interp.idr
@@ -1,5 +1,8 @@
 module Main
 
+import Data.Vect
+import Data.Fin
+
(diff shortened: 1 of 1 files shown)


jfdm:
Hi cantora, thanks for the bump. Under the version of Idris that the example was written `Fin` and `Vect` were included in the Prelude (IIRC), and the examples, and tutorial were not updated to reflect that. The tutorial is included along the distribution in the `doc` folder. Please feel free to update the tutorial if you are up to it.

PR #2451, opened 2015-07-19
make light of `Data.Vect` and `Data.Fin` imports in the interp example

cantora:
the interp documentation section assumes that these types are in the prelude.


david-christiansen:
Thanks for this!

PR #2454, opened 2015-07-21
Abort compilation if compiling non-statically known ffi calls

### LINKED ISSUE #2043: Javascript callbacks registered but not called
Using idris 0.9.17 the following example doesn't work for me with the node or javascript code generators. There is something registered, using node I can see the process waiting for the timeout but the callback is not being called.

```haskell
module Main

setTimeout : (() -> JS_IO ()) -> Int -> JS_IO ()
setTimeout f ms = foreign FFI_JS "setTimeout(%0,%1)"
    (JsFn (() -> JS_IO ()) -> Int -> JS_IO ())
    (MkJsFn f) ms

main : JS_IO ()
main = setTimeout (\_ => putStr "Hello, World") 2000
```

The same thing happens when I try to register something with the browser's "addEventListener" method (something succesfully registered but callback not called).

melted:
fixes #2043, the other reported issue.

### DIFF
--- a/src/IRTS/CodegenJavaScript.hs
+++ b/src/IRTS/CodegenJavaScript.hs
@@ -1331,6 +1331,7 @@ translateBC info bc
   | TAILCALL n            <- bc = jsTAILCALL info n
   | FOREIGNCALL r _ (FStr n) args
                           <- bc = jsFOREIGN info r n (map fcall args)
+  | FOREIGNCALL _ _ _ _   <- bc = error "JS FFI call not statically known"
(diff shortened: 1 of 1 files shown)


melted:
As I wrote on the issue: The second issue is easily fixed, sadly there's no way short of "error" to propagate an error in JS codegen, as even the compile is a pure function from the AST to Text. So I added an error as is used in oither places there.

raichoo:
I will quite possibly turn this into a Monad Transformer Stack once I lay my hands on this thing again. I've got some ideas on how to increase the performance of the backend during compilation.

PR #2461, opened 2015-07-23
Fix #1134 by saving same info in ibc on check

### LINKED ISSUE #1134: idris --client bug?
I'm experiencing issues with `idris-vim` that I could trace back to `idris --client` seemingly buggy functionality in my system (haven't tried in other environments, my environment is described at the end of this issue).  

**Short version:** Commands like `:ac` issued within the _repl_ work as expected, but when issued from `idris --client` (from the same dir as the file and where the _repl_ is being executed) it misbehaves, in this case it doesn't add the `xs ys` patterns for the parameters (just like @apriori reported in https://github.com/idris-hackers/idris-vim/issues/10). 

**Long (possibly replicable) version:**  As a minimal test case: 

1. create a file called `vsimple.idr` in some `~/[code-dir]` with one line content:
  `vsimple : Num a => Vect n a -> Vect n a -> Vect n a`
2.   `~/[code-dir]$ idris vsimple.idr` (enter *repl*)
3. `*vsimple> :ac 1 vsimple`
4. get expected result : `vsimple xs ys = ?vsimple_rhs`
5. without leaving the *repl*, open a separate terminal window and from same directory as before try: `~/[code-dir]/$ idris --client :ac 1 vsimple`
6. I get `vsimple = ?vsimple_rhs` (problem!)
7. Now I can try as many times, and in any order, the steps 3 and 5 and I always get consistently the same results. So the *repl* gives me the expected answer and `idris --client` does not.

Interestingly, there's  one way to make the *repl* give me the same unexpected output as `idris --client` is giving me. All I have to do is: 

1.   `~/[code-dir]$ rm vsimple.ibc`
2.  `~/[code-dir]$ idris --check vsimple.idr` (after this I get an *ibc* file that is much smaller than the one that I had just deleted)
3. from the *repl* terminal window: `*vsimple> :r`
4. from the *repl* terminal window: `*vsimple> :ac 1 vsimple`
5. now the *repl* gives me: `vsimple = ?vsimple_rhs`

My environment: 

- Ubuntu 14.04
- Idris versions tried: tagged 0.9.12 and head (0.9.12-git:62705e4)
- Idris is built from a fresh cabal sandbox each time
- Idris binary is in the sandbox build path and accessed through a symlink from `/usr/bin/idris`, which is what's in my `$PATH` (assuming this is relevant since it was asked in https://github.com/idris-hackers/idris-vim/issues/11)
- let me know if there's anything else relevant in this context

Thanks!

melted:
Otherwise addClause will fail as the ibc lacks that info

### DIFF
--- a/src/Idris/Elab/Clause.hs
+++ b/src/Idris/Elab/Clause.hs
@@ -594,7 +594,7 @@ elabClause info opts (cnum, PClause fc fname lhs_in_as withs rhs_in_as wherebloc
         rep <- useREPL
         when rep $ do
           addInternalApp (fc_fname fc) (fst . fc_start $ fc) (delabMV i clhs) -- TODO: Should use span instead of line and filename?
-          addIBC (IBCLineApp (fc_fname fc) (fst . fc_start $ fc) (delabMV i clhs))
+        addIBC (IBCLineApp (fc_fname fc) (fst . fc_start $ fc) (delabMV i clhs))
 
         logLvl 5 ("Checked " ++ show clhs ++ "\n" ++ show clhsty)
         -- Elaborate where block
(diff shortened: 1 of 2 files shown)


melted:
This is a problem when --check creates an ibc that is then loaded to a repl.

PR #2463, opened 2015-07-23
Pass parameters to codegens with new cmdoption --cg-opt

### LINKED ISSUE #1482: Pass parameters to codegens
Implement a way to pass parameters down to the codegenerators that are separated out.

melted:
Fixes #1482


melted:
A problem here may be that the same mechanism is used to pass flags to other idris backend as the C backend uses for gcc flags. So what is done differs by backend.

PR #2468, opened 2015-07-25
Fix Multiplicative/Additive docs

benjamin-hodgson:
They had `(*)` and `(+)` the wrong way round :wink:


melted:
thanks for the fix!

PR #2469, opened 2015-07-25
Add Stackage Configuration for easy building

lenary:
Using a nightly stackage config to build on GHC 7.10, and there are some
dependencies that are more advanced than the most recent LTS stackage
config. This should just build using `stack build` in the Idris-dev root.

I had to add cheapskate as an "extra dependency" because it's not in any 
stackage configurations. This may be worth revisiting later.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ test/*[0-9][0-9][0-9]/*.exe
 tags
 TAGS
 src/Version_idris.hs
+.stack-work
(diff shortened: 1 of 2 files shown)


melted:
If Travis is happy I guess it's fine for us old-fashioned people.

lenary:
So I haven't changed the travis config to use stack yet, but at some point it might be a good idea, but I'll let the core team adopt/try stackage before I propose those changes.

Issue #2460, opened 2015-07-23
Feature Request: Same type multiple record fields in one line


xged:
One would like to write:

    record Person where
      constructor MkPerson
      name, surname, middleName : String

In Haskell: `data Person = Person {name, surname, middleName :: String}`

david-christiansen:
This seems like a nice extension to me. Want me to point you to the right spot in the Idris source to implement it?

xged:
I am leaving the implementation to somebody with a higher level of skill then mine ;)

david-christiansen:
Fair enough :-)

benjamin-hodgson:
I have opened a pull request (#2470) with this feature in it :smile:

david-christiansen:
#2470 fixed this. Awesome!

PR #2470, opened 2015-07-26
Multiple record fields on one line (#2460)

benjamin-hodgson:
This pull request addresses issue #2460.

You can now write this:
```
record Person where
  constructor MkPerson
  name, surname, middleName : String
```
which is equivalent to:
```
record Person where
  constructor MkPerson
  name : String
  surname : String
  middleName : String
```

Please review my code! I'm open to any feedback. I changed the parser a little and added a test for the new feature. I also updated the tutorial to reflect the new syntax.

### DIFF
--- a/docs/tutorial/typesfuns.rst
+++ b/docs/tutorial/typesfuns.rst
@@ -349,7 +349,7 @@ following:
     $ idris vbroken.idr --check
     vbroken.idr:9:23:When elaborating right hand side of Vect.++:
     When elaborating an application of constructor Vect.:::
-        Type mismatch between 
+        Type mismatch between
                 Vect (k + k) a (Type of xs ++ xs)
         and
                 Vect (plus k m) a (Expected type)
@@ -442,7 +442,7 @@ names, ``n`` and ``a``, which are not declared explicitly. These are
 Implicit arguments, given in braces ``{}`` in the type declaration,
 are not given in applications of ``index``; their values can be
 inferred from the types of the ``Fin n`` and ``Vect n a``
-arguments. Any name beginning with a lower case letter which appears 
+arguments. Any name beginning with a lower case letter which appears
 as a parameter or index in a
 type declaration, but which is otherwise unbound, will be automatically
 bound as an implicit argument. Implicit arguments can still be given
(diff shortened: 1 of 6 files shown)


melted:
Looks good! I wonder if it would be desirable to allow for trailing commas. When autogenerating stuff, it always saves a bit of work when you can just trail a separator :-)

benjamin-hodgson:
@melted Good idea. I can't think of any downsides to allowing a trailing comma. I'll just adjust [`commaSeparated`](https://github.com/idris-lang/Idris-dev/pull/2470/files#diff-2bc041c01060d2611cec3e843b9c78acR440) to use `sepEndBy1` instead of `sepBy1`.

david-christiansen:
What about trailing commas in other contexts, such as pairs, multiple-argument function types, and the like? I think that we should be consistent if we want to allow this kind of thing, and I'm not 100% convinced that it's a good idea (e.g. what if we want to add tuple sections at some point?).

david-christiansen:
@edwinb, what do you think about the trailing commas bit?

xged:
*Allowing trailing commas* has to be a separate issue.

david-christiansen:
@xged We don't typically have such a high degree of formal process in Idris development, so it's fine to have it here. But I don't want to press "merge" until we've got (a reasonable facsimile of) consensus.

melted:
I guess we can just skip the trailing comma, it's only a convenience, we should have better ways to generate idris code than building strings anyway.

ahmadsalim:
I think that since Idris does not have lexical macros (like C), I am unsure how large a problem it would be in practice. I believe that any reasonable un-parser, textual transformation tool, or programming language will allow you to intersperse commas between your field names anyway.
Although, this is of course only my opinion.

benjamin-hodgson:
It sounds like you folks are by-and-large against the idea of allowing trailing commas. Especially considering that it'd be a lot easier to _add_ support for trailing commas in the future than it would be to _remove_ support (which would potentially break previously-valid programs), I'm going to go ahead and remove it from this pull request.

david-christiansen:
@benjamin-hodgson Thanks so much with your patience here, and for improving Idris!

benjamin-hodgson:
@david-christiansen The pleasure is mine! :wink:

PR #2477, opened 2015-07-28
Remove |splitAt| from let-binding section

jeremy-w:
|splitAt| is still introduced and discussed in the case expressions section.
Its appearance in the let-binding section looks like an accident.


jfdm:
Cheers.

PR #2478, opened 2015-07-28
Update bst module reference to Btree

jeremy-w:
Also regularizes capitalization of the module name (btree => Btree).


jfdm:
Cheers

PR #2484, opened 2015-07-28
Sync the documentation with opChars in ParseHelpers.hs

melted:
hash and backslash was missing, and underscore isn't an operator char


jfdm:
I've restarted the build as it looked to have failed randomly.

PR #2492, opened 2015-07-31
Refinement & reorganization of Data.Matrix

BlackBrane:
Changes:

* __Reorganize Data.Matrix__ – Now only functions that don't involve any numeric operations are in `Data.Matrix.idr` itself, so things involving row/collumn manipulation. I've made separate modules `Matrix.Numeric` and `Matrix.Algebraic`, with identical functionality using operations from Num and algebraic classes (Ring, etc) respectively.

  There are a variety of cases where it might be easier to use one or the other version, like depending on how some code you're working with is implemented, or if you have a use-case that isn't numerical at all. But if anyone is set against this kind of duplication let me know.

* __New functions__ – I've added a few common matrix functions whose need I've encountered:
   * `det` – determinant for square matrices
   * `blockDiag` – combine two matrixes to make a new matrix in block diagonal form
   * `concatMatrix` – Flatten a marix of matrixes
   * `diag_` – make a square matrix from a vector of elements to go along the diagonal, with all other entries zero.

  If anyone is wondering I have tested them for correctness. 

* __Small naming improvement__ – This is just a minor namespace convenience issue. Last time around I added `sum`, `product` and `power` to be the algebraic class versions of the existing functions for Num. I've found that these always need to be namespace qualified, since those functions are in the Prelude, so it's just better to call them `sum'`, `product'` and `pow'`. (`pow'` in particular because `pow` is the Num-generic counterpart whereas `power` is a function on Nats.)


melted:
Thanks!

Melvar:
`sum'` appears to be just the same as `concat`, yes?

BlackBrane:
Oh yeah, you're right. They're exactly the same.

I think `sum'` is probably still good to have as a synonym, since it's a much more suggestive word in circumstances where the Monoid instance has nothing to do with concatenation. I'm sure I'd be thrown off seeing `concat` in that context.

Melvar:
Yes, the name is good for context. Perhaps it should be defined as concat, and not duplicating concat’s definition, though?

BlackBrane:
Yeah that sounds like a good idea, just to make clear to anyone reading this code that `concat` is also a thing. I've just snuck this correction into my other current (also very minor) open PR #2495.

PR #2494, opened 2015-08-02
Make --mkdoc follow the sourcedir

uwap:
If you had an ipkg with a sourcedir defined before using the --mkdoc option would fail saying that some source files are missing.
This is because it didn't follow the sourcedir.

The fix here is to add the sourcedir to the list of import directories.


melted:
Thanks for the fix!

uwap:
No problem. I found another bug on my way and I wasn't able to fix it yet. I will try to investigate it further and if I have enough information about it either make a pull request or post an issue =)

Issue #1836, opened 2015-01-05
Setup.hs not compatible with cabal-install 1.22.0.0


mietek:
Idris compiles and installs successfully using _cabal-install_ 1.22.0.0.  However, at runtime, there is a problem:
```
$ idris --libdir
/app/share/$abi/idris-0.9.15.1/
```

`$abi` looks like an unexpanded variable.  I would be expecting `x86_64-linux-ghc-7.8.4` in its place.

The problem does not occur when using _cabal-install_ 1.20.0.3.

mietek:
Looking at `dist/build/autogen/Paths_idris.hs`, both `libdir` and `datadir` seem wrong:
```
libdir     = "/app/lib/$abi/idriskey"
datadir    = "/app/share/$abi/idris-0.9.15.1"
```

Thanks to @david-christiansen for suggesting this may be a Cabal issue.

ahmadsalim:
This doesn't seem to be an issue with the latest `cabal-install` so I am closing this.

PR #2496, opened 2015-08-03
Upgrade to new Travis infrastructure

### LINKED ISSUE #2321: Cabal constraint solving hits the backjump limit
No problems with 0.9.18-git:db84bf8 on Debian stable (3.16.0-4-amd64 #1 SMP Debian 3.16.7-ckt9-3~deb8u1 (2015-04-24) x86_64 GNU/Linux) but attempts at installing the latest commit fail with
```
nicola@cirrus:~/src/Idris-dev-5f33e4a51a68339fcb813f5217c5f85c783afa7b$ make
cabal install 
Resolving dependencies...
cabal: Could not resolve dependencies:
trying: idris-0.9.18.1 (user goal)
trying: unordered-containers-0.2.5.0/installed-11c... (dependency of
idris-0.9.18.1)
next goal: hashable (dependency of
unordered-containers-0.2.5.0/installed-11c...)
rejecting: hashable-1.2.1.0/installed-7e8... (conflict: hashable =>
text==0.11.3.1/installed-e38..., idris => text>=1.2.1.0 && <1.3)
rejecting: hashable-1.2.1.0/installed-e5c..., 1.2.3.2, 1.2.3.1, 1.2.3.0,
1.2.2.0, 1.2.1.0, 1.2.0.10, 1.2.0.9, 1.2.0.8, 1.2.0.7, 1.2.0.6, 1.2.0.5,
1.2.0.4, 1.2.0.3, 1.2.0.2, 1.2.0.1, 1.2.0.0, 1.1.2.5, 1.1.2.4, 1.1.2.3,
1.1.2.2, 1.1.2.1, 1.1.2.0, 1.1.1.0, 1.1.0.0, 1.0.1.1, 1.0.1.0, 1.0.0
(conflict: unordered-containers => hashable==1.2.1.0/installed-7e8...)
Backjump limit reached (change with --max-backjumps).
Makefile:11: recipe for target 'install' failed
make: *** [install] Error 1
nicola@cirrus:~/src/Idris-dev-5f33e4a51a68339fcb813f5217c5f85c783afa7b$
```

### LINKED ISSUE #2364: Idris doesn't compile on GHC 7.10 because of prelude exporting <$>

### LINKED ISSUE #2443: Upgrade to new Travis infrastructure
Travis has new infrastructure and complains that we're on the old one. Upgrading to the new one would get us the ability to cache dependencies and faster built startup generally, so I think it's worth doing sooner rather than later.

The guide on how to do it is at: http://docs.travis-ci.com/user/migrating-from-legacy/

4e6:
# Travis build using new container-based infrastructure and caching

close #2443 Upgrade to new Travis infrastructure
close #2364 Idris doesn't compile on GHC 7.10 because of prelude exporting <$>
close #2321 Cabal constraint solving hits the backjump limit

## build process
A brief explanation of travis build steps:

+ `before_install`: setup build environment. Add current `ghc` and `cabal` executables on `PATH`. Print debug information
+ `install`: initialize shared SANDBOX (which is cached between builds). Install dependencies in the shared SANDBOX
+ `before_script`: generate idris distribution and unpack it to WORKDIR. Copy SANDBOX to WORKDIR (since cabal is not able to cleanup after the `copy` step and we want to cache only dependencies and not the results of each idris compilation)
+ `script`: build idris and run tests using the copy of sandbox

## build matrix
Idris is building via cabal `1.20` and ghc `7.6.3`, `7.8.4`, `7.10.2`. Since Travis doesn't support ghc `7.10` yet, build uses custom [hvr-ghc](https://launchpad.net/~hvr/+archive/ubuntu/ghc) PPA to install appropriate versions of cabal and ghc. All builds are defined explicitly in the `.travis.yml` as `env` list of environment variables `CABAL`, `GHC` and `TESTS`. Also d896abc fixes #2364 to be able to build with ghc `7.10` and `CI` (`-Werror`) flag

## cabal
Atm Idris can not be built with the latest cabal `1.22` due to #1836, this build uses cabal `1.20`

## cache
All builds share the same sandbox for dependencies which is cached between builds. This allows to speedup `install` step, since all dependencies are already installed in the sandbox.

## `install` options changes:
+ added `--force-reinstalls`. Since sandbox is shared between buids. We should be able to reinstall dependencies to proceed the build, if some deps changed.
+ removed `--max-backjumps=-1`. With the progress on #2135 Building Against LTS Haskell, seems that related #2321 is not an issue anymore.

## idris-llvm
llvm backend build was dropped. Currently it is outdated and failing to build against present Idris version. Since it was moved to separate repo at [idris-hackers/idris-llvm](https://github.com/idris-hackers/idris-llvm), I believe it should be tested separately.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,21 +21,28 @@ env:
   - CABAL="1.20" GHC="7.6.3"  TESTS="lib_doc"
   - CABAL="1.20" GHC="7.6.3"  TESTS="test_c"
   - CABAL="1.20" GHC="7.6.3"  TESTS="test_js"
+  - CABAL="1.20" GHC="7.6.3"  TESTS="test_llvm"
   - CABAL="1.20" GHC="7.8.4"  TESTS="doc"
   - CABAL="1.20" GHC="7.8.4"  TESTS="lib_doc"
   - CABAL="1.20" GHC="7.8.4"  TESTS="test_c"
   - CABAL="1.20" GHC="7.8.4"  TESTS="test_js"
+  - CABAL="1.20" GHC="7.8.4"  TESTS="test_llvm"
   - CABAL="head" GHC="7.10.1" TESTS="doc"
   - CABAL="head" GHC="7.10.1" TESTS="lib_doc"
   - CABAL="head" GHC="7.10.1" TESTS="test_c"
   - CABAL="head" GHC="7.10.1" TESTS="test_js"
+  - CABAL="head" GHC="7.10.1" TESTS="test_llvm"
 
 matrix:
+  fast_finish: true
   allow_failures:
+    - env: CABAL="1.20" GHC="7.8.4"  TESTS="test_llvm"
+    - env: CABAL="1.20" GHC="7.6.3"  TESTS="test_llvm"
     - env: CABAL="head" GHC="7.10.1" TESTS="doc"
     - env: CABAL="head" GHC="7.10.1" TESTS="lib_doc"
     - env: CABAL="head" GHC="7.10.1" TESTS="test_c"
     - env: CABAL="head" GHC="7.10.1" TESTS="test_js"
+    - env: CABAL="head" GHC="7.10.1" TESTS="test_llvm"
 
 cache:
   directories:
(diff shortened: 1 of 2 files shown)


melted:
I'm a bit afraid of the runtime of running it all on three GHC versions. I wouldn't mind cutting it down, say running just test_c on GHC 7.6, that should be enough to verify compatibility.

4e6:
First, a word about build time measurement. 
Travis is able to run builds in parallel. And in the end it shows CPU time (sum of build times) instead of actual passed time (time end - time start), see travis-ci/travis-ci#1486. Turned out that [this branch with warmed up cache](https://travis-ci.org/4e6/Idris-dev/builds/73947844) builds roughly the same time as current [master](https://travis-ci.org/idris-lang/Idris-dev/builds) - about 2 hrs 45 min CPU time. Longer build times (with cold cache) should not affect main (this) repository and pull requests since cache (if empty) is inherited from master branch [according to travis docs](http://docs.travis-ci.com/user/caching/#Fetching-and-storing-caches); but could affect forked ones (2-3 first builds will be slower until cache warms up)

GHC 7.6 builds simply remained from original build definition. I agree, 3x4 builds looks like overkill, and idea to left only `test_c` (if nobody is interested in other builds against 7.6) sounds reasonable to me.

4e6:
Uh... I just found an issue with GHC 7.10 build. I better close this one for now (I fill it will take quite some time to resolve) and reopen when fixed.

PR #2502, opened 2015-08-07
Bump dependency upper bounds

### LINKED ISSUE #2500: Please update dependency constraints
According to http://packdeps.haskellers.com/feed?needle=idris, idris doesn't support recent versions of several important packages, which makes it hard to compile the package in a distribution like NixOS. Most constraints can be bumped without any problem, only `zlib` actually needs code changes to support the current 0.6.x version.

david-christiansen:
Also, it looks like zlib is no longer used, so I removed it entirely
rather than bump it.

Fixes #2500.

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -904,7 +904,7 @@ Library
                 , filepath < 1.5
                 , fingertree >= 0.1 && < 0.2
                 , haskeline >= 0.7 && < 0.8
-                , lens >= 4.1.1 && < 4.12
+                , lens >= 4.1.1 && < 4.13
                 , mtl >= 2.1 && < 2.3
                 , network < 2.7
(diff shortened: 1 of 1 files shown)


david-christiansen:
Hm, looks like #2433 is having some issues with a similar change. Let's see what Travis does here.

melted:
The zlib thing is real. I can reproduce it locally, with intermittent segfaults when trying to compress the ibc files. I suppose we should report it to zip-archive and zlib and see if they can figure it out. First I'll try git hEAD on those and see if there's an unreleased fix.

LeifW:
The lens constraint has been removed, the utf8-string constraint change has been applied in another merged PR, that leaves the vector constraint bump and the zlib constraint removal. I have zlib 0.6.1.1 installed locally I'll test witih.

PR #2503, opened 2015-08-08
[WIP] Propagate more implicits through instance definitions

bkoropoff:
This causes a regression in effects003 as written.  I'm not sure if I'm on the right track here.

cc @david-christiansen 

-------

The old code generated a lot of extraneous implicit arguments
and depended on elaboration to sort things out, but this led
to confusing failures in more complex scenarios (e.g. using
associated types).  Example code which now works:

```idris
%default total

class Project (t : Type) where
  output : Type
  project : t -> output

instance Project t => Project (() -> t) where
  output = () -> output {t}
  project f = project . f
```

### DIFF
--- a/src/Idris/AbsSyntax.hs
+++ b/src/Idris/AbsSyntax.hs
@@ -973,10 +973,11 @@ setTypeCase t = do i <- getIState
 -- Dealing with parameters
 
 expandParams :: (Name -> Name) -> [(Name, PTerm)] ->
+                Maybe [Name] -> -- If present, names not in this list are only bound to placeholders
                 [Name] -> -- all names
                 [Name] -> -- names with no declaration
                 PTerm -> PTerm
-expandParams dec ps ns infs tm = en tm
+expandParams dec ps free ns infs tm = en tm
   where
     -- if we shadow a name (say in a lambda binding) that is used in a call to
     -- a lifted function, we need access to both names - once in the scope of the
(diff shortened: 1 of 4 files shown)


david-christiansen:
Associated types are often a bit brittle as things currently stand - unfortunately! There's a bit of design work to be done to make them work entirely right.

It's difficult for me to evaluate this code, or attempt to diagnose it, because neither the original nor the changes are well-described, and it's not very obvious when reading. Would you mind using your fresh understanding of it to insert some descriptive comments?

bkoropoff:
The regression seems to be a general problem with where clauses:

```idris
data Trivial : () -> Type where
     MkTrivial : Trivial ()

-- Works
foo : Trivial s -> String
foo MkTrivial = "Trivial"

-- Doesn't work
bar : Trivial s -> String
bar = inner where inner MkTrivial = "Trivial"
```

I think it has something to do with `s` being considered accessible within the where clause.  My patch causes this to now manifest in instance definitions because implicits from constraints are now propagated.

bkoropoff:
The issue seems to be that parameters added to clauses in where blocks are always bound to a variable.  This causes unification failures if matching on another parameter refines it.  Is it reasonable to not bind the parameter if it isn't used free in the RHS of the clause?

bkoropoff:
I think I got everything working.

bkoropoff:
Nevermind, I guess I had stale bytecode files that masked some bugs.

ahmadsalim:
Is this PR being actively being worked on or would it be possible to close it?

ahmadsalim:
@bkoropoff I will close the PR for now since it seems that there has been no active work on it, and the code has bitrotted.

If you want to continue working on this you are welcome to reopen the PR or send a new one.

Thank you for the effort so far!

PR #2504, opened 2015-08-08
Update testing section of tutorial

jeremy-w:
- Fix example maths.ipkg syntax (`module` => `package`, `module =` => `modules =`)
- Add captions to the `Math.NumOps` and `Test.NumOps` code blocks to show the expected file names
- Rewrite intro paragraphs; mostly just rewording, though the stress on how it's totally up to the test author to handle reporting is new.

### DIFF
--- a/docs/tutorial/testing.rst
+++ b/docs/tutorial/testing.rst
@@ -4,13 +4,21 @@
 Testing Idris Packages
 **********************
 
-As part of the integrated build system a simple testing framework is provided.
-This framework will collect a list of named functions and construct an Idris executable in a fresh environment on your machine.
-This executable lists the named functions under a single ``main`` function, and imports the complete list of modules for the package.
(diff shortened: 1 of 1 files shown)


jfdm:
It looks mostly fine, thanks for catching the typo's. However, I am not a fan over how you changed the error reporting used by the sample tests. Especially as the reporting style used was not explained. The previous style was more self explanatory.

jeremy-w:
@jfdm: On second thought, I am not happy with that, either!

The style I shifted to is what `rspec` and similar use for reporting success/failure. The output ends up looking like `........F....+....F` so `F`ailures and `+`pending tests stick out like a sore thumb. But it's too terse for a tutorial example. I'll shift it back.

jeremy-w:
@jfdm Should be corrected now.

PR #2508, opened 2015-08-11
Relax parameter bindings when passing free variables to where blocks

bkoropoff:
I'm submitting this separately from #2503 since it still has issues.  This patch in isolation seems to work fine.

---------------

When adding parameters to clauses in a where block, only bind them to
variables if they occur free on the right.  Otherwise, bind them to a
placeholder so they can unify with anything.  This fixes code such
as the following:

```idris
data Trivial : () -> Type where
     MkTrivial : Trivial ()

-- Works
foo : Trivial s -> String
foo MkTrivial = "Trivial"

-- Doesn't work
bar : Trivial s -> String
bar = inner where inner MkTrivial = "Trivial"
```

### DIFF
--- a/src/Idris/AbsSyntax.hs
+++ b/src/Idris/AbsSyntax.hs
@@ -973,10 +973,11 @@ setTypeCase t = do i <- getIState
 -- Dealing with parameters
 
 expandParams :: (Name -> Name) -> [(Name, PTerm)] ->
+                Maybe [Name] -> -- If present, names not in this list are only bound to placeholders
                 [Name] -> -- all names
                 [Name] -> -- names with no declaration
                 PTerm -> PTerm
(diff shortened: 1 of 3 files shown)


edwinb:
The reason parameters are propagated to where blocks is that parameters are genuinely things which never change, and the code will work completely independently of what the parameter is instantiated with. So this patch is really solving the wrong problem - the issue is that 's' isn't actually a parameter because it does matter what it is!

I don't know why Idris has decided that 's' is a parameter, because that isn't a parameter position in 'Trivial'. I'll take a closer look (but not immediately, I'm afraid.)

bkoropoff:
@edwinb Thanks for the explanation.  I'd also appreciate any thoughts you have on the correct way to fix #2503 since it's the original bug I ran in to.

PR #2510, opened 2015-08-11
Add libFFI to extra-deps, and default flags to use gmp and ffi

jfdm:
+ LibFFI is not in stackage and needs to be added as an extra dep.
+ To build Idris with FFI and GMP the flags need to be passed. Best add them to the stack file.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,6 +1,10 @@
-flags: {}
+flags:
+  idris:
+    gmp: true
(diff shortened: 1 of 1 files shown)


jfdm:
We may want to hold merging this. LibFFI is not installing on my Mac machine using stack.

jfdm:
I am removing this PR.

PR #2515, opened 2015-08-12
Corrected "List" to "Vect" in "Maybe" example

Enamex:
Example for using `Maybe` for out of bounds checking, function `list_lookup`.
I assume that was the intention. Not sure how a `List` can be accessed out of bounds, unless I'm missing something.


jfdm:
The original intention is correct. The problem with the function is related to the index given in relation to the list to be indexed. Think about what happens if the index given is greater than the size of the list. What should the function then return.

Enamex:
Actually, I was misreading it but not in this way (was looking for the pattern-matched `Nat` as a part of the `List a` parameter for some reason).

That said, I'm curious how my edit 'failed'. I didn't mess with any formatting.

PR #2516, opened 2015-08-12
Fix indentation in PCAF parser

### LINKED ISSUE #2512: IDE server exits rather than returning error when irrefutable pattern exception encountered
## Summary
Clients of an IDE-mode Idris would expect the server to keep running in the face of bogus file input and simply return an error to show to the user. Instead, the server process will exit.

## Configuration
Idris version 0.9.18 (installed via Homebrew)

ProductName:	Mac OS X
ProductVersion:	10.10.4
BuildVersion:	14E46

## Steps to reproduce
1. Write this to a file:
```
let T = 3
using (G: T)
```
2. To demonstrate the idris executable exiting on attempting to parse, run it: `idris -c file.idr`
2. To demonstrate the server exiting:

```
~/D/D/idris> idris --ide-mode-socket &                                                                                                   [1]
64645
~/D/D/idris> echo "00004b((:load-file \"$PWD/crash_me.idr\") 1)" | nc 0.0.0.0 64645                                                      [1]
000018(:protocol-version 1 0)
idris: src/Idris/ParseHelpers.hs:429:20-46: Irrefutable pattern failed for pattern (x : xs)

'idris --ide-mode-socket &' has ended
```

Note that the `00004b` bit is the length of the following string; it'll vary depending on the path to your file. You can calculate it at the command-line using:

```
dc -e "16 o $(echo ((:load-file \"$FILE_PATH_HERE\") 1) | wc -c) p"
```

then prefix it with zeroes so that it becomes a 6-character length field.

## Workaround

Currently, I have to relaunch the entire Atom editor, since the package does not expect the server process to terminate: [idris-hackers/atom-language-idris:#51](https://github.com/idris-hackers/atom-language-idris/issues/51). It's possible to modify the clients to handle the termination case; I don't know that the clients will be able to provide the error message to the user, though.

david-christiansen:
Fixes #2512. Additionally, the crash on indentation stack underflow is now more informative, should it occur in the future.

### DIFF
--- a/src/Idris/ParseHelpers.hs
+++ b/src/Idris/ParseHelpers.hs
@@ -450,8 +450,9 @@ pushIndent = do pos <- position
 -- | Pops indentation from stack
 popIndent :: IdrisParser ()
 popIndent = do ist <- get
-               let (x : xs) = indent_stack ist
-               put (ist { indent_stack = xs })
+               case indent_stack ist of
+                 [] -> error "The impossible happened! Tried to pop an indentation level where none was pushed (underflow)."
+                 (x : xs) -> put (ist { indent_stack = xs })
 
 
 -- | Gets current indentation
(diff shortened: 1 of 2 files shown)


david-christiansen:
The tests all pass locally for me. Travis seems to be hung up on some other nonsense, so I'll go ahead and just merge it.

PR #2518, opened 2015-08-12
Document interactive theorem proving as deprecated.

jeremy-w:
Puts "DEPRECATED:" in the chapter title and also adds a warning paragraph
at the start of the tutorial chapter.

### DIFF
--- a/docs/proofs/interactive.rst
+++ b/docs/proofs/interactive.rst
@@ -1,6 +1,12 @@
-***************************
-Interactive Theorem Proving
-***************************
+***************************************
+DEPRECATED: Interactive Theorem Proving
+***************************************
+
+.. warning::
+   **Interactive theorem-proving has been deprecated.**
+   Support for tactic-based theorem proving will eventually be removed from
+   Idris.
+   Existing tactic-based proofs should be rewritten to use ``rewrite`` instead.
 
 Idris also supports interactive theorem proving via tactics. This
 is generally not recommended to be used directly, but rather used as a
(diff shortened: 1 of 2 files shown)


david-christiansen:
I think that, perhaps, we should refer them to the section on `Elab` reflection, and not just to the `rewrite` keyword. That will be the eventual replacement for the old-style tactic language. It's almost ready too - I just need to finish implementing `induction` and it will have all the features.

david-christiansen:
@edwinb What do you think about this declaration of deprecation?

jeremy-w:
@david-christiansen I'm trying to rewrite it as you requested, but for some reason, Sphinx doesn't seem to like a cross-reference in a content block. It's showing the literal `ref:foo` in the admonition content block; the same markup in the body text works fine and substitutes in the "Elaborator Reflection" section title with a link there as expected. :\

jeremy-w:
@david-christiansen Got it! The issue turned out to be that the bold markers `**` were somehow preventing the markup from being handled correctly. Yay edge cases. New commits incoming.

david-christiansen:
Thanks for this! I do want to make sure that @edwinb is on board with this change in advertising before we merge it, though.

edwinb:
I think this is a good idea to make clear. For various reasons I don't want to remove tactics for a while (mostly that there's quite a bit of research going on that uses them...) but it's better if we encourage people to use the elaborator reflection in the end.

david-christiansen:
@edwinb Didn't you say on IRC that you'd prefer to mark them as deprecated for the release after the upcoming one? If that's the case, then we should probably merge this one immediately after the release.

PR #2519, opened 2015-08-13
Add "choice"

david-christiansen:
This is called asum in Haskell. I think it's appropriate for the Prelude because I've defined it myself on a number of occasions. What do you all think?

### DIFF
--- a/libs/prelude/Prelude/Applicative.idr
+++ b/libs/prelude/Prelude/Applicative.idr
@@ -5,6 +5,7 @@ import Builtins
 import Prelude.Basics
 import Prelude.Bool
 import Prelude.Classes
+import Prelude.Foldable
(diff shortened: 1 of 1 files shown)


jfdm:
I see no obvious reason to say no but I am not familiar with `asym` nor what it is supposed to do. After reading the docs for `asym` I still see no obvious reason to say no, but maybe request better docs that 'like x in haskell' and maybe a small example.

jeremy-w:
@jfdm: If you have a left-biased alternative operator `<|>`, then `choice` performs left-biased choice from a list of alternatives. If you were writing a parser expression like `expr = literal <|> keyword <|> funcall`, you could rewrite this using `choice` as `expr = choice [literal, keyword, funcall]`.

PR #2522, opened 2015-08-13
Fix substitution when applying syntax rules

### LINKED ISSUE #1283: A few things don’t work inside syntax
Using a syntax rule with a do block on the RHS causes a crash because [`Idris.ParseExpr.extension where update where upd`](https://github.com/idris-lang/Idris-dev/blob/master/src/Idris/ParseExpr.hs#L175-L185) is missing a case for `[]`. This is how I found this problem.

Additionally, [`Idris.ParseExpr.extension where update`](https://github.com/idris-lang/Idris-dev/blob/master/src/Idris/ParseExpr.hs#L155-L187) does not have a case for `PDisamb`, so using a syntax rule that mentions a syntax-bound variable within a `with NS expr` expression causes a “no such variable” error, because the said variable isn’t being replaced when the syntax rule is applied. The same appears to be true of `PRefl`, `PEq`, `PRewrite`, `PIdiom`, `PProof` and `PTactics`. I could probably fix these, but there are a few more constructors I can’t identify immediately, so I decided to report first in case one of the rest also needed to be `update`d under.

Melvar:
Fixes #1283

### DIFF
--- a/src/Idris/ParseExpr.hs
+++ b/src/Idris/ParseExpr.hs
@@ -215,6 +215,8 @@ extension syn ns rules =
             upd ns (DoLetP fc i t) = DoLetP fc (update ns i) (update ns t)
     update ns (PIdiom fc t) = PIdiom fc $ update ns t
     update ns (PMetavar fc n) = PMetavar fc $ updateB ns n
+    update ns (PProof tacs) = PProof $ map (updTactic ns) tacs
+    update ns (PTactics tacs) = PTactics $ map (updTactic ns) tacs
     update ns (PDisamb nsps t) = PDisamb nsps $ update ns t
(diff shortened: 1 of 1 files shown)


Melvar:
Kinda baffled about this test failure.

david-christiansen:
I re-ran the failing test, and now it's fine. Huh.

Melvar:
See perhaps also https://github.com/idris-lang/Idris-dev/commit/46bdfa9b897f2ac487a0f6e313aa73ce2f61a48a . It’s a TCon number varying for some reason it seems.

david-christiansen:
Thanks for all your hard work on this! It looks good to me.

PR #2527, opened 2015-08-14
Update stack configuration to LTS Haskell 3.0

jfdm:
The stack configuration file has been updated to use LTS Haskell 3.0 resolver, and the inclusion of dependencies for the ffi and curses flags.

Flags are called per package either in the stack file or with cmd opts of `--flag <pkg name>:<flag>` before I wanted to explicitly state the flags but I think it best we do this using command line options.

Further to get Idris to use LTS Haskell, lens had to be bumped up to the latest version used in LTS Haskell.

I have tested this against Idris HEAD using the following setup:

+ Ubuntu LTS 14.04.3
+ GHC 7.10.2
+ Cabal-Install 1.22.6.0 with library 1.22.4.0
+ Stack Version 0.1.2.0 with resolver lts-3.0

and command

    stack install --flag idris:FFI --flag idris:GMP

My only concern is building Idris using Stack and the other flags i.e. curses and CI and release, and on other platforms i.e. Mac OS X, and Windows.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ addons:
       - expect
       - cppcheck
       - hscolour
+      # Stack dependencies
+      - libgmp-dev
 
 matrix:
   include:
@@ -40,6 +42,7 @@ cache:
   directories:
     - $HOME/.cabsnap
     - $HOME/.cabal/packages
+    - $HOME/.stack
 
 before_cache:
   - rm -fv $HOME/.cabal/packages/hackage.haskell.org/build-reports.log
@@ -49,6 +52,10 @@ before_install:
   - unset CC
   - export PATH=/opt/ghc/$GHCVER/bin:/opt/cabal/$CABALVER/bin:$HOME/.cabal/bin:$PATH
   - env
+  - mkdir -p ~/.local/bin
+  - export PATH=$HOME/.local/bin:$PATH
+  - travis_retry curl -L https://github.com/commercialhaskell/stack/releases/download/v0.1.4.0/stack-0.1.4.0-x86_64-linux.tar.gz | tar xz -C ~/.local/bin
+
 
 install:
   - cabal --version
@@ -107,3 +114,5 @@ script:
       echo "make -j2 $test";
       make -j2 $test;
     done
+  # Stack test to build Idris, using system ghc.
+  - stack --no-terminal --system-ghc build
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,9 @@ matrix:
       compiler: ": #GHC 7.10.2"
     - env: CABALVER="1.22" GHCVER="7.10.2" TESTS="test_c"
       compiler: ": #GHC 7.10.2"
+    - env: CABALVER="1.22" GHCVER="7.10.2" STACKTEST="stack_test"
+      compiler: ": #GHC 7.10.2"
+
 
 cache:
   directories:
@@ -102,19 +105,23 @@ before_script:
   - cd ..
   - tar -xf Idris-dev/dist/idris*.tar.gz
   - cd idris*
+
 script:
-  - cabal configure -f FFI -f CI
-  - cabal build
-  - cabal copy
-  - cabal register
-  - if [[ "$TESTS" == "test_c" ]]; then
-      cppcheck -i 'mini-gmp.c' rts;
-    fi
-  - for test in $TESTS; do
-      echo "make -j2 $test";
-      make -j2 $test;
-    done
-  # Stack test to build Idris, using system ghc if possible.
-  - if [[ "$GHCVER" == "7.10.2" ]]; then
-      stack --no-terminal --system-ghc build
+  - if [[ "$STACKTEST" == "stack_test" ]];
+    then
+      stack --no-terminal --system-ghc build;
+    else
+      cabal configure -f FFI -f CI;
+      cabal build;
+      cabal copy;
+      cabal register;
+      if [[ "$TESTS" == "test_c" ]];
+      then
+        cppcheck -i 'mini-gmp.c' rts;
+      fi
+      for test in $TESTS;
+      do
+        echo "make -j2 $test";
+        make -j2 $test;
+      done
     fi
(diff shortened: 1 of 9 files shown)


david-christiansen:
I think we should add this to Travis, don't you?

jfdm:
Depends on what you mean by 'add'? Do you mean we should migrate Idris's build and test infrastructure over to stack? Or something else? When it comes to migration and changes in DevOps I want to be as conservative as possible. 

[Now there is documentation on how to use stack for testing on travis](https://github.com/commercialhaskell/stack/wiki/Travis), and it does appear to be straightforward. However, I am not too versed on travis nor how best to move forward with a migration and what that means to ghc version support and the release process for Idris.

A more conservative approach in the mean time might be to use the LTS-3.0 `cabal.config` to ensure that at least we use the curated package set for development

jfdm:
Just thinking about it, when using stack. The default libdir for Idris ended up being in the local install directory for the Idris source. Maybe that is something we need to consider before moving over.

LeifW:
Lens can be removed entirely. We don't use it. Don't care what version it is, just as long as we get trifecta installed. See #2554.
Also, lts-3.2 is out now, FYI.

jfdm:
Ah cheers, I must have branched prior to the merging of #2554. Will remove that line. Also with respect to the bumping of LTS-3.2. I am inclined for the moment to only follow major versions of LTS Haskell. [It is my understanding](https://github.com/fpco/lts-haskell#how-it-works) that the different between LTS 3.0, LTS 3.1, and LTS 3.2 are towards minor changes in a package. I think we should be conservative for now, and follow the Major series (i.e. 3) unless there is an absolutely need to bump up the minor release.

LeifW:
That link makes it sound like we could use whatever the latest 3.X release is to obtain "bug fixes and feature enhancements", and hopefully everything should remain API compatible on 3.X. -  "By following the pattern, you can target the 7.X series of releases and have a high degree of confidence that you will continue to get bug fixes and feature enhancements, without needing to change your code to match upstream API changes."

jfdm:
Just a note that Idris cannot install with LTS Haskell 3.0 nor 3.2. The constraints specified for `annotated-wl-pprint` in the cabal file fall out with the version that stack has. Stackage knows abotu version 0.6.0 but we require 0.7 and greater.

scshunt:
Hopefully 3.5 will contain 0.7.0. It's already in the nightly.

NightRa:
Also, cheapskate is already in the stackage snapshot, so no need to explicitly include it in ```extra-deps```.

scshunt:
Yes, that only changed yesterday. It seems that we can now compile fine against the nightly without any issues.

jfdm:
@scshunt Cheers for the update, I have yet to check my branch against Stackage. My main preference is to get Idris to compile against a LTS Haskell version, rather than a nightly. I will check again ASAP.

jfdm:
i've had a quick look at stackage again. If I understand the version-ing policy correctly, then 'wl-print-annotated` will only get bumped to 0.7.0 n the next Major release: **LTS Haskell 4.0**. Maybe we should target that for the version we want to build Idris against...

david-christiansen:
Why not just point it at one of the Stackage nightlies? Is there any downside to that?

jfdm:
You mean build against a working nightly for now, and then switch out to an LTS version when the changes in the nightly release have propagated down to the LTS version. Obviously, switching to a newer nightly when appropriate.

david-christiansen:
That seems like the easy approach. As far as I can see, the only real advantage of the LTS Haskell over a nightly is that Stack users are likely to have more of the packages in the binary cache, so it'll build a bit faster.

NightRa:
And now with the improved shared binary cache, this is even less costly.

2015-09-16 12:22 GMT+03:00 David Christiansen <notifications@github.com>:

jfdm:
Idris builds once again with Stack using the latest nightly. I have updated the installation instructions as well, taking note that stack support is experiemental. I am happy for this to be merged after the okay from the majority of committers.

david-christiansen:
I see that I forgot to answer a question above about what I meant with Travis. I think that if we support a build tool, we should get it in CI, so we should add a job to the Travis file that triggers a stack build. Is that possible with a reasonable amount of effort? What I don't want to do is increase user pain by claiming to support a build method that suddenly stops working without warning, so it would be good to know if we break it somehow. E.g. with a new library dependency that needs to get onto the curated list.

jfdm:
I agree to the notion of supporting a build tool we should support tests for it. I do not know off hand how easy it is to do, but I also am not sure I can spend much time on this in the next two months. Moreso, how this build tool will work on platforms for all possible configurations is not entirely clear to me at the moment.

 I shall add a WIP tag to this PR. If someone want to build on top of these efforts that is okay by me.

NightRa:
I've seen an example by Michael Snoyman that it's quite simple to add a stack build to Travis.
Although I don't have any experience with Travis so I can't do it myself.
Also, @kmett did a multi-ghc build on travis using stack.

david-christiansen:
I'm not advocating that we exclusively build on Travis with Stack, just that we run it as one more job to make sure we remember to update the config.

david-christiansen:
Prob. it's a mere matter of shell scripting to add one more environment variable to the build matrix.

jfdm:
@david-christiansen Yeah, I gathered that from your comment. My travis foo is none existent, but i may have a look this weekend and work out how to add another target.

kerscher:
It builds against the latest LTS on Linux:
```yaml
flags: {}
packages:
- '.'
extra-deps:
- annotated-wl-pprint-0.7.0
- cheapskate-0.1.0.4
- libffi-0.1
resolver: lts-3.11
```
My system is similar to the one @jfdm has:

* Ubuntu LTS 14.04.3
* Stack 0.1.6.0 (e22271f5ce9afa2cb5be3bad9cafa392c623f85c)

jfdm:
I am deprecating this PR in favour of #2845 for updating the stack configuration. Updating travis, and improved docs can come later.

PR #2538, opened 2015-08-20
javascript: support prim_eqPtr

raichoo:


Melvar:
Is that really supposed to be `==` and not `===`?

raichoo:
Type checking is already done at that point. Also `===` is somehow a pointless exercise IMHO since it just yields false more often.

PR #2546, opened 2015-08-25
Updated Logging Effect.

jfdm:
+ Fixed namespace of default logger so default is called from module `Effect.Logging`
+ Reworked implementation of default logger to use a record as the
  underlying resource, and added new methods to initialise and
  change the logging levels and logging categories.
+ Simple logger also has method to adjust logging levels.


jfdm:
Ah, I see the error, I forgot to return the result of the effectful function, will fix.

jfdm:
Right, as far as I am aware I am happy for the effectual logger improvements to be merged now. If someone could do a numpty check then please do so.

Melvar:
It would be nice if you could make your `Show` instance parenthesize properly by defining `showPrec`.

Melvar:
Weird that this doesn’t compile in Travis … I assume it works locally, since it also did for me.

jfdm:
@Melvar, this doesn't work locally for me anymore. It once did, but I think a recent change to how instances work may have changed behaviour. I honestly do not know...it would be nice to have this in the upcoming release.

PR #2547, opened 2015-08-26
Fix show for negative numbers

Melvar:
Fixes the parenthesization when showing negative numbers that are arguments to a constructor.

The extra complexity in the `Show` instance then upset primitives002 by not letting the show operations get inlined, so now the backends were doing the converting, and they do it differently than the Haskell primitive. While taking a look, I noted that the actual primops under scrutiny were getting inlined (still) and thus executed by the evaluator. Thus the test did not test what it was supposed to test. I had to fix it to make things work again, which revealed that the JS backend should never have passed the test. Getting sane output that both backends agreed on involved adjusting the C backend to show greater precision and fixing all the comparison primops in the JS backend.


david-christiansen:
It looks to me like you've made the primitive operation that converts `Double` to `String` drop the trailing `.0` from the representation. Is this what we want? As far as I can see, only JavaScript does this - at least Haskell, Python, and Julia all output `0.0` as `"0.0"`.

Melvar:
C and JS agree on this, and it’s not obvious to me how to change this easily in either, so I simply went with what the two tested backends say. If you wish, I will slap on extra string manipulation to stick on the `.0` in each backend’s runtime.

I’m just now looking at the paper describing the proper algorithm for this, but an Idris implementation will require at least one new primitive, which I will now research how to implement in C and JS if possible.

david-christiansen:
On my machine, the `"%f"` format specifier to `printf`prints `0.000000`. Still not `0.0`, but it's also not `0`.

Melvar:
`%f` doesn’t use exponential format and it will always output the same number of digits after the decimal point, meaning that very small-magnitude numbers, including zero, will render as `0.0000000000000000` even once the precision is set to sufficient levels. `%g`, which is currently used, actually uses its precision for the total number of significant digits, which I consider much better for our use case. Compare the output of `%.16f` and `%.16g` on `1.0e-17`, and perhaps also on `1.0e17`.

In all, I would say `%f` causes more problems than not having `.0` on integral floats.

Melvar:
Also consider that these primops never added the `.0` before; this fact was only masked because the test actually used the evaluator’s implementation, which is Haskell’s show implementation which does add it. So strictly speaking it’s not worse than before.

david-christiansen:
You're right about that! So to sum up the state of Idris + this patch:

1. The C and JS backends now have more similar behavior than before
2. They continue to disagree with the evaluator's implementation of the primop
3. The tests are testing what they should, which is exposing the disagreement with the evaluator rather than hiding it
4. Showing negative numbers now properly parenthesizes

Is that an accurate summary?

Melvar:
Well, it is the implementation of 4. that exposed this because apparently now those show calls aren’t inlinable anymore, even if the test had not been fixed as per 3..

To 1. they still disagree on exactly what range they will use exponential notation for, but that is not exposed by this test, which is strictly supposed to be about the *presence* of the prims in a given backend. The test also does not care about trailing digits on long numbers, and there might potentially be some disagreement there as well.

This trouble definitely has me thinking about implementing this in Idris code rather sooner than I would otherwise have cared.

david-christiansen:
All right, I think this is worth merging then.

PR #2561, opened 2015-08-28
Check for TTY when setting colour

melted:
This means that when doing a redirect to file, one doesn't have to turn
off colour manually.

@david-christiansen not anything to do with the Windows commit, just a tweak I saw now that I don't live in a monochrome world.


david-christiansen:
This seems like a good idea to me.

PR #2569, opened 2015-09-02
Fix #2241 by adding a customized display for records documentation

### LINKED ISSUE #2241: Records should have their own documentation display

GuglielmoS:
Fix #2241

Records documentation is now reported with a customized display that lists parameters, constructor and projections.
The changes are visible in the REPL (with :doc) or in the HTML documentation (--mkdoc). 

Examples of code:
```idris
import Data.Vect

record Person where
    constructor MkPerson
    firstName : String
    middleName : String
    lastName : String
    age : Int

record SizedClass (size : Nat) where
    constructor SizedClassInfo 
    students : Vect size Person
    className : String
    
record Prod a b where
    constructor Times
    fst : a
    snd : b
```
Example of documentation:
```
> :doc Person
Record Person

Constructor:
    MkPerson : (firstName : String) ->
        (middleName : String) ->
        (lastName : String) -> (age : Int) -> Person
        
        
Projections:
    firstName : (rec : Person) -> String
        
        
    middleName : (rec : Person) -> String
        
        
    lastName : (rec : Person) -> String
        
        
    age : (rec : Person) -> Int
        
        
> :doc SizedClass
Record SizedClass

Parameters:
    size : Nat

Constructor:
    SizedClassInfo : (students : Vect size Person) ->
        (className : String) -> SizedClass size
        
        
Projections:
    students : (rec : SizedClass size) -> Vect size Person
        
        
    className : (rec : SizedClass size) -> String
        
        
> :doc Prod 
Record Prod

Parameters:
    a : Type, b : Type

Constructor:
    Times : (fst : a) -> (snd : b) -> Prod a b
        
        
Projections:
    fst : (rec : Prod a b) -> a
        
        
    snd : (rec : Prod a b) -> b
        
        

```

### DIFF
--- a/src/Idris/AbsSyntax.hs
+++ b/src/Idris/AbsSyntax.hs
@@ -397,6 +397,10 @@ addClass n i
                       _ -> i
         putIState $ ist { idris_classes = addDef n i' (idris_classes ist) }
 
+addRecord :: Name -> RecordInfo -> Idris ()
+addRecord n ri = do ist <- getIState
+                    putIState $ ist { idris_records = addDef n ri (idris_records ist) }
+
 addAutoHint :: Name -> Name -> Idris ()
 addAutoHint n hint =
     do ist <- getIState
--- a/src/Idris/AbsSyntaxTree.hs
+++ b/src/Idris/AbsSyntaxTree.hs
@@ -177,6 +177,7 @@ data IState = IState {
     idris_implicits :: Ctxt [PArg],
     idris_statics :: Ctxt [Bool],
     idris_classes :: Ctxt ClassInfo,
+    idris_records :: Ctxt RecordInfo,
     idris_dsls :: Ctxt DSL,
     idris_optimisation :: Ctxt OptInfo,
     idris_datatypes :: Ctxt TypeInfo,
@@ -336,7 +337,7 @@ idrisInit = IState initContext S.empty []
                    emptyContext emptyContext emptyContext emptyContext
                    emptyContext emptyContext emptyContext emptyContext
                    emptyContext emptyContext emptyContext emptyContext
-                   emptyContext
+                   emptyContext emptyContext
                    [] [] [] defaultOpts 6 [] [] [] [] emptySyntaxRules [] [] [] [] [] [] []
                    [] [] Nothing [] Nothing [] [] Nothing Nothing [] Hidden False [] Nothing [] []
                    (RawOutput stdout) True defaultTheme [] (0, emptyContext) emptyContext M.empty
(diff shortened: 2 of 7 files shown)


david-christiansen:
This is wonderful! Thanks so much!

What does the output look like for a record that has documentation? E.g.
```
||| A foo.
||| @ the thing in the foo
record Foo a where
  ||| Build me a foo!
  constructor MkFoo
  ||| Make a thing from a nat.
  ||| @ x the natural number
  blagh : (x : Nat) -> a
```

david-christiansen:
Also, don't forget to add yourself to the list of contributors, and to put this in the changelog :)

GuglielmoS:
The output for Foo is the following:
```
> :doc Foo
Record Foo
    A foo.

Parameters:
    a : Type

Constructor:
    MkFoo : (blagh : Nat -> a) -> Foo a
        Build me a foo!
        Arguments:
            (implicit) a : Type  -- the thing in the foo
            
            blagh : Nat -> a  -- Make a thing from a nat.
            
Projections:
    blagh : (rec : Foo a) -> (x : Nat) -> a
        Make a thing from a nat.
        
```
Could this be ok? Because by looking at it now, I think that the documentation of the parameter 'a' might be useful if shown in the parameters section too and not only in the constructor where 'a' is an implicit argument. On the other hand having it in two places might be redundant (this already happens with the projections and constructor arguments, indeed I haven't found an automatic way to differentiate the two cases with a proper comment).

david-christiansen:
I think it makes sense to be a bit redundant with the parameters - after all, the user might ask for just the docs for a projection, for instance. But it does seem like we're missing some of the docs, and I'm not sure if that's a parser issue (ie them not getting saved) or an output issue. For example, the docs for `blagh`'s argument `x` are missing, and the record type constructor should really show the documentation for  `a` in the parameters section. I don't have time right now to see if the info is getting threaded through, but I'll get back.

GuglielmoS:
Ok, I think to have solved the missing parameters documentation problem. The fact is that when I look up the documentation I have to use the parameter's plain name instead of the fully qualified one with namespaces.
Regarding the other missing documentation related to the projections arguments, I haven't found a solution yet. Could it be that something gets lost during record elaboration (e.g while creating the projection functions)?

david-christiansen:
Right, the parameter name should be local, so no namespaces.  It's quite possible that elaboration of the projection arguments fails to save the docs properly. In any case, this is an improvement, and I'd be willing to merge it with issues created for the parts that still need doing. Or do you want to do it?

GuglielmoS:
Currently I fixed the parameters documentation problem and improved HTML generation by adding the list of parameters with their respective documentation.
As for the other missing documentation, I'll try to look more closely at record elaboration. However, as you said, I think that it's better to create an issue so that people which have worked on it previously can see it.

david-christiansen:
OK, great, I just tested it out and it's a major improvement! Thanks again!

david-christiansen:
Please add yourself to the CONTRIBUTORS list if you want to be there as well :-)

GuglielmoS:
Of course, and thanks for your support :)

PR #2574, opened 2015-09-02
Docs cleanup

enolan:


enolan:
Travis failed because the tests were too slow. The docs built fine, and this doesn't touch anything but docs.

david-christiansen:
This is really great! I had no idea we had so many typos and spelling inconsistencies. Thanks!

Issue #2194, opened 2015-04-26
backend executables are confusing


osa1:
So apparently users should not run `idris-c`, `idris-javascript` etc. manually, and I think this is weird, because Idris is installing all those executables to users' systems but expecting them to not run those..

As a new user, here's how it caused problems to me:

* `idris` executable always runs the REPL, even if I point it a file. So I thought it's just for repl.
* Seeing all those `idris-c`, `idris-javascript` etc. executables, I thought those should be for compiling Idris programs.
* I wasted some time trying to figure out why `idris-c` is doing linking wrong, and not generating some symbols that are used in generated C files.
* Note that `idris-c`, `idris-javascript` etc. don't have `--help`.
* People on IRC channel pointed out that I'm doing it wrong, instead I should be using `idris` executable at all times.

It seems to me like `idris-c` etc. executables should be removed and backends should be used as libraries/plugins.

david-christiansen:
Thanks for a valuable report! It's always good to see how real users encounter the system.

The reason to have them as separate executables is to make it easier to maintain them outside of the main compiler tree.

Also, it should be possible to call `idris-c` directly, if you provide it with the correct options.

It seems to me that, rather than going for some complicated plugin system, we could either make the backend executables dump helpful messages unless called with a particular flag, or we could install them in Idris's libraries directory instead of where Cabal is putting executables. I prefer the former.

david-christiansen:
I suppose another way to fix this is to have all the backends accept a flag `--yep-this-is-the-main-idris-executable`. If it's not provided, then they ask the user to call the main `idris`.

suhr:
Consider renaming `idris-*` to `idris-codegen-*` to avoid “oh this should compile my code to ‹langname›!” confusion.

PR #2579, opened 2015-09-04
Avoid confusion between the compiler and codegens

suhr:
Two things are done

- `idris-‹codegen name›` is now `idris-codegen-‹codegen name›`
- Added flag `--yes-really` to avoid *accidental* usage of codegens instead of the compiler

Resolves https://github.com/idris-lang/Idris-dev/issues/2194.

### DIFF
--- a/codegen/idris-codegen-c/Main.hs
+++ b/codegen/idris-codegen-c/Main.hs
@@ -21,8 +21,9 @@ data Opts = Opts { really :: Bool,
                    interface :: Bool,
                    output :: FilePath }
 
-showUsage = do putStrLn "A code generator which is intended to be called by the compiler, not by a user.\n"
-               putStrLn "Usage: idris-codegen-c --yes-really <ibc-files> [-o <output-file>]"
+showUsage = do putStrLn "A code generator which is intended to be called by the compiler, not by a user."
+               putStrLn "To call the code generator manually, pass the --yes-really option.\n"
+               putStrLn "Usage: idris-codegen-c [--yes-really] <ibc-files> [-o <output-file>]"
                exitWith ExitSuccess
 
 getOpts :: IO Opts
--- a/codegen/idris-codegen-c/Main.hs
+++ b/codegen/idris-codegen-c/Main.hs
@@ -21,7 +21,8 @@ data Opts = Opts { really :: Bool,
                    interface :: Bool,
                    output :: FilePath }
 
-showUsage = do putStrLn "Usage: idris-codegen-c --yes-really <ibc-files> [-o <output-file>]"
+showUsage = do putStrLn "A code generator which is intended to be called by the compiler, not by a user.\n"
+               putStrLn "Usage: idris-codegen-c --yes-really <ibc-files> [-o <output-file>]"
                exitWith ExitSuccess
 
 getOpts :: IO Opts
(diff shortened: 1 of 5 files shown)


codegen/idris-codegen-c/Main.hs
```diff
@@ -16,17 +16,19 @@ import Paths_idris
...
 import Util.System
 
-data Opts = Opts { inputs :: [FilePath],
+data Opts = Opts { really :: Bool,
+                   inputs :: [FilePath],
                    interface :: Bool,
                    output :: FilePath }
 
-showUsage = do putStrLn "Usage: idris-c <ibc-files> [-o <output-file>]"
+showUsage = do putStrLn "Usage: idris-codegen-c --yes-really <ibc-files> [-o <output-file>]"
```
david-christiansen: Perhaps the usage message should state that the idea is that this gets called by Idris rather than users?

codegen/idris-codegen-c/Main.hs
```diff
@@ -44,9 +46,9 @@ c_main opts = do runIO setupBundledCC
...
 main = do opts <- getOpts
-          if (null (inputs opts)) 
+          if (null (inputs opts))
              then showUsage
-             else runMain (c_main opts)
-
-
-
+             else if (not $ really opts)
+                     then do putStrLn "This is not what you may expect it is."
```
david-christiansen: I think that this string should be a bit more informative. Perhaps "This code generator is intended to be called by the Idris compiler. Please pass Idris the '--codegen' flag to choose a backend."

david-christiansen:
This is really nice! I added a few comments about the strings that we show users that I think will prevent even more confusion.

suhr:
I suppose the usage message still ~might be useful, so I left it.

edwinb:
What problem is the --yes-really flag solving? So far I've only found it annoying because I have to implement it for existing code generators...

I agree it's a good idea to rename the executables, and we should have a better error/usage message when they are invoked with something that's not an .ibc and no -o, but I don't think hiding behaviour behind an obscure option is the right way to deal with this.

david-christiansen:
The problem that it's intended to solve is to give users a good message when running a codegen directly. Though perhaps just checking for a .ibc and -o is enough for that - this came out of an IRC discussion some time ago.

edwinb:
I don't know what the exact problem was, or how people were trying to
run the code generators, but it seems to me that the --yes-really flag
is a step too far...

I should do this soon, because I don't know if codegen authors will have
changed things to cope with this flag yet.

david-christiansen:
The problem was that people who wanted to generate e.g. Javascript thought they needed to call `idris-javascript`, which was causing all sorts of confusing messages. The idea here was to provide a description of what was going wrong and prevent new users from trying this, while still allowing backends to be separate executables. But yes, it would be better to change it back quickly if it's going back!

bamboo:
FWIW, [idris-cil](https://github.com/bamboo/idris-cil) has been [updated](https://github.com/bamboo/idris-cil/commit/c50593fc166210f0c1dae0c4639c357ae2c8ed94) to the new design but I'd be happy to remove `--yes-really`. Yes, really.

jfdm:
@bamboo Feel free to add your CodeGen to the [wiki page for Code Generators](https://github.com/idris-lang/Idris-dev/wiki/Codegen). We do not keep track of working external code gens, and I think we should start...

melted:
I have made a patch that I will upload to #2604 as soon as I have checked that all the codegens compile fine.

bamboo:
@jfdm thanks for the suggestion. [Done](https://github.com/idris-lang/Idris-dev/wiki/Codegen/_compare/c820d188336f033ea3c2b4fc8ffdac73ba042b89...832a9e165789121dad0c0443ba208c3c7bdc8ec4).

PR #2589, opened 2015-09-06
Tweak show instances in Logging effect

Melvar:
@jfdm I’m not sure what you intended with the `Show` instance for `LogRes`.

### DIFF
--- a/libs/effects/Effect/Logging/Category.idr
+++ b/libs/effects/Effect/Logging/Category.idr
@@ -26,7 +26,7 @@ record LogRes (a : Type) where
   getCategories : List a
 
 instance (Show a) => Show (LogRes a) where
-  show (MkLogRes l cs) = unwords ["Log Settings:", show l, show cs]
+  showPrec d (MkLogRes l cs) = showCon "MkLogRes" $ showArg l ++ showArg cs
 
 instance Default (LogRes a) where
   default = MkLogRes OFF Nil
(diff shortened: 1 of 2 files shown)


jfdm:
I know exactly what I was doing, I was providing a *canonical 'string' representation*.

However, I find the tone of your comment in this PR rather rude. I know you raised concerns in my original PR for the logging effect and that I failed to respond to them. Regardless, this PR is not informative as to why you have tweaked the show instance in the way you have done so. More so the comment is highly condescending with no need to be given in that manner. Had you said:

> The `Show` instance for `LogRes` does not represent the canonical form seen with items already in the `libs`.

Then I would have been okay with your change, but you didn't...

Further by changing the Show instances for the `LogLevel` data structure and not seeing how this data structure is used in conjunction with its show instance you failed to see that the tests for this effect fail to pass.

Melvar:
I’m sorry. Obviously I did rather badly at both communicating and setting up a contribution.

The comment here was supposed to be something along the lines of a question whether you had a specific intention that the `LogRes` Show instance not use the actual constructor name. I intended to back out the pull request if this was the case, but it would have been more polite for me to have contacted you by some other channel before making it in the first place. Putting it up like this and expecting you to merely give a nod was presumptuous.

Further, I indeed failed to consider that there were tests for the Logging effect that might need to be changed accordingly.

Finally, I made an embarassing type error even in my changed code. Both of the latter would have been caught if I had obeyed my own standards for testing.

jfdm:
Thank you.

For the record, I do not mind PRs that alter my contributions as long as there is a reasonable context and explanation for the need.

PR #2595, opened 2015-09-08
Make test totality010 work when ncurses is enabled

Melvar:
Otherwise, I get escape sequences in the output, making the test fail when I run it locally.


david-christiansen:
Thanks!

PR #2596, opened 2015-09-09
Make chr check for the correct range

Melvar:
Unicode codepoints go up to `U+10FFFF`, so the check was missing a `0`.


david-christiansen:
Thanks!

PR #2598, opened 2015-09-10
Put Elab in Prelude and deprecate old-style tactics

david-christiansen:
This PR deprecates the old style of tactic proofs with a warning (but code still compiles, and they still work fine). It also brings elaborator reflection into the Prelude, so it doesn't need importing.


test/effects003/hangman.idr
```diff
@@ -223,3 +223,7 @@ were caught by the type checker:
 
 main : IO ()
 main = run runGame
+
+-- Local Variables:
+-- idris-load-packages: ("effects")
```
Melvar: Did you add this on purpose?

david-christiansen: I did. It lets the code work interactively in idris-mode when debugging test failures.

david-christiansen:
Hmm, I think I'd like to make it easier to remove tactic script deprecation warnings before I merge this. I'll add another command line argument.

david-christiansen:
OK, once Travis is done, I think this is good enough for a merge.

shlevy:
Is there a replacement for default implicits using elab scripts?

Melvar:
`default` takes an expression, it doesn’t need to be a tactic script. A `tactics` block is an expression, and so is an invocation of `%runElab`, so you can use `{default (%runElab (do …)) prf : …} ->` and so on.

shlevy:
Thanks!

PR #2600, opened 2015-09-10
Remove +, -, and * on Fin

david-christiansen:
There are a number of reasonable type signatures for these, and there
doesn't seem to be a reason to prioritize the particular versions in the
library. They generated more questions than actual use.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -16,6 +16,7 @@ such, the (-) operator and 'abs' have been moved there from 'Num'.
 * A special version of (-) on 'Nat' requires that the second argument is
 smaller than or equal to the first. 'minus' retains the old behaviour,
 returning Z if there is an underflow.
+* The (+), (-), and (*) operations on Fin have been removed.
 * New Logging Effects have been added to facilitate logging of effectful
   programmes.
 
--- a/libs/base/Data/Fin.idr
+++ b/libs/base/Data/Fin.idr
@@ -42,7 +42,7 @@ finToNatInjective FZ     FZ     Refl = Refl
 finToNatInjective (FS m) FZ     Refl impossible
 finToNatInjective FZ     (FS n) Refl impossible
 finToNatInjective (FS m) (FS n) prf  =
-  cong (finToNatInjective m n (succInjective (finToNat m) (finToNat n) prf)) 
+  cong (finToNatInjective m n (succInjective (finToNat m) (finToNat n) prf))
 
 instance Cast (Fin n) Nat where
     cast x = finToNat x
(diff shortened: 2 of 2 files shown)


david-christiansen:
I haven't merged this myself due to doubt about how much others want it. Are there any objections to this removal? Any code that it would break? (@nicolabotta, I'm looking your way)

NightRa:
What are alternative formulations for (+) and (*)?
For the time being I see alternatives only to (-)

nicolabotta:
David, no objections to the removal from me! Best, Nicola

david-christiansen:
@NightRa The bound on the type in the version in the library is too loose. It should be something like Fin n -> Fin m -> Fin (n + m - 1), but that's inconvenient in practice. There are other equivalent formulations, but not really anything canonically right, so I want to let people define their own as needed (which is rare anyway).

@nicolabotta thanks!

nicolabotta:
David,

removing +, -, and * on Fin made no problems, as I was expecting. But
Nat.(-) has been redefined

  (-) : (m : Nat) -> (n : Nat) -> {auto smaller : LTE n m} -> Nat

which does break a lot of code, I guess. In our framework for sequential
decision problems, at least, is not anymore possible to type check even

https://github.com/nicolabotta/SeqDecProbs/blob/master/frameworks/14-/NatProperties.lidr

My question is: why is |(-)| defined in prelude/Prelude/Nat.idr at all?

The module itself does not contain any call to |(-)| and does not prove
any property of this function.

My suggestion would be to simply remove the definition of |(-)| from
prelude/Prelude/Nat.idr.

Best,
Nicola

david-christiansen:
The reason why it's defined in the Prelude is that lots of users expect (entirely reasonably) to be able to use infix subtraction with common datatypes.

The old `(-)` on `Nat` is now available as `minus` - it should be a quick find and replace to change `a - b` to ``a `minus` b``.

nicolabotta:
David Christiansen <notifications@github.com> wrote:

> The reason why it's defined in the Prelude is that lots of users
> expect (entirely reasonably) to be able to use infix subtraction with
> common datatypes.

But the problem is that the new infix subtraction has a very different
signature than the old one! Thus, many users of the old version are
actually left with broken code. Also, it is not clear to me why the new
version

(-)   : (m : Nat) -> (n : Nat) -> {auto smaller : LTE n m} -> Nat

does require "m" and "n" to fulfill "LTE n m". It does not appear to use
this restriction to provide any stronger guarantees on the result than
the original version used to provide.

In this sense, the new implementation seems a step back: it requires
(much) more while delivering quite the same.

Best,
Nicola

PR #2601, opened 2015-09-10
Update README.md

slaven-s:
Hi,

I added section "Books" and a link to the Type-Driven Development with Idris written by Edwin Brady  for Manning Publications. We believe this book is a great resource and all members of community will benefit from this information. If you find this inappropriate, please let me know where I can publish it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,3 +76,7 @@ where some interesting projects are being hosted.
 
 For those interested in contributing to Idris directly we kindly ask that
 prospective developers please consult the [Contributing Guide](CONTRIBUTING.md) first.
+
+## Books
+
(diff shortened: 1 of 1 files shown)


jfdm:
Cheers for the contribution. It might be better to future proof the text by removing the link to chapter one, and leave the link to the actual book website.

That being said, i do not know if the README file is the best place to publicise a link to the book. Maybe adding it to a list of resources for learning Idris on the wiki, or getting a link on the website might be better...

david-christiansen:
I worry that the link URL doesn't look particularly permanent. Perhaps it might be better to link to the book's homepage instead?

david-christiansen:
I think the README is perfectly fine here - not with the goal of publicizing it per se, but rather with the goal of getting people to what will be better than the tutorial.

david-christiansen:
It seems that this exact same pull request is being indiscriminately made to every Idris-related repo on Github. I consider this spam, and will close the ones on the idris-hackers organization and my own account.

PR #2613, opened 2015-09-13
Substring as primitive

david-christiansen:
Introduce a new primitive for taking substrings. This can't really be done efficiently with `strTail` and `strCons`, so I think it deserves to exist as a primitive. Implementations are included for the evaluator as well as the C and JS backends. Other backends will need to implement it.

I'd appreciate a review of the C code that implements this. As far as I can tell, it does the right thing, but this is my first time manually dealing with UTF8.


david-christiansen:
In particular, for code review, I'd appreciate a look at https://github.com/david-christiansen/Idris-dev/blob/substr/rts/idris_rts.c#L597-L605 and https://github.com/david-christiansen/Idris-dev/blob/substr/rts/idris_utf8.c#L81-L104. Thanks!

melted:
It looks mostly good to me, the advance function makes the assumption that the string is well-formed and not terminated in the middle of a char, that seems reasonable, but if it skips past the zero it will hop randomly in the memory until it encounters another zero. Here I would much refer Pascal-style strings with length so it would be easy to to do proper bounds checking, but of course we don't have that. I don't know if this is an issue worth taking seriously. If it was, every byte would have to be looked at.

What about adding add some error at the end of the if-statement? If it encounters an invalid byte now, it will just iterate in place until i has gone down to zero.

rts/idris_utf8.c
```diff
@@ -78,6 +78,32 @@ unsigned idris_utf8_index(char* s, int idx) {
...
+        if ((*str & 0x80) == 0) {
+            str += 1;
+        }  else if ((*str & 0xe0) == 0xc0) {
+            str += 2;
+        } else if ((*str & 0xf0) == 0xe0) {
+            str += 3;
+        } else if ((*str & 0xf8) == 0xf0) {
+            str += 4;
+        } else if ((*str & 0xfc) == 0xf8) {
+            str += 5;
```
Melvar: Just to note, UTF-8 may only contain sequences up to four long because Unicode only admits that many codepoints.

rts/idris_rts.c
```diff
@@ -594,6 +594,16 @@ VAL idris_strIndex(VM* vm, VAL str, VAL i) {
     return MKINT((i_int)idx);
 }
 
+VAL idris_substr(VM* vm, VAL offset, VAL length, VAL str) {
+    char *start = idris_utf8_advance(GETSTR(str), GETINT(offset));
+    char *end = idris_utf8_advance(start, GETINT(length));
+    Closure* newstr = allocate(sizeof(Closure) + (end - start) +1, 0);
+    SETTY(newstr, STRING);
+    memcpy(newstr -> info.str, start, end - start);
+    end[1] = '\0';
```
Melvar: Isn’t end here a pointer into the original string, and thus this mutates the original string?

david-christiansen:
Thanks for the look, @melted. I'm always paranoid when messing with the RTS!

Yeah, that probably should get changed. What do you think - crash on invalid byte, or just skip it and keep going? I suppose the place to worry is if interacting with C code, which might not respect the encoding on a string. I tend to prefer crashing over corruption, but it seems like it might also just work OK to skip a byte and not really corrupt things.

rts/idris_rts.c
```diff
@@ -594,6 +594,16 @@ VAL idris_strIndex(VM* vm, VAL str, VAL i) {
     return MKINT((i_int)idx);
 }
 
+VAL idris_substr(VM* vm, VAL offset, VAL length, VAL str) {
+    char *start = idris_utf8_advance(GETSTR(str), GETINT(offset));
+    char *end = idris_utf8_advance(start, GETINT(length));
+    Closure* newstr = allocate(sizeof(Closure) + (end - start) +1, 0);
+    SETTY(newstr, STRING);
+    memcpy(newstr -> info.str, start, end - start);
```
Melvar: It looks like `newstr->info.str` isn’t initialized here.

rts/idris_rts.c
david-christiansen: Nice catch! Thanks!

rts/idris_utf8.c
david-christiansen: Here I was going for consistency with the rest of the file. This won't hurt anything, will it? Otherwise I think we should remove it from all the functions.

rts/idris_rts.c
david-christiansen: I suppose it worked due to the initialization above on line 600...

rts/idris_utf8.c
Melvar: Strictly speaking a string containing these is invalid, but this doesn’t seem like the right place to validate strings. It would be nice if we made sure to validate all strings as they are constructed from “external” data, and then removed these cases or made them crash with a `runtime error: invalid string` or something similar.

rts/idris_rts.c
david-christiansen: You're right, I misunderstood the second argument to `allocate`. This is why I'm asking for review here :-)

rts/idris_utf8.c
david-christiansen: I suppose that's a bit out of the scope of this change - but it might be a good thing in the long run.

Melvar:
So far, you still seem to be writing through an uninitialized pointer.

david-christiansen:
It's true that I don't explicitly write anything to it before filling it with `memcpy`. Is your concern stylistic or is there something that's a direct error there?

Melvar:
So, on further inspection, it looks to me like this: You allocate a section of memory, using the first `Closure`-sized part as a `Closure`. The allocation has zeroed the memory. Part of this memory is the bytes representing the pointer. The pointer is thus NULL. You then pass this pointer to memcpy, which dereferences it, which should reliably generate a segfault. Does this not happen?

david-christiansen:
Ah, I think you're right. Thanks! The test program that I wrote actually worked, which bothers me, as did a few others. Perhaps it was just memory getting re-used that would have corrupted other data after a while. Thanks for the help! I'll let this lay here a bit in case something else that I did was not so good. My C has gotten rusty enough that I don't trust myself not to destroy things!

Melvar:
At any rate, the other such string-creating functions make sure to set the pointer to the rest of the allocated chunk just after the `Closure`.

Did you make sure to use non-static data in your test program?

david-christiansen:
ah, right, tricksy primitives getting partially evaluated :-)

david-christiansen:
Thanks so much for the help here, @Melvar!

Melvar:
I have an interest in stuff working.

david-christiansen:
Nevertheless, I appreciate it :)

david-christiansen:
OK, so it seems that the issue with the test was UTF-8 being generally problematic in the JS backend, so I moved the unicode bits to a separate test and only ran the ASCII subset of UTF-8 in the JS test.

david-christiansen:
I think this is ready to merge now. What do you think, @melted and @Melvar?

melted:
Looks good to me. I'll let you press the button though.

david-christiansen:
I'll rebase 149d63a onto #2615 and see if that makes everything happy once I get a bit of time.

Melvar:
I’m not that happy with the Unicode test being separate, fearing that it might make it seem non-mandatory.

david-christiansen:
@Melvar That's why I want to do the rebase - I think @melted fixed it. I just need to do other stuff just this moment, and can hack Idris later.

Melvar:
He did, but you didn’t mention folding the new test back into the old in the process of rebasing, so I decided to make a note.

david-christiansen:
Thanks!

david-christiansen:
#2617 has the rebase, and the JS backend passes on it for me locally. I also squashed the fixes for the dumb mistakes. Closing this one.

Thanks again!

PR #2617, opened 2015-09-14
Substring as primitive

david-christiansen:
Introduce a new primitive for taking substrings. This can't really be done efficiently with strTail and strCons, so I think it deserves to exist as a primitive. Implementations are included for the evaluator as well as the C and JS backends. Other backends will need to implement it.

Supersedes #2613, with cleaned up history and tests.

### DIFF
--- a/rts/idris_utf8.c
+++ b/rts/idris_utf8.c
@@ -78,6 +78,32 @@ unsigned idris_utf8_index(char* s, int idx) {
    return top;
 }
 
+char* idris_utf8_advance(char* str, int i) {
+    while (i > 0 && *str != '\0') {
+        // In a UTF8 single-byte char, the highest bit is 0.  In the
+        // first byte of a multi-byte char, the highest two bits are
+        // 11, but the rest of the bytes start with 10. So we can
+        // decrement our character counter when we see something other
+        // than 10 at the front.
+
+        // This is a bit of an overapproximation, as invalid multibyte
+        // sequences that are too long will be treated as if they are
+        // OK, but it's always paying attention to null-termination.
+        if ((*str & 0xc0) != 0x80) {
+            i--;
+        }
+        str++;
+    }
+    // Now we've found the first byte of the last character. Advance
+    // to the end of it, or the end of the string, whichever is first.
+    // Here, we don't risk overrunning the end of the string because
+    // ('\0' & 0xc0) != 0x80.
+    while ((*str & 0xc0) == 0x80) { str++; }
+
+    return str;
+}
+
+
 char* idris_utf8_fromChar(int x) {
     char* str;
     int bytes = 0, top = 0;
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,8 @@ Language updates
 * Improved unification by implementing a pattern unification rule
 * The syntax `{{n}} quotes n without resolving it, allowing short syntax
   for defining new names. `{n} still quotes n to an existing name in scope.
+* A new primitive operator prim__strSubstr for more efficient extraction of
+  substrings. External code generators should implement this.
 
 Library updates
 ---------------
--- a/idris.cabal
+++ b/idris.cabal
@@ -550,6 +550,7 @@ Extra-source-files:
                        test/primitives001/run
                        test/primitives001/*.idr
                        test/primitives001/expected
+                       test/primitives001/input
                        test/primitives002/run
                        test/primitives002/expected
                        test/primitives003/run
--- a/libs/prelude/Prelude/Strings.idr
+++ b/libs/prelude/Prelude/Strings.idr
@@ -308,12 +308,15 @@ length : String -> Nat
 length = fromInteger . prim__zextInt_BigInt . prim_lenString
 
 ||| Returns a substring of a given string
-||| @index The (zero based) index of the string to extract. If this is
-||| beyond the end of the String, the function returns the empty string.
-||| @len The desired length of the substring. Truncated if this exceeds
-||| the length of the input.
-substr : (index : Nat) -> (len : Nat) -> String -> String
-substr i len = pack . List.take len . drop i . unpack
+|||
+||| @ index The (zero based) index of the string to extract. If this is
+|||         beyond the end of the string, the function returns the empty
+|||         string.
+||| @ len The desired length of the substring. Truncated if this exceeds
+|||       the length of the input.
+||| @ subject The string to return a portion of
+substr : (index : Nat) -> (len : Nat) -> (subject : String) -> String
+substr i len subject = prim__strSubstr (cast i) (cast len) subject
 
 ||| Lowercases all characters in the string.
 |||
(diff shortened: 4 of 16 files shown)


rts/idris_utf8.c
```diff
@@ -78,6 +78,36 @@ unsigned idris_utf8_index(char* s, int idx) {
    return top;
 }
 
+char* idris_utf8_advance(char* str, int i) {
+    // The pointer is not at the beginning of a character - find the next one
+    while ((*str & 0xc0) == 0x80) { str++; }
```
mbj: Will this run past the buffer in case a specially crafted string is made out of bytes that always pass the check?

rts/idris_utf8.c
```diff
@@ -78,6 +78,36 @@ unsigned idris_utf8_index(char* s, int idx) {
...
+
+    // Scan characters i times, but bail on end of string
+    while (i > 0 && *str != '\0') {
```
mbj: Lets say a cstring is passed in of size N < 5, and the first byte tells the scanner to advance more than N bytes: This will pass by the check for `*str != '\0'` and overrun the buffer.

rts/idris_utf8.c
```diff
@@ -78,6 +78,36 @@ unsigned idris_utf8_index(char* s, int idx) {
    return top;
 }
 
+char* idris_utf8_advance(char* str, int i) {
```
mbj: @david-christiansen Sorry for raising my voice here, without being asked. I flagged to issues that IMO lead to buffer over reads below. That can result in serious security issues.

These are made under the assumption that the UTF-8 in the buffer is un-validated. I did not check the call sides.

One way to resolve this is to pass in the buffers size, and make sure you do not pass by its maximum size.

rts/idris_utf8.c
mbj: No because the string is null terminated (that will not pass the check). Sorry for the noise.

rts/idris_utf8.c
melted: Yeah, I noted that in the last review. a non utf-8 string (say a truncated string) can skip past the end.  Having looked at it some more, I think it should be changed. There is nothing that guarantees that what comes in from idris_readStr is a valid utf-8 string.

I would just take strlen at the beginning and check that it isn't exceeded. It's not the most performant way, but we can save microoptimization for later.

mbj: Isnt there the option to get the byte length of the buffer from the idris side?

> Yeah, I noted that in the last review.

Sorry for the duplication.

rts/idris_utf8.c
```diff
@@ -78,6 +78,36 @@ unsigned idris_utf8_index(char* s, int idx) {
...
+    // The pointer is not at the beginning of a character - find the next one
+    while ((*str & 0xc0) == 0x80) { str++; }
+
+    // Scan characters i times, but bail on end of string
+    while (i > 0 && *str != '\0') {
+        if ((*str & 0x80) == 0) {
+            str += 1;
+        }  else if ((*str & 0xe0) == 0xc0) {
```
mbj: There is an extra whitespace between the `}` and the `else`.

rts/idris_utf8.c
melted: come to think of it, it's probably easiest to do this:

```c
while (i > 0 && *str != 0) {
    if ((*str & 0x80)== 0 ||
        (*str & 0xc0) == 0xc0) {
        i--;
    }
}
```

melted: > Isnt there the option to get the byte length of the buffer from the idris side?

The length isn't stored with the string so it would just do a strlen, I think my code snippet is better than checking the length anyway (because a stride might pass the length and then you have to get the logic just right to avoid accessing unallocated memory).

> Sorry for the duplication.

I thought it wasn't such a big deal (it works right on correct string), but your warning got me thinking, so thanks for setting me straight.

david-christiansen:
@mbj Thanks, I think you're right here. I'll fix it up :-)

rts/idris_utf8.c
david-christiansen: @mbj please don't be sorry - I explicitly asked for help making this correct, because C-level bit-munging isn't the thing in the world that I'm best at. I appreciate it!

david-christiansen: @melted I don't think that will quite work - it cuts off the end of the character because i hits 0 at the beginning of the multibyte sequence. But I'll do something similar.

jfdm:
In future, I think we should add some more care when merging a PR that was under going review/query until all participants in the review have explicitly provided a consensus. This goes double for something that may introduce a security issue. It was not clear from reading the comments that all participants had signed off on it. If there was discussion in the IRC then I missed it. Regardless something like that should be signed off in a more recorded manner.

mbj:
@david-christiansen I was woundering on how to improve the tests for the C implemented primopts. Are there any active / existing attempts to use something like the ALF fuzzer agains this? http://lcamtuf.coredump.cx/afl/

Or any other kind of fuzzer that exercises the primopts, that one could even be backend agnostic? Is there a port of QuickCheck to Idris?

jfdm:
@mbj @david-christiansen did have a *port* of quickcheck for Idris---see his repositories on github. It has bitrotted IIRC, and I do not know about its completeness in terms of functionality. Further, another questions arises if we are to use it for testing primitives, should this library also be included in the provided set of libraries? It adds questions over maintenance and housekeeping overheads...

david-christiansen:
I got a bit stuck on my Quickcheck port, largely because I couldn't make type class resolution work sufficiently well to do generators for dependent types, and then I got busy with other things. The solution is really to provide a combinator library for building generators, and then use reflection to automate that application in common cases, but that's going to be an after-PhD project.

Having a better story on testing of primitives would be a very nice thing to have, and if you're looking for a way to make Idris development nicer @mbj, it would be a greatly appreciated contribution! It would also make it easier for new backed authors. Getting afl-fuzz rigged up would also be very nice.

david-christiansen:
@jfdm I think it's fine to have a library that we use for testing, but don't necessarily expose as part of the prelude. After all, the test suite contains things like a Hangman game! Also, a lib for doing random testing of primitives can be much simpler than a lib for Quickcheck in general, as primitive operators are all first-order and primitive types don't have any complicated indexing structure.

david-christiansen:
Oh, one more thing - I think that it might be good to wait a reasonable amount of time for an "OK" from participants in discussions like this, @melted - I do think that my updates took care of the critiques, but I was wrong about that before :-)

Thanks again @Melvar, @melted, and @mbj for review and discussion of this patch!

jfdm:
@david-christiansen For the minimal primitive testing library I agree with having a specialised 'non-exposed', in the loosest sense of non-exposed, library for testing. Indeed @mbj a better story would be good to have. Especially with afl-fuzz.

mbj:
I think that a "primopts API fuzzer" in Idris, is easy to make. As it was said before: The type signatures are easy, and it does not require a full QuickCheck port. And will help all people who develop backends, independent of language.

`afl-fuzz` will help in a different dimension, which is IMO orthogonal to an input fuzzer.

Dunno if I've the time to help. We'll see.

PR #2618, opened 2015-09-15
syntax guide: discuss modules, imports.

cbiffle:
This discusses the previously-undocumented "import X as Y" facility.
This seemed like a good time to discuss the high-level file structure,
so I added a bit of that.

### DIFF
--- a/docs/reference/syntax-guide.rst
+++ b/docs/reference/syntax-guide.rst
@@ -4,6 +4,89 @@ Syntax Guide
 
 Examples are mostly adapted from the Idris tutorial.
 
+Source File Structure
+---------------------
+
+Source files consist of:
+
+1. An optional :ref:`syntax-module-headers`.
+2. Zero or more :ref:`syntax-imports`.
+3. Zero or more declarations, e.g. :ref:`syntax-variables`,
+   :ref:`syntax-data-types`, etc.
+
+For example:
+
+.. code:: idris
+
+    module MyModule   -- module header
+
+    import Data.Vect  -- an import
+
+    %default total    -- a directive
+
+    foo : Nat         -- a declaration
+    foo = 5
+
+.. _syntax-module-headers:
+
+Module Header
+~~~~~~~~~~~~~
+
+A file can start with a module header, introduced by the ``module`` keyword:
+
+.. code-block:: idris
+
+  module Semantics
+
+Module names can be hierarchical, with parts separated by ``.``:
+
+.. code-block:: idris
+
+  module Semantics.Transform
+
+Each file can define only a single module, which includes everything defined in
+that file.
+
+Like with declarations, a :ref:`docstring <syntax-comments>` can be used to
+provide documentation for a module:
+
+.. code-block:: idris
+
+  ||| Implementation of predicate transformer semantics.
+  module Semantics.Transform
+
+.. _syntax-imports:
+
+Imports
+~~~~~~~
+
+An ``import`` makes the names in another module available for use by the current
+module:
+
+.. code-block:: idris
+
+    import Data.Vect
+
+All the declarations in an imported module are available for use in the file.
+In a case where a name is ambiguous --- e.g. because it is imported from
+multiple modules, or appears in multiple visible namespaces --- the ambiguity can be resolved using :ref:`syntax-qualified-names`.  (Often, the compiler can
+resolve the ambiguity for you, using the types involved.)
+
+Imported modules can be given aliases to make qualified names more compact:
+
+.. code-block:: idris
+
+    import Data.Vect as V
+
+Note that names made visible by import are not, by default, re-exported to
+users of the module being written.  This can be done using ``import public``:
+
+.. code-block:: idris
+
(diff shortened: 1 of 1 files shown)


docs/reference/syntax-guide.rst
```diff
@@ -4,6 +4,72 @@ Syntax Guide
...
+
+Source files consist of:
+
+1. An optional :ref:`syntax-module-headers`.
+2. Zero or more :ref:`syntax-imports`.
+3. The file's content, which is a mix of :ref:`syntax-variables`,
+   :ref:`syntax-data-types`, :ref:`syntax-directives`, etc.
+
+.. Should that last bullet become more specific?  Terminologically I'm not even
+   certain what the top-level entities in an Idris file are properly called.
```
david-christiansen: They should be called "declarations". I'd phrase it:

```
3. Zero or more declarations (e.g. pattern-matching functions or data types)
```

docs/reference/syntax-guide.rst
```diff
@@ -4,6 +4,72 @@ Syntax Guide
...
+
+.. _syntax-module-headers:
+
+Module Header
+~~~~~~~~~~~~~
+
+The module header appears at the top of the file.  Each file can contain only
+one module, but content within the file may be broken into multiple namespaces.
+
+Unlike most Idris declaration headers, the module header does not end in the
```
david-christiansen: Shouldn't it mention the `module` keyword here? I'd phrase it

A module header consists of the keyword `module` followed by the name of the module. Hierarchical module names are specified with `.` as a separator. Unlike some other languages, Idris modules headers are not terminated with the keyword `where`.

docs/reference/syntax-guide.rst
```diff
@@ -4,6 +4,72 @@ Syntax Guide
...
+The module's name can be used to *qualify* variables that are ambiguous, such
+as when a variable with the same name is defined in two imported modules.  Note
+that qualification is usually not required, as long as the compiler can resolve
+the ambiguity using types.
+
+Modules can be given aliases to make qualified uses more compact::
+
+    import Data.Vect as V
+
+Note that names made visible by import are not, by default, re-exported to
```
david-christiansen: `import` should have double backticks around it, as it's a keyword being referred to literally.

docs/reference/syntax-guide.rst
```diff
@@ -4,6 +4,72 @@ Syntax Guide
...
+
+Imports
+~~~~~~~
+
+An *import* makes the names in another module available for use by the current
+module::
+
+    import Data.Vect
+
+The module's name can be used to *qualify* variables that are ambiguous, such
```
david-christiansen: It's not technically the module's name that's used, it's the variable's namespace. This can come from the module name, but sub-namespaces can also result from `namespace` blocks in the source code, `record` declarations, and I think a couple more places. Perhaps say that modules induce namespaces, and namespaces can be used to qualify names? Additionally, qualified names only require the _suffix_ of their namespace that makes them unique.

docs/reference/syntax-guide.rst
```diff
@@ -4,6 +4,72 @@ Syntax Guide
...
+~~~~~~~~~~~~~
+
+The module header appears at the top of the file.  Each file can contain only
+one module, but content within the file may be broken into multiple namespaces.
+
+Unlike most Idris declaration headers, the module header does not end in the
+keyword ``where``.
+
+A :ref:`docstring <syntax-comments>` can be placed before the module header to
+serve as module-level documentation.
```
david-christiansen: It doesn't serve as documentation, it _is_ documentation. Perhaps something like:

"Just like declarations, modules can have REPL-accessible documentation. To document a module, place a :ref:`docstring <syntax-comments>` prior to the module header.

david-christiansen:
This is really nice! Thanks! I added a couple of comments - what do you think?

docs/reference/syntax-guide.rst
cbiffle: Great, will fix.

docs/reference/syntax-guide.rst
cbiffle: Sounds great.

docs/reference/syntax-guide.rst
cbiffle: Good call.

docs/reference/syntax-guide.rst
cbiffle: Wow, that last bit about only requiring the suffix has blown my mind.  That's a great feature, I will definitely add it.

cbiffle:
Thanks, David.  I'll incorporate your feedback later today.

david-christiansen:
Thanks for doing this!

docs/reference/syntax-guide.rst
cbiffle: Say, @david-christiansen , can a *namespace* be `import`-ed, or is it only modules?

cbiffle:
Changes incorporated, and I added a discussion of qualified names.  PTAL.

PR #2628, opened 2015-09-17
WIP: sticking string length in the RTS

david-christiansen:
This is a WIP pull request so that the skilled C hackers can get critique in early. It's not done yet!

The idea is that an explicit length field means we're not dependent on null-termination. This has, in my view, two benefits:
1. Less chance of accidental buffer overruns from malformed UTF-8 strings
2. It becomes possible to use \0 inside of strings, giving the type `String` the semantics "sequence of code points" rather than "sequence of nonzero code points" which is at least elegant and possibly even useful.

A null terminator is retained to avoid marshalling issues or excess copying when calling C functions.

This pull request does the following:
- [X] Add a length field to the runtime representation of strings, tracking their size in bytes
- [X] Refactor string allocations to do all object initialization in one place, to help avoid mistakes
- [X] Modify the UTF-8 functions to not access memory outside the size size of the string
- [X] Crash with informative messages on partial string operations (like `strTail ""`)
- [X] Validate the output of primitive conversions, like `prim__floatToStr` to make sure that library code is returning what we expect
- [ ] Stop using C string functions that assume null-termination (partially complete)
- [ ] Check for `NULL` return from all `malloc` calls (partially complete)
- [X] Pass the test suite
- [ ] Validate that strings from calls to C code are actually real UTF-8 (I may punt on this or implement a poor version of it until we get a proper Unicode library integrated)
- [X] Validate that purported characters (represented as `Int` under the hood) are actually in the Unicode range before consing onto a string
- [X] Add `assert`s about the byte size invariants
- [ ] Make the C codegen actually allow `'\0'` in the middle of Idris strings, and display them appropriately using the print functions in the library
- [ ] Factor out checks for `NULL` after `malloc` into a function

Potential things to do:
1. Also cache the length in characters (would make some things more elegant, but adds overhead and complexity and may not be worth it)
2. More thorough Unicode testing

### DIFF
--- a/rts/idris_gmp.c
+++ b/rts/idris_gmp.c
@@ -384,12 +384,11 @@ VAL idris_castBigStr(VM* vm, VAL i) {
 
     // Compute the appropriate size in bytes for a C string holding the result
     size_t max_len = mpz_sizeinbase(GETMPZ(GETBIG(vm, i)), 10) + 2;
-    char *temp = (char *)malloc(max_len);
+    char *temp = (char *)calloc(max_len, sizeof(char));
     if (temp == NULL) {
         fprintf(stderr, "Couldn't allocate %lu bytes for bigint as string", (unsigned long)max_len);
         exit(EXIT_FAILURE);
     }
-    memset(temp, '\0', max_len);
     mpz_get_str(temp, 10, GETMPZ(GETBIG(vm, i)));
 
     unsigned int j;
--- a/rts/idris_gmp.c
+++ b/rts/idris_gmp.c
@@ -6,6 +6,7 @@
 #endif
 #include <stdlib.h>
 #include <string.h>
+#include <ctype.h>
 
 // TMP HACK! Require this much space in the heap before a GMP operation
 // so it doesn't garbage collect in the middle.
@@ -374,11 +375,37 @@ VAL idris_castFloatBig(VM* vm, VAL f) {
 }
 
 VAL idris_castStrBig(VM* vm, VAL i) {
-    return MKBIGC(vm, GETSTR(i));
+    // TODO: validate that the string is actually an integer
+    return MKBIGC(vm, (char *) GETSTR(i).str_data);
 }
 
 VAL idris_castBigStr(VM* vm, VAL i) {
-    char* str = mpz_get_str(NULL, 10, GETMPZ(GETBIG(vm, i)));
-    return MKSTR(vm, str);
+
+
+    // Compute the appropriate size in bytes for a C string holding the result
+    size_t max_len = mpz_sizeinbase(GETMPZ(GETBIG(vm, i)), 10) + 2;
+    char *temp = (char *)malloc(max_len);
+    if (temp == NULL) {
+        fprintf(stderr, "Couldn't allocate %lu bytes for bigint as string", (unsigned long)max_len);
+        exit(EXIT_FAILURE);
+    }
+    memset(temp, '\0', max_len);
+    mpz_get_str(temp, 10, GETMPZ(GETBIG(vm, i)));
+
+    unsigned int j;
+    for (j = 0; j < max_len; j++) {
+        if (temp[j] != '\0' && !(isdigit(temp[j])) && temp[j] != '-') {
+            fprintf(stderr, "Fatal error: invalid output from sprintf on an int!");
+            fprintf(stderr, "It was [%c] at index [%u] in [%s].\n", temp[j], j, temp);
+            exit(EXIT_FAILURE);
+        }
+    }
+
+
+    // TODO: validate that mpz_get_str created valid UTF-8
+    VAL res = MKSTR(vm, (utf8_byte *)temp);
+    free(temp);
+
+    return res;
 }
 
--- a/rts/idris_rts.c
+++ b/rts/idris_rts.c
@@ -1,4 +1,6 @@
 #include <assert.h>
+#include <limits.h>
+#include <ctype.h>
 
 #include "idris_rts.h"
 #include "idris_gc.h"
@@ -149,10 +151,10 @@ int space(VM* vm, size_t size) {
 }
 
 void* idris_alloc(size_t size) {
-    Closure* cl = (Closure*) allocate(sizeof(Closure)+size, 0);
+    Closure* cl = (Closure*) allocate(sizeof(Closure) + size, 0);
     SETTY(cl, RAWDATA);
     cl->info.size = size;
-    return (void*)cl+sizeof(Closure);
+    return (void*)cl + sizeof(Closure);
 }
 
 void* idris_realloc(void* old, size_t old_size, size_t size) {
@@ -231,29 +233,51 @@ VAL MKFLOAT(VM* vm, double val) {
     return cl;
 }
 
-VAL MKSTR(VM* vm, const char* str) {
-    int len;
-    if (str == NULL) {
-        len = 0;
-    } else {
-        len = strlen(str)+1;
-    }
-    Closure* cl = allocate(sizeof(Closure) + // Type) + sizeof(char*) +
-                           sizeof(char)*len, 0);
+Closure *idris_allocate_string(size_t size, int outerlock) {
+    Closure* cl = allocate(sizeof(Closure) + // object header
+                           sizeof(SizedString) + // string bookkeeping
+                           sizeof(utf8_byte) * size + // the string data
+                           sizeof(utf8_byte), // the null terminator
+                           outerlock);
     SETTY(cl, STRING);
-    cl -> info.str = (char*)cl + sizeof(Closure);
-    if (str == NULL) {
-        cl->info.str = NULL;
-    } else {
-        strcpy(cl -> info.str, str);
-    }
+    cl -> info.str = (SizedString *) ((uint8_t *)cl + sizeof(Closure));
+    cl -> info.str -> size = size;
+    cl -> info.str -> str_data = (utf8_byte*)cl + sizeof(Closure) + sizeof(SizedString);
+    // We keep the strings null-terminated to allow easier C interop without copying
+    cl -> info.str -> str_data[size] = '\0';
     return cl;
 }
 
-char* GETSTROFF(VAL stroff) {
+VAL MKSTR(VM* vm, const utf8_byte* str) {
+    if (str == NULL) {
+        // We need to preserve NULLness so that things like getenv
+        // work with the FFI as presently specified.
+        Closure* cl = idris_allocate_string(0, 0);
+        cl -> info.str -> str_data = NULL;
+        return cl;
+    } else {
+        size_t len = idris_utf8_unitlen(str);
+        Closure* cl = idris_allocate_string(len, 0);
+        if (str != NULL) {
+            memcpy(cl -> info.str -> str_data, str, len);
+        }
+        assert(cl -> info.str -> size == len);
+        return cl;
+    }
+}
+
+SizedString GETSTROFF(VAL stroff) {
     // Assume STROFF
     StrOffset* root = stroff->info.str_offset;
-    return (root->str->info.str + root->offset);
+    SizedString with_offset = {
+        .str_data = (utf8_byte *)(root ->str->info.str->str_data) + (root->offset),
+        .size = root->str->info.str->size - root->offset
+    };
+    return with_offset;
+}
+
+char *idris_get_c_str(SizedString str) {
+    return (char *) str.str_data;
 }
 
 VAL MKPTR(VM* vm, void* ptr) {
@@ -281,13 +305,9 @@ VAL MKFLOATc(VM* vm, double val) {
     return cl;
 }
 
-VAL MKSTRc(VM* vm, char* str) {
-    Closure* cl = allocate(sizeof(Closure) + // Type) + sizeof(char*) +
-                           sizeof(char)*strlen(str)+1, 1);
-    SETTY(cl, STRING);
-    cl -> info.str = (char*)cl + sizeof(Closure);
-
-    strcpy(cl -> info.str, str);
+VAL MKSTRc(VM* vm, SizedString* str) {
+    Closure* cl = idris_allocate_string(str->size, 1);
+    memcpy(cl -> info.str -> str_data, str->str_data, str->size);
     return cl;
 }
 
@@ -340,7 +360,7 @@ VAL MKB64(VM* vm, uint64_t bits64) {
 void PROJECT(VM* vm, VAL r, int loc, int arity) {
     int i;
     for(i = 0; i < arity; ++i) {
-        LOC(i+loc) = r->info.c.args[i];
+        LOC(i + loc) = r->info.c.args[i];
     }
 }
 
@@ -382,7 +402,11 @@ void dumpVal(VAL v) {
         printf("] ");
         break;
     case STRING:
-        printf("STR[%s]", v->info.str);
+        printf("STR(%lu)[%.*s]\n", v->info.str->size, (int)v->info.str->size, v->info.str->str_data);
+        break;
+    case STROFFSET:
+        printf("STROFF(%lu)=\n\t", v->info.str_offset->offset);
+        dumpVal(v->info.str_offset->str);
         break;
     case FWD:
         printf("FWD ");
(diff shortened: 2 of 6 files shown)


david-christiansen:
Comments, critique, and questions are most certainly welcome!

david-christiansen:
And please, even when the tests pass, nobody click "Merge" - when it's ready and I've got buy-in from the right people, I'll do so. Thanks!

david-christiansen:
@edwinb, I'd like to hear what you think, at least about the idea.

@Melvar, @mbj, @daira, @melted: you were very helpful last time I poked at some C - if you have time and energy, I'd appreciate further feedback!

mbj:
I really like the intention of this PR. The cstring data format and its in-line signalling are a major bug class source. Avoiding this construct eliminates a bug class. So big +1.

I do not have the time to do a dense review. Nor followup with the old discussions, been sick the last weeks and need to cure (AKA sleeping). BTW if anyone of you goes to StrangeLoop: I'll be in the Idris workshop by @edwinb would be cool to meet more people from here in person.

melted:
I'll go through it.

> And please, even when the tests pass, nobody click "Merge" - when it's ready and I've got buy-in from the right people, I'll do so. Thanks!

I wonder if there's any way to disable the merge button temporarily, except by introducing an intentional merge conflict.

mbj:
> I wonder if there's any way to disable the merge button temporarily, except by introducing an intentional merge conflict.

Githup repos in organizations have the concept of "protected branches", these right now prevent merges when *any* configured integration is red (example Travis).

There could be a custom integration that adds some kind of "review overlay", it would only report green when all assigned people did a special comment like `[agreeed]` etc. Its true such a system requires coding + external state keeping. But maybe such an overlay bot already exists.

rts/idris_gmp.c
```diff
@@ -374,11 +375,37 @@ VAL idris_castFloatBig(VM* vm, VAL f) {
...
+
+
+    // Compute the appropriate size in bytes for a C string holding the result
+    size_t max_len = mpz_sizeinbase(GETMPZ(GETBIG(vm, i)), 10) + 2;
+    char *temp = (char *)malloc(max_len);
+    if (temp == NULL) {
+        fprintf(stderr, "Couldn't allocate %lu bytes for bigint as string", (unsigned long)max_len);
+        exit(EXIT_FAILURE);
+    }
+    memset(temp, '\0', max_len);
```
melted: can use calloc to allocate zeroed memory

rts/idris_gmp.c
```diff
@@ -374,11 +375,37 @@ VAL idris_castFloatBig(VM* vm, VAL f) {
...
+    mpz_get_str(temp, 10, GETMPZ(GETBIG(vm, i)));
+
+    unsigned int j;
+    for (j = 0; j < max_len; j++) {
+        if (temp[j] != '\0' && !(isdigit(temp[j])) && temp[j] != '-') {
```
melted: Why can't we trust that we get digits from GMP or minigmp?

rts/idris_rts.c
```diff
@@ -411,102 +435,186 @@ void idris_memmove(void* dest, void* src, i_int dest_offset, i_int src_offset, i
...
         fprintf(stderr, "Fatal Error: ClosureType %d, not an integer type", ty);
         exit(EXIT_FAILURE);
     }
 
-    SETTY(cl, STRING);
-    return cl;
+    // avoid invalid UTF-8: all ASCII digits are acceptable as UTF-8 characters.
+    int j;
+    for (j = 0; j < 21; j++) {
+        if (temp[j] != '\0' && !(isdigit(temp[j]))) {
```
melted: this section does exactly the same as the one above, should make a function validateNumber out of it, though I suspect the other place s uneccesary. Also, that 21 shouldn't be hardcoded like that. Magic constants should be defined in one place with a comment about what they represent.

rts/idris_rts.c
```diff
@@ -411,102 +435,186 @@ void idris_memmove(void* dest, void* src, i_int dest_offset, i_int src_offset, i
...
-    return cl;
+    char temp[32] = { 0 };
+    // snprintf will put a null at the end, so 31 chars are available for the float
+    snprintf(temp, 32, "%.16g", GETFLOAT(i));
+
+    // avoid invalid UTF-8: all ASCII digits, plus period, plus, and
+    // minus, are acceptable as UTF-8 characters.
+    int j;
+    for (j = 0; j < 21; j++) {
+        if (temp[j] != '\0' &&
```
melted: Magic constants, also mixup of 21 and 32

rts/idris_rts.c
```diff
@@ -411,102 +435,186 @@ void idris_memmove(void* dest, void* src, i_int dest_offset, i_int src_offset, i
...
 }
 
 VAL idris_castStrFloat(VM* vm, VAL i) {
-    return MKFLOAT(vm, strtod(GETSTR(i), NULL));
+    SizedString str = GETSTR(i);
+
+    // First, we need to alloc a C string that's long enough (to get null-termination)
+    char *temp = (char *) malloc(str.size + 1);
+    if (temp == NULL) {
+        fprintf(stderr, "Couldn't allocate temp buffer of size %lu for int->string conversion", str.size);
```
melted: I don't know about the usability of checking for failure on malloc, everybody has overcommit. Linux failure mode is to freak out when you try to use the memory. It will never fail to allocate memory as long as there is address space.

rts/idris_rts.c
```diff
@@ -411,102 +435,186 @@ void idris_memmove(void* dest, void* src, i_int dest_offset, i_int src_offset, i
...
+
+
+    return MKSTR(vm, (utf8_byte *)temp);
 }
 
 VAL idris_castStrFloat(VM* vm, VAL i) {
-    return MKFLOAT(vm, strtod(GETSTR(i), NULL));
+    SizedString str = GETSTR(i);
+
+    // First, we need to alloc a C string that's long enough (to get null-termination)
```
melted: hey, don't we already have null termination?

rts/idris_rts.c
```diff
@@ -411,102 +435,186 @@ void idris_memmove(void* dest, void* src, i_int dest_offset, i_int src_offset, i
...
+        exit(EXIT_FAILURE);
+    }
+    // Then grab the UTF8 bytes out of the Idris string. We don't have
+    // to worry about nulls in the middle because we'll later check to
+    // see if the whole string was used.
+    memcpy(temp, str.str_data, str.size);
+    temp[str.size] = '\0';
+
+    char *end;
+    double v = strtod(temp, &end);
```
melted: I like scanf better, say `scanf("%lf", &v)`, it returns 1 (for nbr of read values) on success.

rts/idris_rts.c
```diff
@@ -411,102 +435,186 @@ void idris_memmove(void* dest, void* src, i_int dest_offset, i_int src_offset, i
...
 }
 
 VAL idris_castStrInt(VM* vm, VAL i) {
     char *end;
-    i_int v = strtol(GETSTR(i), &end, 10);
-    if (*end == '\0' || *end == '\n' || *end == '\r')
-        return MKINT(v);
+    SizedString str = GETSTR(i);
+
+    // First, we need to alloc a C string that's long enough (to get null-termination)
```
melted: wasn't it nullterminated in the VAL?

rts/idris_rts.c
```diff
@@ -556,62 +673,83 @@ VAL idris_strTail(VM* vm, VAL str) {
             root = root->info.str_offset->str;
         }
 
-        cl->info.str_offset->str = root;
-        cl->info.str_offset->offset = offset+idris_utf8_charlen(GETSTR(str));
+        cl -> info.str_offset -> str = root;
+        cl -> info.str_offset -> offset = offset + idris_utf8_charlen(string);
```
melted: no space around pointer arrows.

rts/idris_rts.c
```diff
@@ -556,62 +673,83 @@ VAL idris_strTail(VM* vm, VAL str) {
...
         return cl;
     } else {
-        char* nstr = GETSTR(str);
-        return MKSTR(vm, nstr+idris_utf8_charlen(nstr));
+        SizedString nstr = GETSTR(str);
+        unsigned int tailIdx = idris_utf8_charlen(nstr);
+        assert(tailIdx <= nstr.size);
+        Closure *cl = idris_allocate_string(nstr.size - tailIdx, 0);
+        // Use memcpy here instead of MKSTR because MKSTR assumes
+        // C-style strings, and 0 is a sensible Unicode character that
```
melted: wouldn't it be reasonable then that MKSTR takes a length argument?

rts/idris_utf8.h
```diff
@@ -1,24 +1,38 @@
...
+#include <stdint.h>
+#include <stdlib.h>
+
 /* Various functions for dealing with UTF8 encoding. These are probably
    not very efficient (and I'm (EB) making no guarantees about their
    correctness.) Nevertheless, they mean that we can treat Strings as
    UFT8. Patches welcome :). */
 
+typedef uint8_t utf8_byte;
+
```
melted: not a fan of this typedef. uint8_t is always the correct type for it, and nobody will be confused by utf8 being represented by an array of uint8_t.

rts/idris_utf8.h
```diff
@@ -1,24 +1,38 @@
...
+// Pascal-style strings with their lengths.
+typedef struct {
+    size_t size;
+    utf8_byte* str_data;
+} SizedString;
```
melted: Could save some indirection by appending str_data as a zero-length array. It won't be computable at compile time so lose sizeof, but the length is in there so easy to substitute.

melted:
MUST NOT PRESS BIG GREEN BUTTON... MUST NOT PRESS BIG GREEN BUTTON...

rts/idris_utf8.h
david-christiansen: The reason I did it was to document the intention at use sites - that this particular `uint8` is intended as part of a string. Is that not reasonable?

rts/idris_gmp.c
david-christiansen: We probably can. Really, my goal is to ensure that all in-memory strings are valid UTF-8, so I want to do this kind of validation for each source of strings.

rts/idris_rts.c
mbj: On embedded devices checking for malloc fails is critical. If idris ever wants to support such environments it needs to check each malloc return. Ideally idris uses a centralized `idris_malloc` or something more sophisticated to centralize detection and handling of the issue.

rts/idris_gmp.c
david-christiansen: Thanks, fixed.

rts/idris_gmp.c
david-christiansen: This might go wrong, for example, if we accidentally specified the wrong base.

rts/idris_rts.c
david-christiansen: `MKSTR` is really responsible for turning actual C strings into Idris RTS strings. It's used various places in the codegen, along with `MKINT` and friends. It might be feasible to plumb lengths through that, however, and that will probably be necessary in the end to allow `'\0'` as part of a string.

rts/idris_rts.c
david-christiansen: I think that, for running on very small systems, we'd likely need to do the RTS differently, and these malloc fail checks are the least of our worries! But I still prefer to do it, even if it's highly unlikely to happen on Linux, because it should increase the portability to new systems that we haven't anticipated.

melted: As it exits on allocation failure it could be made into a function and there wouldn't be a need to handle the error on every call site.

Från: "David Christiansen" <notifications@github.com>
Skickat: ‎2015-‎09-‎18 08:40
Ämne: Re: [Idris-dev] WIP: sticking string length in the RTS (#2628)

In rts/idris_rts.c:
>  }
>  
>  VAL idris_castStrFloat(VM* vm, VAL i) {
> -    return MKFLOAT(vm, strtod(GETSTR(i), NULL));
> +    SizedString str = GETSTR(i);
> +
> +    // First, we need to alloc a C string that's long enough (to get null-termination)
> +    char *temp = (char *) malloc(str.size + 1);
> +    if (temp == NULL) {
> +        fprintf(stderr, "Couldn't allocate temp buffer of size %lu for int->string conversion", str.size);
I think that, for running on very small systems, we'd likely need to do the RTS differently, and these malloc fail checks are the least of our worries! But I still prefer to do it, even if it's highly unlikely to happen on Linux, because it should increase the portability to new systems that we haven't anticipated.

david-christiansen: That's certainly the case.

rts/idris_utf8.h
david-christiansen: The cost of this is that the code gets more complex. Given that the `str_data` array is always allocated immediately adjacent to the `SizedString`, it should be in the same cache line. Would we really win much there to make it worth the more convoluted code?

rts/idris_rts.c
david-christiansen: You're right! This is obsolete - I was going to get rid off the null termination, but that didn't happen here. I'll fix it!

rts/idris_rts.c
david-christiansen: That won't return a pointer to the end of what was read, though, will it? We need that to check if the whole string was consumed.

rts/idris_rts.c
david-christiansen: Def. needs fixing. Thanks!

rts/idris_rts.c
david-christiansen: I was looking at validating it in a function, but the different types admit different characters in their string representations. Making that into a nice function would make things crazier, in my view. I did remove the magic numbers though.

rts/idris_rts.c
david-christiansen: OK, magic numbers replaced.

rts/idris_rts.c
david-christiansen: The downside is that it fails to allow for custom messages - but I suppose I could make a macro that passes `__LINE__` to make it so we can find where it happened.

rts/idris_rts.c
david-christiansen: it is, but the size field is the size of the actual data, not including the `'\0'` at the end for C

rts/idris_rts.c
melted: I don'rt see why you'd need a custom message if malloc fails. It's a system thing, the only thing that could be of interest would be the size of the allocation. It's an extreme corner case anyhow, and it's very unlikely the system will live long enough to print anything if small allocs fail.

Many programs don't bother to handle alloc failure at all (and OpenBSD's malloc just aborts the program). It's an extreme corner case, the least we can do is to handle it in one place and not sprinkle error handling for it all over.

rts/idris_rts.c
daira: I came to the same conclusion as @melted -- we do similar validation in about 4 places IIRC, so it is definitely worth factoring out. It is ok if the set of characters accepted is an overapproximation as long as it is a subset of US-ASCII.

rts/idris_rts.c
daira: +1 for centralizing checks for malloc failures.

rts/idris_rts.c
```diff
@@ -411,102 +435,193 @@ void idris_memmove(void* dest, void* src, i_int dest_offset, i_int src_offset, i
...
+    // see if the whole string was used.
+    memcpy(temp, str.str_data, str.size);
+    temp[str.size] = '\0';
+
+    char *end;
+    double v = strtod(temp, &end);
+    Closure* answer;
+    // If we used the whole string, return the double, otherwise
+    // return 0.0. This is the meaning of the primitive in the
+    // evaluator.
```
daira: I think this is error-prone and should be reconsidered.

david-christiansen:
Thanks for the feedback, @daira! As I get some free time where I need to de-stress from PhD writing, I'll go through and take care of details. Appreciate it!

rts/idris_rts.c
Melvar: It is, however, the current semantic of the primitive. Changing that is a separate issue.

daira: Filed #2758.

ahmadsalim:
@david-christiansen Are you still planning to complete this when you have time?

ahmadsalim:
@david-christiansen Or let me rephrase, would it be possible to partition tasks into a `nice-to-have` and a `necessary-to-complete` section, and then finishing up the `necessary-to-complete` parts sooner, while converting the others to issues?

I think this feature is useful, and it would be sad if this work was never finished because of some `nice-to-have` requirements like allowing `\0` mid string.

rts/idris_gmp.c
```diff
@@ -374,11 +375,36 @@ VAL idris_castFloatBig(VM* vm, VAL f) {
 }
 
 VAL idris_castStrBig(VM* vm, VAL i) {
-    return MKBIGC(vm, GETSTR(i));
+    // TODO: validate that the string is actually an integer
```
ericqweinstein: Can we use `isnumber` for this? (`$ man isnumber`)

rts/idris_rts.c
```diff
@@ -411,102 +435,193 @@ void idris_memmove(void* dest, void* src, i_int dest_offset, i_int src_offset, i
...
+    // to worry about nulls in the middle because we'll later check to
+    // see if the whole string was used.
+    memcpy(temp, str.str_data, str.size);
+    temp[str.size] = '\0';
+
+    i_int v = strtol(temp, &end, 10);
+    Closure* answer;
+    // If we used the whole string, return the int, otherwise return 0. This is
+    // the meaning of the primitive in the evaluator.
+    if (end == temp + str.size)
```
ericqweinstein: V. small (and possibly paranoia-related and not in keeping with the style guide), but is it beneficial to include braces here, even for the single line? (Here I'm thinking of [#gotofail](https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/).)

```c
if (end == temp + str.size) {
  answer = MKINT(v);
} else {
  answer = MKINT(0);
}

File: rts/idris_rts.c [hunk 16]
```diff
@@ -411,102 +435,193 @@ void idris_memmove(void* dest, void* src, i_int dest_offset, i_int src_offset, i
...
+    char *end;
+    double v = strtod(temp, &end);
+    Closure* answer;
+    // If we used the whole string, return the double, otherwise
+    // return 0.0. This is the meaning of the primitive in the
+    // evaluator.
+    if (end == temp + str.size)
+        answer = MKFLOAT(vm, v);
+    else
+        answer = MKFLOAT(vm, 0.0);
```
ericqweinstein: Ditto above q re: braces

rts/idris_rts.c
```diff
@@ -540,9 +658,15 @@ VAL MKSTROFFc(VM* vm, StrOffset* off) {
...
+    SizedString string = GETSTR(str);
+    // Crash with a message when taking the tail of an empty string
+    if (string.size < 1) {
+        fprintf(stderr, "Tried to take the tail of an empty string.\n");
+        exit(EXIT_FAILURE);
+    } else if (space(vm, sizeof(Closure) + sizeof(StrOffset))) {
     // If there's no room, just copy the string, or we'll have a problem after
     // gc moves str
-    if (space(vm, sizeof(Closure) + sizeof(StrOffset))) {
+//        printf("space\n");
```
ericqweinstein: Should we remove this commented-out line?

rts/idris_utf8.c
```diff
@@ -3,17 +3,25 @@
...
-     if ((s[i] & 0xc0) != 0x80) j++;
+size_t idris_utf8_unitlen(const utf8_byte* s) {
+    return strlen((const char*) s);
+}
+
+unsigned int idris_utf8_strlen(SizedString s) {
+   unsigned int i = 0, j = 0;
+   while (i < s.size) {
+     if ((s.str_data[i] & 0xc0) != 0x80) j++;
      i++;
```
ericqweinstein: Ditto above q re: braces; might improve readability here?

david-christiansen:
@ahmadsalim I would like to finish this!  If you want to partition bits up, let me know, but otherwise I just need to find a weekend afternoon without other plans.

ericqweinstein:
I'm also happy to help out here!

david-christiansen:
@ericqweinstein Thanks! Please feel free - C is far from my first language, and it's easy to get things wrong, so I appreciate your help!

ahmadsalim:
@ericqweinstein Did you ever take a look at this issue?

ericqweinstein:
@ahmadsalim I haven't had time to look yet, but could take a look toward the end of this week

ahmadsalim:
@ericqweinstein That would be very nice if possible! Thanks.

ericqweinstein:
@ahmadsalim Sure thing! Sorry for the delay, busy month/semester thus far. Three-day weekend though, so I'll take a look by Monday night!

ahmadsalim:
@ericqweinstein That sounds great :). I know that you (and many other people) are busy, which is why I try to nudge a bit once in a while.

ericqweinstein:
@ahmadsalim Just wanted to let you know that I'm still interested in this, but due to the work i've got on my plate now, I don't know when I'll be able to get to it—I'll update here if/when things change!

ahmadsalim:
@ericqweinstein OK, I understand. Thanks for looking into it anyway, and please do tell us if you ever come to finish it :)

ahmadsalim:
@david-christiansen Any ideas what we should do about this PR? It has been open for a long time :).

ahmadsalim:
I will close the PR since it has been opened for a long time and no one seems to have had time to fix the issues. If people are interested in fixing the issues please reopen the PR with the required code updates.

PR #2632, opened 2015-09-18
Add --package contrib to tutorial

ExNexu:
I think the information on how to use the contrib package should be in the tutorial as it is a very useful addition.


melted:
Thanks!

Note about Travis: sugar003 failed for javascript, again, clearly irrelevant. Merging.

Issue #1476, opened 2014-08-19
Small syntactic improvements when working with dependent pairs


ahmadsalim:
There are currently a few forms of dependent pairs (`Sigma`, `Exists`, `Subset`), depending on which argument gets erased.
I suggest to extend the current dependent pair syntax to support the new types as follows:

*  `(a : A ** B)` still acts as syntax for `Sigma`
* `(a : A .** B)` acts as new syntax for `Exists` (alternatively `(.a : A ** B)` as @david-christiansen suggests)
* `(a : A **. B)` acts as new syntax for `Subset`

The mnemonic is that the dot is on the side that is getting erased.
Obviously, a question that comes up if there should be a fourth type of dependent pair representing a some kind of dependent "and" where both arguments are erased: `(a : A .**. B)`.

Naturally, there should also be equivalent to these for constructing and destructing relevant values, like `(a0 ** b0)` is used today.

Another part of this suggestion aims to make it easier to use nested dependent pairs, like it is easy to use nested ordinary pairs. I believe a desired feature would be if one could write something like:
`(a : A ** b : B ** C)` and get `(a : A ** (b : B ** C))` for free similar how to pairs are today.

Naturally, it might be better if one could combine the different types of dependent pairs e.g.
`(a : A .** b : B ** c : C **. D)` becoming `(a : A .** (b : B ** (c : C **. D)))`.

cbiffle:
Hello from a year later!

I was surprised this evening to learn that the `(x ** y)` syntax is specific to `Sigma`, and that there is still no (documented) syntax for `Exists` or `Subset`.

ahmadsalim:
Maybe, I should try to implement this proposal at some point since there is some interest. 
There is still no special syntax for Exists or Subset as far as I remember.

cbiffle:
If there were a good way of telling, as a programmer, that the second term in a Sigma such as `(result ** ProofOf result)` were erased, I don't think I would have much interest in `Subset`.  I find its naming and notation a little awkward.  But the application I'm working on produces *very large* proof terms that tend to appear in that position, and I'm currently going through rewriting it to explicitly use `Subset` so that I don't have to worry about using O(n^2) memory.

We can almost get to your original proposal using syntax extensions.  The only thing we can't appear to do is "pun" the same syntax for both types and patterns.  I'm not sure I consider this a bad thing -- Sigma's punning of type and value syntax was confusing for me as a n00b, and still seems error prone.

[...]

PR #2636, opened 2015-09-19
Allowed multiple bindings in dependent pairs

ahmadsalim:
One can now write (a1 : t1 ** ... ** an : tn ** t) instead of (a1 : t1
** (... ** (an : tn ** t))) and similarly with values (a1 ** .. ** an).
This makes the dependent pair syntax mimic the normal pair syntax.

This improves on the second part of #1476. I will still have to do the first part (with mapping to different kinds of pairs), but that will come later since that requires some more complex changes.


ahmadsalim:
It may also be nice to update the pretty printer, but I had no idea how to. @david-christiansen Any suggestions? :)

david-christiansen:
This is a good idea! Before merging, I think it needs the following:
 - [ ] Update pretty-printer (I'll show you where to do it at ITU on Monday if you have a few minutes)
 - [ ] Update CHANGELOG to mention the change
 - [ ] Put it in the test suite, so we make sure nobody breaks it, testing both patterns, constructors, and type syntax

A test could be something like:

```
thing : (a : Type ** n : Nat ** Vect (S n) a) -> (a : Type ** a)
thing (_ ** _ ** (x :: xs)) = (_ ** x)
```

ahmadsalim:
I have resolved the last two points, and we can talk on Monday regarding the first :).

david-christiansen:
The test needs to have an `expected` file and a `run` script in order to be part of the test suite, and it needs to be in the `idris.cabal` file to be shipped with Idris (`cabal sdist` ignores everything else). Ideally, it would also be added on to another of the tests of pattern matching and dependent pairs, rather than being a whole new test, because each little test has some constant overhead that we'd like to reduce, to make the tests go faster.

See you Monday!

ahmadsalim:
OK, I have changed `test0015` to use the new syntax in places where it had nested dependent pairs before (of course there are plenty of other places where normal dependent pairs are used).
I have also squashed the commits.

david-christiansen:
I sat down to throw together a pretty-printer for this, but I think that the parser code is incorrect.
The following works:
```
import Data.Vect

thing : (a ** n ** Vect (S n) a)  -> (a ** a)
thing (_ ** _ ** (x :: xs)) = (_ ** x)
```
but the following does not:
```
import Data.Vect


thing : (a : Type ** n ** Vect (S n) a)  -> (a ** a)
thing (_ ** _ ** (x :: xs)) = (_ ** x)
```

which fails with this error:
```
     When checking type of Main.thing:
     When checking an application of constructor Prelude.Nat.S:
             Type mismatch between
                     Type (Type of n)
             and
                     Nat (Expected type)
```

Also, adding further type annotations makes it into a parse error. The following code:
```
thing : (a : Type ** n : Nat ** Vect (S n) a)  -> (a ** a)
thing (_ ** _ ** (x :: xs)) = (_ ** x)
```

yields the following parse error, pointing at the type signature:
```
 `-- ./dpair.idr line 5 col 11:
      error: expected: "$",
         "&&", "*", "*>", "+", "++", "-",
         "->", ".", "/", "/=", "::", "<",
         "<$>", "<*", "<*>", "<+>", "<<",
         "<=", "<|>", "=", "==", ">",
         ">=", ">>", ">>=", "\\\\", "`",
         "||", "~=~",
         ambiguous use of a left-associative operator,
         ambiguous use of a non-associative operator,
         ambiguous use of a right-associative operator,
         end of bracketed expression,
         function argument,
         matching application expression
     thing : (a : Type ** n : Nat ** Vect (S n) a)  -> (a ** a) 
```

Shouldn't this code work?

ahmadsalim:
@david-christiansen Ah, I didn't consider that the you should be able to not annotate things in the type. I have to check the parser when I get the time. 

Thanks for starting on the pretty printer, I was planning to do this in the weekend, but if you have done it I only have to check the parser I guess.

src/Idris/ParseExpr.hs
```diff
@@ -452,14 +452,19 @@ bracketed' open syn =
...
                lchar ')'
                return $ PTrue (spanFC open (FC f start (l, c+1))) TypeOrTerm
-        <|> try (do (ln, lnfc) <- name
-                    colonFC <- lcharFC ':'
-                    lty <- expr syn
-                    starsFC <- reservedOpFC "**"
+        <|> try (do (ln, lnfc, colonFC, lty, starsFC):bnds <- some $ do
+                        (ln, lnfc, colonFC) <- try $ do
+                          (ln, lnfc) <- name
+                          colonFC <- lcharFC ':'
```
ahmadsalim: Basically, it expects that we get a list (a : ty1 ** a : ty2 ** .. ). It should be possible to make the type optional, but I am concerned about ambiguity of the expression.

src/Idris/ParseExpr.hs
```diff
@@ -452,14 +452,19 @@ bracketed' open syn =
...
+                          return (ln, lnfc, colonFC)
+                        lty <- expr syn
```
ahmadsalim: Actually, I guess this has to be changed to `expr' syn` to avoid ambiguity. I will try changing that.

ahmadsalim:
If I fix the other parsing errors, would it be OK that we only allowed either long form `(a : Type ** n : Nat ** Vect (S n) a)` or short form syntax `(a ** n ** Vect (S n) a)`?

src/Idris/ParseExpr.hs
ahmadsalim: Also, I found out that it will parse `Type **` as `Type* *`, I should probably change how `Type*` is parsed.

src/Idris/ParseExpr.hs
david-christiansen: When I tested it, expressions like `(x ** List x)` worked, but expressions like `(x : Type ** List x)` didn't. This sounds like you're saying the opposite, though...

src/Idris/ParseExpr.hs
david-christiansen: Yeah, that should probably be a bit different :)

david-christiansen:
We need both formats, in my opinion. Only the version in types allows its arguments to be named, after all.

david-christiansen:
Also, the type `(a ** a)` is the same as `(a : _ ** a)`, while the constructor application `(a ** a)` is not a binding form - that is, it is a dependent pair where both projections are references to an in-scope variable `a`.

ahmadsalim:
@david-christiansen Of course, I meant more that it would be easier (for now at least), if we didn't allow mixing, e.g. `(a ** b : Type ** (a, b))` or `(a : Type ** b ** (a, b))`. If you feel strongly otherwise, I could try taking a look, but it requires some refactoring of the grammar :).

ahmadsalim:
I have fixed the other issues, I believe.

david-christiansen:
So you mean that if one has an annotation, all must have an annotation, but that all previous syntax is allowed/

ahmadsalim:
Yes, exactly (for now).

david-christiansen:
TBH I'm a bit worried about requiring consistent annotation. It seems like a strangely arbitrary rule that is likely to confuse users. What do people say when you ask around on IRC?

ahmadsalim:
OK, I will take a look  a bit later then. I have an idea of how it would be possible to allow mixed annotation.

ahmadsalim:
Thanks @david-christiansen for contributing the pretty printer.

edwinb:
Is this still a WIP, or now done? It's certainly something we want!

david-christiansen:
As far as I know it is still WIP.

ahmadsalim:
@edwinb Yes, sorry for lacking behind a bit, but I have been having some trouble building Idris on El Capitan and I have been a bit too busy at my work to use too much time getting things to work correctly.

I will try to complete this soon, most likely in the Weekend or around the holidays.

ahmadsalim:
@edwinb @david-christiansen Things should be fixed now, hopefully the tests will pass.

david-christiansen:
I'll take a peek at this post-PhD-defense, if @edwinb hasn't checked it out first.

ahmadsalim:
@david-christiansen Thanks. 

As a side note, it seems to see `(a ** n ** Vect n a)` as a term instead of a type (and the same with `(a ** (n ** Vect n a))`, so it isn't a parsing problem) but I was unsure if that was intentional or not.

jfdm:
Rem just a bump to @ahmadsalim and @david-christiansen and @edwinb Is this good to go, or is there more actions required before this can be merged?

I for one wouldn't mind seeing some documentation added either to the tutorial or reference manual. But such a request by no means would be a block to this getting merged.

ahmadsalim:
@jfdm One may want to address my latest comment (about the semantics), but the parsing part should be correct.

edwinb:
This looks fine to me now, I've tried a few examples and they behave as I expect them to (varying whether there are annotations, checking pretty printing, etc). @ahmadsalim dependent pairs are elaborated as values unless the goal type of the elaboration is Type. It's possible that we should use the disambiguation machinery instead, since this has improved a lot since the first implementation of dependent pairs, but that can be saved for another time.

ahmadsalim:
Thanks for the merge

@edwinb  I actually annotate them with the value or type annotation when it is clear parsing (because things are parsed differently), but it seems that it is largely ignored.

ahmadsalim:
Actually, for the example I gave above I can see that this would be hard to distinguish, but in the case like

```idris
(a : _ ** n : _ ** Vect a n)
```

edwinb:
Yes, the elaborator looks to see if there is a placeholder, rather than to see if the type was written, because that's all the parser records.

    (a ** n ** Vect n a)

comes out of the parser as

    (a : _ ** n : _ ** Vect n a)

ahmadsalim:
@edwinb Yes, I understand that.

What I meant was that I do set the pun if it the latter case in the parser, see:

https://github.com/idris-lang/Idris-dev/pull/2636/files#diff-b14a7a874e9d59c29a9ad4777f5419a7R494

ahmadsalim:
So if I know something must have definitely have been a type when parsed the `IsType` pun is set in the parsed expression, and if I know something must definitely have been a term then `IsTerm` is set.

PR #2649, opened 2015-09-23
Library of tactics for proof automation with Elab

david-christiansen:
I'm collecting all the various tactics that I've defined in my various
projects that have proven to be useful and making them into an
includable library, both as an example of Elab reflection and to have a
semi-standard vocabulary.

### DIFF
--- a/libs/pruviloj/Pruviloj.idr
+++ b/libs/pruviloj/Pruviloj.idr
@@ -1,0 +1,3 @@
+module Pruviloj
+
+import public Pruviloj.Core
--- a/test/pruviloj001/pruviloj001.idr
+++ b/test/pruviloj001/pruviloj001.idr
@@ -1,0 +1,77 @@
+-- Demonstrate that partial tactics can produce total programs and
+-- that the basic parts of Pruviloj are doing what they should.
+
+import Pruviloj
+import Pruviloj.Induction
+
+%default total
+
+||| Try some simplification and rewriting heuristics, then attempt to
+||| solve the goal
+partial
+auto : Elab ()
+auto = do compute
+          attack
+          try $ repeatUntilFail intro'
+          hs <- map fst <$> getEnv
+          for_ hs $
+            \ih => try (rewriteWith (Var ih))
+          hypothesis <|> search
+          solve
+
+
+||| ; in Coq (roughly)
+andThen : Elab (List TTName) -> Elab a -> Elab ()
+andThen tac1 tac2 =
+    do hs <- tac1
+       for_ hs $ \h =>
+         inHole h $ ignore tac2
+
+||| Common pattern of proofs by induction.
+partial
+mush : Elab ()
+mush =
+    do n <- gensym "j"
+       intro n
+       try intros
+       induction (Var n) `andThen` auto
+
+plusAssoc : (j, k, l : Nat) -> plus (plus j k) l = plus j (plus k l)
+plusAssoc = %runElab mush
+
+plusZeroR : (k : Nat) -> plus k Z = k
+plusZeroR = %runElab mush
+
+plusSRightS : (left : Nat) -> (right : Nat) -> S (left + right) = left + (S right)
+plusSRightS = %runElab mush
+
+multOneLNeutral : (right : Nat) -> 1 * right = right
+multOneLNeutral = %runElab mush
+
+multOneRNeutral : (left : Nat) -> left * 1 = left
+multOneRNeutral = %runElab mush
+
+plusMinusLeftCancel : (left : Nat) -> (right : Nat) -> (right' : Nat) ->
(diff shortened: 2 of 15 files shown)


libs/pruviloj/Pruviloj.idr
```diff
@@ -0,0 +1,115 @@
...
+ignore : Functor f => f a -> f ()
+ignore x = map (const ()) x
+
+||| Do nothing
+skip : Applicative f => f ()
+skip = pure ()
+
+||| Attempt to apply a tactic. If it fails, do nothing.
+try : Elab a -> Elab ()
+try tac = tac *> pure () <|> pure ()
```
Melvar: Perhaps `ignore tac` instead of `tac *> pure ()`? Also the function is generic to any `Alternative`.

libs/pruviloj/Pruviloj.idr
```diff
@@ -0,0 +1,115 @@
...
+
+||| Solve the goal using the most recent applicable hypothesis
+hypothesis : Elab ()
+hypothesis =
+    do hyps <- map fst <$> getEnv
+       choiceMap (\n => do fill (Var n); solve)
+                 hyps
```
Melvar: This seems like it could look nicer with `flip choiceMap`, i.e. something like

```idris
       flip choiceMap hyps $ \n => do
         fill (Var n)
         solve
```

If this is a frequent idiom, a dedicated combinator may be appropriate, as with `traverse`/`for`.

libs/pruviloj/Pruviloj.idr
```diff
@@ -0,0 +1,115 @@
...
+                                       (n' ::) <$> go body
+        go _ = return []
+
+||| Run a tactic inside of a particular hole, if it still exists. If
+||| it has been solved, do nothing.
+inHole : TTName -> Elab () -> Elab ()
+inHole h todo =
+  if h `elem` !getHoles
+    then do focus h; todo
+    else skip
```
Melvar: This could use `when` in place of `if`.

david-christiansen:
Thanks for the style notes, @Melvar!

test/pruviloj001/pruviloj001.idr
```diff
@@ -0,0 +1,75 @@
...
+          traverse_ (\ih => try (rewriteWith (Var ih))) hs
+          hypothesis <|> search
+          solve
+
+
+||| ; in Coq (roughly)
+andThen : Elab (List TTName) -> Elab a -> Elab ()
+andThen tac1 tac2 =
+    do hs <- tac1
+       traverse_ (\h => inHole h $ ignore tac2) (reverse hs)
```
david-christiansen: Remember to remove this `reverse` - it's not needed

david-christiansen:
I think this is ready for an initial merge. I will then update it as I find myself using the same derived tactics/tacticals in other contexts.

jfdm:
Erm quick question: The name `pruviloj` I think is a very bad choice without more context. As someone not familiar with elaboration nor tactics, searching for the term brings up nothing.

david-christiansen:
I think the name makes about as much sense as "Parsec" or "Twitter" :)

david-christiansen:
In other words, because this represents one way to do things and not The Way to do things, it's not getting a generic name. The idea is that it is unique and googleable.

jfdm:
Erm 'Parsec' and 'Twitter' both convey more information to the lay audience in their name w.r.t to their intended usage than `pruviloj`. `Parsec` has `Parse` to indicate parsing, and "Twitter" incorporates the notion of birds twittering. `pruviloj` implies nothing about intended use. I think we should err on the side of constructive naming than distinctive naming.

david-christiansen:
Your opinion is noted. I'll add a README explaining the lib and the name.

jfdm:
I still honestly think that `pruviloj` is still a bad choice given the context of its distribution. I'm not trying to bike-shed the issue, rather I think a more generic descriptive name is more appropriate for going into `libs/`. TBH if this wasn't going into the `libs/` directory and wasn't going to be distributed with Idris then I would not care, but it is so I do. 

I am well aware that the contents of `libs/` does not present the 'standard' way of doing X in Idris, and that it presents one possible construction. However, the names have nonetheless been simply descriptive, and are clearly indicative w.r.t their intended use. Further, given the package's placement as the 'default' and 'distributed', having a more accessible name for new users would be beneficial.

As an aside,the usability of the name for those not familiar with esperanto might be trying. When typing the name `pruviloj`, I have made mistakes every single time I have typed it. As this PR has been merged, I am going to file an issue over the name.

david-christiansen:
I asked around before I called it that, and nobody else objected, and many appreciated it getting a non-generic name. But if it does cause problems for more than just you, of course it could be renamed.

PR #2655, opened 2015-09-23
"Can't find name" message was missing a space

benjamin-hodgson:
I keep seeing this error message (I'm too lazy to actually read my code before I compile it). The mild annoyance of the obvious typo has compounded over many occasions, and finally it reached the tipping point for me to actually go ahead and make the one-character change.


david-christiansen:
Thanks!

PR #2668, opened 2015-09-26
Improvements to ipkg files for dependancies.

jfdm:
iPKG files have a new option `pkgs` which takes a comma-separated list
of package names that the idris project depends on. This reduces bloat
in the `opts` option with mutliple package declarations.

So rather than have:

```
package foo

opts = "-p lightyear -p containers -p tests -p pruviloj -p effects -p x --log 4"
```

we can now say:

```
package foo

opts = "--log 4"

pkgs = lightyear, containers, tests, pruviloj, effects, x
```

The file `PkgParser.hs` file was also cleaned up a little.

### DIFF
--- a/src/Pkg/PParser.hs
+++ b/src/Pkg/PParser.hs
@@ -2,13 +2,6 @@
 #if !(MIN_VERSION_base(4,8,0))
 {-# LANGUAGE OverlappingInstances #-}
 #endif
---  ------------------------------------------------------------- [ PParser.hs ]
---  Module    : PParser.hs
---  Copyright : (c) The Idris Community.
---  License   : see LICENSE
---
---  The parser for Idris ipkg files.
---  -------------------------------------------------------------------- [ EOH ]
 module Pkg.PParser where
 
 import Text.Trifecta hiding (span, charLiteral, natural, symbol, char, string, whiteSpace)
@@ -112,5 +105,3 @@ pClause = do reserved "executable"; lchar '=';
              ts <- sepBy1 (fst <$> iName []) (lchar ',')
              st <- get
              put st { idris_tests = idris_tests st ++ ts }
-
---  -------------------------------------------------------------------- [ EOF ]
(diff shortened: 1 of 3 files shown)


src/Pkg/PParser.hs
```diff
@@ -101,3 +113,4 @@ pClause = do reserved "executable"; lchar '=';
              st <- get
              put st { idris_tests = idris_tests st ++ ts }
 
+--  -------------------------------------------------------------------- [ EOF ]
```
david-christiansen: What is the point of this comment?

src/Pkg/PParser.hs
```diff
@@ -2,7 +2,13 @@
 #if !(MIN_VERSION_base(4,8,0))
 {-# LANGUAGE OverlappingInstances #-}
 #endif
-
+--  ------------------------------------------------------------- [ PParser.hs ]
```
david-christiansen: We don't have this kind of header anywhere else in the source repository - why should it be introduced with this change?

david-christiansen:
This is a nice feature to add - thanks for doing it!

I added a couple questions as line annotations.

src/Pkg/PParser.hs
```diff
@@ -2,7 +2,13 @@
...
+--  Module    : PParser.hs
+--  Copyright : (c) The Idris Community.
+--  License   : see LICENSE
+--
+--  The parser for Idris ipkg files.
```
david-christiansen: This is a good thing to add - but shouldn't it be a [Haddock module description](https://www.haskell.org/haddock/doc/html/ch03s03.html) instead of a comment?

jfdm:
i'll remove the 'extra' comments, they are just styling and visual aides I use when working with most source files.

david-christiansen:
Looks great! Once Travis is done, I'll merge and then go add support to `idris-mode`.

@archaeron: FYI, this is incoming, and the Atom mode should be made to understand it.

archaeron:
awesome! I like this a lot more.
is there a timeline on this?

Issue #1448, opened 2014-07-31
Behaviour when working with packages and setting options.


jfdm:
Currently when building packages from the command line, all options not included in the `ipkg` file are ignored. For example, given the command below:

```sh
$> idris --log 3 --build foobar.ipkg
```
The logging level is not set to 3. This default behaviour is counter intuitive in the cases that someone wishes to quickly add options when developing packages. For example: setting logging levels, setting verbosity output, setting warning flags *et cetera*.

I think the default behaviour should be that the options used when building a package should be the union of those specified in an `ipkg` file, and those specified on the command line. Furthermore, in the event that two options clash, I think that the options from the command line should have precedence. 

This should be flagged as low hanging fruit/be a mini project.

jfdm:
Going through some old issues and noticed that Issue #499 might actually be a symptom of this.

Issue #499, opened 2013-09-25
The --ibcsubdir flag doesn't generate the output uniformly


gabrielmc:
While trying to refactor the build to output everything into 'dist' (for #311), I noticed that `--ibcsubdir` doesn't actually work correctly when passed as a flag for building packages. For instance:

```bash
mkdir tmp
cd effects
sed -i '' -e 's#lib"#lib --ibcsubdir ../tmp"#' effects/effects.ipkg
idris --build effects.ipkg
```
will put the top-level Effects.ibc in place, but generate a new tmp directory in the effects directory for everything else.

I think this actually raises a larger issue of whether packages should be built by a different executable.

jfdm:
@gabrielmc sorry this has taken so so so so so so long to address. However, the root of the problem is that idris package build tooling does not use the option `ibcsubdir`, and appears to ignore it completely. Thus, there is a straightforward fix to the problem you highlighted. Update the package build tooling to be aware of the option. I'll try and get round to it when I have time...

PR #2671, opened 2015-09-27
Improve interaction between Command Line Options and Package File Options.

jfdm:
Package Options are now influenced by command line options.

This PR addresses in part: #1448, and #499.

When building packages preapproved command line options (if
list. The allowed options are:

+ Logging levels
+ Warn partial
+ Warn reach
+ Location of IBC subdir
+ Location of import dir.

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -57,7 +57,7 @@ runIdris opts = do
                        runIO $ exitWith ExitSuccess
        case opt getPkgCheck opts of
            [] -> return ()
-           fs -> do runIO $ mapM_ (checkPkg (WarnOnly `elem` opts) True) fs
+           fs -> do runIO $ mapM_ (checkPkg opts (WarnOnly `elem` opts) True) fs
                     runIO $ exitWith ExitSuccess
        case opt getPkgClean opts of
            [] -> return ()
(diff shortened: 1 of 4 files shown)


david-christiansen:
Sounds good!  You're on a roll today!

What's the reasoning behind silently removing invalid options when merging option lists, as opposed to halting and informing the user? It seems like the latter is less likely to be surprising, but there may be some reason that I'm not seeing.

jfdm:
Well, I wasn't sure how best to report any arising 'merge conflicts', given how the code is currently structured. My knowledge of haskell and the idris internals is murky.

jfdm:
Also I will add, I also wasn't sure what should constitute a merge conflict.

david-christiansen:
It would probably require a bit of a refactor - what I would do is put `buildPkg` and `mergeOptions` into the `Idris` monad, instead of the current respective `IO` and pure contexts, and then use `filterM` to conditionally remove elements, throw errors, or keep them.

david-christiansen:
I think it's fine for package options to override certain options - that should make sense. But there's a call to `filter` that silently discards some options - that's what seems like it should return an error to me.

david-christiansen:
(instead of using the `Idris` monad for error handling, `mergeOptions` could also return `Either Err [Opt]`, and then the callers could deal with the `Err` how they wanted - this might be cleaner, because `buildPkg` could stay in `IO` if that's where it really belongs)

jfdm:
Both options sound good. I was thinking of going the `Either` route as I wasn't sure if the package commands should be integrated into the Idris monad.

jfdm:
Well that went horrible wrong.

PR #2674, opened 2015-09-28
Fixed parsing error in ipkg files

jfdm:
+ Options was not updated to preserve any previosuly set options from.


jfdm:
This should fix the stupid parsing error in parsing ipkg files. I have added a test that should be suitable for future reference.

david-christiansen:
Thanks for catching it!

Issue #2642, opened 2015-09-21
Allow non-parenthesized infix operators after :t


david-christiansen:
It would be nice for `:t +` to mean the same thing as `:t (+)`.

RE https://github.com/idris-hackers/atom-language-idris/issues/66

PR #2677, opened 2015-09-29
Parse operators as expressions in the REPL

zgrannan:
This pull requests allows infix operators to be parsed as expressions in the REPL. Previously, the REPL did not parse infix operators as expressions.

This should resolve https://github.com/idris-lang/Idris-dev/issues/2642. However this applies to any repl command that takes an expression as an argument, not just the `:type` command.

Before:
```
Idris> :t ++
(input):1:4: error: expected: dependent type signature,
    end of input
:t ++<EOF> 
   ^   
```

After:
```
Idris> :t ++
Prelude.List.(++) : List a -> List a -> List a
Prelude.Strings.(++) : String -> String -> String
```

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -15,6 +15,7 @@ Jason Dagit
 Guglielmo Fachini
 Simon Fowler
 Google
+Zack Grannan
 Cezar Ionescu
 Heath Johns
 Irene Knapp
--- a/idris.cabal
+++ b/idris.cabal
@@ -535,6 +535,9 @@ Extra-source-files:
                        test/interactive009/input
                        test/interactive009/*.idr
                        test/interactive009/expected
+                       test/interactive010/expected
+                       test/interactive010/input
+                       test/interactive010/run
 
                        test/io001/run
                        test/io001/*.idr
--- a/src/Idris/REPLParser.hs
+++ b/src/Idris/REPLParser.hs
@@ -163,10 +163,9 @@ pCmd :: P.IdrisParser (Either String Command)
 pCmd = choice [ do c <- cmd names; parser c 
               | (names, _, _, parser) <- parserCommandsForHelp ++ parserCommands ]
      <|> unrecognized 
-     <|> nop 
+     <|> nop
      <|> eval
     where nop = do eof; return (Right NOP)
-          eval = exprArg Eval ""
           unrecognized = do
               P.lchar ':'
               cmd <- many anyChar
@@ -194,6 +193,11 @@ noArgs cmd name = do
 
     emptyArgs <|> failure
 
+eval :: P.IdrisParser (Either String Command)
+eval = do
+  t <- P.fullExpr defaultSyntax
+  return $ Right (Eval t)
+
 exprArg :: (PTerm -> Command) -> String -> P.IdrisParser (Either String Command)
 exprArg cmd name = do
     let noArg = do
@@ -201,9 +205,9 @@ exprArg cmd name = do
         return $ Left ("Usage is :" ++ name ++ " <expression>")
 
     let justOperator = do
-        (op, fc) <- P.maybeWithNS P.operatorFC False []
+        (op, fc) <- P.operatorFC
         eof
-        return $ Right $ cmd (PRef fc [] op)
+        return $ Right $ cmd (PRef fc [] (sUN op))
 
     let properArg = do
         t <- P.fullExpr defaultSyntax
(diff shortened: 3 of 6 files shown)


david-christiansen:
This seems useful to me. There's a few things preventing a merge right now, some of which I can check myself by building and running when I get a few minutes, and some of which should be changed in the PR.

1. How does it affect ordinary REPL evaluation? What happens when you just type `++` at the Idris REPL?
2. Please add the new test to the relevant section of `idris.cabal`. Otherwise, it isn't included in the output of `cabal sdist` and won't actually run on non-git-checkout copies of the Idris source tree (such as Travis).
3. Please add tests for a few REPL commands showing that it works for them, or that it doesn't. E.g. `:doc` and `:printdef` would be nice.
4. The `run` script should specify `--quiet`, `--nocolour` and `--consolewidth 80` to prevent non-essential things from breaking the test cases on platforms without color support, with different curses implementations, and if we update the banner.

Thanks so much!

david-christiansen:
Also, I don't recall you having submitted anything before, but I may be mistaken. Remember to add yourself to CONTRIBUTORS :-)

Melvar:
To me, this feels like papering over the problem by introducing a glaring inconsistency in name handling.

zgrannan:
@david-christiansen Thanks for the feedback. I've updated the PR to address points 2-4.

With regards to your first point, this PR should only change the behavior of normal repl evaluation when an operator is entered alone. The following are some examples

```
Idris> ++
(input):Can't infer argument a to (++)
Idris> /
prim__divFloat : Double -> Double -> Double
Idris> Nat.-
(-) : (m : Nat) -> (n : Nat) -> LTE n m -> Nat
```

david-christiansen:
@Melvar: Which problem do you see this a papering over? What would you think of as a real solution to that problem?

It concerns me a bit that this allows the typing of invalid Idris expressions at the REPL. For things like `:t` looking up the declared type of an unqualified name, it seems like a straightforward extension. But not only does this make grammatically invalid expressions get checked, it also makes qualified names that don't follow the Idris rules get parsed (e.g., `Nat.-` is not allowed and must be written `Nat.(-)`).

david-christiansen:
I removed the tag because this is not a WIP pull request - it's intended for merging as-is. I'm not sure we should do it though, because of the inconsistencies it introduces in allowed name syntax.

jfdm:
Bit of a meta comment here about PR workflows. My intention was that we need to differentiate between PRs whose inclusion is up for debate, with a means that is easily visible to those with commit rights. Rather than use a new tag, I used WIP as it was convenient. Maybe we should add a set of review labels i.e. `Needs Review`, `Under Review`, `Ready to Merge`, to indicate such a workflow.

david-christiansen:
That seems overly bureaucratic. One should simply read the comments before pressing the green button, and if there's no consensus, take that into account.

jfdm:
'simply read the comments' i think is an indicator that their needs to be a more defined process. but this thread is not the place for this discussion.

david-christiansen:
Agreed that another forum is better - please send a mail to the list!

Melvar:
My understanding is that the original problem is that it is nontrivial for editors to parse pieces out of source files. I expect that the instance starting this could have been fixed in the editor by checking if one had selected something operator-shaped, but this wouldn’t help if, for example, one used the type query on a syntax keyword. I also don’t know if it would parse qualified operator names correctly, which would be another complication to add. So, perhaps it needs to be made easier to use Idris for this parsing. I believe one could already gather the relevant things out of the hilighting info, but if that is too complex to implement in general, something like a `type-at`, `classify-at`/`what-is` or even a `help-at` command might be the way to go.

david-christiansen:
@Melvar the highlighting info contains all the relevant information already, so Emacs users aren't bugged by getting this, and neither will others when the rest of the editors catch up. But when on the command line, all those parens are also a bit irritating to type - writing `:doc ++` is much easier than `:doc (++)`. The question is really whether we should special-case infix `names`, just as we special-case names in general for various commands that begin with a colon. I don't think we should treat them as valid expressions.

zgrannan:
@david-christiansen 
Agree with you about the handling of infix expressions in regular repl evaluation. I'll work on updating the PR to disallow the special case for names in regular expressions as well as to disallow expressions such as `Nat.-` later tonight.

zgrannan:
I've made the above-mentioned changes.

Melvar:
See, I don’t find writing `:doc ++` detectably easier than `:doc (++)`, though just now it occurs to me that the parens are well-placed on the keyboard layout I use. At any rate, every other place that takes a Name does not accept `++`, including binders, syntax that takes Names, and Name quotations, so I would say that `++` is not syntactically a Name at all. I guess if you phrase it as commands now taking bare operators, a separate thing from Names, it sounds somewhat more palatable.

melted:
I interpret the discussion as finding this acceptable and it seems useful to me.

PR #2679, opened 2015-09-29
Decidable.Equality: properties of decEq.

cbiffle:
These trivial properties can be proven for any implementation of DecEq.
They come in handy for discharging `with (decEq x y)` clauses using
`rewrite`.

There are several areas for potential discussion here.  Off the top of my head:

- Should properties like this go into `Decidable.Equality` or some sub-module like `Decidable.Equality.Properties`?  Personally, I'd argue either for the former, or for `Decidable.Equality` to re-export `...Properties`, for discoverability reasons.

- Naming.  I've used the camel-case-separated-by-underscores pattern that was discussed on the mailing list recently.  This is unusual but not without precedent in the standard library.

I'm wide open to suggestions on both points.

### DIFF
--- a/libs/prelude/Decidable/Equality.idr
+++ b/libs/prelude/Decidable/Equality.idr
@@ -8,6 +8,7 @@ import Prelude.Either
 import Prelude.List
 import Prelude.Nat
 import Prelude.Maybe
+import Prelude.Uninhabited
 
 --------------------------------------------------------------------------------
 -- Utility lemmas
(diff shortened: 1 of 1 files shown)


cbiffle:
(I've amended this after noting that this module is not `%default total` -- I've added `total` to the properties I defined.)

david-christiansen:
I'm sorry to have not replied to your invitation to discussion yet - I'm really quite swamped.

WRT point 1: I agree that properties should just be in the main namespace, so that `:search` can find them.

WRT point 2: the only thing I _really_ care about is some kind of consistency. The underscores are fine, but perhaps we should rewrite everything that way and then be vigilant about future contributions? @ahmadsalim did this with constructor names a few releases back, and it's really made things much nicer.

Melvar:
Now that I look closer at this, are there problems with using the more generic
```
decNot : (d : Dec a) -> Not a -> (x ** d = No x)
```
? In the case of `decEq_refl` I can see that the generalization would require tedious plumbing through of Refls, but with `decEq_not` I’m not sure.

edwinb:
Sorry for the lack of discussion on this PR... 

We probably need some written conventions for naming in the Prelude. I don't especially care what those conventions are, as long as we use them consistently. Personally I'd prefer camel case here (especially since the constructor is called 'Refl' not 'refl')

In the past we've said that new contributions to the Prelude should start in contrib/ and be added if they are proved to be useful. There's a case for bypassing this when adding properties and lemmas, though.

edwinb:
Discussion appears to have stalled here. I'm inclined to suggest that for now properties of DecEq should go into contrib, and that we should put some proper thought into how to deal with this sort of thing in the Prelude.

For example, in addition to your basic functions, we also have to think about where properties go, if they should be in separate modules, and where views go (I would find a library of views on Lists and Vects very useful, for example).

ahmadsalim:
Since there has not been any recent work or response from the original poster in over 10 days I am going to close this PR. 

@cbiffle If you are interested in continue working on the properties, you are more than welcome to reopen the PR or make a new one.

In any case thank you for your effort.

Issue #2639, opened 2015-09-20
Safe String to Int function.


NightRa:
There should definitely be a safe function for parsing Ints from Strings in the standard library.
```idris
parseInt : String -> Maybe Int
```

The existance of an unchecked Cast from String to Int in Idris is really bewildering to me, and in my opinion, it shouldn't even exist.

```idris
x : Int
x = cast "hello"
> x = 0
```

markuspf:
Cast is a convenience hack. Use a parser instead.

NightRa:
Are you saying one needs a whole parser library to read ints, say from a console, safely?
I believe it deserves its place outside a parser library too.

NightRa:
I see this as a design mistake. Why offer the wrong, illogical, hacked solution by default and not the right thing by default?

markuspf:
What, other than a parser, would a solution use?

There is no deeper design behind Cast. Cast is a convenience, and maybe a mistake.

We'd like to see a better solution, someone will have to contribute it though.

suhr:
What about generic `parse` function for strings? Something like http://doc.rust-lang.org/std/primitive.str.html#method.parse

jfdm:
I wouldn't say cast is a 'hack' it does it's job fine. It is just that people have to explicitly provide the conversion routines. What people are talking about is the provision of `Read` instances that do the parsing. Remember that `read (show a) === a`.

suhr:
It *is* a hack as long as it provides invalid casts.

NightRa:
@markuspf Not a whole parser library, just a simple recursion / loop.
I maybe could contribute something.
"Casts" should not be losing information and should be from types that have an obvious embedding in the other. 
For Isomorphisms, casts can be made both ways.

Prisms cover what we generally want here, but restrict the implementation quite a bit.
I've been thinking about section/retraction pairs to do this sort of thing, which do allow multiple strings to be parsed to the same result, in contrast to prisms.

But first of all, there needs to be a proper function 
```idris
parseInt : String -> Maybe Int
```
The ```Cast String Int``` should be deprecated, 
and implemented for the time being as ```maybe 0 id . parseInt```

edwinb:
Casts as implemented in the Cast type class may lose information, e.g. casting a Double to an Int, or an Integer to an Int, or an Int to a Nat. It's not in the library documentation that they may lose information, and it should be.

[...]

PR #2680, opened 2015-10-01
Safe String to Int function. #2639

eklavya:


libs/prelude/Prelude/Strings.idr
```diff
@@ -340,6 +340,20 @@ toUpper x with (strM x)
...
 
+||| Convert a number string to int.
+|||
+||| ```idris example
+||| parseInt "123"
+||| ```
+parseInt : String -> Maybe Int                                                                                                                                                       
+parseInt s with (strM s)                                                                                                                                                             
+  parseInt ""             | StrNil         = Nothing                                                                                                                                 
+  parseInt (strCons x xs) | (StrCons x xs) = parseIntAux (unpack xs) (ord x - 48) where
```
david-christiansen: This magic number 48 seems a bit difficult to read - why not write `ord '0'`?

david-christiansen:
Thanks for the contribution! Do we want this in `prelude` as-is? It seems to me that as the 1.0 release approaches, we should be a bit conservative about the contents of the default namespace. I'd rather see this in `base`, behind an `import Data.Int` or something like that.

libs/prelude/Prelude/Strings.idr
eklavya: Yeah, sorry I meant to change that but forgot while doing the indentation check :tongue:

eklavya:
So in which file do you want it?

melted:
Travis is failing, saying:
```
Type checking ./Prelude/Strings.idr
./Prelude/Strings.idr:348:10:
When checking type of Prelude.Strings.parseInt:
No such variable Maybe
````

**Comment by eklavya** [comment]
Just waiting for the location, where this change should be included, to be decided to make the changes.

**Comment by Melvar** [comment]
Is there a reason it doesn’t return an arbitrary instance of `Num`? What about a derived version for `Neg` which allows a sign?

**Comment by eklavya** [comment]
@Melvar I am not sure if I understood that. Are you saying parseInt should be:
```Idris
parseInt : String -> Num
```
I can look for a leading - to parse the integer as negative. Would that do?
I am very likely to have completely misinterpreted what you said :)

david-christiansen:
Melvar's got a good point here RE `Num` and `Neg`. What he means is that the signature could be

`parsePositive : Num a => String -> Maybe a`

and that one that accepts negatives could be
`parseInteger : (Num a, Neg a) => String -> Maybe a`

As far as location goes, I think it makes sense in libs/base/Data/Integer.idr.

david-christiansen:
It would also make sense in libs/base/Data/String.idr or something like that.

CraigStuntz:
For the sake of comparison, .NET's int.Parse accepts leading and trailing whitespace, as well as a leading '+'. https://msdn.microsoft.com/en-us/library/b3h1hf19(v=vs.110).aspx

eklavya:
File has been added as libs/base/Data/String.idr

It supports parsing 
- positive nums 
```Idris
parsePositive : Num a => String -> Maybe a
```
- nums
```Idris
parseInteger : (Num a, Neg a) => String -> Maybe a
```

It takes care of '-'/'+' or leading or trailing spaces.

david-christiansen:
Thanks for being persistent! I added two notes. I think that one is a bug, and the other is merely a style issue.

After that, I think it's good to go.

eklavya:
I was almost sure that someone will tell me a shorter way of doing the string trimming (without an extra where) and a way out of nested ifs :tongue: 

Thanks for being patient :)

david-christiansen:
Thank you for your patience while we keep bombarding you with comments and requests!

libs/base/Data/String.idr
```diff
@@ -0,0 +1,49 @@
...
+||| parsePositive "123"
+||| parsePositive {a=Int} " +123"
+||| ```
+parsePositive : Num a => String -> Maybe a
+parsePositive s = parsePosTrimmed (trim s) where
+  parsePosTrimmed s with (strM s)
+    parsePosTrimmed ""             | StrNil         = Nothing
+    parsePosTrimmed (strCons x xs) | (StrCons x xs) = 
+      if (x == '+') 
+        then map fromInteger (parsePosAux (unpack xs) (cast {from=Int} 0))
```
david-christiansen: Why do you need the cast here? Does Idris give an error if you just write `0`?

libs/base/Data/String.idr
```diff
@@ -0,0 +1,49 @@
...
+        else map fromInteger (parsePosAux (unpack xs)  (cast (ord x - ord '0')))
+          where
+            parsePosAux : (List Char) -> Integer -> Maybe Integer
```
david-christiansen: There shouldn't be parens around `List Char` here, because application binds tighter than `->`

libs/base/Data/String.idr
```diff
@@ -0,0 +1,49 @@
...
+            parsePosAux []        acc = Just acc
+            parsePosAux (c :: cs) acc = 
+              if (c >= '0' && c <= '9') 
+                then parsePosAux cs ((acc * (cast {from=Int} 10)) + (cast ((ord c) - (ord '0'))))
```
david-christiansen: Why do you need that `cast`? It would seem that an integer literal should be enough.

libs/base/Data/String.idr
david-christiansen: If it's due to not enough type information, then IMO a cleaner solution is adding a type annotation to `parsePosTrimmed`.

libs/base/Data/String.idr
```diff
@@ -0,0 +1,49 @@
...
+  parseIntTrimmed s with (strM s)
+    parseIntTrimmed ""             | StrNil         = Nothing
+    parseIntTrimmed (strCons x xs) | (StrCons x xs) = 
+      if (x == '-') 
+        then map (\y => negate (fromInteger y)) (parseIntegerAux (unpack xs) (cast {from=Int} 0))
+        else if (x == '+') 
+          then map fromInteger (parseIntegerAux (unpack xs) (cast {from=Int} 0))
+          else map fromInteger (parseIntegerAux (unpack xs) (cast (ord x - ord '0')))
+            where
+              parseIntegerAux : (List Char) -> Integer -> Maybe Integer
```
david-christiansen: `List Char` doesn't need these parens.

libs/base/Data/String.idr
```diff
@@ -0,0 +1,49 @@
...
+              parseIntegerAux []        acc = Just acc
+              parseIntegerAux (c :: cs) acc = 
+                if (c >= '0' && c <= '9') 
+                  then parseIntegerAux cs ((acc * (cast {from=Int} 10)) + (cast ((ord c) - (ord '0'))))
```
david-christiansen: Why `cast {from=Int}` here?

libs/base/Data/String.idr
```diff
@@ -0,0 +1,49 @@
+module Data.String
+
+||| Convert a positive number string to a Num.
+|||
+||| ```idris example
+||| parsePositive "123"
```
david-christiansen: The syntax for examples doesn't work quite that way. Each example needs to be in its own block to have it render properly when viewing docs:
`````
||| ```idris example
||| parsePositive "123"
||| ```
||| ```idris example
||| parsePositive {a=Int} " +123"
||| ```
`````

david-christiansen:
One final style note: we don't yet have formal style guidelines for the Idris code repo, but most of the code in the library already has the `where` keyword at the beginning rather than end of lines, and type annotations on each `where`-block definition.

One of these days I'll make a guideline and a tool to check it.

david-christiansen:
And I'm sorry to be so slow replying to things - a million things going on right now!

eklavya:
I could have sworn the compiler complained about the int literals. It seems to hate me!

I got that "if" and "where" style from the tutorials or docs somewhere. So glad it's not the norm (in case of if).

eklavya:
No, it's not.

david-christiansen:
I seem to have been foolish. Thanks for bearing with me!

eklavya:
@david-christiansen you are too kind, thanks for bearing with ME! :)

PR #2686, opened 2015-10-02
Restructed vector add example and fixed error in code snippet.

cmcl:
read_vec was missing an explicit cast to the computation context
when running parseNumber. I also moved this code snippet to just
below where it is described in the text (removing the spurious
link to a non-existent listing).

The remaining change was just a typo fix in another snippet.

### DIFF
--- a/docs/effects/simpleeff.rst
+++ b/docs/effects/simpleeff.rst
@@ -382,7 +382,7 @@ definition:
 
     vadd : Vect n Int -> Vect n Int -> Vect n Int
     vadd []        []        = []
-    vadd (x :: idris xs) (y :: ys) = x + y :: vadd xs ys
+    vadd (x :: xs) (y :: ys) = x + y :: vadd xs ys
 
 Using , we can set up a program so that it reads input from a user,
 checks that the input is valid (i.e both vectors contain integers, and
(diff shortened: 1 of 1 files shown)


docs/effects/simpleeff.rst
```diff
@@ -410,10 +410,28 @@ following type:
...
-indicate the end of input, is shown in Listing [readvec]. This uses a
-variation on ``parseNumber`` which does not require a number to be
-within range.
+is going to input. We can use ``-1`` to indicate the end of input:
+
+.. code-block:: idris
+
+    read_vec : Eff (p ** Vect p Int) [STDIO]
+    read_vec = do putStr "Number (-1 when done): "
+                  case the (Maybe Int) $ run (parseNumber (trim !getStr)) of
```
david-christiansen: Is this really the nicest way to provide the missing type information, as opposed to (say) giving `run` an explicit type argument? And shouldn't the surrounding text also say something about _why_ we need to do this?

cmcl:
I'm not sure what you mean by, "give run an explicit type argument"; do you mean in the call or in its definition? If the former, that is more or less what this achieves. If the latter, its definition is polymorphic with respect to the computation context. It is not entirely clear to me why Idris cannot infer the correct type from the case branches; all I know is that without it, the program does not typecheck.

edwinb:
Sorry for leaving this a while. The thing that bothers me is that I think needing the explicit type is a regression - this certainly used to work or it wouldn't be in the tutorial!

edwinb:
Oh, actually, I can see why this happens, and I would expect it. Previously, it got the type by accident due to a flaw in the way type classes were resolved. Idris doesn't look at the case branches to infer the type, only the expected return type and the type of the thing being inspected, because of the way the elaborator is set up.

For the moment, the code snippet should at least be something that works! I'll see if I can find a way to improve case elaboration in cases like this though.

PR #2689, opened 2015-10-02
Allow certain package options to be overridden from console.

jfdm:
This PR addresses in part: #1448, #499.

When building packages a preapproved command line options (if
list such that they. The allowed options are:

+ Logging levels
+ Warn partial
+ Warn reach
+ Location of IBC subdir
+ Location of import dir.

Idris also reports an error if invalid command lines options are
specified when used to override package specified options.

The code will also facilitate specific reporting of the offending
flags, if we choose to want such functionality.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -40,6 +40,9 @@ Tool updates
 * iPKG files have a new option `pkgs` which takes a comma-separated list
   of package names that the idris project depends on. This reduces bloat
   in the `opts` option with mutliple package declarations.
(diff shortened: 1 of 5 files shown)


jfdm:
This is not building as the filter for command line options is filtering out the existing packages. I forgot the set of options that was filtered was also the default options. I fill fix this when I next get the chance.

jfdm:
I am killing this as the branches have become to divergent, and it is probably better to start afresh in this.

PR #2690, opened 2015-10-02
Fix typo

justjoheinz:


jfdm:
Cheers thanks for fixing the typo's. It is always annoying to see them, but even better seeing them fixed.

PR #2691, opened 2015-10-03
pure -> pureM in depeff tutorial

mitchellwrosen:


david-christiansen:
I'm not totally up on the effects library right now - could you say a couple of words about why this change is better? Does the old one throw an error?

mitchellwrosen:
Yeah, you need to use `pureM` here rather than `pure` because the out-list off effects depends on the value returned.

```
pure : a -> EffM m a xs xs
pureM : (val : a) -> EffM m a (xs val) xs
```

edwinb:
The version in the tutorial is correct. The reason is that it's actually doing the opposite of what you expect! The function you're writing has a general type, calculated from the value that's being returned (it's an 'if' expression, branching according to your return value), whereas the value you're giving is specific so the type 'pure' will give is too specific. Going from 'xs val' to 'xs', therefore, allows you to go from specific to general, rather than the other way around.

Otherwise, yes, this is a fix that needed making. Thanks!

PR #2692, opened 2015-10-03
last -> list

mitchellwrosen:


jfdm:
Cheers. Typo's are always annoying to have in the documentation. Thanks for finding it and fixing it..

PR #2693, opened 2015-10-03
length (letters w) -> length w

mitchellwrosen:


mitchellwrosen:
Oops, didn't see that `letters` is defined later.

PR #2695, opened 2015-10-04
Implement support for C callbacks

melted:
This adds support for wrapping up Idris functions so they can be passed to C as callbacks.

It also has functionality to get the function pointer in case they need to be stuffed into structures.

Functions in IO doesn't work yet. That will need to be fixed before it can be merged.

### DIFF
--- a/test/ffi007/ffi007.c
+++ b/test/ffi007/ffi007.c
@@ -1,0 +1,21 @@
+#include <stdio.h>
+
+typedef int (*callback)(int);
+typedef int (*cb2)(char *);
+
+void test_ffi(callback cb)
+{
+    printf("%d\n", cb(3));
+}
+
+void test_ffi2(cb2 cb)
+{
+    printf("%d\n", cb("hello"));
+}
+
+void test_ffi3(void (*cb)(void))
+{
+    printf("Before calling callback\n");
+    cb();
+    printf("After calling callback\n");
+}
--- a/test/ffi007/ffi007.idr
+++ b/test/ffi007/ffi007.idr
@@ -1,0 +1,35 @@
+module Main
+
+%include c "ffi007.h"
+
+getInt : Int -> Int
+getInt _ = 8
+
+test : IO ()
+test = foreign FFI_C "test_ffi" (CFnPtr (Int -> Int) -> IO ()) (MkCFnPtr getInt)
+
+strLen : String -> Int
+strLen s = cast $ length s
+
+test2 : IO ()
+test2 = foreign FFI_C "test_ffi2" (CFnPtr (String -> Int) -> IO ()) (MkCFnPtr strLen)
+
+-- Do IO
+printIt : () -> ()
+printIt _ = unsafePerformIO $ do
+    putStrLn "In an Idris callback"
+
+test3 : IO ()
+test3 = foreign FFI_C "test_ffi3" (CFnPtr (() -> ()) -> IO ()) (MkCFnPtr printIt)
+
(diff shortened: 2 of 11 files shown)


david-christiansen:
Cool!

Melvar:
What is the role of `CVoid`? Why do functions need to be wrapped in `CFnPtr`? Are you planning to add support to the executor as well?

melted:
CVoid is there because a function taking just () as an argument will be reduced, so there's nothing to wrap in that case. So we need it to create wrappers taking nothing.

CFnPtr and those parts are modeled on the Javascript FFI for function arguments.

melted:
As for the executor, that should be possible using libffi, ffi_prep_closure can be used to build them at runtime, and then there can be just one target function that can sort out the arguments and execute the idris, data can also be passed on the side via a pointer for user data. The only problem is that isn't included in the Haskell bindings, which really only supports calling functions, it's unmaintained and has has no releases in six years so that likely won't change.

david-christiansen:
We could call libffi using Haskell's C FFI directly if we had to, or we could update the bindings and take over maintenance of them. It would be nice to keep the executor and the C FFI as close as possible, though maybe we should also just bite the bullet and split them. There's already a couple of differences regarding handling of pointers, after all, and this can be highly surprising.

Melvar:
I am beginning to put together a sort of picture how the implementation in the executor could work.

test/ffi007/ffi007.idr
```diff
@@ -0,0 +1,33 @@
+module Main
+
+%include c "ffi007.h"
+
+getInt : Int -> Int
+getInt _ = 8
+
+test : IO ()
```
mbj: Lets named that one `test1` for consistency with `test2` etc.

test/ffi007/ffi007.c
```diff
@@ -0,0 +1,21 @@
+#include <stdio.h>
+
+typedef int (*callback)(int);
+typedef int (*cb2)(char *);
+
+void test_ffi(callback cb)
```
mbj: Lets name that one `test_ffi1` for consistency with the other ffi test functions.

Melvar:
Say, instead of taking CVoid, couldn’t one take a universally quantified type? That ensures its behavior can’t depend on the argument, but doesn’t feel as troublesome as a function taking an empty type actually being executed.

melted:
I'm going to revisit this now. I wonder if it would be a great hardship to require that the generated C program is linked against libffi. Then we could use that for the wrappers.

melted:
I have removed CVoid as it isn't needed.

Sending IO functions still doesn't work, the equivalent of unsafePerformIO needs to be done and it's too late to add unsafePerformIO to the program when we're in codegen. I have added a helpful error message that mentions unsafePerformIO.

jfdm:
Cool. However, two things:

+ The CHANGELOG needs updating.
+ It *might be* useful to take the ffi007 test and write a brief tutorial/guide on how to get the callbacks working.
+ It *might also be* useful to add some text to the reference manual.

melted:
Yes, those are good points. I'm working on documentation. 

Skickat: ‎2016-‎02-‎07 11:11
Ämne: Re: [Idris-dev] Implement support for C callbacks (#2695)

Cool. However, two things:
The CHANGELOG needs updating. 
It might be useful to take the ffi007 test and write a brief tutorial/guide on how to get the callbacks working. 
It might also be useful to add some text to the reference manual.

PR #2699, opened 2015-10-06
Simplify reflection and Raw

david-christiansen:
Some constructors of TT will never occur during elaboration, but are
instead a part of the optimization process. Thus, they have been removed
from reflection, to simplify the interface.

Likewise, I discovered that RForce is no longer produced anywhere in
Idris, so it was also deleted from both the reflection API and the
internals.


david-christiansen:
In other words, YAY DEAD CODE DELETION!

PR #2718, opened 2015-10-10
Pruviloj and reflection updates

david-christiansen:


Melvar:
I’m not sure why you’re prefixing the type name to the projections. Why force people to write (and read) `tyDeclName` every time when `name` would normally suffice, and `TyDecl.name` works when it doesn’t?

david-christiansen:
I tend to do it because I think it makes things more readable, rather than less readable. It could be that I'm alone in this viewpoint, however, in which case they should be renamed now before anything becomes dependent on them. I'll think about it for a bit.

Melvar:
I personally frown on duplicating namespace information in the local name. Users have much better control over use of namespaces, so doing it takes away a degree of freedom.

david-christiansen:
On the other hand, when not actively reading the code in Emacs with semantic annotations, it can make it much more difficult to know which version was selected. I'll puzzle a bit more over it and see what things look like locally with more overloading.

Melvar:
I also feel a bit like it’s a haskellism. There I believe it’s done because Haskell is so severely allergic to name clashes, and it doesn’t have namespaces that aren’t whole-file modules. Now that I think about it, in Haskell if one qualifies a name, one qualifies it by provenance and not by origin the way we do in Idris, but you could conceivably want to provide two things that could have the same name out of the same module, which would thus be fatal in Haskell.

Melvar:
I also can’t help thinking: If not for this, what is the stuffing of accessors into an extra namespace for?

david-christiansen:
All right, you convinced me :-)

Update incoming.

Issue #2721, opened 2015-10-11
Allow dashes in package names


shlevy:
Unless there's some ambiguity introduced when doing so? I note that every project owned by @idris-lang has a dash in its name :wink:

shlevy:
Also in `executable` names

jeremy-w:
+1 from me. I'd expect the changes for generated executable names wouldn't be terribly invasive.

jeremy-w:
Executable names:

- `PParser.hs:pClause` parses the executable name as an `iName`.
- `ParseHelpers.hs:iName`'s docstring says that it "Parses an identifier with possible namespace as a name". This is kind of a weird choice for an executable parser! But you can indeed name your executable like "executable.with.dots.in.it".

It looks like adding a parser for a valid filename, and using that instead of `iName`, would fix this.

jeremy-w:
I have the executable name bit almost working, except I'm leaving the quote marks around the executable name. `cabal install` takes so long that I won't be able to resolve and test this tonight.

jeremy-w:
I think I've fixed it, but it remains to add a test to the test suite. Work so far is available [on this branch](https://github.com/jeremy-w/Idris-dev/tree/jeremy-w/ipkg-allow-arbitrary-executable-names-2721).

david-christiansen:
This has annoyed me for a while, but never enough to do anything about it. Thanks for taking initiative!

shlevy:
@edwinb Is this something you'd prefer not to be in the issue tracker?

ahmadsalim:
I think this has been fixed in #2728.

jfdm:
i tried this the other day, and a package name with a hyphen was rejected... I am inclined to reopen this issues if other people are seeing this error.

ahmadsalim:
@jfdm You have to add quotes around the name.

jfdm:
ah okay, we should update the docs...

ahmadsalim:
@jfdm Yeah, I had to look at the source to figure it out :).

jeremy-w:
Ah, changelog should have noted it, but other docs likely still need updating. IIRC at the time I had to look at the source to figure out how IPKG worked in general, and expected eventually to convert that knowledge into proper docs.

Maybe we should open a new issue to update the IPKG docs about this?

jfdm:
[...]

PR #2728, opened 2015-10-13
IPKG: Relax requirement that executable name be a valid identifier [#2721]

jeremy-w:
- Preserves support for current format like `executable = some.namespaced.identifier` (test/pkg002)
- Adds new format using a quoted string like `executable = "some-name-with-hyphens and spaces"` (test/pkg003)
    - New format rejects names that have a directory component like `invalid/filename`. (test/pkg004)

TODO:

- [x] Address ["illegal tuple constraint" warning](https://github.com/idris-lang/Idris-dev/pull/2728#issuecomment-147666654) under GHC 7.6/7.8
- [ ] Replace `iName`/`identifier` with a quote-free flavor of filename parsing
- [x] Add check that it's a valid path (so no bogus characters) [using `System.FilePath.isValid`](https://github.com/idris-lang/Idris-dev/pull/2728#issuecomment-147690603)
- [ ] Add entry to CHANGELOG

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -41,7 +41,7 @@ Tool updates
   datatype
 * iPKG files have a new option `pkgs` which takes a comma-separated list
   of package names that the idris project depends on. This reduces bloat
-  in the `opts` option with mutliple package declarations.
+  in the `opts` option with multiple package declarations.
 
 Miscellaneous updates
 ---------------------
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -166,4 +166,4 @@ To help increase the chance of your pull request being accepted:
 * [General GitHub Documentation](https://help.github.com/).
 
 
-Adapted from the most excellent contributing files from the [Puppet project](https://github.com/puppetlabs/puppet) and [Factroy Girl Rails](https://github.com/thoughtbot/factory_girl_rails/blob/master/CONTRIBUTING.md)
+Adapted from the most excellent contributing files from the [Puppet project](https://github.com/puppetlabs/puppet) and [Factory Girl Rails](https://github.com/thoughtbot/factory_girl_rails/blob/master/CONTRIBUTING.md)
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -35,6 +35,7 @@ Adam Sandberg Eriksson
 Seo Sanghyeon
 Benjamin Saunders
 Alexander Shabalin
+Jeremy W. Sherman
 Timo Petteri Sinnemäki
 JP Smith
 startling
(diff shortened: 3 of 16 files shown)


jeremy-w:
This is the first non-documentation PR I've issued.

Two questions:

- Should I add an item in CHANGELOG describing this change?
- Should I add myself to CONTRIBUTORS?

melted:
GHC 7.6 and 7.8 are failing on Travis with the error:
```
src/Pkg/PParser.hs:70:13:
    Illegal tuple constraint: MonadicParsing m
      (Use ConstraintKinds to permit this)
    In the type signature for ‘filename’:
      filename :: (MonadicParsing m, HasLastTokenSpan m) => m String
```

melted:
You should definitely add yourself to the contributors list! It should probably go in the changelog as it's a user-visible if minor change.

david-christiansen:
I'm still not entirely convinced about the rules for unquoted identifiers. It seems to me that we should either make it mandatory to quote filenames (which breaks backwards compatibility, unfortunately) or we should accept at least filenames matching [a-zA-Z0-9_.+-]+, and preferably also non-ASCII letters like æ, ø, å, á, ĉ, ł, İ and ideally also non-Latin scripts. Perhaps @Melvar can weigh in on the precise Unicode classifications that make sense here :-)  After all, most systems today support non-ASCII filenames. But I think a minimum for non-quoted names should be ASCII letters, numbers, underscore, dash, dot, and plus.

Melvar:
Well, ordinary Idris identifiers already allow Letter followed by Letters, Marks, and Numbers. `+` is a Math Symbol. `_.-` are Punctuation (Connector, Other, and Dash respectively). `"` is also Other Punctuation, so if including whole Categories, you’d need to exclude it specifically.

Separately, does this code check for all the extra disallowed characters on Windows?

david-christiansen:
No, it doesn't. It probably should call `isValid` from `System.FilePath`, which will be the right function on Windows or Unix. Probably the parser should accept one or more non-whitespace characters or a string literal, then first check that `takeFileName` doesn't change the name, that it's non-empty, and that it passes `isValid`.

jeremy-w:
Parsing it as unquoted non-whitespace characters or a double-quoted string and then validating the text afterwards using isValid seems like the best way to leverage what System.FilePath already knows about in/valid filenames.

Does that sound a good solution to you?
--
Jeremy W. Sherman
http://jeremywsherman.com/

david-christiansen:
Well, I think we need to validate three things:

1. It's non-empty (but this can be implicit in the parser definition)

2. It has no directory components (i.e. it's a bare filename) - this is to prevent overwriting files by setting `executable = "~/.authinfo"` or something

3. It's a valid filename on the current system

PR #2736, opened 2015-10-15
Reference docs: correct sentence about type syntax.

enolan:
Types don't actually need to be capitalised, unlike Haskell.

Someone should write a section on how implicit arguments work, but for
now this is at least less wrong.

### DIFF
--- a/docs/reference/syntax-guide.rst
+++ b/docs/reference/syntax-guide.rst
@@ -112,7 +112,7 @@ Types
 
 In Idris, types are first class values. So a type declaration is the
 same as just declaration of a variable whose type is ``Type``. In Idris,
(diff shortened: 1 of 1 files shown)


edwinb:
I would rather not make it explicit in the tutorial that types need not have a capital letter because in practice it can cause complications if they don't... there's now a warning and it might turn into an error. So I actually slightly prefer the old way, even if it isn't exactly right.

The tutorial is really intended as a quick start guide, not a complete reference document, so it doesn't cover everything or go into a huge amount of detail. We do need to have a reference document, though.

enolan:
This *is* the reference document. :) It's in docs/reference not docs/tutorial

edwinb:
Oh, so it is, silly me!

PR #2738, opened 2015-10-15
Renaming Dependent Pairs

jfdm:


jfdm:
i'm happy for this to get merged in, and in future deprecated the old names. Existing code should work.

edwinb:
Personally I think these are better names, but I'm about to do a release so I think it's best to wait until after that to merge into master.

It also occurs to me that the Unique things don't really belong in the prelude yet. They are too experimental and nobody is really actively working with them. So I'll move them at some stage.

jfdm:
Okay, I will do the changes but unfortunately I can only get to them in around a months time. I have frozen my dev env untill after the big day, breaking changes that affect my code were made to idris and I do not want to upgrade my code just yet.

david-christiansen:
@edwinb You're OK with a merge now, right?

jfdm:
Right, I am able to work on this again. I have rebased, and I am adding `%deprecate` annotations for the old names.

jfdm:
The build is failing due to the new name for dependent pairs (`DepPair`) being used in `meta002`. I tried to work out what went wrong but i am not familiar with the test itself. Maybe @david-christiansen can have a quick nosey.

jfdm:
Finally this can be merged assuming the tests pass...they should do I checked locally.

PR #2739, opened 2015-10-15
Declare a dependency on gmp when the gmp flag is set.

### LINKED ISSUE #2499: Undeclared dependency on libgmp library
Idris depends on libgmp when build with appropriate flags, but the Cabal does not declare an `extra-libraries: gmp` stanza.

shlevy:
Fixes #2499

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -993,6 +993,7 @@ Library
      cpp-options:   -DIDRIS_FFI
   if flag(GMP)
      build-depends: libffi < 0.2
+     extra-libraries: gmp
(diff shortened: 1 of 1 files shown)


shlevy:
Any problems with this PR?

melted:
Thanks for the PR!

jfdm:
@shlevy the reason your PR took so long to be merged is that everyone who could have merged it was probably busy and just missed it. It sadly does happen, but in these cases a bump should be sufficient to trigger someone to act. If there were problems someone would say.

PR #2741, opened 2015-10-15
Pruviloj-related updates

david-christiansen:


Melvar:
It’s too bad I didn’t see this PR in time: The doc for `apply` has not changed to describe its new signature, and indeed describes the old one in some detail. Same with `matchApply`.

david-christiansen:
It'll get fixed :-) thanks for the notice.

PR #2748, opened 2015-10-17
Remove apparently unused iif directory

shlevy:


melted:
Let's merge this, we can always find the files in the history if someone misses them.

PR #2753, opened 2015-10-19
Keep :set options in sync.

mrmonday:
This change will keep tab completion for `:set`, `:help`, and parsing in sync
as new options are added. Documentation has also been updated with all options.


melted:
Thanks!

PR #2755, opened 2015-10-20
Update Mystery Word example code

Legogris:
* Make it typecheck (previous version of game function didn't)
* Make it more readable (if instead of case over bool, explicit precondition on NewWord)

### DIFF
--- a/docs/effects/hangman.rst
+++ b/docs/effects/hangman.rst
@@ -108,16 +108,15 @@ effect signature:
          Guess : (x : Char) ->
                  sig MysteryRules Bool
                      (Mystery (Running (S g) (S w)))
-                     (\inword =>
-                            Mystery (case inword of
-                                 True => (Running (S g) w)
-                                 False => (Running g (S w))))
+                     (\inword => if inword
+                                 then Mystery (Running (S g) w)
+                                 else Mystery (Running g (S w)))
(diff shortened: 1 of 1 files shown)


docs/effects/hangman.rst
```diff
@@ -108,16 +108,15 @@ effect signature:
...
+                     (\inword => if inword
+                                 then Mystery (Running (S g) w)
+                                 else Mystery (Running g (S w)))
          Won  : sig MysteryRules () (Mystery (Running g 0))
                                     (Mystery NotRunning)
          Lost : sig MysteryRules () (Mystery (Running 0 g))
                                     (Mystery NotRunning)
          NewWord : (w : String) -> 
-                   sig MysteryRules () h (Mystery (Running 6 (length (letters w))))
+                   sig MysteryRules () (Mystery NotRunning) (Mystery (Running 6 (length (letters w))))
```
mrmonday: Did you mean to make this change? Surely `NewWord` should work for both `NotRunning` and `Running`? (though I guess it isn't important for this particular version of the code).

Legogris: Yes, I just thought it made the code more understandable for new readers. And it doesn't really make sense to change the word in the middle of a running game of hangman. It's not vital for making it run, though.

mrmonday: That make sense - I hadn't read through the full example.

PR #2759, opened 2015-10-21
PParser.hs: filename: Report the specific failed check on error

jeremy-w:
This makes the error message more specific and so more useful to the
person trying to debug their iPKG file.

This is an addendum to #2728. I've been sitting on it for a week or so as I tried to work out how to improve the reported span, but it benefits no-one hiding on my hard-drive. :)

### DIFF
--- a/src/Pkg/PParser.hs
+++ b/src/Pkg/PParser.hs
@@ -15,6 +15,7 @@ import Idris.CmdOptions
 import Control.Monad.State.Strict
 import Control.Applicative
 import System.FilePath (takeFileName, isValid)
+import Data.Maybe (isNothing, fromJust)
 
 import Util.System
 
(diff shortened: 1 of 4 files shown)


src/Pkg/PParser.hs
```diff
@@ -79,18 +80,33 @@ filename = (do
         -- Through at least version 0.9.19.1, IPKG executable values were
         -- possibly namespaced identifiers, like foo.bar.baz.
         show <$> fst <$> iName []
-    if isValidFilename filename
+    let errorMessage = filenameErrorMessage filename
+    if isNothing errorMessage
       then return filename
-      else fail "a filename must be non-empty and have no directory component")
+      else fail $ fromJust errorMessage)
```
Melvar: This `if isNothing`/`fromJust` use looks like it should be replaced with a `case` or a call to `maybe`.

melted:
I saw that the pkg tests options were broken, so I fixed up them in https://github.com/idris-lang/Idris-dev/pull/2761, which means this will PR have to be rebased on that.

jeremy-w:
@melted: Thanks for fixing that. I carried forward the existing pkg test settings, but the expected output is far more readable with those settings. :)

melted:
Rerunning travis...

melted:
Thanks for the PR!

Melvar:
Whoops, looks like we’ve got color escapes in a test output!

PR #2765, opened 2015-10-25
Define void with an elab script, not elim_for

shlevy:
Eliminator generation is deprecated, so let's remove this use of a generated eliminator.

### DIFF
--- a/libs/prelude/Prelude.idr
+++ b/libs/prelude/Prelude.idr
@@ -37,6 +37,8 @@ import public Language.Reflection.Errors
 %default total
 
 -- Things that can't be elsewhere for import cycle reasons
+-- See comment after declaration of void in Builtins.idr
+-- for explanation of this definition's location
 %runElab (defineFunction $ DefineFun `{void} [])
 
 decAsBool : Dec p -> Bool
--- a/libs/prelude/Prelude.idr
+++ b/libs/prelude/Prelude.idr
@@ -37,6 +37,8 @@ import public Language.Reflection.Errors
 %default total
 
 -- Things that can't be elsewhere for import cycle reasons
+%runElab (defineFunction $ DefineFun `{void} [])
+
 decAsBool : Dec p -> Bool
 decAsBool (Yes _) = True
 decAsBool (No _)  = False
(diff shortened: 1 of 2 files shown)


libs/prelude/Prelude.idr
```diff
@@ -37,6 +37,8 @@ import public Language.Reflection.Errors
 %default total
 
 -- Things that can't be elsewhere for import cycle reasons
+%runElab (defineFunction $ DefineFun `{void} [])
```
david-christiansen: I'm not a fan of the definition being so far from the declaration - why does it need to be that way?

I can see that Builtins exists prior to the existence of Language.Reflection.Elab, which means it can't be there, but what about moving the declaration of Void to Prelude.idr?

shlevy: `Prelude.Uninhabited` seems to get in the way of that.

david-christiansen: Can `void` be declared in there? It should be able to import `Language.Reflection.Elab`, I would think...

shlevy: `Nat` imports it.

shlevy: Might it help to have a comment in `Builtins.idr` that the definition is in `Prelude.idr`?

shlevy: bump

david-christiansen: Sorry, I'm really busy right now! Yes, I think that having comments pointing between the two _and_ summarizing the dependency issue should be ok. After all, `void` doesn't compute :-)

shlevy: Hmm, another possible option is to move the definition of `void` to Uninhabited and have Uninhabited import Nat rather than the other way around. Testing...

shlevy: Eh that will be a huge reorg, I'll just do the comment for now :grin:

shlevy:
@david-christiansen OK, added comments.

david-christiansen:
Thanks for doing this, and sorry it took me so long to get it merged.

Issue #2743, opened 2015-10-16
Proposal: Annotate prelude with %used


lenary:
I've been doing some work on my erlang backend, and it seems possible to confuse the erasure system into such a way that it thinks `List.::` doesn't take any arguments once it's got to code generation.

I can only assume this is because the uses of cons it inspects is full of proof-terms, not values, and thus every part can be erased. I guess this is partly a bug in erasure, that it doesn't analyse later uses of cons to check if they are full of values.

Given the `%used` annotation exists, is it worth using everywhere in the prelude? I'm happy to make the code changes, just wanted to get a go-ahead before I did. 

I guess if `%used` is still considered to be a hack, rather than a feature, then this is probably a bad idea.

david-christiansen:
I'm a bit confused by your issue - is the problem that no uses of the head or tail arguments to `::` are runtime relevant, causing their erasure, which breaks some assumptions about `List` in the Erlang backend, or is the problem that there are indeed runtime-relevant uses of the head and the tail, but they are getting erased anyway?

lenary:
So, after further discussion with @ziman, it turns out that there were various things leading to the issue.

- It was an example program, all lists were being passed to a foreign function as their only "usage"
- Erasure can't actually look *through* a foreign function, because of a few reasons. Firstly, if it assumed all constructor arguments were used, it would keep type information around (the first argument of `::`). Secondly, if it assumes none are used, it would keep no arguments around, also not quite what is expected. What is expected is that the last 2 arguments are kept, which is not generally deducible. 

This leads me to think annotating all data constructors in the prelude with the correct `%used` annotations would be useful, and would allow backends to assume their presence when used in foreign function interfaces, including if a backend wanted to special-case any prelude items.

I *don't* think that we need to annotate anything outside of prelude, because there isn't quite the same assumption about presence in that code as there is for code inside the prelude.

Upon further inspection, some declarations of the form `data List e = Nil | (::) e (List e)` may need to be changed so I can name the two arguments, so changing to the form below. Currently running tests to see if this breaks anything.

```
||| Generic lists
%elim data List : (elem:Type) -> Type where
  ||| Empty list
  Nil : List elem
  ||| A non-empty list, consisting of a head element and the rest of
  ||| the list.
  (::) : (e : elem) -> (l : List elem) -> List elem

-- Name hints for interactive editing
%name List xs, ys, zs, ws

-- Usage hints for erasure
%used (::) e
%used (::) l
```

[...]

PR #2769, opened 2015-10-26
Annotate the Prelude with %used for backends

lenary:
This is my answer to #2743.

Not all the tests pass. There's a funny interaction with '%name' that prevents interactive things passing, and of course the Quasiquotation tests fail because some of the code has changed. I'd appreciate some help deciding which of those tests are now wrong, and which are really failures.


melted:
I don't think I'm able to help with the quasiquoter. I wonder about the primitives002 test though, it just seems to crash and burn.

lenary:
@melted I think the quasiquoter expectations just need to be updated, but would like some confirmation from @david-christiansen (or someone else who knows the system well enough to know whether i've broken it).

As for primitives002, I'll take a look, though I'm somewhat surprised I broke it, given I don't think I have code that interacts with it.

david-christiansen:
The quasiquote test outputs shown in Travis are fine - just update the output with the new expected information. The quoted forms show the names bound in Pi, which you've modified from an MN to a UN, and this is reflected in the output. Just go into the test directory and type `./run > expected`.

lenary:
Ok, so, tests pass, though there's the weird "killed…" message in the output, which I think is from something else entirely.

edwinb:
I'm slightly scared by the "killed" message. I've restarted the tests on the off chance it was some one off travis thing, but the fact that there's a message but the tests report success is a bit worrying. I'd like to at least work out where the message came from before merging...

lenary:
It seems to be coming from this: https://github.com/idris-lang/Idris-dev/blob/master/test/meta002/Deriving.idr

The same 'killed...' message is happening in the master branch builds: https://travis-ci.org/idris-lang/Idris-dev/jobs/88681037#L1396

lenary:
I'm going to open a new ticket for that "killed…" message, as it's totally orthogonal to this pull request. I've been doing a little analysis, and think I've nailed down the cause of it.

david-christiansen:
I don't think there's anything blocking this, but I don't want to push "merge" due to not being entirely sure about the erasure mechanism and backends. If @ziman or @edwinb says "go", I'm happy to do it.

ziman:
I don't think this will break anything.

edwinb:
Done. Sorry for slowness.

PR #2772, opened 2015-10-29
Added deleteAt, replaceAt and insertAt for Lists

DanielWaterworth:


melted:
Thanks for the PR!

david-christiansen:
I'm not actually sure that we should have merged this - these operations don't seem like they would be generally applicable and useful enough to go in the Prelude.

The reason that I'm skeptical about how useful this is is that in order for the auto implicit to work, the list would have to be available as a literal (or at least the elements up to the one to be deleted). In that case, one could just leave it out.

I'd be happy to stick this in `contrib`, but I think we should revert the merge to Prelude.

PR #2773, opened 2015-10-29
Added Always and Elem types

DanielWaterworth:


jfdm:
Documentation would be nice for this. Also are `Always`, and `Elem` appropriate names here. 
For instance, in [this implementation of an `Always` data structure](https://github.com/jfdm/idris-containers/blob/master/Data/Sigma/DList.idr) An alternate name was used.

DanielWaterworth:
I found these types in various tutorials so I just gave them the names they had there.

jfdm:
Other names used are `All` and `Any` (cf. Agda's) implementation, regardless, if there is a chance to use better names, then we shoudl go for it.

DanielWaterworth:
Sure. Incidentally, DList isn't a good name since, at least for me, it has connotations of https://hackage.haskell.org/package/dlist-0.7.1.2

DanielWaterworth:
In fact, https://github.com/edwinb/idris-demos/blob/master/Intro/elem.idr

jfdm:
I will not discuss naming issue's here over the naming of `DList`, those should be brought up elsewhere. Besides reading the documentation and type makes the purpose of DList clear.

Second, with the version you pointed to for `Elem` I still do not think the name is appropriate. For example, in prelude we have an inconsistency wrt to naming of list test functions:

+ `elem`
+ `isSuffix`
+ `isPrefix`

This is I was going to have a data type to represent proof of an elements existance in a list, then is `IsElem` not a more appropriate name. Just because there is demonstrably evidence of its use before, doesn't mean that was the best name to use.

DanielWaterworth:
`IsElem` is fine with me. I also think `All` is better than `Always`.

Melvar:
I would suggest adjusting these to match the `Vect` versions that are already present in the library, namely see `Data.Vect` and `Data.Vect.Quantifiers`. Certainly given the existence of other structures with corresponding concepts, these should not be directly under `Data` in my opinion.

jfdm:
Agreed, that would be better since this primarily affects Lists.

DanielWaterworth:
I didn't even realize they existed. Definitely a good idea.

DanielWaterworth:
However, `Elem` is different than `Any`. `Elem` says a particular item is a member of a list, `Any` says that a particular property holds for any member of a list.

Melvar:
Yes, I pointed at `Data.Vect` for `Elem`.

DanielWaterworth:
Sorry, my bad.

libs/contrib/Data/Always.idr
```diff
@@ -0,0 +1,39 @@
+module Data.Always
+
+import Data.Elem
+
+%default total
+
+data Always : (x -> Type) -> List x -> Type where
```
david-christiansen: `x` is a bit of an unconventional name for a type variable here - typically, we use `a`, `b`, `c` or something descriptive like `elem` to range over types.

Also, I'd expect the name `Always` to refer to the similar predicate over streams, rather than the List version.

libs/contrib/Data/Always.idr
```diff
@@ -0,0 +1,39 @@
+module Data.Always
```
david-christiansen: Shouldn't this be in Data.List or Data.List.All/Data.List.Always? Similar structures could be defined for lots of list-like structures, such as Vect and friends.

libs/contrib/Data/Always.idr
```diff
@@ -0,0 +1,39 @@
...
+
+index : Elem x xs -> Always f xs -> f x
+index Here (x :: _) = x
+index (There elem) (_ :: xs) = index elem xs
+
+replaceAt : Elem x xs -> f x -> Always f xs -> Always f xs
+replaceAt Here element (_ :: xs) = element :: xs
+replaceAt (There n) element (x :: xs) = x :: replaceAt n element xs
+
+replaceAndIndex : {elements:Always f xs} -> index elem (replaceAt elem x elements) = x
```
david-christiansen: There should be spaces around the colon in a type annotation. Also, the compiler is unlikely to solve the `Always f xs` unless you make the argument `auto`.

What have you used this for in the past?

libs/contrib/Data/Always.idr
```diff
@@ -0,0 +1,39 @@
...
+replaceAndIndex {elem = Here} {elements = Nil} impossible
+replaceAndIndex {elem = Here} {elements = (_ :: _)} = Refl
+replaceAndIndex {elem = There elem} {elements = Nil} impossible
+replaceAndIndex {elem = There elem} {elements = (_ :: _)} = replaceAndIndex
+
+replaceAndIndexElsewhere : {elements:Always f xs} -> (elem1 = elem2 -> Void) -> index elem1 (replaceAt elem2 x elements) = index elem1 elements
```
david-christiansen: See above - there should be spaces around the colon, and it should probably be an `auto` implicit.

libs/contrib/Data/Always.idr
```diff
@@ -0,0 +1,39 @@
...
+replaceAndIndexElsewhere f {elem1 = Here} {elem2 = Here} = void (f Refl)
+replaceAndIndexElsewhere _ {elem1 = Here} {elements = Nil} impossible
+replaceAndIndexElsewhere _ {elem1 = There _} {elements = Nil} impossible
+replaceAndIndexElsewhere _ {elem1 = Here} {elem2 = There _} {elements = _ :: _} = Refl
+replaceAndIndexElsewhere _ {elem1 = There _} {elem2 = Here} {elements = _ :: _} = Refl
+replaceAndIndexElsewhere f {elem1 = There _} {elem2 = There _} {elements = _ :: _} = replaceAndIndexElsewhere (f . cong)
+
+replaceOrder : {elements:Always f xs} -> (elem1 = elem2 -> Void) -> replaceAt elem2 x2 (replaceAt elem1 x1 elements) = replaceAt elem1 x1 (replaceAt elem2 x2 elements)
```
david-christiansen: Spaces and auto :-)

libs/contrib/Data/Elem.idr
```diff
@@ -0,0 +1,49 @@
+module Data.Elem
```
david-christiansen: I think this should be in `Data.List.Elem`, because similar structures make sense for other datastructures as well.

libs/contrib/Data/Elem.idr
```diff
@@ -0,0 +1,49 @@
...
+
+position : Elem xs x -> Nat
+position Here = 0
+position (There p) = 1 + position p
+
+positionInBounds : (prf : Elem x xs) -> InBounds (position prf) xs
+positionInBounds Here = InFirst
+positionInBounds (There p) = InLater (positionInBounds p)
+
+lookupElemEq : (prf:Elem x xs) -> index (position prf) xs {ok=positionInBounds prf} = x
```
david-christiansen: There should be spaces around colons in type signatures.

david-christiansen:
Thanks for the contribution!

I stuck some comments on the code, many reiterating earlier discussion here. I think that, to merge, it should have the libraries moved under Data.List, Always should be renamed to All, and every exported name should have a docstring with at least a sentence saying what it does.

david-christiansen:
I'm going to close this due to no activity for about a month. Please feel free to re-open.

PR #2774, opened 2015-10-29
List quantifiers

DanielWaterworth:


melted:
The Travis check fails. It's because this adds a file but it isn't added to the sources list in idris.cabal, so it gets left out when building a source tarball.

Issue #2666, opened 2015-09-25
Invalid syntax accepted


mrmonday:
As far as I can tell, the following code uses invalid syntax:

```idris
foo : Int -> Bool
foo 1 | 2 | 3 = True
foo _ = False
```

But compiles successfully.

edwinb:
The syntax is valid (it's the syntax for a with block clause, and clauses are parsed independently) but the elaborator should reject it because there's no expected patterns from a with block. As it is, it's just silently rejecting the additional arguments.

This might be a good one for someone to look at who is interested in learning a bit about how the elaborator works. We should probably take a closer look reasonably soon, though, since a similar syntax is the kind of thing people coming from Haskell expect to work (as guards).

ahmadsalim:
I know get:

```idris
Type checking ./test.idr
test.idr:2:5:When checking left hand side of foo:
unexpected patterns outside of "with" block
```
which is a suitable error.

PR #2775, opened 2015-10-29
Fix issue #2666 (Invalid syntax accepted)

bravit:


jfdm:
Can a test be added to show this catch in action?

jfdm:
Also adding more to the commit message, so that it is self contained and does not necessarily need referencing to the github issue tracker would be nice to see as well.

bravit:
Done.

melted:
Thanks!

PR #2776, opened 2015-10-29
Ioexcept golf

LeifW:
Simplify some expressions in IOExcept.
 It seemed weird to be to be implementing <*> in terms of >>=, even though it's at a different level... Tried to make it look more like "apply the thing to the inner thing using the outer thing's version of that", like is the case with map and pure.

Use ioe_fail helper method where appropriate in Catchable and Exception Effect.

Add *> as a shortcut for do a; b on Effects
    
Like Haskell's >>.
Was surprised *> wasn't already available on Effects, until I realized
their >>= wasn't from Monad.

### DIFF
--- a/libs/base/Control/IOExcept.idr
+++ b/libs/base/Control/IOExcept.idr
@@ -10,8 +10,7 @@ instance Functor (IOExcept e) where
 
 instance Applicative (IOExcept e) where
      pure x = IOM (pure (pure x))
-     (IOM f) <*> (IOM a) = IOM (do f' <- f; a' <- a
-                                   return (f' <*> a'))
+     (IOM f) <*> (IOM a) = IOM [| f <*> a |]
 
 instance Monad (IOExcept e) where
      (IOM x) >>= k = IOM (do x' <- x;
@@ -21,15 +20,10 @@ instance Monad (IOExcept e) where
                                   Left err => return (Left err))
 
 ioe_lift : IO a -> IOExcept err a
-ioe_lift op = IOM (do op' <- op
-                      return (Right op'))
+ioe_lift op = IOM $ map Right op
 
 ioe_fail : err -> IOExcept err a
-ioe_fail e = IOM (return (Left e))
+ioe_fail e = IOM $ pure (Left e)
 
 ioe_run : IOExcept err a -> (err -> IO b) -> (a -> IO b) -> IO b
-ioe_run (IOM act) err ok = do act' <- act
-                              case act' of
-                                   Left e => err e
-                                   Right v => ok v
-
+ioe_run (IOM act) err ok = either err ok !act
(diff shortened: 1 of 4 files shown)


libs/base/Control/IOExcept.idr
```diff
@@ -10,8 +10,7 @@ instance Functor (IOExcept e) where
 
 instance Applicative (IOExcept e) where
      pure x = IOM (pure (pure x))
-     (IOM f) <*> (IOM a) = IOM (do f' <- f; a' <- a
-                                   return (f' <*> a'))
+     (IOM f) <*> (IOM a) = IOM $ (<*>) <$> f <*> a
```
Melvar: This could be `(IOM f) <*> (IOM a) = IOM [| f <*> a |]` I think.

LeifW: Yep. Wasn't sure which was preferred. I was worried that variant might not be so obvious that "it's defined by applying <*> to the inner in terms of the outer <*>" e.g. that pattern you see in the map and pure impls above. But it does look tidier.

libs/base/Control/IOExcept.idr
```diff
@@ -2,34 +2,25 @@ module Control.IOExcept
...
 
 instance Monad (IOExcept e) where
-     (IOM x) >>= k = IOM (do x' <- x;
-                             case x' of
-                                  Right a => let (IOM ka) = k a in
-                                                 ka
-                                  Left err => return (Left err))
+     (IOM x) >>= f = IOM $ x >>= either (pure . Left) (io . f)
```
LeifW: I guess just `(\(IOM io) => io)` instead of `io` here would work too, and avoid making `IOException` into a record.

libs/base/Control/IOExcept.idr
david-christiansen: I think the version with the idiom brackets is much more readable, personally.

libs/base/Control/IOExcept.idr
```diff
@@ -2,34 +2,25 @@ module Control.IOExcept
 
 -- An IO monad with exception handling
 
-data IOExcept : Type -> Type -> Type where
-     IOM : IO (Either err a) -> IOExcept err a
+record IOExcept err a where
+     constructor IOM
+     io : IO (Either err a)
```
david-christiansen: I'm not a fan of this projection name. Other, similar types use something like `runIOExcept`.

LeifW: updated

libs/base/Control/IOExcept.idr
LeifW: updated

LeifW:
Outstanding comments addressed via force-pushed updates.

Issue #2779, opened 2015-10-30
Clean up deprecated FFI mechanisms


jmitchell:
In the Type Provider docs `sizeOfDevT` calls `mkForeign`, which apparently isn't available. Was it part of a deprecated FFI system?

There are a few other [lingering uses](https://github.com/idris-lang/Idris-dev/search?utf8=%E2%9C%93&q=mkForeign).

Unless the new FFI has a drop-in replacement for `mkForeign`, cleanup will involve additional functions and types.

david-christiansen:
Yes, this is part of the old FFI. Thanks for catching it!

jmitchell:
I've started work on this (see ebbae35323e8328757641eda2423fb867589c62b). There's a test to prove type providers and the new FFI are compatible. I can patch up the tutorial soon as well.

PR #2788, opened 2015-11-03
Fix/master/cleanup deprecated ffi

jmitchell:
See #2779 for context.

### DIFF
--- a/test/tutorial007/expected
+++ b/test/tutorial007/expected
@@ -1,0 +1,1 @@
+sizeof(size_t): 8 bytes
--- a/test/tutorial007/expected
+++ b/test/tutorial007/expected
@@ -1,1 +1,1 @@
-sizeof(size_t): 8 bytes
+Pass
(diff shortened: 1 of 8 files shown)


test/tutorial007/expected
```diff
@@ -0,0 +1 @@
+sizeof(size_t): 8 bytes
```
Melvar: This looks like a problematic assumption. Tests are supposed to work everywhere.

jmitchell: Oops, thanks for catching that.

edwinb:
The test failure here is just due to my having changed the interface for readFile to return an error if it fails. Therefore also due to us being slow to merge - sorry about that! If you can fix that then I'll merge. Thanks!

jfdm:
I am closing this as #3070 takes this PR and addresses the changes to Idris that happened mid PR. Thanks @jmitchell for the original hard work.

PR #2791, opened 2015-11-05
[WIP] Well-founded recursion

shlevy:
This is the start of work to add a richer notion of well-founded recursion to the language, inspired by Coq's standard library. The next step is to teach the compiler about the notions added here, which I can think of doing in 2 ways:

1. Trust instances of `prove_smaller` like `assert_smaller`, *only if* every recursive call passes the argument we're recursing on through `prove_smaller` with the same well-founded relation
2. Add syntax that desugars into a call to `recurse` (and remove `prove_smaller` and `smaller`).

I'm pretty sure 1 is safe (though I haven't proven it), but 2 may be easier to swallow. Let me know if this is something we want and which approach is preferred if so.

### DIFF
--- a/libs/prelude/Prelude/WellFounded.idr
+++ b/libs/prelude/Prelude/WellFounded.idr
@@ -1,0 +1,41 @@
+||| Well-founded recursion
+module WellFounded
+
+import Builtins
+
+%access public
+%default total
+
+||| An element is accessible for a relation
+||| @ R The relation
+||| @ x The element that is accessible for `R`
+data Accessible : (R : a -> a -> Type) -> (x : a) -> Type where
+  ||| An element is accessible if every element that relates to it is accessible
+  MkAccessible : ((y : _) -> r y x -> Accessible r y) -> Accessible r x
+
+||| The type of well-founded relations
+||| @ R The relation that is well-founded
+wellFounded : (R : a -> a -> Type) -> Type
+wellFounded r = (x : _) -> Accessible r x
+
+||| Recurse over a well-founded relation
+recurse : (wellFounded r) ->
+          .{p : a -> Type} ->
+          ((x : a) -> ((y : a) -> r y x -> p y) -> p x) ->
+          (x : a) -> p x
+recurse {r} {p} wf rec x = recurse' x (wf x) where
+  recurse' : (x : a) ->
+             (Accessible r x) ->
+             p x
+  recurse' x (MkAccessible base) = rec x (\y => \r => recurse' y (base y r))
+
+||| Prove one term smaller than another over a relation
+|||
+||| If used to satisfy the termination checker, all recursive calls must go through
+||| `prove_smaller` with the same relation.
+prove_smaller : .(wellFounded r) -> .(x : a) -> (y : a) -> .(r y x) -> a
+prove_smaller _ _ y _ = y
+
+||| Automatically prove one term smaller than another over a relation
+smaller : .(wellFounded r) -> .(x : a) -> (y : a) -> .{auto pf : r y x} -> a
+smaller wf x y {pf} = prove_smaller wf x y pf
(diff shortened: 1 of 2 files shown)


shlevy:
@edwinb feedback requested

shlevy:
(It may be possible to do all of this with `syntax` extensions, but if so it would involve a lot of reflection and be pretty hairy)

david-christiansen:
We already have a lot of this stuff in the `contrib` package, which is where I think it should stay until it's been made convenient or been more widely used. But it would be good to get some nice syntax for using it!

shlevy:
@david-christiansen Are we OK with compiler syntax referencing stuff in non-prelude libs?

shlevy:
Ah, I like the type-class usage in contrib though, that makes this cleaner

david-christiansen:
In particular, see:

 * https://github.com/idris-lang/Idris-dev/blob/master/libs/contrib/Control/WellFounded.idr
 * https://github.com/idris-lang/Idris-dev/blob/master/libs/contrib/Data/Nat/DivMod/IteratedSubtraction.idr

david-christiansen:
I think that the thing you should try first for making handy syntax for this is to use a combination of syntax rules and Elab scripts, rather than adding it directly to the compiler. I think that it will be similarly hairy to do this in Haskell as it would be in Idris, but having it be library code makes the maintenance burden much lower.

shlevy:
How can I handle multiple clauses in a syntax rule?

david-christiansen:
You can't right now :-) One possibility is to have a decoration on each RHS, another is that we extend Elab to be able to look up the clauses of functions so that you can rewrite it after-the-fact.

david-christiansen:
Another thing you might look into is a DSL for writing well-founded recursion, and then use either a function or Elab to interpret as calls to recursion over the accessibility predicate.

shlevy:
If we did the latter, wouldn't the pre-rewritten definition fail the totality checker before we can get in there?

david-christiansen:
Yes. That's OK - partial functions are allowed in Idris.

david-christiansen:
It might look like this:
```
partial divMod : (a, b : Nat) -> DivMod a b
divMod ....

divMod' : (a, b : Nat) -> DivMod a b
makeWellFounded divMod divMod'
```
where `divMod` has the annotations you're describing, and this process might also produce some proof obligations as a side effect.

shlevy:
Hmm, and we'd have to somehow make reasonable syntax for passing in explicit proofs of the relation when it can't be inferred.

shlevy:
Ah, or we can just live with the proof obligations, right.

david-christiansen:
You don't need to pass them in - you can use the `metavar` tactic to produce a hole to be filled later, just like the provisional definition syntax.

shlevy:
I think I understand, but can you point me to the "provisional definition syntax" so I can verify?

david-christiansen:
http://docs.idris-lang.org/en/latest/tutorial/provisional.html

david-christiansen:
Hmm, that needs updating to use "extract lemma" and explicit proofs.

shlevy:
How hard would it be to extend the elaborator to look up function clauses? That seems like possibly the best option here.

david-christiansen:
Not very. If you want to do it, you can look at how `:printdef` works - it's the same datastructure that should be used. Otherwise I can do it after I get married next week :-)

david-christiansen:
What I'd do is have a signature like:
```
lookupFunctionDef : TTName -> Elab (List FunDefn)
```
and then a derived operation
```
lookupFunctionDefExact : TTName -> Elab FunDef
```
that doesn't admit overloading.

After this, it's mostly a matter of following the existing patterns in the code.

This is something I will do eventually, because I want to derive Bove-Capretta predicates automatically :-)

shlevy:
OK, will see if I can find time to look into this, thanks!

shlevy:
@david-christiansen I'm guessing I need something analogous to https://github.com/idris-lang/Idris-dev/commit/d97715efd18d0517b0a8792e90329d198f0b9377 to propagate through `idris_patdefs`? Is there no way to look this stuff up through the `Context` of the proof state?

david-christiansen:
I think I answered your question elsewhere (IRC or other Github bits). Let me know if not.

ahmadsalim:
The PR seems to have a nearly completed component.
@shlevy Would it be possible to clean up the finished part into contrib and separate the remaining usability enhancements into their own issues?

Thanks!

shlevy:
@ahmadsalim https://github.com/idris-lang/Idris-dev/blob/master/libs/contrib/Control/WellFounded.idr :smile:

ahmadsalim:
@shlevy Great, thanks! I will close this then.

PR #2814, opened 2015-11-12
Implement missing idris_mkFileError for windows rts files

Heather:
- work around error: undefined reference to `idris_mkFileError'
 - minor code formatting corrections (use 4 spaces, no tailoring spaces)

(tested)

### DIFF
--- a/rts/idris_heap.c
+++ b/rts/idris_heap.c
@@ -7,11 +7,11 @@
 
 
 /* Used for initializing the heap. */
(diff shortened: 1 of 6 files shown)


melted:
Thanks!

I think we should separate out the parts of idris_stdfgn.c that actually differ between posixy platforms and windows. The duplication is uneccessary. But the fix shouldn't wait for that.

Heather:
@melted that's could be in some TODO list, it will be handy to have such

PR #2815, opened 2015-11-13
idris_stdfgn.c: use conditional compilation

Heather:
since there is only one function which is different for windows realization

(btw Makefile contains tabs mixed with spaces)

related: https://github.com/idris-lang/Idris-dev/pull/2814
(tested)


melted:
Thanks, that looks good.

Tabs and spaces aren't interchangable in makefiles.  For stupid reasons they require leading tabs.

PR #2818, opened 2015-11-14
Add doc comment and deprecation warning for Float

enolan:


jfdm:
Looks like some test output needs updating. Also there appears to be a lot of deprecation warnings being generated. Is that because `Float` is used, or because the type synonym is being evaluated and deprecate warning on the use of Double...

enolan:
Oh durr, I should've ran the testsuite. Assume doc only changes couldn't possibly break anything, but of course the deprecation does. I'll look at this tomorrow.

enolan:
Works now :)

david-christiansen:
Thanks for doing this! I'll update the reflection bits.

PR #2820, opened 2015-11-17
rmFile: don't even try removing not existing files

Heather:
Often when I run idris --clean (specially automated) I see warnings due trying to remove files which are not present.
I wish to not see those annoying messages

### DIFF
--- a/src/Util/System.hs
+++ b/src/Util/System.hs
@@ -16,7 +16,6 @@ import System.Directory (getTemporaryDirectory
                         , removeFile
                         , removeDirectoryRecursive
                         , createDirectoryIfMissing
-                        , doesFileExist
                         )
 import System.FilePath ((</>), normalise)
 import System.IO
@@ -66,7 +65,7 @@ isATTY = do
 withTempdir :: String -> (FilePath -> IO a) -> IO a
 withTempdir subdir callback
   = do dir <- getTemporaryDirectory
-       let tmpDir = (normalise dir) </> subdir
+       let tmpDir = normalise dir </> subdir
        removeLater <- catchIO (createDirectoryIfMissing True tmpDir >> return True)
                               (\ ioError -> if isAlreadyExistsError ioError then return False
                                             else throw ioError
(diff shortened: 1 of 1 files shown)


Melvar:
That’s racy. Isn’t the proper way to catch the specific exception and not print the message in that case?

src/Util/System.hs
```diff
@@ -76,10 +76,13 @@ withTempdir subdir callback
...
 
 rmFile :: FilePath -> IO ()
-rmFile f = do putStrLn $ "Removing " ++ f
-              catchIO (removeFile f)
-                      (\ioerr -> putStrLn $ "WARNING: Cannot remove file "
-                                 ++ f ++ ", Error msg:" ++ show ioerr)
+rmFile f = do
+  fileExists <- doesFileExist f
+  when fileExists
+   $ do putStrLn $ "Removing " ++ f
```
david-christiansen: This dollar should really be on the previous line to fit in with the rest of the codebase. Something like:
```
when fileExists $
  do putStrLn $ "Removing" ++ f
     ...
```

david-christiansen:
I also agree with Melvar that this has a potential race condition. However, in this case, the race condition seems to be that someone else might delete the file in the meantime, resulting in a harmless warning being shown that the file couldn't be found for deletion. Or am I missing something?

Melvar:
I just consider it bad style; feel free to accept it regardless.

Heather:
@david-christiansen maybe you're right - I can try to rewrite it with catching specific exception but then is there case for printing `"Removing " ++ f` or should it be printed in case of success?

david-christiansen:
I think it should print it in case of success, but in past tense: `"Removed " ++ f`.

Heather:
@david-christiansen @Melvar may you take a look now - this way "Removed..." prints either with success or error :/ should I change catch to try there and then check for Left Right?

Heather:
Done

Heather:
@david-christiansen @Melvar any problems with it?

david-christiansen:
Hi @Heather, sorry to be slow here! I've been very busy with wedding followup and other things.
Thanks for improving Idris!

PR #2824, opened 2015-11-23
parsing doubles from strings

eklavya:
Since we have a way to parse integers from strings, it would be nice to have a way to parse doubles from strings for complete number parsing coverage.


david-christiansen:
Hello!

Would you please update the commit message and the PR text to describe the reason for these changes? Also, please use capital letters at the beginning of commit messages.

Thanks!

PR #2825, opened 2015-11-25
Minimal Enum instance for Char

Garciat:


Garciat:
Perhaps this should be placed in `libs/prelude/Prelude/Chars.idr` ?

Melvar:
`Enum` is a problem all by itself. It wants conversion to and from `Nat`, but `Char`, unlike `Nat`, is finite. Properly we’d need separate handling for enumeration of finite vs min-bounded countable vs max-bounded countable vs unbounded countable types, but of course that’d be a lot more work so nobody is doing it.

edwinb:
We probably do need to rethink Enum. But in the absence of doing that, this is useful, thanks for the contribution!

PR #2827, opened 2015-11-26
Elab lookup function def

shlevy:
Enable looking up function definitions from elab scripts.

TODO:
- [x] Make FunDefn polymorphic so the lookup can return TT
- [x] Save type info for pattern bindings in idris_patvars
- [x] Fix Impossible clauses


shlevy:
@david-christiansen

david-christiansen:
Thanks! I'll take a peek soon. I've had more non-Idris things to do than expected the last while.

david-christiansen:
One thing that would make the PR easier to evaluate would be entries in the test suite that exercise it.

shlevy:
It's not done yet :smile: There's no real new feature to test

david-christiansen:
Ah, OK!

I think the existing Idris-level representation of cases will work just fine here.

david-christiansen:
So looking over the code so far, I think that it might be a good idea to make a little record type that holes the datatypes and pattern definitions together, and thread that through the elaborator, rather than having the two be separate. They're both used similarly and in similar places, so having that be one item that's appropriately named would probably keep the code more readable.

What do you think?

shlevy:
It actually doesn't look necessary any more to do the threading through, we have access to the Idris state where we need it. I'm still working on it, but I think I can finish this up without changing ProofState at all.

shlevy:
Going to try to dig into this myself, but running into a problem where I can't go from the `([Name], Term Term)` representation of a clause to the `(Term, Term)` representation because the types of `P` are erased. I'm guessing they're recoverable from context, though.

shlevy:
@david-christiansen I've pushed what I've got so far. Currently fails on functions with impossible clauses due to the "imp" var inserted by the elaborator and on functions without impossible clauses because the types of `P` terms are erased. Not sure if I'll have time in the near future to fix this, but I'll leave this open.

shlevy:
@david-christiansen It turned out the type info we wanted was already erased by the time `idris_patvars` is inserted into, so propagating it along wasn't helpful. I switched to the polymorphic approach, and what I just pushed works, but it gives TT with `Erased` types for P bindings and P vars. I'm not sure how to do anything useful with it, so I can't write a meaningful test.

My latest attempt was:

```idris
import Pruviloj.Core

forgotToHole : Raw -> Elab Raw
forgotToHole (RBind name b body) = do
  b' <- for b forgotToHole
  body' <- forgotToHole body
  pure $ RBind name b' body'
forgotToHole (RApp f arg) = do
  f' <- forgotToHole f
  arg' <- forgotToHole arg
  pure $ RApp f' arg'
forgotToHole (RConstant Forgot) = do
  n <- gensym "forgottenTy"
  pure $ RBind n (Hole RType) (Var n)
forgotToHole tm = pure tm

roundtrip : TTName -> TTName -> Elab ()
roundtrip old new = do
  DefineFun _ clauses <- lookupFunDefnExact old
  clauses' <- for clauses (\(MkFunClause lhs rhs) => do
    lhs' <- forget lhs >>= forgotToHole
    rhs' <- forget rhs >>= forgotToHole
    pure $ MkFunClause lhs' rhs')
  defineFunction (DefineFun new clauses')

plus' : Nat -> Nat -> Nat
%runElab (roundtrip `{plus} `{plus'})
```

Which fails with

```console
While running an elaboration script, the following error occurred:
Type mismatch between
        ([__])
and
        Nat
```

If I skip the `forgotToHole` stuff, I get:

```
While running an elaboration script, the following error occurred:
INTERNAL ERROR: [__] is not a Type
```

david-christiansen:
Sorry to be slow on feedback here - I've had non-Internet stuff keeping me from giving it a proper go. I'll get into it as soon as I can!

shlevy:
@david-christiansen Pushed a test that bypasses the issue above for now, removed the WIP in the title. Let me know if you think the Forgot issue should be resolved before merging this.

david-christiansen:
I need to play with it a bit to figure out exactly what the Forgot issue is. After all, the type annotation on a P shouldn't be surviving `forget`, which should be translating it to `Var` which has no annotation.

david-christiansen:
Ah, I see, the issue is that the pattern variable types are set to `Erased` rather than the actual variable type. I think we should fix this prior to a merge. I still think the right solution is to keep the information from earlier in elaboration - but it will likely require going back a couple more steps. Nevertheless, that will make it have a good interface.

david-christiansen:
And once again, thanks so much for doing this!

shlevy:
Do you know off hand where the erasure is happening? Won't this end up making IBC files and in-memory AST a lot bigger, if we stop erasing all types in all function definitions?

shlevy:
Also, FWIW if `defineFunction` used the `([TTName], Raw, Raw)` representation of clauses then I think this wouldn't be an issue, as there'd be no introduced pattern bindings.

david-christiansen:
I don't think it will have a major impact on space usage, but I can check before and after. After all, it's only a couple of extra binders.

That representation is not reasonable for `defineFunction` because it needs to type check the results, which requires types on the binders.

shlevy:
@david-christiansen OK, I think the last commit should do it!

david-christiansen:
I opened up my computer to do this, and lo and behold! you had done it already! I'll give it a test-through.

david-christiansen:
This is looking really nice!

The only issue I'm discovering so far is that impossible clauses in submitted definitions are required to be type-correct, rather than type-incorrect. I'll see what I can do.

shlevy:
@david-christiansen If you're looking at that, you're probably going to need to fix https://github.com/idris-lang/Idris-dev/blob/master/src/Idris/Elab/Term.hs#L1800 not to be an infinite loop :smile:

david-christiansen:
I've got a fix locally. I'm running the test suite an extra time, and then I'll push the merged version.

Thanks again!

david-christiansen:
OK, I merged it in with my updates for impossible cases. This is great stuff!

PR #2831, opened 2015-11-27
Add Data.List.Quantifiers

david-christiansen:
This module is modeled after Data.Vect.Quantifiers, presenting a
basically-compatible API. I believe that this supersedes #2773.


david-christiansen:
I just saw #2774. Sorry. I'll make that pass Travis instead.

PR #2832, opened 2015-11-27
Disable the DecEq deriving test, due to slowness

### LINKED ISSUE #2785: Meta002 test gets killed by travis-ci.org
As evidenced by this message, the meta002 test seems to get killed by travis-ci when running. (Parallel builds interleave output, but the "meta002 passed" is always printed after this killed message).

https://travis-ci.org/idris-lang/Idris-dev/jobs/86099125#L1091

After looking into this further, it seems to have originated in the builds for #2749 (the build linked, but appears in every test including meta002 since then), it doesn't appear in travis build 4111.

I have run the test on my own machine, trying to get out some figures. Travis will kill some processes, in certain cases: http://docs.travis-ci.com/user/common-build-problems/#My-build-script-is-killed-without-any-error and it seems this is triggered. I ran the following, and it showed me:

```
$ /usr/bin/time -lp idris -p pruviloj --check Deriving.idr
real        58.89
user        47.99
sys          5.25
4464189440  maximum resident set size
         0  average shared memory size
         0  average unshared data size
         0  average unshared stack size
   1629379  page reclaims
       407  page faults
         0  swaps
        31  block input operations
         3  block output operations
         0  messages sent
         0  messages received
      5785  signals received
       124  voluntary context switches
    143867  involuntary context switches
```

That figure for Max RSS is in bytes, so, about 4.4 GB of memory is used at peak by this script. Travis will kill a process if it goes above 3GB.

I can't get the other two files in the test to actually run, but maybe we need to do something like "set +e" or something in the ./run script (this will cause the whole script to error if any of it does, you can "set -e" to remove this again later in a script).

david-christiansen:
Fixes #2785.

### DIFF
--- a/test/meta002/Deriving.idr
+++ b/test/meta002/Deriving.idr
@@ -1,6 +1,8 @@
 ||| Test some deriving features
 module Deriving
 
+-- NB: test disabled due to excess memory consumption
+
 import Pruviloj
 import Pruviloj.Derive.DecEq
 
--- a/test/meta002/run
+++ b/test/meta002/run
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 idris $@ -p pruviloj --nocolour --check --consolewidth 70 Tacs.idr
 idris $@ -p pruviloj --nocolour --check --consolewidth 70 AgdaStyleReflection.idr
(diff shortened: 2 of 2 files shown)


david-christiansen:
Also RE #2769.

PR #2836, opened 2015-12-01
For discussion: Add function to forcefully tighten Fin bound

Kazark:
If tightening the bound would not work with the value, the value is
decremented. I called it `crimp` because I wanted to gave a name that
clearly distinguished it from the existing `strengthen` function.

I don't have a rule of three on this function. How do you determine what
functions are useful to add to the base library? Do you try to apply a
rule of three? I have an intuition that this is a useful variation of `strengthen`,
but then again, if you don't have a rule of three...

Honestly I'm submitting this pull request to learn more than anything.
I am still early in my understanding of Idris, but I'm already eager to
contribute. Don't feel bad about closing this pull request if you don't
like it. Any feedback/guidance about future contributions would also
be welcome.

### DIFF
--- a/libs/base/Data/Fin.idr
+++ b/libs/base/Data/Fin.idr
@@ -86,6 +86,15 @@ last : Fin (S n)
 last {n=Z} = FZ
 last {n=S _} = FS last
 
+||| Forcefully tighten the bound on a Fin.
+||| If tightening the bound would not work with the current value,
+||| decrement the value such that it is within the tightened bound.
+crimp : Fin (S (S n)) -> Fin (S n)
+crimp x =
+  case (strengthen x) of
+       Left _ => last
+       Right x' => x'
+
 total FSinjective : {f : Fin n} -> {f' : Fin n} -> (FS f = FS f') -> f = f'
 FSinjective Refl = Refl
 


david-christiansen:
Hello Kazark,

Thanks for the contribution! What do you mean by "rule of three"?

Is this a function that you've found yourself needing in a real program?

Thanks again!

Kazark:
Hi David, thanks for the response! Yeah, I should have given a little more context. I've started building a [text editor in Idris](https://github.com/PolyglotSymposium/void.idr) as a way to learn the language. So I don't know whether that qualifies as a "real" program, given that it is not much more than a breakable toy currently, but it wasn't just something I made up in isolation, if that's what you mean.

However, here's the use case, and I think there may be other analogous situations in other real programs: one of the ways I'm playing with the type safety is by making it impossible for a cursor to be outside the bounds of the text. So when you move the cursor from one line of text to a shorter line of text, you want to force the cursor to fit on the shorter lines.

By "rule of three" I'm referring to the idea that you should have three concrete examples before introducing an abstraction.

Kazark:
Another function I have, with a related concept, for when you move the cursor from one line of text to a shorter line of text, you want to force the cursor to fit on the shorter lines:

```idris
adjustColumnIndex : Nat -> Fin (S n)
adjustColumnIndex {n} prevColumn =
  case natToFin prevColumn (S n) of
       Nothing => last
       Just x => x
```

The concept is the same and so I wonder if I have not yet hit on the correct abstraction here.

david-christiansen:
This seems reasonable, at least semantically. I'm not convinced that `Fin k` is what you want, as opposed to a `Subset Nat (\n => LT n k)` which would take much less space at runtime.

In any case, why not call it `truncate` instead? And could you please stick it in `contrib` for now?

Kazark:
I don't know anything about this `Subset` construct that you bring up, so I better learn about that before I charge ahead with this. :) Also I'll keep growing my editor code until I feel like I have a firmer grasp on this.

Kazark:
Since I haven't gotten this, I'm going to close this until I do so it's not just sitting around hanging open.

Issue #2835, opened 2015-12-01
Cannot declare instance overlapping with private instance


Melvar:
An instance declared in a module with `%access private` is not usable from outside the module, yielding `Can't resolve type class`. However, it is not possible to declare an instance with the same head; that yields an `Overlapping instance` error. The result is that if one library declares a private instance, it is impossible to declare a similar one in a consumer of the library.

This problem bit me just now because Pruviloj declares `instance Show FunClause` privately in `Pruviloj/Derive/Eliminators.idr`, meaning that `:doc Show` lists `Show FunClause`, but the instance can’t actually be found and I can’t output my clauses for debugging purposes.

david-christiansen:
I think that the correct solution to this is to have unnamed instances always be public. That respects the global nature of instances. What do you think?

enolan:
That's a weird extra thing to have to learn.

It's a reasonable expectation that one can set any declaration to private, without having to change the declaration itself.

You say that instances have a "global nature", but that's only the case if you don't allow their import and export to be controlled.

Melvar:
By “global nature”, do you mean that the Idris internals keep unnamed instances in a single list for purposes of instance search? Or the way unnamed instances are internally named as something not in the namespace they’re actually declared in?

ahmadsalim:
What was the conclusion?

Melvar:
To summarize: Currently, it is possible to apply visibility modifiers to unnamed implementations. This is taken into account by instance search, but not by the overlap check or `:doc` listing implementations.

I see these options (feel free to propose others):

1.  Continue allowing visibility modifiers to affect unnamed implementations, and change the overlap check and `:doc` to take this into account. (I favor this one because I find it most consistent.)

2.  Make unnamed implementations always `public export`, forcing users to display their method implementations. (Making them always `export` would make for example `(+)` unusable in the type of `Vect.(++)` if I am not mistaken, so I am not listing that as an option.)

3.  Allow applying both `export` and `public export` to unnamed implementations, but not `private`.

For options 2 and 3, a warning or error, respectively, would be required if an implementation is defined with the modifier `private`, without any access modifier under the effects of `%access private`, or without any access modifier with no `%access` directive anywhere preceding it (as `private` is the default IIRC). The same, mutatis mutandis, applies to option 2 and `export`.

ahmadsalim:
@Melvar Thanks.

PR #2838, opened 2015-12-02
Replace private Show FunClause with public one

Melvar:
The existence of the private one prevented defining one in any consumer of Pruviloj as noted in #2835, but it should presumably be in `Language.Reflection.Utils` like the other `Show` instances for reflection data.


david-christiansen:
Agreed. Thanks!

Issue #2840, opened 2015-12-02
doOp not implemented ((LASHR ITBig,[L 2,L 3]))


norm2782:
I'm getting the following compile error while using Integer types:

`idris: doOp not implemented ((LASHR ITBig,[L 2,L 3]))`

I'm still working on a minimal example, but I figured I'd post it here already :)

norm2782:
Perhaps not minimal, but small nonetheless:

```
a : Integer
a = let d = 123 in (d << 8) .|. (d >> 8)

main : IO ()
main = putStrLn (show a)
```

`idris: doOp not implemented ((LSHL ITBig,[L 0,L 1]))`

PR #2841, opened 2015-12-03
Add support for logical and, or, shl, shr for big ints

norm2782:
This solves #2840 and related issues for big integers.
Please note that arithmetic and logical right shifts (`bigAShiftRight` and `bigLShiftRight`) are currently implemented identically in `idris_gmp.c`; both use the `mpz_fdiv_q_2exp` function. This is not perfect, so please have a look at that :)


edwinb:
Thanks for this! Sorry for the slow response.

PR #2845, opened 2015-12-05
updated stack yaml

dysinger:
This P/R updates the stack.yaml so you can see how to build with flags & a stable resolver.  Uncomment flags & extra-deps if you want the optional dependencies.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,6 +1,13 @@
-flags: {}
+resolver: lts-3.16
 packages:
 - '.'
+flags:
+  # idris:
+  #   FFI: true
+  #   GMP: True
+  #   curses: True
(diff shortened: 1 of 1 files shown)


jfdm:
@dysinger Cheers for the PR. 

In case you were not aware PR #2527 was in line (many moons ago) to add updated stack support, along side adding a separate stack test to the travis build. However, I could not find the time to sort out getting travis to work as intended; the point was to use Travis to detect when the version of LTS/Stackage failed to build Idris.

Now I am going to completely drop PR #2527 in favour of this. Tis better to get Idris building on an LTS. If you know how to get Travis to have in its build a means to *only* build Idris that would be awesome.

Cheers.

PR #2849, opened 2015-12-09
Bug fix for infinite loop while executing a callback in the Javascript runtime

aaronc:
The `NULL` `BC` operation, which is supposed to clear a register (according to the docs) was translated as `delete` in the Javascript backend. This worked fine for clearing registers from the stack, but did not work well while trying to clear `i$RT`. This cause an infinite loop in `i$ffiWrap` which was checking the return value should have been cleared but continued to hold its old value (essentially ignoring the delete operation). Callbacks work fine now in my tests.

### DIFF
--- a/src/IRTS/CodegenJavaScript.hs
+++ b/src/IRTS/CodegenJavaScript.hs
@@ -604,7 +604,7 @@ jsBASETOP _ 0 = JSAssign jsSTACKBASE jsSTACKTOP
 jsBASETOP _ n = JSAssign jsSTACKBASE (JSBinOp "+" jsSTACKTOP (JSNum (JSInt n)))
 
 jsNULL :: CompileInfo -> Reg -> JS
-jsNULL _ r = JSDelete (translateReg r)
+jsNULL _ r = JSClear (translateReg r)
 
 jsERROR :: CompileInfo -> String -> JS
 jsERROR _ = JSError
@@ -1277,4 +1277,3 @@ toFType (FApp c [_,ity])
 toFType (FApp c [_,fty])
     | c == sUN "JS_FnT" = toFnType fty
 toFType t = error (show t ++ " not yet defined in toFType")
-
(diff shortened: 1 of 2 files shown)


edwinb:
Thanks for this. The JS back end really needs some attention - there isn't anyone actively maintaining it at the minute (possibly not even using it). I'm wondering if it really ought to be factored out like the other back ends, and have its own test suite so that this kind of thing will be caught.

aaronc:
So it looks quite likely that I'll be actively using both the Java and Javascript backends and making small patches as I go.

My concern with factoring the JS backend out into its own repo is that this may actually end up in it getting less test coverage and end up being viewed as a second-class citizen. It doesn't look like the Java backend is getting much test coverage at all but that backend is actually in quite good shape (I would say comparable to the JS backend). I have to say that when I first looked at Idris, I thought that because the Java backend was in its own repo that this meant that it didn't really work...

One of my biggest concerns with these backends initially was performance (https://groups.google.com/forum/#!topic/idris-lang/koqBAPlDWDo). The JS backend introduces significant overhead to do tail call optimization - I would say on the order of at least 10x from my benchmarks. After I fixed this bug, however, and was actually able to do some tests doing "real" things, I found that this overhead doesn't translate into unacceptable performance at runtime. I have a React animation benchmark that I've done in Clojurescript (animating dozens to hundreds of points simultaneously). The reduction in frame rate doing this same animation with Idris and React is negligible compared to doing it with Clojurescript and React. I also did some performance benchmarks for the Java backend (comparing to this time to my go to language Clojure) and in all cases Idris was comparable if not often faster (sometimes significantly) than Clojure.

So at this point, I would say that for someone trying to use these backends correctness and feature completeness are the highest priority and that performance is "good enough" for now. Also, overall the current state of both backends is in my assessment quite good (better than I was expecting). I'm thinking that if I continue using them in the long term, it would be nice to do an additonal IR step that I'll call IR<sub>TCO</sub> (for Tail Call Optimization) that uses loops, inlining and gotos to optimize tail calls as opposing to trampolining or a VM to do tail call optimization. But I don't think this is a super high priority compared to other things...

Anyway, happy to help with maintenance of these backends and discuss the pros/cons of keeping them in separate repos as well as how to keep the test suites up to date.

PR #2852, opened 2015-12-10
Updates to installation docs for devs.

jfdm:
Migrated installation instructions from the README to a separate
file, and included documentation for using stack.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -2,65 +2,70 @@
 
 Copyright (C) 2015, The Idris Community
 
-This file provides information over obtaining and installing Idris.
-
 ## Installing Idris from Hackage
 
-This repository represents the latest development version of the language,
-and may contain bugs that are being actively worked on.
-For those who wish to use a more stable version of Idris please consider
+This repository represents the latest development version of the
+language, and may contain bugs that are being actively worked on.  For
+those who wish to use a more stable version of Idris please consider
 installing the latest version that has been released on Hackage.
-Installation instructions for various platforms can be [found on the Idris Wiki](https://github.com/idris-lang/Idris-dev/wiki/Installation-Instructions).
+Installation instructions for various platforms can be
+[found on the Idris Wiki](https://github.com/idris-lang/Idris-dev/wiki/Installation-Instructions).
 
 ## Installing Development Versions
 
-If you like to work against the latest development version, please consider
-using Cabal Sandboxes to minimise disruption to your local Haskell setup.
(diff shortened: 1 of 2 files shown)


INSTALL.md
```diff
@@ -0,0 +1,87 @@
+# Idris Installation Guide
+
+Copyright (C) 2015, The Idris Community
+
+This file provides information over obtaining and installing Idris.
```
Melvar: Did you mean “information about”?

INSTALL.md
```diff
@@ -0,0 +1,87 @@
...
+
+Idris has an optional buildtime dependency on the C library
+`libffi`. If you would like to use the features that it enables, make
+sure that it is compiled for the same architecture as your Haskell
+compiler (e.g. 64 bit libraries for 64 bit ghc). By default, Idris
+builds without it. To build with it, pass the flag `-f FFI`.
+
+A secondary optional buildtime dependency is on `libGMP`, this allows
+better support for numeric operations. As with `libFFI` Idris builds
+with out it, to enable `GMP` support use the flag `-f GMP`
```
Melvar: The use of “buildtime dependency” here to me suggests the buildtime of idris itself, but if I understand correctly the usage patterns of these two libraries are actually different; libffi is linked into idris itself and called when running idris i.e. when compiling programs with idris but also when using the REPL, while libGMP is linked into the resulting programs and called when running them.

INSTALL.md
```diff
@@ -0,0 +1,87 @@
...
+
+If you are only compiling for installing the most current version, you can
+omit the CI flag, but please make sure you use it if you want to contribute.
+
+## Experimental Support for Building with Stack
+
+[Stack](https://github.com/commercialhaskell/stack) is a new
+cross-platform program for developing Haskell projects, that enhances
+the functionality provided by Cabal. There is experimental support for
+building Idris from source using with stack, that uses the LTS 3.6 resolver.
```
enolan: Should be 3.16, not 3.6.

INSTALL.md
LeifW: Yeah, these are optional deps of the produced idris compiler. I think GMP is always a "buildtime" dep, as GHC itself depends on it (the Idris binary links to it on my system; compiled Idris programs link to it since I built idris with the `-fGMP` flag)

enolan: It's not terribly important, but you can build GHC with GMP or with a
native Haskell integer library. I think the reason was licensing concerns?
So you could conceivably not have GMP when you compiled Idris.

PR #2853, opened 2015-12-10
Gather timing data for tests.

jfdm:
To aid in the managment of the test suit, the perl script now records
the duration of each test for compilation and execution. The resulting
time is then displayed to stdout in the form:

    Duration of <test name> was <time>

When running tests locally, test output can be grep'd and graphs
plotted. A good improvement will be the automatic storing of the
results in a dat file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,9 @@ test_all:
 	$(MAKE) test_llvm
 	$(MAKE) test_java
 
+test_timed:
+	$(MAKE) -C test IDRIS=../dist/build/idris time
+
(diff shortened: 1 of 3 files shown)


edwinb:
Could this be made optional? Or perhaps dumped at the end? I can see it being useful if some test suddenly starts taking longer than usual to run, but otherwise it seems like clutter...

jfdm:
I do not see why not. I will look into making the collection of timings optional, and if so dumped at the end.

PR #2856, opened 2015-12-11
Category based logging for Idris.

jfdm:
Currently logging is based on horizontally defined levels across the
entire language infrastructure. This commit introduces vertical splits.

Logging statements record both the level the message is to be recorded
at and the categories it belongs too. No categories is an alias for
belonging to all categories.

The State record stores both the current logging level, and also the
categories to show.

Logging is decided if a given logging statement is at the right level
and if the categories specified are listed. By default all categories
are allowed, as levels will take care of logging statement visibility.

**TODO**

+ [x] Update IDE Protocol
+ [x] Finalise Logging Categories
+ [x] Implement logging helpers based on categories.
+ [x] Categorise logging statements.
+ [x] Update documentation.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,14 @@
+New in 0.9.21:
+==============
+
+* Idris' logging infrastructure has been categorised. Command line and repl
+  are available. For command line the option `--logging-categories CATS`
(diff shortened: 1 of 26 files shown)


jfdm:
The remaining TODO items have been refactored out to issues #2866 and #2865.

PR #2857, opened 2015-12-12
Add :watch files for changes command to REPL.

LeifW:
While reading through the upcoming Idris book, I came across the following paragraph:
> Note that the REPL does not reload files automatically, as it runs independently of the
> interactive editing in Atom. Therefore, you will need to reload explicitly using the :r command.

That had me thinking, "well, we can automate that".
fsnotify is a cross-platform lib - uses whatever is appropriate for each platform (inotify on Linux, etc).
Was thinking along the lines of SBT's "~compile" command it has in its REPL. Could add support for running other commands on file changes in the future, but I can't think of any others you'd want to do (test?).

If people think this is a good idea, could add this as a configure flag, so as to not require those two extra Haskell libs, and possibly complicate the build for some people.

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -979,6 +979,8 @@ Library
                 , vector-binary-instances < 0.3
                 , zip-archive > 0.2.3.5 && < 0.2.4
                 , safe
+                , fsnotify < 2.2
+                , async < 2.1
   -- zlib >= 0.6.1 is broken with GHC < 7.10.3
   if impl(ghc < 7.10.3)
      build-depends: zlib < 0.6.1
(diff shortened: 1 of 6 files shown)


edwinb:
This sounds a good idea to me. Adding dependencies always scares me slightly though, so can anyone think of a reason why this might not be a good idea (working on Windows, perhaps)? I'll merge if nobody shouts...

LeifW:
Could also make this feature (and its deps) optional w/ a cabal flag, I suppose.

LeifW:
The REPL is using InputT, of which this is running inside, and yet here I'm calling getLine :: IO String. Hope that won't cause any complications. Seemed to work fine in my use.
This call to getLine on a separate thread is just to provide the user an opportunity to break out of the watch file loop by pressing enter.
https://github.com/idris-lang/Idris-dev/pull/2857/files#diff-6915954b3868f5d3475fb228597aa890R700

melted:
This should work on windows, fsnotify is portable.

PR #2861, opened 2015-12-15
Be consistent when giving universe inconsistency errors

ozgurakgun:
This must have been forgotten when improving the error message before.

The error message can originate from two places. This commit makes the error message consistent between the two.


david-christiansen:
Thanks!

PR #2862, opened 2015-12-16
fix typo in tutorial : 'desugars' -> 'displays ... as ...'

xieyuheng:
I am not a native english speaker, feel free to change my sentences.

### DIFF
--- a/docs/tutorial/typesfuns.rst
+++ b/docs/tutorial/typesfuns.rst
@@ -149,10 +149,13 @@ Idris prompt:
 
 .. note::
 
-    Idris automatically desugars the ``Nat`` representation into a
-    more human readable format. The result of ``plus (S (S Z)) (S (S Z))``
-    is actually ``(S (S (S (S Z))))`` which is the Integer 4. This can be
-    checked at the Idris prompt:
+   When displaying an element of ``Nat``,
+   such as ``(S (S (S (S Z))))``, for a user.
+   Idris displays it as 4 -- a more human readable format.
+   The result of ``plus (S (S Z)) (S (S Z))``
+   is actually ``(S (S (S (S Z))))``
+   which is the natural number 4.
+   This can be checked at the Idris prompt:
 
 ::
 
--- a/docs/tutorial/typesfuns.rst
+++ b/docs/tutorial/typesfuns.rst
@@ -149,12 +149,11 @@ Idris prompt:
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


Kazark:
Hi @xieyuheng , I'm not an Idris maintainer, but I will give my two-cents. "Desugar" is a valid English technical word. Are you concerned about the readability of the tutorial? I don't think this is actually an error, and its certainly not a typo.

david-christiansen:
@Kazark Based on an IRC discussion, I believe that the concern was actually that "desugars" has the opposite of the intended meaning. In this case, the inductive representation of the Nat (`S (S Z)`) is being converted to the nice surface syntax (`2`), which is sugaring, not desugaring.

In any case, "displays as" means precisely the same thing as "resugars" or "sugars", but it's simpler and easier to understand.

docs/tutorial/typesfuns.rst
```diff
@@ -149,10 +149,13 @@ Idris prompt:
 
 .. note::
 
-    Idris automatically desugars the ``Nat`` representation into a
-    more human readable format. The result of ``plus (S (S Z)) (S (S Z))``
-    is actually ``(S (S (S (S Z))))`` which is the Integer 4. This can be
-    checked at the Idris prompt:
+   When displaying an element of ``Nat``,
+   such as ``(S (S (S (S Z))))``, for a user.
```
david-christiansen: I'd remove "for a user." here, as Idris can't display it to a non-user.

david-christiansen: Also, make this one sentence rather than having a sentence break here.

docs/tutorial/typesfuns.rst
```diff
@@ -149,10 +149,13 @@ Idris prompt:
...
+   Idris displays it as 4 -- a more human readable format.
```
david-christiansen: While you're at it, how about "Idris displays it as ``` ``4`` ```, which is easier for humans to read."  I think this phrasing is easier for humans to read (har har), and it's also referring to the example rather than to the format that lies implicit in the example.  Note that this comment is me being exceedingly picky :-)

xieyuheng:
thank you christiansen. 

if this PR is ever merged, 
I wish the maintainer can change my changes according to christiansen's advises.

david-christiansen:
@xieyuheng You can update the PR yourself, and then I'll push the "merge" button :-)

Issue #1837, opened 2015-01-08
Ordering of Effects when calling effectual programs.


jfdm:
The effects tutorial mentions that the specification order for a list of effects does not matter. When an effectual function is treated in isolation this is true. However, ordering does matter between effectful functions. For instance, given the following program:

```idris
module EffectfulExpressionCalculator

import Effects
import Effect.State
import Effect.Random
import Effect.Exception

data Expr = Val Integer | Var String | Add Expr Expr | Random Integer

Env : Type
Env = List (String, Integer)

eval : Expr -> { [EXCEPTION String, RND, STATE Env]} Eff Integer
eval (Val x)   = pure x
eval (Add l r) = pure $ !(eval l) + !(eval r)
eval (Random upper) =  rndInt 0 upper
eval (Var x)   = case lookup x (!get) of
  Nothing  => raise $ "No such variable " ++ x
  Just val => pure val

runEval : Env -> Expr -> Maybe Integer
runEval args expr = run (eval' expr)
  where
    eval' : Expr -> {[EXCEPTION String, STATE Env, RND]} Eff Integer
    eval' e = do
      put args
      eval e
```

The program does not compile as the list of effects used is 'out-of-order'.

Can `Effects` be fixed such that ordering does not matter across an effectual program?

edwinb:
On 8 Jan 2015, at 14:02, Jan de Muijnck-Hughes <notifications@github.com> wrote:

> Can Effects be fixed such that ordering does not matter across an effectual program?

Not easily - it would need the ‘SubList’ proof to be cleverer and be able to detect changes in ordering as well as merely missing elements, and we’d need some kind of automated proof search to spot this (it might not work with proof search, depending on how it was done).

It would be nice, at least, if we could add some kind of explicit effect permutation operator, otherwise there’s difficulty in composing effectful programs.

Now that we’ve been using Effects for a few different things, especially with resource tracking, I’m getting a better idea of what the annoyances are and what the requirements should have been if only I’d known what I was doing when I started ;). I’m playing around with a few ideas separately so we’ll see what happens…

jfdm:
Oh yeah, I should have lead with: I knew somewhat about the 'restrictions' but wanted to make public a note about them, as I don't think we had done so yet. This issue was submitted on the back of effects library complaining about my error-ful code not working, and me not realising there were errors in my code.

aaronc:
[...]

PR #2864, opened 2015-12-19
Order-independent implementation of SubList for Effects

aaronc:
This is my first attempt at solving #1837 - Ordering of Effects when calling effectual programs

This implementation has what seems to me to be unnecessary complexity due to needing to figure out how to get the Handler eff m type class resolution to be inferred through-out the Env re-ordering code - envElem and dropEnv is a good example of this. I think that possibly someone else could make this a bit cleaner, but it works in my tests.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,6 @@
+FROM haskell:7
+RUN cabal update
+ADD . Idris-dev
+WORKDIR Idris-dev
+RUN make
+WORKDIR ..
--- a/libs/effects/Effects.idr
+++ b/libs/effects/Effects.idr
@@ -97,31 +97,32 @@ namespace EffElem
     Here : EffElem x a (MkEff a x :: xs)
     There : EffElem x a xs -> EffElem x a (y :: xs)
     
-envElem : SubElem x xs -> Env m xs -> Env m [x]
+total envElem : SubElem x xs -> Env m xs -> Env m [x]
 envElem Z (x :: xs) = [x]
 envElem (S k) (x :: xs) = envElem k xs
 
 ||| make an environment corresponding to a sub-list
-dropEnv : Env m ys -> SubList xs ys -> Env m xs
+%assert_total dropEnv : Env m ys -> SubList xs ys -> Env m xs
 dropEnv [] SubNil = []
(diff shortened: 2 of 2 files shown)


aaronc:
I apologize - I typed in the wrong issue ref number for some reason (brain fart...). This should ref issue #1837

jfdm:
Cheers for looking at this, is it possible for you to add your tests to the test suite demonstrating the new functionality.

aaronc:
So, I have some tests that I'm almost ready to add. I ran into another issue however. `dropEnv` as I've defined it isn't total. There is a missing case clause for `[] (InList i is)`. Of course this case can't exist because the a non-empty list can't be a sub-list of an empty list, but I haven't figured out how to tell Idris this. To get my code working, I've had to add a `%assert_total` directive for `dropEnv`. I'd like to tell Idris this case is `impossible` but it won't accept that as is. Any thoughts on how to solve this?

edwinb:
Thanks for this!

I'd like to play around before I merge into master, though, since I'm about to do a release and I'd like to make sure it works on the effects code I have which really exercises this. In the mean time, if you get around to fixing the conflicts before I do that would be most helpful.

edwinb:
As far as I can tell, this works very well, thanks! Sorry for taking so long to sort out the merge conflicts, but it's done now and I've updated it for the latest export rules and so on.

aaronc:
Oh looks like you merged in the Dockerfile I created too. Was that intentional? This Dockerfile is useful and could be auto-built by dockerhub so that people can quickly pull a docker image for building idris code... That may actually be the quickest way to get an Idris repl up an running on a new machine now that I think of it. If you have docker installed you can try it like this: `docker run -it aaroncr/idris-dev idris` - instant idris repl, no installation required.

Also, melvar had showed me how to avoid the `%assert_total` in `dropEnv` but I didn't get around to pushing the fix yet. Also, he pointed out that `EffElem` can be safely replaced by `Elem` from `Data.List`.

edwinb:
I fixed the dropEnv in a different patch. Didn't notice the Dockerfile
though...

I did notice about Elem too, but that can be fixed easily enough.

PR #2869, opened 2015-12-19
Added flag to display available logging categories.

jfdm:
This commit also removes magic constants used to processing the
logging categories.

### DIFF
--- a/src/Idris/AbsSyntaxTree.hs
+++ b/src/Idris/AbsSyntaxTree.hs
@@ -465,6 +465,14 @@ data LogCat = IParse
             | IIBC
             deriving (Show, Eq, Ord)
 
+strLogCat :: LogCat -> String
+strLogCat IParse    = "parser"
+strLogCat IElab     = "elab"
+strLogCat ICodeGen  = "codegen"
+strLogCat IErasure  = "erasure"
+strLogCat ICoverage = "coverage"
+strLogCat IIBC      = "ibc"
+
 codegenCats :: [LogCat]
 codegenCats =  [ICodeGen]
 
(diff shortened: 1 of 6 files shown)


src/Idris/AbsSyntaxTree.hs
```diff
@@ -474,6 +482,16 @@ parserCats = [IParse]
...
 
+loggingCatsStr :: String
+loggingCatsStr = unlines
+    [ (strLogCat IParse)
+    , (strLogCat IElab)
+    , (strLogCat ICodeGen)
+    , (strLogCat IErasure)
+    , (strLogCat ICoverage)
+    , (strLogCat IIBC)
+    ]
```
Melvar: This just screams for `map strLogCat`.

src/Idris/REPLParser.hs
```diff
@@ -407,12 +407,12 @@ cmd_cats name = do
...
-           <|> try (P.symbol "codegen"  >> return codegenCats)
-           <|> try (P.symbol "coverage" >> return [ICoverage])
-           <|> try (P.symbol "ibc"      >> return [IIBC])
-           <|> try (P.symbol "erasure"  >> return [IErasure])
+    pLogCats = try (P.symbol (strLogCat IParse)    >> return parserCats)
+           <|> try (P.symbol (strLogCat IElab)     >> return elabCats)
+           <|> try (P.symbol (strLogCat ICodeGen)  >> return codegenCats)
+           <|> try (P.symbol (strLogCat IErasure)  >> return [ICoverage])
+           <|> try (P.symbol (strLogCat ICoverage) >> return [IIBC])
+           <|> try (P.symbol (strLogCat IIBC)      >> return [IErasure])
```
Melvar: These seem not to match.

src/Idris/AbsSyntaxTree.hs
jfdm: A more constructive comment  might make me more inclined to change it to map. However, non-constructive comments won't.

shlevy: What was unconstructive about @Melvar's comment? I can see it being considered a bit abrasive/rude (though I'm not sure I'd see it that way myself), but it's very clear on what it suggests should be done to fix the problem identified. And it seems a bit odd to decide not to make a very quick improvement to your code because the person who suggested it did so rudely.

Issue #1879, opened 2015-02-01
Access modifier rule seems silly


puffnfresh:
If we write the following:

```idris
module A

a : Nat
b : String
```

We have `a` and `b` exported and available to use. If we decide that `b` should be private, we would write:

```idris
module A

a : Nat

private
b : String
```

Now we have *nothing* exported. WAT

The default switches from public to private as soon as you include an access modifier somewhere. We should probably change that so that public (or even private) is *always* the default.

mbj:
I agree that the default should not switch implicitly, as a new Idris user I was a bit confused, through this behavior is well documented. I personally would like to see private as the default since it results in a very explicit definition of the public interface.

With that private default reading someone else code would result in not having to scan for modifiers that switch the default anymore. Reducing mental overhead on reviews.

david-christiansen:
I would be in favor of not changing the default to `private` as soon as one thing is `public`. I understand the thought process behind why it is the way it is, but it seems to be a frequent source of surprise for both new and experience users.

edwinb:
I've flagged this "Important" because it means I'll get around to fixing it sooner rather than later that way. Unless someone else gets there first.. (If you do, please remember to update the tutorial!)

edwinb:
I've finally got around to doing this.

PR #2870, opened 2015-12-19
Make Data.String function available.

imuli:
Working around bug #1879. (everything was private)

### DIFF
--- a/libs/base/Data/String.idr
+++ b/libs/base/Data/String.idr
@@ -1,4 +1,5 @@
 module Data.String
+%access public
 
 private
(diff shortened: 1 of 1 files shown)


imuli:
libs/pruviloj/Pruviloj/Derive/{DeqEq,Eliminators}.idr are also all-private, but I'm not sure what is up with them.

melted:
Thanks for the PR!

Melvar:
@imuli Those two are not all-private, they each export one thing (as abstract) and explicitly hide everything else with `%access private`.

Issue #2819, opened 2015-11-16
All the benchmarks are bitrotted


enolan:
None of them compile. We should really make them part of the test suite. Here's the output when trying to build them:

```
enolan at Behemoth in ~/mystuff/code/Idris-dev/benchmarks [float-double*]
$ ./build.pl
Building trivial / sortvec
Removing sortvec.ibc
WARNING: Cannot remove file sortvec.ibc, Error msg:sortvec.ibc: removeLink: does not exist (No such file or directory)
Removing /home/enolan/mystuff/code/Idris-dev/benchmarks/trivial/sortvec
WARNING: Cannot remove file /home/enolan/mystuff/code/Idris-dev/benchmarks/trivial/sortvec, Error msg:/home/enolan/mystuff/code/Idris-dev/benchmarks/trivial/sortvec: removeLink: does not exist (No such file or directory)
Type checking ./sortvec.idr
sortvec.idr:7:8:When checking type of Main.insert:
No such variable Vect
Building quasigroups / qgsolve
Removing Solver.ibc
WARNING: Cannot remove file Solver.ibc, Error msg:Solver.ibc: removeLink: does not exist (No such file or directory)
Removing Parser.ibc
WARNING: Cannot remove file Parser.ibc, Error msg:Parser.ibc: removeLink: does not exist (No such file or directory)
Removing Main.ibc
WARNING: Cannot remove file Main.ibc, Error msg:Main.ibc: removeLink: does not exist (No such file or directory)
Removing /home/enolan/mystuff/code/Idris-dev/benchmarks/quasigroups/qgsolve
WARNING: Cannot remove file /home/enolan/mystuff/code/Idris-dev/benchmarks/quasigroups/qgsolve, Error msg:/home/enolan/mystuff/code/Idris-dev/benchmarks/quasigroups/qgsolve: removeLink: does not exist (No such file or directory)
Type checking ./Solver.idr
Solver.idr:10:6:
When checking right hand side of Cell with expected type
        Type

When checking an application of type constructor Prelude.Maybe.Maybe:
        No such variable Fin
Solver.idr:13:11:When checking type of Solver.MkBoard:
No such variable Vect
Solver.idr:15:12:When checking type of Solver.emptyBoard:
No such variable Board
Building fasta / fasta
Removing fasta.ibc
WARNING: Cannot remove file fasta.ibc, Error msg:fasta.ibc: removeLink: does not exist (No such file or directory)
[...]

PR #2883, opened 2016-01-03
2819 fix benchmarks

paulkoerbitz:
I went over the benchmarks and fixed the compile time problems mentioned in #2819. Benchmarks now run fine. Still missing is running the benchmarks in travis after the build has been successful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -107,3 +107,4 @@ script:
       echo "make -j2 $test";
       make -j2 $test;
     done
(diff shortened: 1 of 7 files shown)


jfdm:
@paulkoerbitz Thanks for this, it is always nice to see support code fixed. I am going to merge this now, travis can come later. One idea for improving the benchmark suite further could be to use Make, as the test suite uses it. This means that the benchmarks could be called from the master makefile. This might also help with travis integration.

PR #2885, opened 2016-01-06
The old code doesn't work

DreamLinuxer:
The type of `SizedClass.set_students` is `Vect size Person -> SizedClass size -> SizedClass size`, so this function must use the constructor `SizedClassInfo`.


edwinb:
Hi, sorry for the slow response to this. I'll merge this for the sake of the tutorial not being wrong, but I suspect the type of set_students is wrong here and so we need to have a look at how records work - I'm sure this code used to work because otherwise it would never have been in the tutorial!

DreamLinuxer:
Yeah, the behavior of record is changed, it would be nice if the old code works.

PR #2892, opened 2016-01-14
Fix typo in nested record update.

thisfred:
Complete Idris newbie here, but I *believe* this should be `val` rather than `x`


edwinb:
You're right, thank you!

PR #2895, opened 2016-01-15
Publish functions needed to compute the index of Bits

ziman:
These functions are necessary for type signatures to work properly because the constructor of `Bits` refers to them, which breaks my `UTF8` module.

Also, I don't think there's reason to hide them — they define part of the interface.


Melvar:
*Huh*. I was entirely sure that `Bits` was declared `abstract` (making the data constructor private), but I can’t find that to have ever been the case.

PR #2896, opened 2016-01-18
Greatly expanded file modes

sheganinans:
I threw in the binary versions of all the modes in as well for completeness, I'm not sure how well they'll play with String and can be removed if needed. But I definitely need Append at the least and would not like to have to depend on a private fork.


sheganinans:
I threw in the binary versions of all the modes in as well for completeness, I'm not sure how well they'll play with String. But I definitely need Append at the least.

Melvar:
My opinion on this: The binary versions should certainly not be there since we don’t even have a type for binary data; nor should the exclusive versions since not everyone is set up for C11. If and when the latter are to be enabled, they should probably not be mashed into the file mode.

For reference, Haskell’s `System.IO` only provides `ReadMode`, `WriteMode`, `AppendMode`, and `ReadWriteMode`. Binary files are opened with a different function.

sheganinans:
Yes, I'm perfectly fine dropping the binary versions until there is a proper Binary data. You are right the exclusive versions could go in a separate function and data structure, for people who have C11. Here's the updated version. I think that having AppendPlus and WritePlus there for completeness. I'm fine with a little bike shedding.

edwinb:
I agree that it's best without the binary versions for now. I will put some thought into having a proper Binary primitive for the C backend though - probably done as an external primitive like Ptr is rather than truly primitive (which Buffer was before we took it out due to not being maintained...) - since it'll be useful for all kinds of things.

Melvar:
I don’t like those `Plus` names much. Looking at the `fopen` docs for the actual semantics, I would propose to clarify the names as follows:
```
r  = Read
r+ = ReadWrite
w  = WriteTruncate
w+ = ReadWriteTruncate
a  = Append
a+ = ReadAppend
```
Indeed I found it surprising that opening a file for writing would truncate it without this being obvious from the mode.

sheganinans:
I like the names, glad to change them. But also I just want to mention that renaming Write to WriteTruncate (which is something I agree we should do) would break existing code. Thoughts?

sheganinans:
3 ways I see this working.

1) The right way (tm)?: Go through and switch all the existing code to this version, breaking upstream code.
2) Add Write back in, while also keeping WriteTruncate, and only document WriteTruncate and encourage it's use, eventually deprecate Write to get to choice 1.
3) Rename WriteTruncate back to Write.

How I see it:
Choice 1 would probably be the best way in the long run.
Two sounds acceptable in the short term, as a way to get us to choice 1.
I think that choice 3 is the easiest, but also least desirable.

I think going route 2 with our eyes set on 1 is the right choice here. Thoughts?

david-christiansen:
It's rare that anyone pattern-matches these constructors - mostly they're provided by a user, and then interpreted by the underlying library. So I think it should be enough to add:
````
%deprecated
Write : Mode
Write = WriteTruncate
````
for a version or two.

**Comment by Melvar** [comment]
I agree with @david-christiansen there.

**Comment by melted** [comment]
@david-christiansen's snippet is wrong, it's `%deprecate` and not `%deprecated`. Which is why Travis is unhappy.

**Comment by david-christiansen** [comment]
That's what I get for quickly dashing off a response without actually trying it! Thanks @melted :-)

**Comment by sheganinans** [comment]
@melted Thanks!

**Comment by sheganinans** [comment]
Hmm, that still doesn't work.

    ./Prelude/File.idr:94:7: error: expected: declaration,
    end of input, string
    Write : Mode 
           ^

**Comment by melted** [comment]
That's likely because @david-christiansen meant to write `Type` instead of `Mode`. :-)

**Comment by sheganinans** [comment]
@melted Nope, it's supposed to be Mode. Just that %deprecate goes under the definition and it contains it's Name and a String.

**Comment by Melvar** [comment]
Now it occurs to me that the `ModeX` type is superfluous, since you’re expressing you want exclusion by calling the other function.

**Comment by sheganinans** [comment]
@Melvar You are right, I reflected the code to match that.

**Comment by sheganinans** [comment]
Updates?

**Comment by sheganinans** [comment]
Apologize for this horrible new commit. But I had to add it to get around a merge conflict. Here's what I did:

```shell
git remote add upstream https://github.com/idris-lang/Idris-dev.git
git pull upstream master
git merge
get commit -am "file modes extended"
```

I'm not quite sure how to squash this into a single nice commit of just my changes and none of the other commits..

sheganinans:
If it's preferred I can close this PR and make a new one based on the latest revision.

melted:
You can change the set of commits by force-pushing to the branch. Then the PR is automatically updated. Very useful when squashing commits.

Från: "Aistis Raulinaitis" <notifications@github.com>
Skickat: ‎2016-‎02-‎03 12:18
Ämne: Re: [Idris-dev] Greatly expanded file modes (#2896)

If it's preferred I can close this PR and make a new one based on the latest revision.

melted:
Thanks for the PR!

sheganinans:
Thanks everyone!

jfdm:
Not to be a bore, but can the CHANGELOG be updated to reflect what has been changed with this PR.

sheganinans:
I can do that if nobody's done it yet.

jfdm:
@sheganinans Please do, and also add your self to CONTRIBUTORS.

PR #2897, opened 2016-01-20
Add an example of binding without type annotation

ForNeVeR:
Honestly that confused me a bit, so I think that explicit example without any type annotation would be nice.


Melvar:
I’m surprised that works, since `:let` takes top-level declarations, and those generally require a type.

edwinb:
It'll work fine as long as it's something that could also be inferred on its own at the REPL, and I've often found this form useful.

PR #2900, opened 2016-01-22
Remove escapes from expected output in pkg004

melted:
grumble

### DIFF
--- a/test/pkg004/expected
+++ b/test/pkg004/expected
@@ -1,5 +1,5 @@
-Uncaught error: user error ([1mtest.ipkg[0m:[1m10[0m:[1m1[0m: [91merror[0m: filename
+Uncaught error: user error (test.ipkg:9:1: error: filename
     must contain no directory
     component, expected: space
-[1m[94m<EOF>[0;1m[0m 
-[92m^[0m     )
+<EOF> 
+^     )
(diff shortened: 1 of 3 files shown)


Melvar:
This appears to have removed the escapes from the expected output, while changing the call to idris so that the escapes are actually written, so that the result is still the two being inconsistent.

melted:
Yhe other options do nothing when Idris is in pkg mode. So they didn't do anything, it is the ones in the ipkg file that matters. Even better, idris doesn't output ANSI codes when stdout isn't a tty these days, so the --nocolour option isn't necessary anymore.

Melvar:
Something about that statement is incorrect, since on my system this test fails on Idris HEAD, with the difference being precisely color.

melted:
What OS do you have? Travis is happy with this.

The tty check is implemented at https://github.com/idris-lang/Idris-dev/blob/master/src/Util/System.hs#L63 and is used at https://github.com/idris-lang/Idris-dev/blob/9e8930d494325ff68d0aea0aacb01271b81187b0/src/Idris/REPL.hs#L1699

melted:
Ah, of course. That's useless because the package functionality doesn't come via idrisMain. I'll make a fix of some sort.

Melvar:
I’m on Linux and build with I think all the options.

melted:
The escapes comes from trifectas error messages which are pretty-printed. So make them plain to standardize it, as it is outside the options we have to control printing. https://github.com/idris-lang/Idris-dev/pull/2917

PR #2902, opened 2016-01-22
Use only one line of output per test

melted:
Remove unused var

### DIFF
--- a/test/runtest.pl
+++ b/test/runtest.pl
@@ -17,7 +17,6 @@ my @idrOpts    = ();
 #  2. Stack
 #
 sub set_path {
-    my ($test_dir,) = @_;
     my $sandbox = "../.cabal-sandbox/bin";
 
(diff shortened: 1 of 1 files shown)


Melvar:
Did you test this when running tests in parallel? I’m guessing that will be a huge mess now.

melted:
Emmm, parallel tests should work fine outputwise for an embarrassing reason, my changes made them serial. I didn't think parallel testing was a thing, the test targets don't use it. I'll fix things later tonight.

Från: "Melvar Chen" <notifications@github.com>
Skickat: ‎2016-‎01-‎22 18:19
Ämne: Re: [Idris-dev] Use only one line of output per test (#2902)

Did you test this when running tests in parallel? I’m guessing that will be a huge mess now.

Melvar:
The way to invoke testing in parallel is just with `make`’s `-j` option.

PR #2903, opened 2016-01-22
Add C heap

ziman:
C heap is another heap for Idris that is mark-and-sweep-collected, rather than Cheney-collected. This means that C code can attach finalizers to allocated objects and those will be called once they are no longer referenced from the main heap.

The idea is that the main heap, also called the "FP heap", copies only live data, which works great for FP-style allocation where lots of constructors are rapidly allocated and discarded. However, this makes it impossible to attach finalizers to objects. This is exactly where the "C heap" comes in by tracking every object, live or dead.

Seems to work. Used by https://github.com/ziman/idris-bytes and transitively by https://github.com/ziman/text and no problems have been observed during the one day of testing.

Questions to address before merging:
- correctness
- perhaps testing and benchmarks
- suitability of naming and API
- documentation


rts/idris_rts.h
```diff
@@ -112,6 +114,27 @@ struct VM_t {
...
+ *
+ * Although not enforced in code, CData is meant to be opaque
+ * and non-RTS code (such as libraries or C bindings) should
+ * access it only in the following three ways:
+ *
+ *   CData cd = cdata_allocate(...);
+ *     or
+ *   void * ptr = malloc(...);
+ *   CData cd = cdata_manage(ptr, ...);
+ *
```
david-christiansen: inserting "followed by" here would make this comment more readable. This should also appear on the readthedocs site.

david-christiansen:
I think that this seems like a reasonable idea. I do think it needs the following before a merge:
 1. Documentation that describes how to use it, visible on the readthedocs site
 2. Tests that make sure finalizers work, and preferably also one that exercises the feature moderately intensely

docs/reference/misc.rst
```diff
@@ -118,7 +225,7 @@ Here is an example:
       (a * b * c) ={ sym (multAssociative a b c) }=
       (a * (b * c)) ={ cong (multCommutative b c) }=
       (a * (c * b)) ={ multAssociative a c b }=
-      (a * c * b) ={ cong {f = (*b)} (multCommutative a c) }=
+      (a * c * b) ={ cong {f = (* b)} (multCommutative a c) }=
```
ziman: Vim's reSt syntax plugin makes everything italic from this point to the end of the file. Hotfix: add space.

melted:
I so want to have finalizers when I'm playing around with the FFI. Another thing that would be great for using callbacks would be something like Foreign.StablePtr in Haskell, so it would be possible to send an idris value through C code disguised as a void pointer.

ziman:
Can anyone take the `WIP` label off this PR, please?

(Unless there are objections.)

jfdm:
@ziman label removed. Will do a numpty check once the tests have finished, then look to merge.

test/primitives006/run
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+cc -c -O2 -o array.o array.c `idris --include`
```
melted: can't hardcode cc and idris here. 

Use `${CC:-cc}` and `${IDRIS:-idris}`

melted:
This works, well, I get a seg fault, but it builds:
```
#!/usr/bin/env bash
${CC:-cc} -c -O2 -o array.o array.c $(${IDRIS:-idris} --include)
${IDRIS:-idris} $@ -o load-test load-test.idr --nocolour --warnreach
./load-test
rm -f load-test *.o *.ibc
find -name \*.ibc -delete
```

melted:
So, the test is allocating a zillion 64MB arrays that are never freed because the gc is only run if the idris heap is running out of space and there is no notion about the C heap triggering a gc at all. It's not real memory, but a system without overcommit won't let you allocate too many virtual pages.

ziman:
Yes, this is an awkward problem because allocation in C code cannot trigger GC. I'll see where I can sneak GC in.

melted:
How about just accounting for the size of the C heap when deciding to do gc in allocate? Because some Idris objects are bound to be allocated while the program is running, and waiting for that might not be too bad. It would need a growth factor though, so gc isn't run all the time when it crosses a threshold size.

ziman:
Yes, I'm doing something along these lines.

defanor:
`test/primitives006/load-test.idr` appears to be missing in `idris.cabal`, and Travis uses `cabal sdist`.

ziman:
It's the JS configs that fail. We probably should disable this test for JS.

rts/idris_rts.c
```diff
@@ -272,6 +286,22 @@ char* GETSTROFF(VAL stroff) {
...
+    c_heap_insert_if_needed(vm, &vm->c_heap, item);
+    return cl;
+}
+
+VAL MKCDATAc(VM* vm, CHeapItem * item) {
+    Closure* cl = allocate(sizeof(Closure), 1);
+    SETTY(cl, CDATA);
+    cl->info.c_heap_item = item;
+    c_heap_insert_if_needed(vm, &vm->c_heap, item);
+    return cl;
```
melted: The heap insert may trigger a gc and cl wouldn't be a valid pointer after that. Simplest is to allocate the closure after the heap insert is done.

test/primitives006/load-test.idr
```diff
@@ -0,0 +1,48 @@
...
+  BA.pokeInt (63*1024*1024) i arr
+  -- read number from offset 63M
+  j <- BA.peekInt (63*1024*1024) arr
+  -- count matches
+  alloc (x + if i == j then 1 else 0) (i - 1)
+
+main : IO ()
+main = do
+    -- First allocate 4096 64M arrays to break the C heap if there's a bug
+    alloc 0 4096 >>= printLn
```
melted: Can you dial this down to 1024 for people with small C stacks? We need a better story for the stacks on the C backend :-(

rts/idris_rts.c
```diff
@@ -272,6 +286,22 @@ char* GETSTROFF(VAL stroff) {
     return (root->str->info.str + root->offset);
 }
 
+VAL MKCDATA(VM* vm, CHeapItem * item) {
+    Closure* cl = allocate(sizeof(Closure), 0);
+    SETTY(cl, CDATA);
+    cl->info.c_heap_item = item;
+    c_heap_insert_if_needed(vm, &vm->c_heap, item);
```
melted: ditto

melted:
I moved the c_heap_insert line to the top of `MKCDATA` to make it work

melted:
I submitted a PR to ziman/c-heap with those changes, it would be good to finally get this in.

melted:
Also, this inspired me to decouple the idris stack from the c stack. I implemented it, but still have some debugging to do.

melted:
Argh, I forgot to commit 'expected'. I'll fix that separately.

PR #2906, opened 2016-01-23
Changed the '[static]' annotation to '%static', for uniformity with the rest of the language.

ahmadsalim:


jfdm:
So I am happy to commit this if no one objects. Annotation uniformity is good.

PR #2907, opened 2016-01-23
Make a haskell test runner

melted:
This is designed to be drop-in compatible with runtest.pl.


melted:
It's failing on Travis because amazingly idris doesn't depend on the regex libs so they don't get installed. :(
I have put it in as a dependency just to see if there's any further roadblocks. If I have to change the cabal file, I should just add it as a proper test suite there so we can run the tests with cabal test.

melted:
Two problems:
  - GHC 7.6 doesn't have setEnv, I think I'll just conditionally import the one from Posix for that version.
  - This needs regex-compat and noting a dependency in cabal for something that isn't in the cabal file is hacky. I'm still astounded that there isn't a single regex in Idris.´

Otherwise it is fine.

melted:
Does anybody have any objections to this? @edwinb? @david-christiansen ? Anyone else? I'll leave it open a couple of days, so no hurry.

david-christiansen:
Sorry, I clicked wrong on my phone! Didn't mean to close it!

I think this seems like a reasonable idea, in order to reduce dependencies when installing on non-Unix-like systems.

melted:
Time to get a bigger phone :-)

david-christiansen:
Ok, time to stop using Github from the phone. 2 questions: does parallel execution of tests still work? And why not delete the Perl test runner immediately?

david-christiansen:
Hehe

melted:
Parallel execution still works as it runs this in parallel instead of the perl test script, it's still parallel in the Makefile. I want to move the parallelism into this later so we can keep state while still running the tests in parallel. But for now it works exactly like runtest.pl.

I can remove the perl test runner, I just thought I'd do it when this goes in. It was convenient for me to have it available while working on this :-)

jfdm:
Sorry to be late to the party, but did anyone check that the test runner works when the project is built using stack....because it currently doesn't.

The test runner fails as it cannot find 'Text.Regex' even though it is a project dependency, the project it is used for (the test runner) is not picked up by stack.

melted:
I did check with stack and it worked.

Skickat: ‎2016-‎01-‎26 15:36
Ämne: Re: [Idris-dev] Make a haskell test runner (#2907)

Sorry to be late to the party, but did anyone check that the test runner works when the project is built using stack....because it currently doesn't.
The test runner fails as it cannot find 'Text.Regex' even though it is a project dependency, the project it is used for (the test runner) is not picked up by stack.

jfdm:
it doesn't for me.

jfdm:
See issue #2919 for more information.

jfdm:
Also can you double check that you do not have `regex-compat` installed using cabal. I think that is most likely the reason why running `make test_c` will work even if you used stack.

melted:
Ok, I must have contaminated the ghc stack pulled down with regex, maybe I tried to install something other than idris too. I'll fix this later today.

I'm so tempted to use a regex in idris somewhere :)

I'd like to make the script known to cabal some way, either as a conditional exe oe as a test suite. But I  don't want a rebuild of idris to be triggered by some one wanting to run tests. 

Skickat: ‎2016-‎01-‎26 15:46
Ämne: Re: [Idris-dev] Make a haskell test runner (#2907)

it doesn't for me.

jfdm:
@melted I have documented a 'fix' in #2919.

Issue #2905, opened 2016-01-23
Static implicits do not work


ziman:
Partial evaluation seems to wipe values passed to implicit arguments. This makes static implicit arguments not work. (i am not sure why class constraints work despite this.)

Test program:

``` idris
addni : Nat -> [static] {n : Nat} -> Nat
addni  Z    {n = n} = n
addni (S m) {n = n} = S (addni m {n = n})

addne : Nat -> [static] (n : Nat) -> Nat
addne  Z    n = n
addne (S m) n = S (addne m n)

main : IO ()
main = do
  printLn $
    addni (S (S (S (S Z)))) {n = S (S (S Z))}
    + addne (S (S (S (S Z)))) (S (S (S Z)))
```

With `--log 3`, we can see that `addni` specialises to `PE_addni_bd895c42`, whose first clause is:

``` idris
PE_addni_bd895c42 Z = addni Z {n = _}
```

This clause does not typecheck:

```
Couldn't specialise: bits.idr:12:1:Can't infer argument n to addni
```

Which is no surprise since the value of `n` has been erased because `n` is implicit, despite being the static argument.

Implicits seem to be erased in `mkPE_TermDecl`, which contains `eraseImps` in its where block. Unfortunately, `PArg` does not contain information about static-ness (why?), and therefore the easy fix "erase implicits unless static" is not possible because the information is not available at that point.

I'm not sure how to fix this. Why do we erase implicits at all?

ziman:
The actual problem is that delaboration-elaboration did not preserve the implicit. Preparing a patch.

ziman:
The key change of the fix is the line that reconstructs the forgotten implicit:
https://github.com/idris-lang/Idris-dev/compare/master...ziman:fix-peval?expand=1#diff-36e6808bf2c73df436a854a1f62916f0R233

ziman:
This fixes the above test program but not `Data.Bits`. :(
```
Couldn't specialise: No location:When checking an application of PE_or_580f5e8d:
        Type mismatch between
                Bits n (Type of (2arg))
        and
                Bits 5 (Expected type)
```

ziman:
This is taken care of by commenting out this line:
https://github.com/idris-lang/Idris-dev/compare/master...ziman:fix-peval?expand=1#diff-36e6808bf2c73df436a854a1f62916f0L268
Now `Data.Bits` specialises correctly! \o/

So actually both implicit-erasing passes have to go away. Do we need them?

[...]

PR #2911, opened 2016-01-24
Add [static] + fix erasure warnings in Data.Bits

ziman:
The bit width is generally runtime-relevant so we have to make the implicits bound to avoid warnings. I also made the bit widths `[static]` to get the functions specialised.

However, specialisation will not work until the following issues are resolved:
- static implicits don't work — https://github.com/idris-lang/Idris-dev/issues/2905
- `Nat` functions in `Prelude` won't specialise because they don't have `[static]` annotations

I reckon this patch is worth merging now anyway.


jfdm:
@ziman If this can be merged, you want to fix up the conflicts and then merge it?

PR #2914, opened 2016-01-24
Fix the "stack build" for the current NixOS unstable

DmitryTsygankov:
The "stack build" command does not currently work on NixOS, since the library & header paths are all weird.
The current version of stack supports Nix integration, which allows the build process to locate the libraries.
Added the required Nix package dependencies, and updated the INSTALL.md with the new instructions.
Bumped up the stack resolver version, since I somehow had issues building with 3.16, and also, why not?


melted:
Thanks for the PR!

PR #2917, opened 2016-01-25
Convert trifecta's pretty-printed error messages in Pkg

melted:
Because Pkg is entirely outside the mechanism that control the output
in other parts of Idris, the easiest solution is to scrub the error
messages from trifecta, which will otherwise contain ANSI escapes or not
on different systems.


melted:
@Melvar can you test this?

melted:
The immediate motivation is to make pkg004 work everywhere.

Issue #2909, opened 2016-01-24
Watch command doesn't work


etorreborre:
I am not sure what to include in this report to reproduce the issue. Basically I upgraded Idris from `0.9.2` to `0.10` using `brew upgrade`. When I open a file `:l tdd.idr` then `:watch` it nothing happens if I type new contents into the file, even when I had expressions which don't compile:
```idris
*tdd> :watch
tdd.idr
Watching for .idr changes in ["tdd.idr"], press enter to cancel.
```

melted:
I see something similar on Windows. I have tracked down one bug, that it's case-sensitive. Easy enough to fix. I will continue later.

LeifW:
I can reproduce this on Mac OSX.

The relevant code is here: https://github.com/idris-lang/Idris-dev/blob/master/src/Idris/REPL.hs#L691

I added some debug prints (via Debug.Trace), and see what's going on now. I set dir(s) to watch by calling `takeDirectory` (to get the base dir path) on the set of file inputs as reported by Idris.
If the file's in the current dir, the watch dir is just ".",  otherwise, it's something like "/tmp", if I'm editing "/tmp/tdd.idr"

The edit events come back as happening to "/private/tmp/tdd.idr". Which I don't respond to, as that doesn't look like one of the files I was initially watching ("/tmp/tdd.idr"). Sounds like a bug with the Mac bindings for this fsnotify lib, if you ask me; I said I was watching "/tmp", so why are change events in there coming back as "/private/tmp"? I'll look into that, but we could probably hack up a workaround for now.

LeifW:
One workaround that seems to work for now: If I say `:l /private/tmp/tdd.idr` (instead of `:l /tmp/tdd.idr`), then I'm watching `/private/tmp/tdd.idr`, and the change events still come back as `/private/tmp/tdd.idr`, so everything works.

LeifW:
Sounds like "/tmp" is a symlink into "/private/tmp" on OSX. Some things I'm thinking: we can canonicalize all paths (maybe things are already coming back canonicalized in the events, as above?) - use absolute paths (rather than ".", as above), and resolve symlinks. On case-insensitive file systems we could normalize everything to one casing or use a case-insensitive comparison...

LeifW:
[...]

PR #2918, opened 2016-01-26
Canonicalize watched files

LeifW:
Canonicalize the set of loaded files for comparing watch events to.
Resolves `.`, `..`, and symlinks, though docs say symlink resolution doesn't
work on Windows.
For #2909

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -981,7 +981,7 @@ Library
                 , zip-archive > 0.2.3.5 && < 0.2.4
                 , safe
                 , fsnotify < 2.2
-                , async < 2.1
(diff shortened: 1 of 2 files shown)


melted:
GHC 7.6 and 7.8 fail with

```
src/Idris/REPL.hs:695:35:
    Not in scope: ‘traverse’
    Perhaps you meant one of these:
      ‘reverse’ (imported from Data.List),
      ‘T.reverse’ (imported from Data.Text)
```

LeifW:
Yeah, was thinking to change that to mapM when I get home.

PR #2920, opened 2016-01-26
Renamed `Classes.idr` to be consistent with new doctrine.

jfdm:


melted:
```
Type checking ./Data/String.idr
./Data/String.idr:84:26-28:
When checking right hand side of Data.String.case block in case block in Data.String.parseDouble, wfe at ./Data/String.idr:81:19 at ./Data/String.idr:83:23 with expected type
        Maybe (Double, Double, Integer)
Can't find implementation for Neg a```

jfdm:
So the reason for the error message was that the proof search was failing. This was due to hard coding the location of some interface definitions to the old path. The commit has been amended and (fingers crossed) that should be the issue resolved.

PR #2921, opened 2016-01-26
Remove the troublesome dependency on regex-compat

### LINKED ISSUE #2919: Test Infrastructure does not work for with stack.
The new test runner cannot be used when the project is built using stack. 

Doing:

```sh
stack install
make test_c
```

will fail with 

```sh
$ make test_c
make -C test IDRIS=../dist/build/idris/idris test
make[1]: Entering directory `/home/jfdm/projects/idris/source/test'

runtest.hs:16:8:
    Could not find module ‘Text.Regex’
    Perhaps you meant Text.Read (from base-4.8.2.0)
    Use -v to see a list of the files searched for.
make[1]: *** [runtest] Error 1
make[1]: Leaving directory `/home/jfdm/projects/idris/source/test'
make: *** [test_c] Error 2
```
Although, the `idris.cabal` file stipulates that the required dependency should be installed, due to stack's sandboxing efforts this won't be picked up when building the test runner. The fix for using stack would be to build the  test runner using:

```sh
stack runghc runtest.hs <args>
```

I do not know what the appropriate fix would be when cabal sandboxes are being used.

melted:
Resolves #2919

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -983,9 +983,6 @@ Library
                 , fsnotify < 2.2
                 , async < 2.1
 
-  -- regex-compat needed for the test/runtest.hs script, which isn't built by cabal
-  build-depends: regex-compat < 1.0
-
   -- zlib >= 0.6.1 is broken with GHC < 7.10.3
   if impl(ghc < 7.10.3)
      build-depends: zlib < 0.6.1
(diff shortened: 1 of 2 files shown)


melted:
What this doesn't resolve is the thornier question of how to do scripts with dependencies when there can be sandboxes and other stuff and yet do it generally. Sometimes you just want to build something without all the machinery. I'm thinking of making a little helper for that, that plugs into the current cabal and stack environment but otherwise behaves like ghc --make. But that is a promplem for another project than Idris.

PR #2922, opened 2016-01-27
Integrate test suite with cabal

jfdm:
Thanks to @melted the test suite is haskell based. Let's try and get it running using cabal...

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -1043,6 +1043,24 @@ Executable idris
   ghc-prof-options: -auto-all -caf-all
   ghc-options:      -threaded -rtsopts -funbox-strict-fields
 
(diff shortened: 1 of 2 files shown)


jfdm:
A dirty hack to facilitate the running of the test suite with cabal and by extension stack.

melted:
Thanks!

Now we just have to make sure to not create a `test` directory in test. Maybe we could add a command line flag for cabal to pass to it.

jfdm:
I think rather than adding a cabal flag, look at redoing the test suite in Tasty (see #2927) This will avoid the need for the dirty hack, and also remove the need for environment variable trickier as cabal would give us the location of the generated exe, IIRC.

PR #2933, opened 2016-01-31
Util cleanup

LeifW:
Remove some stuff related to looking up env vars for overriding Idris's libdir.

Also removed some java backend (maven) stuff. The Java backend currently uses the `getMvn` function exported from here; will submit a PR against the Java backend adding it to there.


melted:
Looks good!

PR #2935, opened 2016-02-01
Update interp.rst

bftjoe:
Change Int to Integer and use printLn


jfdm:
Apologies for not getting to this sooner. 

Thanks for the PR.

PR #2938, opened 2016-02-01
Fix stack --nix on OS X

projedi:
On OS X Idris depends on hfsevents which in turn depends on Cocoa and CoreServices.

I had to make a separate .nix file to conditionally depend on said frameworks.

Also, unfortunately, you need to specify the resolver in both stack.yaml and .nix shell file.

### DIFF
--- a/stack-shell.nix
+++ b/stack-shell.nix
@@ -1,0 +1,28 @@
+with (import <nixpkgs> {});
+
+let
+  # MUST match resolver in stack.yaml
+  resolver = haskell.packages.lts-4_2.ghc;
+
+  native_libs = [
+    libffi
+    zlib
+    ncurses
+    gmp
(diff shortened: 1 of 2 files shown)


melted:
Thanks for the PR!

LeifW:
Sorry about that - the dep on hfsevents was added in https://github.com/idris-lang/Idris-dev/pull/2857, for the :watch command in the REPL. Could make it an optional feature via a configure flag if that dep is a pain.
Thanks for fixing this!

projedi:
Not a problem at all! If you use Nix on OS X, you probably already have them. And less build options the better in my opinion. What's slightly concerning is the need to specify native dependencies for all your haskell dependencies (and their dependencies). No idea if (or, indeed, how) stack is going to address it.

PR #2941, opened 2016-02-01
Don't rewrite binders in quoted terms under syntax

david-christiansen:
Quoted terms should keep their binders intact, so we need to track the
quote level.


jfdm:
Before I merge this. @david-christiansen is this PR good to go?

david-christiansen:
Sorry about that, I should have merged it long ago.

Issue #2936, opened 2016-02-01
Handling Ctrl-C in REPL


projedi:
Currently sending Ctrl-C to the idris REPL will cause it to exit. This doesn't seem to be the usual behavior of interpreters which is to interrupt running operation if any, or reset the user input. Is current REPL behavior the desired one?

PR #2942, opened 2016-02-02
Do not quit the REPL on Ctrl-C

projedi:
Addresses #2936 with some hacky code to get things started.

I've only tried on the main REPL loop (as opposed to `:prove` or some other subloop) and it seems to "just work" as expected:

```
Idris>           (pressing ^C)
Interrupt
Idris> 123       (pressing ^C)
Interrupt
Idris> fib 40    (pressing Enter and then ^C)
^Cuser interrupt
Idris>           (pressing ^D)
Bye bye
```

### DIFF
--- a/src/Idris/Prover.hs
+++ b/src/Idris/Prover.hs
@@ -271,6 +271,21 @@ undoElab prf env st [] = ifail "Nothing to undo"
 undoElab prf env st (h:hs) = do (prf', env', st') <- undoStep prf env st h
                                 return (prf', env', st', hs)
 
+runWithInterrupt
+  :: ElabState EState
+  -> Idris a -- ^ run with SIGINT handler
+  -> Idris b -- ^ run if mTry finished
+  -> Idris b -- ^ run if mTry was interrupted
+  -> Idris b
+runWithInterrupt elabState mTry mSuccess mFailure = do
+  ist <- getIState
+  case idris_outputmode ist of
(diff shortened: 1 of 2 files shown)


jfdm:
@projedi I've labelled this as WIP. Just let us know when you want the labelled removed.

projedi:
I've decided to go for the easiness of implementation. Now Ctrl-C during REPL input is handled as feeding an empty line; Ctrl-C during computation is interrupted and drops you back to the appopriate REPL (i.e. Ctrl-C during `:eval fib 40` in the prover REPL will drop you into the prover REPL - not the main one). Quitting the subREPLs can still be done by Ctrl-D.

@jfdm I consider this ready for review, could you remove the label, please?

src/Idris/Prover.hs
```diff
@@ -271,6 +271,15 @@ undoElab prf env st [] = ifail "Nothing to undo"
 undoElab prf env st (h:hs) = do (prf', env', st') <- undoStep prf env st h
                                 return (prf', env', st', hs)
 
+ctrlC e mTry mSuccess mFailure = do
```
melted: This could use a better name and a type signature.

projedi: True, renamed to `runWithInterrupt`.

melted:
Thanks for the PR!

projedi:
Thanks for merging!

PR #2943, opened 2016-02-02
Primitive memory operations

melted:
Also add a Storable interface akin to Haskells.

Problem: I can't make prim_pokeN return IO () without the compiler optimizing it out. I implemented it like 
```
prim_poke8 : Ptr -> Int -> Bits8 -> IO ()
 prim_poke8 ptr offset val = MkIO (\w => prim_io_bind (prim_io_return (
     prim__poke8 (world w) ptr offset val)) (\x => prim_io_return ())) 
```
But if I do that it just disappears in the compilation. So I added a bogus Int return and then it works fine. What to do?


Melvar:
You don’t seem to have added the test to `idris.cabal`, so it won’t be being executed I believe.

Also, `Int` is not necessarily large enough to represent every offset.

Melvar:
I just noticed that you put in the `sizeof` and `alignof` functions in the Haskell way, taking a dummy argument. However the way to call those is with `undefined`, which we don’t have. I believe this requires a different design.

melted:
I thought about sizeOf a bit, and yes as it is now it requires a value of that type which differs from Haskell. But then I realized that it might not be all that bad as it is. Because then we can have storables of varying size (they would have to provide a way to read in the value by encoding the length or null-termination or something). I put it in contrib because I expect it to evolve.

Melvar:
A major use-case of `sizeOf` is allocating arrays and reading at offsets, both of which require that the size returned be uniform.

melted:
That's true. That might belong one abstraction level above.

shlevy:
Could `sizeOf` be implemented with elab reflection somehow?

melted:
I will merge this as is, Data.Storable is in contrib, so we can come up with a better interface later.

PR #2950, opened 2016-02-05
Minor haskell code readability improvements for AST modules

Heather:
replace if else with list comprehension syntax, use pipes `.` instead of applieng `map` two times. used to indent `=` because it was easier for myself to read/see difference between values over AST patterns, not sure if that could be helpful for others


jfdm:
Cheers @Heather. Readable code is easier code to maintain. As long as the changes don't change the expected operation (i.e. the tests pass) I see no reason to not merge this.

Heather:
@jfdm thanks

melted:
And all the tests pass!

PR #2952, opened 2016-02-07
cabal: Remove redundant ghc-options

bgamari:
-rtsopts and -threaded only apply to executables yet they were applied
to a library.


melted:
Thanks for the PR!

PR #2953, opened 2016-02-08
Some new C FFI features

melted:
This adds some new features to the C FFI that match what's in the Haskell FFI.

Prefixing the name with `&` gives a pointer to a C global. 

```
errno : IO Ptr
errno = foreign FFI_C "&__errno" (IO Ptr)
```

Putting `%wrapper` in the name argument gives a function pointer to a wrapped Idris function (useful for APIs that take a struct with a bunch of functions). It's functionality matches "wrapper" in Haskell, but the illegal C name makes name clashes go away.

Putting `%dynamic` in the name argument calls a C function from a function pointer. This is useful for APIs that provide structures with C functions (popular in object oriented C) or otherwise returns C function pointers. ~~There is an ugliness here, as there is no great way to generate temp names in the pure codegen, this thing allows a name after `%dynamic`, like `"%dynamic example"`, so that it can be used without name clashes if used several times in a declaration.~~ The type signature in dynamic should take an extra Ptr up front for the function pointer.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -15,6 +15,14 @@ Updated export rules
 * Exported types can only refer to other exported names
 * Publicly exported definitions can only refer to publicly exported names
 
+Improved C FFI
+--------------
+
+* Idris functions can now be passed as callbacks to C functions or wrapped
+in a C function pointer.
+* C function pointers can be called.
+* Idris can access pointers to C globals.
+
 Minor language changes
 ----------------------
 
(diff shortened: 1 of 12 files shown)


jfdm:
Might want to ensure the change log is updated too. Also it might be a good idea to place the PR comment in the last commit message so that it is **part** of the Idris git repo and not GitHub's infrastructure.

melted:
I need to remember to update the change log.

Of course the %dynamic uglkiness isn't even neccessary, I can just omit the local and fold it into the usage site with a completely horrible cast. I will fix that.

melted:
It's time to document stuff!

PR #2959, opened 2016-02-09
Add flatten to Monad.

ziman:
Per request on the mailing list. Stdlib compiles, tests pass.

### DIFF
--- a/libs/prelude/Prelude/Monad.idr
+++ b/libs/prelude/Prelude/Monad.idr
@@ -21,12 +21,16 @@ interface Applicative m => Monad (m : Type -> Type) where
 
     -- default implementations
     (>>=) x f = join (f <$> x)
-    join x = x >>= (\y => y)
+    join x = x >>= id
 
 ||| For compatibility with Haskell. Note that monads are **not** free to
 ||| define `return` and `pure` differently!
 return : Monad m => a -> m a
 return = pure
+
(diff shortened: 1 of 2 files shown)


david-christiansen:
Why did you choose to call it `flatten` rather than `join`? Most of the library adopts names from Haskell.

ziman:
I didn't choose the name, I just moved the function into the interface.

Shall we rename it and add an alias, like with `return`?

david-christiansen:
I think so, unless someone else has a good reason for this naming.

Melvar:
IIUC it was named that way because at some point there were Algebra classes including `Joinsemilattice` which defined `join`. I am in favor of using `join` for the monad meaning now that those are out of the way.

sellout:
Do we need to leave `flatten = join` somewhere to avoid breaking existing user code? And why use `(\y => y)` rather than `id`?

Otherwise, I strongly approve.

david-christiansen:
There should probably be a deprecated top-level defn along the lines of:
```
%deprecated
flatten : Monad m => m (m a) -> m a
flatten = join
```

jfdm:
@david-christiansen Do you not mean:

```
flatten : Monad m => m (m a) -> m a
flatten = join
%depreated flatten "..."

david-christiansen:
Probably. I seem to be constitutionally unable to remember that syntax.

PR #2965, opened 2016-02-11
Fix missing consolewidth in test

### LINKED ISSUE #2964: basic017 breaks my terminal - Curses problem
Running `.stack-work/dist/x86_64-linux/Cabal-1.22.5.0/build/regression-and-sanity-tests/regression-and-sanity-tests basic017` in tmux either makes my terminal stop updating anything other than the last line or makes it as though the terminal were half as tall as it is. The test fails as well. Outside of tmux, it runs properly, but I still get a test failure. This is the error:
```
Using stack work dir at ../.stack-work/dist/x86_64-linux/Cabal-1.22.5.0/build/idris
basic017 finished...FAILURE
1c1,3
Uncaught error: user error (Curses[-1]:endwin)
---
> basic017a.idr:11:8:When checking type of Main.append:
> When checking an application of Main.Vect:
>         No such variable n

----
1 tests run: 0 succesful, 1 failed, 0 updated.

Failed tests:
basic017
```

The left side of the diff above is the same as the content of `test/basic017/output`, regardless of tmux. It runs successfully, giving the current output, if I `cd` into `tests/basic017` and do `./run`. And if I do `./run > log`. Running it with `make basic017.test` instead of from Stack's build of the test runner gives the same error as above.

Why would whether or not it's run from the test runner change the behavior of curses? And why would curses even be an issue when it's redirected to a file instead of the terminal? And what does the content of basic017 have to do with curses anyway?

This is 56122b5

melted:
Fixes #2964

### DIFF
--- a/test/basic017/run
+++ b/test/basic017/run
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-${IDRIS:-idris} $@ basic017.idr --check --nocolor
-${IDRIS:-idris} $@ basic017a.idr --check --nocolor
+${IDRIS:-idris} $@ basic017.idr --check --nocolor --consolewidth 80
+${IDRIS:-idris} $@ basic017a.idr --check --nocolor --consolewidth 80
 rm -f *.ibc


david-christiansen:
Thanks!

PR #2966, opened 2016-02-11
No need to require bash to run tests

melted:
I had to install bash on my FreeBSD server to run the tests, and that felt unnecessary, it's not like it's actually using any bash feature.

### DIFF
--- a/test/basic001/run
+++ b/test/basic001/run
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 ${IDRIS:-idris} $@ reg005.idr -o reg005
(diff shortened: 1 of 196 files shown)


Melvar:
Are you sure? Did you check every test? I thought there was at least one using arrays.

melted:
We'll see, Travis uses dash for sh, I'm pretty sure that sh is bash locally for me.

melted:
Looking good.

Melvar:
Yep, `primitives002` does use arrays, because it was by far the easiest way for me to fix it. `reg039` and `basic010` also do to attempt shellescaping with their timeout thing.

melted:
well, then I put this on ice

PR #2968, opened 2016-02-12
Get more data from bigint when truncating

### LINKED ISSUE #2967: basic011 failing on windows
```
basic011 finished...FAILURE
1,10c1,10
< 3774905B97FA18D6
< 3774905B97FA18D6
< F4996317B78A192F
< 76596D9293646134
< 3774905B97FA18D6
< 3774905B97FA18D6
< 3774905B97FA18D6
< 3774905B97FA18D6
< 3774905B97FA18D6
< 3774905B97FA18D6
---
> 6EA049D797FA18D6
> 6EA049D797FA18D6
> FDA01C93B78A192F
> AD85270E93646134
> 6EA049D797FA18D6
> 6EA049D797FA18D6
> 6EA049D797FA18D6
> 6EA049D797FA18D6
> 6EA049D797FA18D6
> 6EA049D797FA18D6
```

melted:
Systems with 32-bit longs only got 32 bits even when truncating to 64 bits.

Fixes #2967

### DIFF
--- a/rts/idris_gmp.c
+++ b/rts/idris_gmp.c
@@ -488,3 +488,16 @@ VAL idris_castBigStr(VM* vm, VAL i) {
     return MKSTR(vm, str);
 }
 
+// Get 64 bits out of a big int with special handling
+// for systems that have 32-bit longs which needs two limbs to
+// fill that.
+uint64_t idris_truncBigB64(const mpz_t bi) {
+    if (sizeof(mp_limb_t) == 8) {
+        return mpz_get_ui(bi);
+    }
+    int64_t out = mpz_get_ui(bi);
+    if (bi->_mp_size > 1 ) {
+        out |= (uint64_t)bi->_mp_d[1] << 32;
+    }
+    return out;
+}
--- a/rts/idris_gmp.h
+++ b/rts/idris_gmp.h
@@ -43,6 +43,8 @@ VAL idris_bigOr(VM* vm, VAL x, VAL y);
 VAL idris_bigShiftLeft(VM* vm, VAL x, VAL y);
 VAL idris_bigShiftRight(VM* vm, VAL x, VAL y);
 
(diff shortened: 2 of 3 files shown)


Melvar:
Are those underscored fields part of the public interface of GMP?

melted:
it's part of gmp.h, so they should be fair game.

melted:
though I see a problem here, the sizeof comparison should be with mp_limb_t because it's possible to compile the real gmp with different typedefs for that.

Melvar:
Not necessarily fair game; they might only be there to make the type usable without indirection. I thought I heard prefixing with `_` was a common way to mark struct fields as private, lacking a way to do this at the language level.

Also, according to the docs on `mpz_get_ui`, it will truncate the absolute value of the integer, rather than pretending it uses two’s complement which is what the Evaluator’s version of the prim appears to do.

melted:
It's easy enough to make an opaque struct in C. You just have a declaration without a body in the header and the full struct somewhere else. If it's in the header it's part of the interface.

As for the behavior, it doesn't seem great, but I made it the same as before. Putting the sign in would actually be worse for what caused #2967 to appear, there's a recent bug that creates these bigints and then makes Bits64 from them. The previous behavior was to just make the Bits64 directly.

Melvar:
If you make a struct opaque, my understanding is you can’t have local variables of it, call functions with it by value, or do anything else that requires the size of the struct be known.

Melvar:
In fact, these fields are detailed in the documentation in the chapter _Internals_ which begins with  a bold warning, “This chapter is provided only for informational purposes and the various internals described here may change in future GMP releases. Applications expecting to be compatible with future releases should use only the documented interfaces described in previous chapters.”

melted:
The struct had the same fields in 1996, I think we can take that risk. Better risk brokenness in the indefinite future than accept it now.

PR #2972, opened 2016-02-14
A higher level C FFI

melted:
This isn't ready yet, but it's beginning to take shape. Putting it up here for feedback. Just don't comment on the test yet, it's a playground for now :-)


libs/prelude/Builtins.idr
```diff
@@ -213,3 +213,8 @@ export data ManagedPtr : Type
 
 %extern prim__peekPtr : prim__WorldType -> Ptr -> Int -> Ptr
 %extern prim__pokePtr : prim__WorldType -> Ptr -> Int -> Ptr -> Int
+
+%extern prim__peekDouble : prim__WorldType -> Ptr -> Int -> Double
+%extern prim__pokeDouble : prim__WorldType -> Ptr -> Int -> Double -> Int
+%extern prim__peekSingle : prim__WorldType -> Ptr -> Int -> Double
+%extern prim__pokeSingle : prim__WorldType -> Ptr -> Int -> Double -> Int
```
ziman: Why are these primitives, rather than FFI calls? How do non-C backends implement these?

melted: They are primitives because I believe that a general mechanism for accessing something memory-like would be useful for most possible backends, even Javascript has the tools to implement them these days. I thought it was reasonable to handle floats the same as integer types.

The cop-out answer for how to implement them is that one doesn't have to. They're external primitives, they're not mandatory (not that internal primitives are mandatory either :-)). But I made the interface Ptr + Offset is that Ptr could represent for example, a memoryview object in Python or a JavaScript ArrayBuffer..

melted:
Because there was interest I cleaned this up (added some doc comments, made stuff private etc) so it's fit to go in. I punted on bounds checking the field access at compile time. There is still a lot that can be done, but it's in contrib so we don't have to be terribly strict about breaking changes.

PR #2973, opened 2016-02-15
Curses fixes for 4 more tests

enolan:
Same problem as in #2964.

### DIFF
--- a/test/reg068/run
+++ b/test/reg068/run
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-${IDRIS:-idris} $@ reg068.idr --check
+${IDRIS:-idris} $@ reg068.idr --check --nocolor --consolewidth 80
 rm -f *.ibc
--- a/test/reg069/run
+++ b/test/reg069/run
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-${IDRIS:-idris} $@ reg069.idr --check
+${IDRIS:-idris} $@ reg069.idr --check --nocolor --consolewidth 80
 rm -f *.ibc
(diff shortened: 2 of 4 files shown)


spearman:
Looks like the original issue was closed, but building this commit I still get the following error when executing `stack build` (same error when building `master`): http://lpaste.net/434370422264299520

enolan:
Okay, something more is going on. We should figure out why the behavior changed in the first place, rather than passing `--nocolor --consolewidth 80` to every single invocation of Idris in the project.

@spearman the error you're getting is from [the makefile for building the prelude](https://github.com/idris-lang/Idris-dev/blob/master/libs/prelude/Makefile#L4-L5). Idris is called without any options for curses. I don't have a problem there. What terminal are you in? I'm in GNOME Terminal which sets $TERM to `xterm-256color`.

spearman:
GNOME Terminal 3.6.2, $TERM is set to `xterm`

I've narrowed it down to somewhere between f2eaf90 and a009be6

spearman:
Looks like f2eaf90 is the last commit i can build without the error

melted:
No-color isn't needed anymore as it checks if the output is a tty. Presumably the  screenwidth shouldn't be detected for fancy formatting if isATTY is false either.

Från: "Echo Nolan" <notifications@github.com>
Skickat: ‎2016-‎02-‎15 06:43
Ämne: Re: [Idris-dev] Curses fixes for 4 more tests (#2973)

Okay, something more is going on. We should figure out why the behavior changed in the first place, rather than passing --nocolor --consolewidth 80 to every single invocation of Idris in the project.
@spearman the error you're getting is from the makefile for building the prelude. Idris is called without any options for curses. I don't have a problem there. What terminal are you in? I'm in GNOME Terminal which sets $TERM to xterm-256color.

enolan:
Withdrawn in favor of #2974.

Issue #2915, opened 2016-01-25
Replace Make with Shake in build system.


enolan:
I propose to get rid of the GNU Make based system and replace it with one based on [Shake](http://shakebuild.com/). Shake is a Haskell EDSL for writing build systems. There are several advantages over Make:
- less implicit behavior
- all the abstraction facilities of Haskell
- no tabs
- automatic prediction of build times and time remaining
- build time profiling

Practically, this plan would make the build system easier to extend and maintain.

The current build system is "irritating" (#311) and contains a lot of redundant code. libs/Makefile is basically five lines repeated five times because Make doesn't even have functions. The libs/*/Makefile files are almost exact copies of each other. There are more examples. We can clean things up a lot with a better system.

There are disadvantages to this plan however. The clean way to use an external library in the build system is with [setup-depends](http://www.well-typed.com/blog/2015/07/cabal-setup-deps/), but that won't be available until Cabal-1.23 is out. It's planned to be released concurrent with GHC 8.0, in mid February. Any PR would have to wait until then. Users who hadn't upgraded would have to `cabal install Cabal cabal-install` or get GHC 8.0 based systems.

I've planned to do this for a while, but would like to get a thumbs up or down before starting.

david-christiansen:
I would certainly like a more maintainable and easier to understand build system!

I've got a couple of questions: How is Shake for Windows support? Will the new Cabal work with older GHCs? Is Shake likely to be maintained for the reasonably foreseeable future?

melted:
Neil Mitchell is a Windows developer, so no worries on that account. Shake even has a nifty progressbar in the task bar on Windows.

enolan:
The Windows support is pretty good, as @melted says. As for the future of Shake, it's been in use by Standard Chartered since 2009, it's used in Stack's Docker image creation and there's an ongoing project to move GHC over to it. I think it'll continue being maintained, even if Neil gets hit by a bus or something.

As for backwards compatibility, "[Cabal's policy is to support being built by versions of GHC that are up to 3 years old.](https://github.com/haskell/cabal/blob/master/HACKING.md)" So we should be fine.

melted:
GHC is close to merging the shake build. See this thread: https://mail.haskell.org/pipermail/ghc-devs/2016-January/011115.html

ahmadsalim:
[...]

PR #2974, opened 2016-02-16
Don't use curses when we're not in a terminal

enolan:


spearman:
Any idea about this `Incompatible ibc version` error? http://lpaste.net/3705792169190096896

I tried `$ stack clean` but the error remained

enolan:
The clean implementation is incomplete. Run `find libs -name "*.ibc" -delete` or `git clean -dfx`. The git command will wipe out everything not tracked by git, the find one will wipe out ibc files in libs, but not anything else. Clean will be better when I get around to #2915.

melted:
Thanks!

Issue #2983, opened 2016-02-21
Clear documentation that type checker doesn't reduce what it can't see is total


david-christiansen:
Right now, the documentation doesn't clearly state that non-total definitions aren't reduced when type checking. As far as I can tell, it's only mentioned in the middle of a moderately-related FAQ entry. We should make this more obvious.

david-christiansen:
Supersedes #2981

xekoukou:
The relevant section on the tutorial is when the Dependent Types are discussed. Probably add a new section before them.

edwinb:
I wonder what would happen if the type checker issued a warning when it encountered a non-total name in a type... given that it's unlikely to be what you wanted.

ahmadsalim:
I think that #3098 actually fixes this.

PR #2988, opened 2016-02-24
Added documentation for codata to the tutorial.

lambda-11235:
I added a section on using codata to the tutorial. I'm not very familiar with Idris yet, so feedback on whether the new documentation is incorrect, or should be moved to another place would be appreciated.

### DIFF
--- a/docs/tutorial/typesfuns.rst
+++ b/docs/tutorial/typesfuns.rst
@@ -720,7 +720,9 @@ infinite loop and cause a stack overflow.
     main = do printLn $ findB (== 1) blue
 
 To fix this we must add explicit ``Inf`` declarations to the constructor
-parameter types. For example, the following outputs "1".
+parameter types, since codata will not add it to constructor parameters of a
+**different** type from the one being defined. For example, the following
+outputs "1".
 
 .. code-block:: idris
 
--- a/docs/tutorial/typesfuns.rst
+++ b/docs/tutorial/typesfuns.rst
@@ -680,7 +680,7 @@ This gets translated into the following by the compiler.
 
 .. code-block:: idris
 
-    codata Stream : Type -> Type where
+    data Stream : Type -> Type where
       (::) : (e : a) -> Inf (Stream a) -> Stream a
 
 The following is an example of how the codata type ``Stream`` can be used to
--- a/docs/tutorial/typesfuns.rst
+++ b/docs/tutorial/typesfuns.rst
@@ -137,7 +137,7 @@ Unlike Haskell, there is no restriction on whether types and function
 names must begin with a capital letter or not. Function names
 (``plus`` and ``mult`` above), data constructors (``Z``, ``S``,
 ``Nil`` and ``::``) and type constructors (``Nat`` and ``List``) are
-all part of the same namespace. By convention, however, 
+all part of the same namespace. By convention, however,
 data types and constructor names typically begin with a capital letter.
 We can test these functions at the Idris prompt:
 
@@ -583,7 +583,7 @@ example for reading and writing files, including:
 
     openFile : (f : String) -> (m : Mode) -> IO (Either FileError File)
     closeFile : File -> IO ()
-    
+
     fGetLine : (h : File) -> IO (Either FileError String)
     fPutStr : (h : File) -> (str : String) -> IO (Either FileError ())
     fEOF : File -> IO Bool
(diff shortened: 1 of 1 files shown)


justjoheinz:
+1, but a small typo: `Codata types are like are like regular ...` right at the beginning.

docs/tutorial/typesfuns.rst
```diff
@@ -661,6 +661,93 @@ without any explicit use of ``Force`` or ``Delay``:
     ifThenElse True  t e = t;
     ifThenElse False t e = e;
 
+Codata Types
+============
+
+Codata types are like regular data types, except that they allow for us to
+define infinite data structures.  More precisely, for a type ``T``, each of its
+constructor arguments of type ``T`` are transformed into a coinductive parameter
+``Inf T``. This makes each of the ``T`` arguments lazy, and allows infinite data
```
jfdm: Am wondering if there is a better way to say 'coinductive parameter' using more simple language.

lambda-11235: @jfdm I debated between calling it a lazily evaluated parameter and coinductive parameter, but I decided the coinductive was more precise. I could change it. Any other suggestions would be helpful as well.

LeifW: 'recursive parameter'?

LeifW: I think I like the precise language. We're already talking about codata here; I think they can get the gist from the context on this that the recursive position is treated a little differently; worst case they can go and look up coinduction.

xekoukou: A)
The explanation of why codata can not allow for infinite mutual recursive
definitions doesn't look good.
Codata is supposed to be sugar for Inf, thus it doesn't make sense for it
to not work.

In the definition of Mutual it specifies that dependent type reduction
happen after type declarations are added. Does codata need to be reduced to
the desugared version? If so, it is very bad from a pedagogical point to
have to understand the implementation of codata to make sense of the rules.

B) There are two more examples that are needed:

1) An example of a type that has one constructor with finite recursion and
another with infinite corecursion.
https://github.com/idris-lang/Idris-dev/blob/da914af83e36d726ba352fb89f5c15689b675540/test/totality004/totality004.idr

2) One also needs to expand the totality definition to add functions that
have codata in their corecursions.

My opinion is that the concept of totality should be removed from the
Theorem proving section and become an independent one.
This would also fix #2983.

docs/tutorial/typesfuns.rst
```diff
@@ -661,6 +661,93 @@ without any explicit use of ``Force`` or ``Delay``:
...
+.. code-block:: idris
+
+    codata Stream : Type -> Type where
+      (::) : (e : a) -> Stream a -> Stream a
+
+This gets translated into the following by the compiler.
+
+.. code-block:: idris
+
+    codata Stream : Type -> Type where
```
LeifW: I think you can just say it translates into `data Stream ...`? I mean, you could write the impl of Stream with `data` + `Inf` instead of `codata` and the end result would be the same?

xekoukou: Ok I made some tests. So for reasons that are not known to me, codata are
sugar for self recursive infinite data structures only.
Whether this definition is good or not, i have no experience to say.

If the word "codata" is used for the general case of infinite data
structures, then it creates confusion.

So I wouldn't create an example of codata with non-self recursion and say
that this doesn't work because it doesn't work by Definition.

LeifW: Are you replying to my comment above? I just meant you can leave your existing example as-is (which has direct self-recursion), only changing the "desugared" form from `codata` to `data` (it already has the `Inf` manually inserted on the recursive param).

xekoukou: I wasn't. (not my example) But you are right. There is a mistake there.

lambda-11235: @LeifW That `codata` in the translated version of Stream was a typo. I changed it to `data`.

ahmadsalim:
@lambda-11235 I have seen that you have made some updates, would you consider this PR largely completed by now?

lambda-11235:
@ahmadsalim I don't know. @xekoukou's points may not be completely satisfied. The changes I made might clarify (A). However, I'm not familiar enough with how totality checking to work on (B2), and I'm not sure how/if I'd want to address (B1). My main goal was to document codata and its use, which I believe I've done, so that people won't use simple Lazy constructor parameters out of ignorance, like I did.

ahmadsalim:
@lambda-11235 I think any documentation is better than nothing, and we should merge things soon if you are not planning to work more on it. If people want to improve later, then that is of course great.

@jfdm What do you think, should we just merge it as is?

xekoukou:
I would like to have a precise definition of codata and coinduction, and
then check that this definition is consistent with the "codata" keyword.

ahmadsalim:
@xekoukou Would it be possible if we possibly after merging this documentation, you could add some of your desired clarification to the text?

I think that would be very helpful, and allows us to progress this PR which has been stuck for over a month.

xekoukou:
Sure.

lambda-11235:
@ahmadsalim @xekoukou That would be nice, as I'm not exactly certain what changes you want made. You may also be able to pull it into another branch besides *master* and make changes there. I don't currently have enough time to work on this in detail.

jfdm:
If it is not ready but want to work on it further we can place in a branch and people can PR and work as they see fit.

ahmadsalim:
@jfdm I think we should do that then

lambda-11235:
Changing target branch.

ahmadsalim:
Thank your for the discussion everyone!

@xekoukou I have made a new branch `documentation/codata`, please add your contributions to that branch if/when you have time. Thanks.

ahmadsalim:
@xekoukou It seems that you haven't taken a look into this yet, if you don't have time/hadn't done anything so far, should I just go ahead and see if I can fix somethings myself?

@jfdm What kind of effort would you suggest for this to be acceptable in to master?

xekoukou:
Please do.

ahmadsalim:
Ping @jfdm !

jfdm:
erm...having a quick look at this. I think best to merge for now, and get working on what, if any, improvements need to be made to the text, together with any other 'issues'. For example, the comment of @xekoukou on where total should be situated in the documents, is an orthogonal issue.

xekoukou:
536372f4cdcf9f83e0d9bc0d050711206790d111 Maybe this is related.

PR #2998, opened 2016-02-28
Add @ to reserved operators and refactor operator/reserved word parsing a bit

enolan:


david-christiansen:
I think I prefer the old definition of `idrisStyle`, that had the named parts. With these changes, it becomes necessary to track down the Haddock for `IdentifierStyle` in order to understand it. Otherwise, things are getting easier to read here. Thanks!

enolan:
Yeah, revised.

melted:
Thanks!

PR #3004, opened 2016-03-02
Update typesfuns.rst - extra documentation for implicit pattern matching - because it was not clear (to me) how pattern matching for implicit works. 

jhegedus42:


david-christiansen:
This commit message doesn't tell me very much. I can see in the diff which file you modified - please write commit messages and pull request descriptions that contain a bit of the "why" in addition to the "what".

Also, while I think this is generally a good thing to have in the docs, this specific example is a bit unfortunate because it conflicts with the intended erasure behavior of a Vect by forcing the Nat to be runtime-relevant. If it also showed the constructors, that wouldn't be a problem anymore because it would have runtime-relevant info to case-split.

edwinb:
I think this example is fine, to be honest. It's not something I'd want to go into in the docs, but my intention with Vect is often to have the length around at runtime, but on the stack (i.e. as an argument to functions) but not the heap (i.e. not stored in the structure itself). I don't think we have a good notation yet for making this clear.

We'll have to think about the right way of annotating erasure intentions. I don't think the current way is very good, though it is the best anyone has thought of so far.

PR #3009, opened 2016-03-03
Curses! This dependency must be removed!

### LINKED ISSUE #3007: Get rid of curses dependency
@melted has found a different, more lightweight way to do what we do with curses: https://hackage.haskell.org/package/terminal-size

melted:
Remove curses dependency

fixes #3007

### DIFF
--- a/custom.mk-alldeps
+++ b/custom.mk-alldeps
@@ -1,1 +1,1 @@
-CABALFLAGS += -f LLVM -f FFI -f curses --disable-documentation --disable-profiling --disable-library-profiling
+CABALFLAGS += -f GMP  -f FFI --disable-documentation --disable-profiling --disable-library-profiling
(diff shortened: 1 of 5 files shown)


custom.mk-alldeps
```diff
@@ -1 +1 @@
-CABALFLAGS += -f LLVM -f FFI -f curses --disable-documentation --disable-profiling --disable-library-profiling
+CABALFLAGS += -f GMP  -f FFI --disable-documentation --disable-profiling --disable-library-profiling
```
david-christiansen: Nice catch!

david-christiansen:
Looking good! Hopefully this will also eliminate the need for all the screen width flags in the tests to avoid Curses vomiting everywhere. Fun title too!

melted:
The screen width in the tests was fixed in #2974

david-christiansen:
Well then! :-)

david-christiansen:
Thanks so much!

PR #3010, opened 2016-03-03
Increase the stack reserved on Windows

melted:
Normal stack reserve is 2MB on 64-bit and it's a bit tight, it's just virtual memory
so could have gone higher than this on 64-bit, but it's a limited resource for 32-bit users.


melted:
I'm working on separating the Idris call stack from the C call stack. But this is a seful band aid in the meantime.

PR #3011, opened 2016-03-03
Simplify getScreenWidth

Melvar:
The extra check for a terminal device should have no effect since it is already
part of retrieving the terminal size.


melted:
well golfed!

PR #3014, opened 2016-03-04
Update stack.yaml: curses flag was removed

enolan:


melted:
Did I miss that? Curses!

PR #3017, opened 2016-03-04
Updated the stack configuration to LTS Haskell 5.5

jfdm:


stack.yaml
```diff
@@ -1,14 +1,9 @@
...
-     GMP: True
-extra-deps: 
-- annotated-wl-pprint-0.7.0
-- cheapskate-0.1.0.4
-- libffi-0.1
-nix:
-   enable: false
-   shell-file: stack-shell.nix
+extra-deps:
+- text-1.2.2.0
```
LeifW: This is already in LTS 5.5 afaik?
https://www.stackage.org/lts-5.5/package/text-1.2.2.0

jfdm: The file was amended using `stack` itself, and tbh I forgot to double check the outpu...will amend shortly

PR #3018, opened 2016-03-05
Add where keyword to UList and BadList in the Uniqueness Types docume…

mcgirr:
…ntation.

Without the `where` keyword/operator, the `UList` example will not type check.


melted:
Thanks for the PR!

PR #3024, opened 2016-03-08
Fixed matchApply to use `Prim__MatchApply`

ahmadsalim:
Instead of `Prim__Apply` as it dit before


david-christiansen:
Nice catch!

ahmadsalim:
Thanks! I was casually looking at prim tactics, and it appeared weird that `Prim__Apply` was used twice.

PR #3025, opened 2016-03-08
Add the ability to define datatypes in Elab

david-christiansen:
Add two new Elab effects:
 * declareDatatype, which adds the type constructor declaration to the context
 * defineDatatype, which adds the constructors to the datatype

See test/meta002/DataDef.idr for an example.

### DIFF
--- a/libs/prelude/Language/Reflection/Elab.idr
+++ b/libs/prelude/Language/Reflection/Elab.idr
@@ -57,11 +57,11 @@ data TyConArg =
   ||| Indices are not uniform
   TyConIndex FunArg
 
-||| A type declaration
+||| A type declaration for a function or datatype
 record TyDecl where
   constructor Declare
 
-  ||| The name of the function being declared.
+  ||| The fully-qualified name of the function or datatype being declared.
   name : TTName
 
   ||| Each argument is in the scope of the names of previous arguments.
@@ -81,6 +81,34 @@ record FunDefn a where
   constructor DefineFun
   name : TTName
   clauses : List (FunClause a)
+
+
+||| A constructor to be associated with a new datatype.
+record ConstructorDefn where
+  constructor Constructor
+
+  ||| The name of the constructor. The name must _not_ be qualified -
+  ||| that is, it should begin with the `UN` or `MN` constructors.
+  name : TTName
+
+  ||| The constructor arguments. Idris will infer which arguments are
+  ||| datatype parameters.
+  arguments : List FunArg
+
+  ||| The specific type constructed by the constructor.
+  returnType : Raw
+
+
+||| A definition of a datatype to be added during an elaboration script.
+record DataDefn where
+  constructor DefineDatatype
+  ||| The name of the datatype being defined. It must be
+  ||| fully-qualified, and it must have been previously declared as a
+  ||| datatype.
+  name : TTName
+
+  ||| A list of constructors for the datatype.
+  constructors : List ConstructorDefn
 
 
 data CtorArg = CtorParameter FunArg | CtorField FunArg
@@ -151,6 +179,8 @@ data Elab : Type -> Type where
 
   Prim__DeclareType : TyDecl -> Elab ()
   Prim__DefineFunction : FunDefn Raw -> Elab ()
+  Prim__DeclareDatatype : TyDecl -> Elab ()
+  Prim__DefineDatatype : DataDefn -> Elab ()
   Prim__AddInstance : TTName -> TTName -> Elab ()
   Prim__IsTCName : TTName -> Elab Bool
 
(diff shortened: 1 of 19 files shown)


libs/prelude/Language/Reflection/Elab.idr
```diff
@@ -83,6 +83,34 @@ record FunDefn a where
   clauses : List (FunClause a)
 
 
+||| A constructor to be associated with a new datatype.
+record ConstructorDefn where
+  constructor Constructor
+
+  ||| The name of the constructor. The name must _not_ be qualified -
+  ||| that is, it should begin with the `UN` or `MN` constructors.
```
shlevy: If only we had some way to guarantee at compile time that only a certain subset of constructors were used :tongue:  (not at all serious)

libs/prelude/Language/Reflection/Elab.idr
```diff
@@ -510,6 +540,17 @@ namespace Tactics
   defineFunction : FunDefn Raw -> Elab ()
   defineFunction defun = Prim__DefineFunction defun
 
+  ||| Declare a datatype in the global context. This step only
```
shlevy: Outside of elab is it possible to define a datatype e.g. in a where clause? If so, any reason this can't be scope-limited too?

src/Idris/Core/Elaborate.hs
```diff
@@ -139,13 +139,14 @@ runElab a e ps = runStateT e (ES (ps, a) "" Nothing)
 execElab :: aux -> Elab' aux a -> ProofState -> TC (ElabState aux)
 execElab a e ps = execStateT e (ES (ps, a) "" Nothing)
 
-initElaborator :: Name -> Context -> Ctxt TypeInfo -> Type -> ProofState
+initElaborator :: Name -> Context -> Ctxt TypeInfo -> Int -> Type -> ProofState
```
shlevy: Can we clean up/document these args somehow? Especially with the `Int` now it's not clear from the signature how I should call this.

src/Idris/Elab/Term.hs
```diff
@@ -2076,11 +2135,42 @@ runElabAction ist fc env tm ns = do tm' <- eval tm
...
+           -- Prepare the constructors
+           ctors' <- mapM (prepareConstructor n) ctors
+           ttag <- do ES (ps, aux) str prev <- get
+                      let i = global_nextname ps
+                      put $ ES (ps { global_nextname = global_nextname ps + 1 },
+                                aux)
+                               str
+                               prev
+                      return i
+           let ctxt' = addDatatype (Data n ttag tyconTy False (map (\(cn, _, cty) -> (cn, cty)) ctors')) ctxt
```
shlevy: Will this fail if the datatype is already defined?

libs/prelude/Language/Reflection/Elab.idr
Melvar: Why this limitation? You can define additionally-qualified data constructors in a normal data declaration, so it makes little sense to disallow it in Elab.

src/Idris/Elab/Term.hs
```diff
@@ -2076,11 +2135,42 @@ runElabAction ist fc env tm ns = do tm' <- eval tm
...
+           set_context ctxt'
+           -- the rest happens in a bit
+           updateAux $ \e -> e { new_tyDecls = RDatatypeDefnInstrs n tyconTy ctors' : new_tyDecls e }
+           returnUnit
       | n == tacN "Prim__AddInstance", [cls, inst] <- args
       = do className <- reifyTTName cls
            instName <- reifyTTName inst
-           updateAux $ \e -> e { new_tyDecls = RAddInstance className instName :
-                                               new_tyDecls e}
+           updateAux $ \e -> e { new_tyDecls = --RAddInstance className instName :
```
shlevy: Doesn't this make `Prim__AddInstance` do nothing?

src/Idris/Core/Elaborate.hs
david-christiansen: Good idea, will do.

libs/prelude/Language/Reflection/Elab.idr
david-christiansen: Where clauses ultimately turn into the global context with names that aren't accessible, and this can be used the same way.

shlevy:
I'm noticing that transporting state from the global state to the elab reflector state and back is pretty hairy and manual. Is there any way we can make it more automated?

libs/prelude/Language/Reflection/Elab.idr
david-christiansen: I thought we had gotten rid of that...

src/Idris/Elab/Term.hs
david-christiansen: Now it will! Force-push incoming :-)

david-christiansen:
@shlevy probably. One thing could be a little wrapper around `elaborate`, but that's called in a few different ways, so that wouldn't quite do it.

libs/prelude/Language/Reflection/Elab.idr
david-christiansen: Or at least agreed that it was a bug to be fixed!

src/Idris/Elab/Term.hs
david-christiansen: This is why I shouldn't code before coffee. Thanks for your sharp eyes!

libs/prelude/Language/Reflection/Elab.idr
Melvar: Well, strictly speaking it also works for *everything else*. Functions, interfaces, interface methods can all be declared with additional qualification.

Melvar: And interface methods can be declared with a different namespace than their parent interface. So at the moment, it’s all uniform.

david-christiansen:
I'm going to go ahead and merge this, and we can look in to the constructor-naming issue in the near future.

PR #3029, opened 2016-03-09
Fix example with lower case parameter.

xash:


enolan:
This is probably not the style we want to encourage. Types and functions that return types are generally uppercase, everything else is lowercase. The preferable fix is keeping `testVec` as it is, and changing `inVect`'s type signature to `IsElem 5 Main.inVect`. Ideally we'd have a sentence or two explaining why this is necessary, referencing the above section about implicits and lower vs upper case.

edwinb:
Closing since #3034 fixes this in the preferred way. Thanks for drawing our attention to it though, @xash

PR #3033, opened 2016-03-10
Move Default from Effects to Prelude

jfdm:
The `Default` interface has uses outside of the effects library. This
commit moves the implementatation from `Effects` and redistributes it
accordingly among `Prelude` and `Base`.

I've tagged this as WIP as when I compiled this locally I get an error compiling base:

```
idris --build base.ipkg
Type checking ./System.idr
When attempting to perform error reflection, the following internal error occurred:
        No such variable Language.Reflection.UNThis is probably a bug. Please consider reporting it at https://github.com/idris-lang/Idris-dev/issues
```

I want to see if this is my issue or something else, before I think about merging this.

### DIFF
--- a/libs/base/Data/Vect.idr
+++ b/libs/base/Data/Vect.idr
@@ -527,6 +527,16 @@ implementation Show a => Show (Vect n a) where
     show = show . toList
 
 --------------------------------------------------------------------------------
+-- Default
+--------------------------------------------------------------------------------
+
+implementation Default a => Default (Vect n a) where
+    default = mkDef _ where
+      mkDef : (n : Nat) -> Vect n a
+      mkDef Z = []
+      mkDef (S k) = default :: mkDef k
+
+--------------------------------------------------------------------------------
 -- Properties
 --------------------------------------------------------------------------------
 
@@ -584,7 +594,7 @@ noEmptyElem : {x : a} -> Elem x [] -> Void
 noEmptyElem Here impossible
 
 Uninhabited (Elem x []) where
-  uninhabited = noEmptyElem 
+  uninhabited = noEmptyElem
 
 ||| An item not in the head and not in the tail is not in the Vect at all
 neitherHereNorThere : {x, y : a} -> {xs : Vect n a} -> Not (x = y) -> Not (Elem x xs) -> Not (Elem x (y :: xs))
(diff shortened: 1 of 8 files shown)


david-christiansen:
I personally don't think `Default` belongs in `Prelude`, as it's not clear that the same default makes sense for every context. At most, I think it should be an optional import from `base`, and perhaps even live in `contrib`. Really, though, I think each library that has a notion of defaultness should probably specify it separately.

jfdm:
@david-christiansen Just to clarify, you believe that having `Default` in the `Prelude` is just to restrictive in terms of providing blanket defaults. Your suggestion in the first instance is to move this to base, and make it an optional import.

What are your thoughts, on separating placement of the implementation from the interface. That is, we keep in the prelude (or another reasonable place) the interface, and within base we provide a set of sane 'default' implementations that if defaults are required they would be required to be imported. 

This allows a library to:

a) see that a 'Default' interface is available; and 
b) have the option of use pre-existing defaults or their own set.

david-christiansen:
Actually, I don't think that would work. Let's say I want Library A and Library B. Each of them wants to use something like Default, but the defaults that make sense in A are not those that make sense in B. If A then imported this collection of pre-existing defaults, and B defined its own, then there would be overlapping implementations visible to client program C that imports A and B, leading to unpredictable behavior. That's why I think that A and B should define their own, separate notions of defaultness rather than attempting to share one.

For instance, which of the following is correct:
```
implementation Default a => Default (Maybe a) where
  default = Just default
```
or
```
implementation Default (Maybe a) where
  default = Nothing
```
?  They both seem reasonable, and each might be correct for a different application.

david-christiansen:
Or even

```
implementation Applicative f, Default a => Default (f a) where
  default = pure default 
```

edwinb:
I'm not even sure Default belongs in Effects - it's a bit of a hack to make it easier to run effectful programs with the unit resource. I'd prefer to find a neater way of making effects run than moving Default to the Prelude.

jfdm:
Okay, I am partially convinced for holing off this PR for now.

david-christiansen:
@jfdm, can I close this then? It can always be reopened if we change our minds, after all.

jfdm:
I shall close it, no use having this PR pollute the PR list.

david-christiansen:
thanks :-)

Issue #2990, opened 2016-02-25
How to apply the `private`, `export` and `public export` accessibility qualifiers?


nicolabotta:
This issue is a follow-up to #2985. The idea is to achieve some shared understanding of how the new `private`, `export` and `public export` accessibility qualifiers can be applied -- possibly in conjunction with `%hide` and `%freeze` directives and with qualified `import public` rules -- to control the accessibility of features of imported modules. I will discuss an example that I think is paradigmatic. Consider the `Foo` module
```
> module Foo

> %default total
> %access public export

> data Foo : Nat -> Type where
>   MkFoo  : {n : Nat} -> Foo (S n)
```
together with its `FooOps` and `FooProps` library modules
```
> module FooOps

> import Foo

> %default total
> %access public export
> -- %access export

> fromSucc : (m : Nat) -> (n : Nat) -> S m = n -> Foo n
> fromSucc m n prf = s2 where
>   s1 : Foo (S m)
>   s1 = MkFoo
>   s2 : Foo n
>   s2 = replace prf s1

> toNat : Foo n -> Nat
> toNat {n} _ = n
```
,
```
> module FooProps

> import Foo
> import FooOps

> %default total
> %access export

> toFromLemma : (m : Nat) -> (n : Nat) -> (p : S m = n) -> toNat (fromSucc m n p) = n 
> toFromLemma m n p = Refl
```
and the application-level program
```
> module Bar

> import Foo
> import FooOps

> %default total

> foo : Foo 1
> foo = fromSucc Z (S Z) Refl
```
The question is: **should `FooOps.fromSucc` and `FooOps.toNat` be qualified as `private`, `export` or `public export`?** Before trying to answer this question, let's observe that:

1) `FooProps` needs to access the implementation of `FooOps.fromSucc` and `FooOps.toNat` for `toFromLemma` to type check.

2) `Bar` does not need to (and, for reasons that I do not want to discuss here, should not) access  the implementation of `FooOps.fromSucc` and `FooOps.toNat`.

Thus, **in absence of other mechanisms** for controlling the accessibility of `FooOps.fromSucc` and `FooOps.toNat` in `FooProps` and `Bar`, **we have to do with a specification that we cannot fulfill: `FooOps.fromSucc` and `FooOps.toNat` have to be qualified as `public export` because of 1) and they have to be qualified as `export`because of 2)!**

[...]

PR #3035, opened 2016-03-11
Initital commit of tutorial on visibility modifiers.

jfdm:
Issue #2990 requests further explaination of how to use Idris'
visibility modifier.  This commit introduces the initial version of a
tutorial to do this.

### DIFF
--- a/docs/guides/index.rst
+++ b/docs/guides/index.rst
@@ -20,3 +20,4 @@ Tutorials submitted by community members.
 
    type-providers-ffi
    theorem-prover
+   visibility-modifiers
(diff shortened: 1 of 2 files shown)


shlevy:
Hmm, I'm not sure it's a good idea to show off `%access export` as the first introduction to exporting an interface. That goes against the purpose of keeping things private by default.

jfdm:
Although, the 'purpose of keeping things private by default'. Is not mentioned directly it still nonetheless provides a solution, and explains the reasons for the decision.

It might be better here to add a note that explains the 'private by default' approach, and how the solution presented keeps to it. Further, one can add an alternate solution in which you mark every thing that is to be exported explicitly with the `export` keyword. 

I'll muse on this some more.

jfdm:
Closed as i'm not going to work on it in the near future, anyone is free to take it on if required.

PR #3043, opened 2016-03-17
Namespace the ClosureType enums.

melted:
Some had generic names like INT and FLOAT that caused collisions.


Melvar:
Huh. Isn’t the collision the other side’s fault, since these don’t leak out of an idris program?

melted:
If the idris program pulls in headers that uses INT and FLOAT there will be compilation errors, as happened to me.

PR #3046, opened 2016-03-18
Comments to Lang.hs to disambiguate the LIR.

xekoukou:
Currently it wasn't clear as to which LIR the backends were getting because internally we use the same data type before and after the lifting process.


jfdm:
Cheers. Having more accurate and clarified code comments are always welcome.

PR #3050, opened 2016-03-20
Make building on Windows a bit easier

enolan:


melted:
I thought only l33t hax0rs should build it :-)

Thatnks!

melted:
I note now that it should probably have a check for cygwin too, but they don't see a degraded experience with this.

PR #3060, opened 2016-03-25
Clarified the formatting in the section on `If`.

jackolantern:


jfdm:
Good catch. Thanks for this.

PR #3064, opened 2016-03-26
Fixes #2932 with better documentation.

### LINKED ISSUE #2932: Custom install and search paths for `idris --build` and `idris --install`
So you could install a package into a custom dir, and then use that dir to build an another package.

jfdm:
The env vars used by idris were not well documented, nor how to
interact with paths and these vars from the Idris CLI. This commit
updates the documentaton.

### DIFF
--- a/docs/reference/compilation.rst
+++ b/docs/reference/compilation.rst
@@ -2,8 +2,7 @@
 Compilation and Logging
 ***********************
 
-
-This section provides highlights of the Idris compilation process, and
+This section provides information about the Idris compilation process, and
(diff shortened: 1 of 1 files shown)


docs/reference/compilation.rst
```diff
@@ -2,8 +2,7 @@
 Compilation and Logging
 ***********************
 
-
-This section provides highlights of the Idris compilation process, and
+This section provides information over the Idris compilation process, and
```
Melvar: I believe that should be `about` rather than `over`.

PR #3067, opened 2016-03-26
Use type/term pun gotten from parsing to elaborate dependent pairs

ahmadsalim:
Instead of the current approach where this is guessed by type being placeholder.

This fixes the issue I discussed in #2636 .

### DIFF
--- a/src/Idris/Elab/Term.hs
+++ b/src/Idris/Elab/Term.hs
@@ -413,14 +413,15 @@ elab ist info emode opts fn tm
                                                  pimp (sUN "B") Placeholder False,
                                                  pexp l, pexp r])
     elab' ina _ (PDPair fc hls p l@(PRef nfc hl n) t r)
-            = case t of
-                Placeholder ->
+            = case p of
+                IsType -> asType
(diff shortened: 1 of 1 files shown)


david-christiansen:
Neat! But please note that unless something has changed since I added it, the pun information does not come from parsing. It is only filled out when the delaborator produces a `PTerm` from a `TT`. So I think there may be some dead code here.

ahmadsalim:
@david-christiansen I added it in parsing :).

See here: https://github.com/idris-lang/Idris-dev/pull/2636/files#diff-b14a7a874e9d59c29a9ad4777f5419a7R494 .

Basically, I add it when I know something is going to be a type or a term (because it can't be parsed otherwise). 

E.g. `(a : _ ** n ** Vect n a)` and `(a : Nat ** a = a)` are known to be types, whereas `(S Z ** [1])` is known to be a term (because it has a non-name on the left). Things like `(a ** a = a)` are still ambiguous and resolved as usual.

david-christiansen:
Ah, nice!

PR #3071, opened 2016-03-27
Fixes #1448, Fix cli and pkg options

### LINKED ISSUE #1448: Behaviour when working with packages and setting options.

jfdm:
When building packages a preapproved command line options (if
list such that they. The allowed options are:

+ Logging levels and categories
+ Warn partial
+ Warn reach
+ Location of IBC subdir
+ Location of import dir.

Idris also reports an error if invalid command lines options are
specified when used to override package specified options. This error
message detaisl the offending flag.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,12 @@ New in 0.12:
 * Improved startup performance by reducing the processing of an already imported
   module that has changed accessibility.
 
+* A limited set of command line options can be used to override
+  package declared options. Overridable options are currently,
+  logging level and categories, default totality check, warn reach,
+  IBC output folder, and idris path.
+
+
 New in 0.11:
 ============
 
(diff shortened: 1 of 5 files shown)


jfdm:
i may want to rewrite that commit message before it gets merged...

jfdm:
So this failed due to the options merging code not taking into account the `codegen` option. I will fix this later tonight, or later this week.

PR #3072, opened 2016-03-27
Add generated doc dirs to .gitignore

melted:


jfdm:
Good catch. I think we should accept this now, but a better future proof solution would be a glob with syntax that ignores all generated doc folders in `libs`. For example: `libs/*/*_doc/`

PR #3080, opened 2016-03-30
Fixes #3075 by adding generalised 'Project' output type.

### LINKED ISSUE #3075: Backend specific flags don't belong in the Idris CLI.
When I was getting #3074 ready I noticed that there are still some backend specific options in Idris' options parser.

For instance:

+ `--mvn` is used to construct a maven project for the Java codegen.
+ `--target` and `--cpu` are used for the the llvm backend.

These backend are not distributed with Idris, and no other backend has such top level flags in the Idris CLI. I think these should be moved to their respective backends, and if need to be called passed using the option `--cg-opt` which passes arguments to the code generator.

jfdm:
The code generator output type `MavenProject` was specific to the Java
codegen. This output type has been generalised to `Project` such that
if a codegen wants to output a 'project' it can do so. An activation
flag `--projectonly` has been added. The existing `--mvn` flag still
exists for backwards compatibility, and will be deprecated in a future
release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,13 @@
 * Remove deprecated options `--ideslave` and `--ideslave-socket`. These options
   were replaced with `--ide-mode` and `--ide-mode-socket` in 0.9.17
 
+* The code generator output type `MavenProject` was specific to the
+  Java codegen. This output type has been generalised to `Project`
+  such that if a codegen wants to output a 'project' it can do so. An
+  activation flag `--projectonly` has been added. The existing `--mvn`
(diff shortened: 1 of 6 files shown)


david-christiansen:
I'm not a big fan of this kind of argument being in the front-end. Why not eliminate the argument instead, and then let --cgopts handle it?

jfdm:
I am actually in agreement with your suggestion. In fact that was my original intention. However, I thought I would take the middle ground, and preserve the 'project' output type.

jfdm:
Closing as #3086 is better.

PR #3081, opened 2016-03-30
Add getEntry to each processing function

melted:
To get better accounting and allow them to specialise

Change the names to something less cryptic


melted:
I'm going to try to optimise the start-up performance. But this PR is just refactoring and renaming to clean up beforehand.

Issue #3068, opened 2016-03-26
Replace custom list predicate `EffElem` in Effects with `Elem` from Data.list


jfdm:
As noted in the conversation, the implementation in Issue/PR #2864 can be simplified by replacing `EffElem` with `Elem` from `Data.List`. This issue is just a reminder for it to be done.

PR #3090, opened 2016-04-02
Replace `EffElem` with `Data.List.Elem`

ericqweinstein:
Proposed fix for https://github.com/idris-lang/Idris-dev/issues/3068. @jfdm, is this what you were thinking?

### DIFF
--- a/libs/effects/Effects.idr
+++ b/libs/effects/Effects.idr
@@ -3,6 +3,7 @@ module Effects
 import Language.Reflection
 import public Effect.Default
 import Data.Vect
+import public Data.List
 
 --- Effectful computations are described as algebraic data types that
 --- explain how an effect is interpreted in some underlying context.
@@ -89,7 +90,7 @@ public export
 data SubElem : a -> List a -> Type where
   Z : SubElem a (a :: as)
   S : SubElem a as -> SubElem a (b :: as)
-  
+
 public export
 data SubList : List a -> List a -> Type where
   SubNil : SubList [] xs
@@ -142,13 +143,6 @@ namespace Env
     Nil  : Env m Nil
     (::) : Handler eff m => a -> Env m xs -> Env m (MkEff a eff :: xs)
 
-namespace EffElem
-  public export
-  data EffElem : Effect -> Type ->
-                 List EFFECT -> Type where
-    Here : EffElem x a (MkEff a x :: xs)
-    There : EffElem x a xs -> EffElem x a (y :: xs)
-    
 total envElem : SubElem x xs -> Env m xs -> Env m [x]
 envElem Z (x :: xs) = [x]
 envElem (S k) (x :: xs) = envElem k xs
@@ -158,7 +152,7 @@ total dropEnv : Env m ys -> SubList xs ys -> Env m xs
 dropEnv [] SubNil = []
 dropEnv [] (InList idx rest) = absurd idx
 dropEnv (y::ys) SubNil = []
-dropEnv e@(y::ys) (InList idx rest) = 
+dropEnv e@(y::ys) (InList idx rest) =
   let [x] = envElem idx e
   in x :: dropEnv e rest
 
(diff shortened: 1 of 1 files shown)


ericqweinstein:
Taking a look later today:

```
Effects.idr:194:52:When checking type of Effects.updateResTy:
Type mismatch between
        Type (Type of Elem _ _)
and
        _ -> _ (Is Elem _ _ applied to too many arguments?)
```

ericqweinstein:
@jfdm Were you looking to replace `EffElem` with `Data.List.Elem` entirely, or redefine the former in terms of the latter? (I guess I'm a bit confused because `EffElem : Effect -> Type -> List EFFECT -> Type`, whereas `Data.List.Elem : a -> List a -> Type`.)

ahmadsalim:
@ericqweinstein I believe it should generally be something in the form of `Elem (MkEff a e) xs`, because that is what the `Here` constructor in `EffElem` checked for. That is maybe why you are currently experiencing issues with compiling things.

ericqweinstein:
@ahmadsalim That makes perfect sense—thanks!

libs/effects/Effects.idr
```diff
@@ -346,7 +340,7 @@ new = New
 -- ---------------------------------------------------------- [ an interpreter ]
 
 private
-execEff : Env m xs -> (p : EffElem e res xs) ->
+execEff : Env m xs -> (p : Data.List.Elem (MkEff a e) xs) ->
```
ahmadsalim: This should be `MkEff res e`, which is why the test fails.

ericqweinstein: My mistake, thanks for catching! Should have this ready later today.

ericqweinstein:
I'm now seeing:

```
./Effect/Exception.idr:32:7:
When checking right hand side of raise with expected type
        EffM m b [EXCEPTION a] (\v => [EXCEPTION a])
No such variable Data.List.Elem
```

Which suggests to me a failure to type check (as `Data.List.Elem` [does exist](https://github.com/idris-lang/Idris-dev/blob/master/libs/base/Data/List.idr#L8)), but I'm not totally sure where I'm afield here. (I'm very new to Idris, so apologies for taking so long with this one.)

ahmadsalim:
@ericqweinstein Hmm, I am unsure why this happens.
Is there any reason why you don't just use `Elem`?

PS: It is totally fine that you take some time on this :). Although, please do compile your code locally before pushing to catch errors quicker.

ahmadsalim:
@ericqweinstein So apparently, it seems that you need to add `import Data.List` in the top of `Effect.Exception` for this to work (and all other files that uses the particular public export function).

jfdm:
You are best to set the import to:

```
import public Data.List
```

This will save you having to add it to all the other effects files.

ericqweinstein:
@ahmadsalim @jfdm Thanks! I'll make the import public and make sure things look good after compiling locally before pushing up again. As for the full qualification, I'd wanted to avoid conflicts with `Data.Vect.Elem` (see [here](https://github.com/idris-lang/Idris-dev/blob/master/libs/base/Data/Vect.idr#L578)), but if there's a preferred way of specifying `Data.List.Elem` without repeating myself each time, I'm happy to do that.

ericqweinstein:
Okay, `effects/` type checks and compiles. Pushing up!

ahmadsalim:
@ericqweinstein Thanks for the contribution!

ahmadsalim:
@ericqweinstein One small request though, would it be possible to investigate whether one could use `Elem` unqualified instead of `Data.List.Elem` everywhere?

I think the code would like nicer without qualification.

ericqweinstein:
@ahmadsalim Of course! I'll do that today.

ericqweinstein:
@ahmadsalim What do you think of `with List Elem` as a solution here?

ahmadsalim:
@ericqweinstein If plain `Elem` does not work then what you have now is fine.

ahmadsalim:
Fixes #3068

ericqweinstein:
@ahmadsalim Yeah, there was a conflict between `Data.Vect.Elem` and `Data.List.Elem`. Thanks for merging!

ahmadsalim:
@ericqweinstein Thank you!

Issue #1235, opened 2014-05-16
Release Patterns


jfdm:
From the meta discussion from the idris developer meeting it was decided that we should be more transparent over releases:

- [ ] A document needs to be created that describes the release process.
- [ ] Leverage the release milestones and issues features on GitHub to track the status of the project for release.
- [ ] Opensource the current ad hoc backwards compatibility checking script.
- [x] As part of the release updated versions of idris-doc generated HTML documentation for the prelude and default library should be made available.

jfdm:
I am assigning this issue as a milestone for the next version of Idris 'v0.9.18' this should be significant time to address the 'issues' affecting release patterns.

jfdm:
I believe salient issues have been addressed.

PR #3093, opened 2016-04-04
Addresses #1235 Add release checklist.

jfdm:
Adds a nomative checklist for the release process.

### DIFF
--- a/RELEASE-CHECKS.md
+++ b/RELEASE-CHECKS.md
@@ -1,0 +1,42 @@
+# Release Checks
+
+Things to be checked, or done before a release is made.
+
+## Final Checks
+
++ [ ] Do the tests run to completion?
++ [ ] Can samples build?
++ [ ] Is `man/idris.1` up-to-date?
+
+## Version Numbers
+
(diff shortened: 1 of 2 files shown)


justjoheinz:
If someone of the "Releasers" has a Mac it would be awesome to update MAc homebrew in one go as well. If you need to know what to do, just ping me. Currently I have been doing that quite frequently if I was aware that a new version is available. Subversions are rarely announced on the mailing list though (which would be another bullet point on the checklist)

RELEASE-CHECKS.md
```diff
@@ -0,0 +1,40 @@
+# Release Checks
+
+Things to be checked or done before a release is made.
+
+## Final Checks
+
++ [ ] Doe the tests run to completion?
```
ericqweinstein: V. small, s/`doe`/`do`/

jfdm:
Typo's addressed. Although, this list shouldn't be seen as a strict barrier to releasing a version. It does provide a note of what needs updating for a release. Take for example: version numbers and their location.

PR #3100, opened 2016-04-08
Fixes #3023 with improved error message.

### LINKED ISSUE #3023: Improve error message "Needs reloading"
Sometime idris complains with the message `Needs reloading ...` and two timestamps, which is defined in `IBC.hs timestampOlder`. From a look at the function it seems to be easy to provide a path of the idris file in question and maybe even give a more explicit hint how to fix the issue. I would PR myself, but I haven't got a working idris developer build.

jfdm:
Improves error message for stale IBC files by adding the source path
to the error message.

### DIFF
--- a/src/Idris/IBC.hs
+++ b/src/Idris/IBC.hs
@@ -394,11 +394,17 @@ process reexp phase archive fn = do
                 processAccess reexp phase archive
 
 timestampOlder :: FilePath -> FilePath -> Idris ()
-timestampOlder src ibc = do srct <- runIO $ getModificationTime src
-                            ibct <- runIO $ getModificationTime ibc
-                            if (srct > ibct)
-                               then ifail $ "Needs reloading " ++ show (srct, ibct)
-                               else return ()
+timestampOlder src ibc = do
+  srct <- runIO $ getModificationTime src
+  ibct <- runIO $ getModificationTime ibc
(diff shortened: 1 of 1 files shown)


ericqweinstein:
@jfdm FWIW, this looks good to me!

jfdm:
I have yet to test this, but regardless, I wish to clean up the error message through better formatting, and also showing the IBC file location. May help in future if Idris is using a 'stale' ibc file.

src/Idris/IBC.hs
```diff
@@ -394,11 +394,17 @@ process reexp phase archive fn = do
...
-                               else return ()
+timestampOlder src ibc = do
+  srct <- runIO $ getModificationTime src
+  ibct <- runIO $ getModificationTime ibc
+  if (srct > ibct)
+    then ifail $ unlines [ "Module needs reloading:"
+                         , unwords ["\tSRC :", show src]
+                         , unwords ["\tModified at:", show srct]
+                         , unwords ["\tIBC :", show src]
+                         , unwords ["\tModified at:", show srct]
```
Melvar: The IBC lines also refer to `src` and `srct`, and `ibc` and `ibct` are not output. Is this a copypaste error?

jfdm: arg, yes

PR #3103, opened 2016-04-09
Makes Network.Socket functions public

ericqweinstein:
Since functions are now package private by default, we need to update older libraries in order to ensure they expose the functionality they're intended to. Please review this commit to ensure nothing that should be private has been exposed!

The below PoC fails with current Idris:

```idr
module Main

import Network.Socket

-- Hacky helper function
fromRight : Either a b -> b
fromRight (Right x) = x

runConn : (Socket, SocketAddress) -> IO ()
runConn (sock, _) = do
  send sock "Hello, world!\n"
  close sock

mainLoop : Socket -> IO ()
mainLoop sock = do
  conn <- accept sock
  runConn $ fromRight conn
  mainLoop sock

main : IO ()
main = do
  sock <- socket AF_INET Stream 0
  bind (fromRight sock) (Just (IPv4Addr 127 0 0 1)) 4242
  putStr "Idris listenin' on 4242...\n"
  listen $ fromRight sock
  mainLoop $ fromRight sock
```

```sh
λ idris -p contrib socket.idr -o socket
socket.idr:9:9:When checking type of Main.runConn:
When checking argument A to type constructor Builtins.Pair:
        No such variable Socket
```

But works with the changes in this PR (though it does generate a warning I'm looking into):

```sh
λ ~/.cabal/bin/idris -p contrib socket.idr -o socket
/var/folders/bl/d5lscdms4bl_2npdn_20wc6h0000gn/T/idris94571.c:438:5: warning: implicit declaration of function 'close' is invalid in C99
      [-Wimplicit-function-declaration]
    close(GETINT(LOC(2)));
    ^
1 warning generated.
λ ./socket
Idris listenin' on 4242...
```

(In another shell)

```sh
λ curl 127.0.0.1:4242
Hello, world!
```

### DIFF
--- a/libs/contrib/Network/Socket.idr
+++ b/libs/contrib/Network/Socket.idr
@@ -8,9 +8,11 @@ module Network.Socket
 %include C "sys/socket.h"
 %include C "netdb.h"
 
-public export ByteLength : Type
+public export
+ByteLength : Type
 ByteLength = Int
 
+export
 interface ToCode a where
   toCode : a -> Int
 
@@ -18,6 +20,7 @@ interface ToCode a where
 |||
 ||| The ones that people might actually use. We're not going to need US
 ||| Government proprietary ones.
+public export
 data SocketFamily =
   ||| Unspecified
   AF_UNSPEC |
@@ -26,11 +29,13 @@ data SocketFamily =
   |||  IP / UDP etc. IPv6
   AF_INET6
 
+export
 implementation Show SocketFamily where
   show AF_UNSPEC = "AF_UNSPEC"
   show AF_INET   = "AF_INET"
   show AF_INET6  = "AF_INET6"
 
+export
 implementation ToCode SocketFamily where
   toCode AF_UNSPEC = 0
   toCode AF_INET   = 2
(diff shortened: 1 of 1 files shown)


ericqweinstein:
(The C code is just missing a function prototype for `close`, I'm just not yet sure where it is.)

ahmadsalim:
I think a lot of the implementations of type classes could be made `export` only.

ahmadsalim:
Actually, you should consider making things `export` only, since their implementations are often not needed outside the module (e.g. for theorem proving).

ahmadsalim:
I think the only things that may need to be `public export` are the datatypes containing things like `AF_NET` which you expect to use outside the module.

ericqweinstein:
@ahmadsalim How's this?

ahmadsalim:
@ericqweinstein I think the exports now seem to be fine.

A very small request: could it be possible to have a newline between the visibility and the method signature? It would make things look a bit more aligned.

ahmadsalim:
Great, thanks!

PR #3106, opened 2016-04-09
Added to_C to Prelude.File

colin-adams:
To expose a file handle to be passed to C routines


colin-adams:
Cancelled. It's not sufficient as I would also need to get a File from the Ptr.

PR #3111, opened 2016-04-12
added calloc to CFFI.Memory

colin-adams:
This fixes some strange bugs I was having with large POSTs, as well as it's primary purpose, which is to reduce the number of times I have to write C code rather than CFFI


melted:
Apparently @edwinb has messed up the js tests again.

melted:
In it goes! Thanks!

PR #3112, opened 2016-04-13
Fixes #2049

### LINKED ISSUE #2049: Use travis to check Documentation build.
We now have sphinx based documentation. In the current pipeline for documentation generation mistakes in sphinx builds for html and pdf only appear once the docs are passed onto readthedocs. Thus mistakes in reST format will only be picked up post merge into master. 

```
Should we add sphinx build testing to travis to pick up faults pre merge?
```

jfdm:
These changes augments the build setup to check the sphinx documentation.

The changes are based upon the configuration used by [PyMSQL
project](https://github.com/pymssql/pymssql).

A problem with this approach is that the matrix for checking the docs
requires Idris to be built. The travis build could be made more modal
so that only the sphinx docs are built.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,8 @@ matrix:
       compiler: ": #GHC 7.10.2"
     - env: CABALVER="1.22" GHCVER="7.10.2" TESTS="test_c"
       compiler: ": #GHC 7.10.2"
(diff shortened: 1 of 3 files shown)


jfdm:
After thinking about this. I do not think it is necessary to build Sphinx docs in travis. Read the docs provides a suitable alternative.

PR #3126, opened 2016-04-23
Added slightly to the C FFI reference, and corrected a code example i…

colin-adams:
…n the type providers tutorial

### DIFF
--- a/docs/tutorial/miscellany.rst
+++ b/docs/tutorial/miscellany.rst
@@ -276,7 +276,7 @@ type ``Int`` will be provided. Otherwise, it will provide the type
     strToType _ = Nat
 
     fromFile : String -> IO (Provider Type)
-    fromFile fname = do str <- readFile fname
+    fromFile fname = do Right str <- readFile fname
                         return (Provide (strToType (trim str)))
 
 We then use the ``%provide`` directive:
--- a/docs/tutorial/miscellany.rst
+++ b/docs/tutorial/miscellany.rst
@@ -277,8 +277,8 @@ type ``Int`` will be provided. Otherwise, it will provide the type
 
     fromFile : String -> IO (Provider Type)
     fromFile fname = do Right str <- readFile fname
-		| Left err => pure (Provide Void)
-	pure (Provide (strToType (trim str)))
+		          | Left err => pure (Provide Void)
+		        pure (Provide (strToType (trim str)))
 
 We then use the ``%provide`` directive:
 
--- a/docs/tutorial/miscellany.rst
+++ b/docs/tutorial/miscellany.rst
@@ -277,7 +277,8 @@ type ``Int`` will be provided. Otherwise, it will provide the type
 
     fromFile : String -> IO (Provider Type)
     fromFile fname = do Right str <- readFile fname
-                        return (Provide (strToType (trim str)))
+		| Left err => pure (Provide Void)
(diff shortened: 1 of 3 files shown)


docs/tutorial/miscellany.rst
```diff
@@ -276,7 +276,7 @@ type ``Int`` will be provided. Otherwise, it will provide the type
     strToType _ = Nat
 
     fromFile : String -> IO (Provider Type)
-    fromFile fname = do str <- readFile fname
+    fromFile fname = do Right str <- readFile fname
```
ahmadsalim: I think it might be a good idea to handle the error like in https://github.com/idris-lang/Idris-dev/blob/b3fdc3e2dc6256b9cb03e4cee89edb1dd97b081d/test/ffi004/test026.idr#L14 .

docs/tutorial/typesfuns.rst
```diff
@@ -860,12 +860,12 @@ Dependent Pairs
 
 Dependent pairs allow the type of the second element of a pair to depend
 on the value of the first element. Traditionally, these are referred to
-as “sigma types”:
+as “sigma types”, so in older code you might see Sigma rather than DPair:
```
ahmadsalim: It may be a good idea to just remove this sentence instead of referring to older code.

colin-adams: Well, I was thinking of the beginner who encounters Sigma or MkSigma, and is puzzled. I'm familiar with Pi and Sigma types, but many newcomers to the language won't be. That's why I chose to add rather than remove stuff here.
I can't see much use of DPair compared with Sigma when i look at current code.

ahmadsalim: Sigma is deprecated, but I see your point. Maybe this could be rephrased so that that is more apparent (that it is deprecated now).

jfdm: I think the problem with older code is that they have yet to migrate explicit instances of `Sigma` and `MkSigma` across to the new names. If the dependent pair sugar is used then it doesn't matter.

We should really deprecate fully `Sigma` in a couple of releases. But I am in favour of removing the reference to old code. Maybe we should promote the wiki page for 'backwards breaking' changes to the reference guide...

docs/tutorial/miscellany.rst
colin-adams: Accepted, and coded - new push to follow soon.

docs/tutorial/typesfuns.rst
colin-adams: OK, I removed that reference.

docs/tutorial/miscellany.rst
```diff
@@ -276,8 +276,9 @@ type ``Int`` will be provided. Otherwise, it will provide the type
     strToType _ = Nat
 
     fromFile : String -> IO (Provider Type)
-    fromFile fname = do str <- readFile fname
-                        return (Provide (strToType (trim str)))
+    fromFile fname = do Right str <- readFile fname
+		| Left err => pure (Provide Void)
+	pure (Provide (strToType (trim str)))
```
ahmadsalim: I think this would look better as:

```idris 
 fromFile fname = do Right str <- readFile fname
                                      | Left err => pure (Provide Void)
                                  pure (Provide (strToType (trim str)))
```

ahmadsalim:
Thanks for the PR!

PR #3129, opened 2016-04-24
Fix string to positive number conversion

eklavya:
Currently "-123" is being parsed as Just 0 for Nats, which is wrong. It should return Nothing for non numbers and negative numbers.


ahmadsalim:
Great, thanks!

mbj:
Nice, that behavior always put me off before.

PR #3130, opened 2016-04-26
Enum for bits

colin-adams:
I encountered an error message in code that I was writing that there was no Enum implementation for Bits8. It seemed easier to implement this than rewrite the code, so ...

### DIFF
--- a/docs/reference/ffi.rst
+++ b/docs/reference/ffi.rst
@@ -179,7 +179,8 @@ compilation error:
     example f = foreign FFI_C "callbacker" (CFnPtr (Int -> ()) -> IO ()) f
 
 The other big limitation is that it doesn't support IO functions. Use
-``unsafePerformIO`` to wrap them.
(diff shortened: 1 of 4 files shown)


jfdm:
@colin-adams thanks for the PR, however, it looks like you have included commits from #3126. Is it possible to separate the duplicate commits? Either than, or we can deprecate #3126, and this PR supersedes #3126.

ahmadsalim:
I prefer if the commits were separate, since they concern different areas.

colin-adams:
Oh. I meant to have branched off master. I'll throw this away and start again. Sorry.

jfdm:
Apologies not necessary, been there, done that, have the t-shirt collection. Just push to the same branch and everything can carry on as normal. Time can be rewritten.

ahmadsalim:
Yeah, don't worry. 
Also, I think that @jfdm meant the last part as in git history can be rewritten (and not philosophically).

PR #3131, opened 2016-04-26
Added Enum implementations for Bits8/16/32/64

colin-adams:


libs/prelude/Prelude.idr
```diff
@@ -214,6 +214,30 @@ Enum Char where
 
   pred c = fromNat (pred (toNat c))
 
+Enum Bits8 where
+  toNat b   = fromIntegerNat (prim__zextB8_BigInt b)
```
ahmadsalim: Thanks for the code.

I think the use of `prim__*` operators in real code should be avoided if possible.
Is it not possible to just use `fromInteger` for the `fromNat` one instead of `prim_trunc*`.

I don't know what the corresponding functions are in the other directions, but maybe you could make relevant cast instances.

colin-adams: I'll take a look.

ahmadsalim: Thanks!

colin-adams: Yes. 
That was easier than I thought. I always get confused by fromInteger.

ahmadsalim: There seems to be a `bitsToInt'` method in the `Data.Bits` module.

ahmadsalim: Actually, have you considered just making the instance work with `Bits n` type in `Data.Bits`?
Maybe that is easier/more useful, but I will leave it up to you.

colin-adams: I (mistakenly, I think) initially tried to create an instance for Bits n, thinking this was where Bits8 came from. But it's not. The code I was writing needed an instance of Bits8, not Bits 8.

colin-adams: Another point - Bits8/16/32/64 are defined in Prelude.idr. So if the instances are to be in the same location, they cannot depend on Data.Bits (base package).

Issue #3083, opened 2016-03-30
After disconecting from wifi, I recieve the error "Unable to connect to a running Idris repl" when using `--client` 


jackolantern:
I am on Arch Linux with kernel at 4.4.5-1

Steps to recreate.
* Connect to wifi
* Start Idris repl
* Run test command `idris --client 1`
* Disconnect from wifi
* Rerun test command and see the error "Unable to connect to a running Idris repl"

If I start a new Idris instance, I see the message that the port is still in use, so the issue is not simply that the server is "dying."  If I bring down the wifi network interface, or reconnect to wifi everything starts to work again.  I glanced at the code for the client command line option, and I don't  see any obvious problems.  Maybe this is a bug with the Haskell library?

LeifW:
I get that too on Arch Linux. I was going to look into the Haskell
networking stuff, or optionally use unix domain sockets or named pipes
instead of TCP/IP where available.

ahmadsalim:
Any Arch Linux user who wants to look into this? And is this something fixable by using named pipes instead of sockets?

PR #3134, opened 2016-04-29
update to date

uuhan:


src/Idris/Parser.hs
```diff
@@ -1569,14 +1570,14 @@ parseProg :: SyntaxInfo -> FilePath -> String -> Maybe Delta ->
 parseProg syn fname input mrk
     = do i <- getIState
          case runparser mainProg i fname input of
-            Failure doc     -> do -- FIXME: Get error location from trifecta
+            Failure ErrInfo{..}     -> do -- FIXME: Get error location from trifecta
```
david-christiansen: In my opinion, `RecordWildCards` makes the code difficult to read, especially when it's a record from an external dependency. It took me a while to realize where `_errDoc` is bound, and that was with the `LANGUAGE` pragma on the same commit reader!

src/Idris/Prover.hs
```diff
@@ -317,8 +318,8 @@ elabloop fn d prompt prf e prev h env
        (d, prev', st, done, prf', env', res) <-
          idrisCatch
            (case cmd of
-              Failure err ->
-                return (False, prev, e, False, prf, env, Left . Msg . show . fixColour (idris_colourRepl ist) $ err)
+              Failure ErrInfo{..} ->
```
david-christiansen: Please don't use `RecordWildCards` in this way with external dependencies.

david-christiansen:
Hi @uuhan,

Thanks for contributing to Idris!

I don't think we can merge this pull request in it's current state, but it looks like it does do something useful, so I would like to see if we can get it into a mergeable condition.

* [ ] Would you please update the pull request with a description of why you made these changes, as well as rewrite the commit messages so that someone reading them can understand what was done and why? It might also make sense to squash the commits in this case, as it all appears to be related to a Trifecta API change.

* [ ] How will this Trifecta API change affect our compatibility with LTS Haskell and/or other Stack snapshots?

* [ ] I also don't think that `RecordWildCards` on an external library is a very good coding style in this case, because it makes the binding site of `_error` very mysterious. Even grepping all of Idris won't show where it comes from!

* [ ] Finally, we don't merge pull requests until the tests pass on Travis.

Thanks for helping keep Idris up to date in the ever-improving world of Hackage :-)

/David

uuhan:
@david-christiansen 
My fault. 
I should not do this pull requst by using the latest  trifecta ( **cabal get trifecta -s**)
The **trifecta** package add a new data structure ErrInfo which enrichs the old error message info. I need to do some source code changes to compile idris.  There is nothing to be changed when using the trifecta package in hackage.

By the way, Idris is not working well in my Archlinux. Problems like in this issue [https://github.com/idris-lang/Idris-dev/issues/3083]. 
I did not do anything to my wifi connection, but it just failed to connect to a running Idris repl.
idris --client 0 always shows "Unable to connect to a running Idris repl"

PR #3139, opened 2016-05-02
Add a lower bound constraint to fsnotify version.

### LINKED ISSUE #3138: Idris fails to build with fsnotify < 0.2
I just tried to install idris-0.11 just by doing `cabal install idris`
And I got the following error:
```
[96 of 99] Compiling Idris.REPL       ( src/Idris/REPL.hs, dist/build/Idris/REPL.o )

src/Idris/REPL.hs:700:22:
    Couldn't match type ‘[Char]’
                   with ‘system-filepath-0.4.13.4:Filesystem.Path.Internal.FilePath’
    Expected type: system-filepath-0.4.13.4:Filesystem.Path.Internal.FilePath
      Actual type: FilePath
    In the second argument of ‘watchDir’, namely ‘dir’
    In the expression:
      watchDir
        mgr
        dir
        (allEvents $ flip S.member inputSet)
        (doAllEvents $ putMVar signal)

src/Idris/REPL.hs:700:53:
    Couldn't match type ‘[Char]’
                   with ‘system-filepath-0.4.13.4:Filesystem.Path.Internal.FilePath’
    Expected type: S.Set
                     system-filepath-0.4.13.4:Filesystem.Path.Internal.FilePath
      Actual type: S.Set FilePath
    In the second argument of ‘flip’, namely ‘inputSet’
    In the second argument of ‘($)’, namely ‘flip S.member inputSet’
```

This seems to be caused by me having fsnotify-0.1.0.3 installed, which uses the now deprecated system-filepath instead of the newer filepath.
Starting with [fsnotify-0.2](http://hackage.haskell.org/package/fsnotify-0.2) it switched to filepath

Just doing `cabal install fsnotify` (which installed fsnotify-0.2.1) fixed the error for me.

But it might be good for idris to explicitly depend on fsnotify >= 0.2

LeifW:
Fixes #3138

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -1072,7 +1072,7 @@ Library
                 , vector-binary-instances < 0.3
                 , zip-archive > 0.2.3.5 && < 0.4
                 , safe
-                , fsnotify < 2.2
+                , fsnotify >= 0.2 && < 2.2
                 , async < 2.2
 
   -- zlib >= 0.6.1 is broken with GHC < 7.10.3


ahmadsalim:
Thanks!

PR #3144, opened 2016-05-04
Changed '%default' to '%totality' with a deprecation warning

ahmadsalim:


ahmadsalim:
We agreed on IRC this change might be a bit hasty and it is better to have an RFC issue to discuss pragmas in general

PR #3147, opened 2016-05-05
Covering

colin-adams:
As written so far - parses, but has no effect.


ahmadsalim:
@colin-adams Could you please rebase your work to not include the `Enum` implementations? Thanks!

Also, consider to use feature branches from master :).

colin-adams:
That's odd. I was careful to check that I was branching from an up-to-date master branch. I don't know how it has gone wrong.

I'll try again.

src/Idris/Parser.hs
```diff
@@ -1396,14 +1396,15 @@ pLangExt = (reserved "TypeProviders" >> return TypeProviders)
...
-Totality ::= 'partial' | 'total'
+Totality ::= 'partial' | 'total' | 'covering'
 @
 
 -}
 totality :: IdrisParser Bool
 totality
         = do reservedHL "total";   return True
       <|> do reservedHL "partial"; return False
+      <|> do reservedHL "covering"; return False
```
ahmadsalim: So your issue is that the result should be changed from a Bool to somekind of Totality enum which is then propagated elsewhere

PR #3150, opened 2016-05-06
Added covering as a possible value for %default

colin-adams:


david-christiansen:
This is great, thanks!

Before I merge this, I'd like to see the test suite extended to make sure that future changes don't break it.

colin-adams:
Any hint as to where I should be looking for the test cases?

david-christiansen:
I think you should write one :-)

The tests are in the `test` directory in the repo. Each test is itself a directory, named after a category. The other tests should serve as a reasonable example.

Once the test is added and works for you, please add it to the Cabal file - look at the others for what to do. This is because the Travis build first does a `cabal srcdist` to make sure that works, and the test should come along there.

colin-adams:
Done, I think.
How do I get the test to run (I've just typed sh ./run and shoved the output into expected so far)

ahmadsalim:
Hi, I think everything looks great!
However, I have a small comment about something that is not obviously linked to this but actually is, that you may want to resolve as well. See my comment on the code!

src/Idris/Parser.hs
```diff
@@ -711,9 +712,10 @@ postulate syn = do (doc, ext)
...
                                    return (doc, ext)
                    ist <- get
-                   let initOpts = if default_total ist
-                                     then [TotalFn]
-                                     else []
+                   let initOpts = case default_total ist of
+                                    DefaultCheckingTotal    -> [TotalFn]
+                                    DefaultCheckingCovering -> [CoveringFn]
+                                    _                       -> []
                    opts <- fnOpts initOpts
```
ahmadsalim: I think the `fnOpts` removes the `initOpts` if an explicit totality is given.
However this now works incorrectly since `CoveringFn` would not be removed if you look at the code.

This hack has seemingly been there for a long time and I think it would be nice if you managed to fix it. I would propose removing `initOpts` from `fnOpts` and instead checking after parsing the options whether there exists an annotation and if not adding the default one (if there exist multiple, it should fail which for some reason or other it doesn't do now).

colin-adams: Where would the check after parsing go?

ahmadsalim: So it would be something like:

```
...
opts <- fnOpts [] -- Ultimately the parameter should be dropped, but I am unsure how for now
acc <- accesbility
opts' <- fnOpts opts
let existingTotality = opts' `intersect` [TotalFn, CoveringFn, PartialFn]
opts'' <- case opts' of
  [] -> 
     case default_total ist of
        DefaultCheckingTotal -> return (TotalFn : opts')
        ... -- The rest of the cases
  [x] -> -- already has one
  _ -> -- has more than one, fail or provide an error message
... -- use opts'' instead of opts' later
```

ahmadsalim: If you want you could also wrap it in a separate function this check.

.gitignore
```diff
@@ -32,6 +32,7 @@ test/output
 test/*[0-9][0-9][0-9]/output
 test/*[0-9][0-9][0-9]/*.exe
 test/runtest.exe
+test/runtest
```
david-christiansen: Hooray!

src/Idris/Parser.hs
colin-adams: I don't understand the code.

ahmadsalim:
OK, I have made a new PR #3151 with my suggested additions.

PR #3154, opened 2016-05-08
Added some needed exports in Network.Socket

gpyh:
Dearly needed, especially `recvFrom` and `sendTo` to use connection-less protocols.


ahmadsalim:
Thanks!

Issue #3170, opened 2016-05-15
Support building idris-0.11.1 on OpenBSD


neduard:
Hello, when attempting to build on OpenBSD, `cabal install` fails with:
```
Downloading idris-0.11.1...
[1 of 1] Compiling Main             ( /tmp/idris-0.11.1-9302/idris-0.11.1/Setup.hs, /tmp/idris-0.11.1-9302/idris-0.11.1/dist/setup/Main.o )
Linking /tmp/idris-0.11.1-9302/idris-0.11.1/dist/setup/setup ...
....
Configuring idris-0.11.1...
*** Parse error in /tmp/idris-0.11.1-9302/idris-0.11.1/rts: Missing dependency operator (../config.mk:1)
*** Parse error: Need an operator in 'endif' (../config.mk:8)
*** Parse error: Missing dependency operator (../config.mk:21)
...
```

Turns out this is caused by using `make` instead of `gmake` in Setup.hs. Adding a check for `defined(openbsd_HOST_OS)` fixes the problem and idris-0.11.1 compiles fine.

PR submitted at #3171

LeifW:
There's also the e.g. `if os(freebsd)` stuff in idris.cabal that defines some CPP options. https://github.com/idris-lang/Idris-dev/blob/master/idris.cabal#L1110

If we have to enumerate all the BSD's separately, maybe we should add NetBSD while we're at it, too?

neduard:
@LeifW Yes, that is a good idea, I have replied to a similar comment on the PR (#3171), basically I'm wondering if the cpp-options can be removed and just use `*_HOST_OS` in `IRTS/System.hs` and `Util/DynamicLinker.hs`.

neduard:
Fixed

PR #3171, opened 2016-05-15
#3170: Add support for OpenBSD and NetBSD

neduard:
PR for issue #3170

### DIFF
--- a/Setup.hs
+++ b/Setup.hs
@@ -38,7 +38,8 @@ idrisCmd local = ".." </> ".." </>  buildDir local </>  "idris" </>  "idris"
 -- Make Commands
 
 -- use GNU make on FreeBSD
-#if defined(freebsd_HOST_OS) || defined(dragonfly_HOST_OS) || defined(openbsd_HOST_OS)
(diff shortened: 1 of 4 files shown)


melted:
Should we put in NetBSD while we're at it?

neduard:
Yes, that is a good idea, I will look into it. Two questions though:
1. Should I change the PR name to "Generalise support for \*BSDs" ?
2. Why do [System.hs](https://github.com/idris-lang/Idris-dev/blob/master/src/IRTS/System.hs#L32) and [DynamicLinker.hs](https://github.com/markuspf/Idris-dev/blob/master/src/Util/DynamicLinker.hs#L28) use the flags `FREEBSD` and `DRAGONFLY` defined in  [idris.cabal](https://github.com/idris-lang/Idris-dev/blob/master/idris.cabal#L1110) instead of the ghc-defined, default, ones of `*_HOST_OS`? Could the `cpp-options:` definitions from `idris.cabal` regarding *BSDs be removed?

melted:
Thanks for the PR!

PR #3173, opened 2016-05-15
Export intermediate representation as JSON

melted:
This adds an option to pass the intermediate representations JSON-encoded to the backend. This makes it possible to build backends in other languages than Haskell.

### DIFF
--- a/codegen/idris-codegen-c/Main.hs
+++ b/codegen/idris-codegen-c/Main.hs
@@ -40,7 +40,7 @@ c_main opts = do runIO setupBundledCC
                  mainProg <- if interface opts
                                 then liftM Just elabMain
                                 else return Nothing
-                 ir <- compile (Via "c") (output opts) mainProg
+                 ir <- compile (Via IBCFormat "c") (output opts) mainProg
                  runIO $ codegenC ir
 
 main :: IO ()
--- a/codegen/idris-codegen-javascript/Main.hs
+++ b/codegen/idris-codegen-javascript/Main.hs
@@ -34,7 +34,7 @@ jsMain :: Opts -> Idris ()
 jsMain opts = do elabPrims
                  loadInputs (inputs opts) Nothing
                  mainProg <- elabMain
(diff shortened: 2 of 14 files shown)


david-christiansen:
This is a great idea!

jfdm:
First off, this is a good idea. Kudos.

However, I have some reservations about the implementation concerning its placement w.r.t. codegen targets and how it is plugged into the compiler. Rather than use a boolean flag can we either:

a) express the JSON output as another codegen target i.e. have `idris-codegen-json`; or alternatively 
b) extend `CodeGen` to provide an additional constructor for portable representations.

Of the two I am more in favour of (a), but am also inclined for (b).

melted:
I don't like a) because it would mean that backends using this wouldn't be
as seamless in use as the current backends. Outputting the json is useful,
and I have a backend that do that, but making Idris unable to run the whole
compilation process would require backends that use this to supply their
own scripts to run it all through. Also it would mean packaging another
huge executable. They compress well, but it feels silly to have a 30 meg
program that does something trivial.

melted:
Well, that's the most absurd Travis failure I've seen.

melted:
I can change the switch to something that parallels `--codegen` and make a new Codegen constructor.

jfdm:
I understand your reservations about a separate JSON codegen executable, but I do not share them. 

However, the second option of extending `CodeGen` might be the best way forward, you may want to parameterise the constructor such that in the event we want to change the format or support multiple formats (e.g. XML or Sexp) this can be achieved more easily.

Alternatively, extending `OutputType`, might be a third option to consider, however this again ties generation of the portable output to the code generator itself rather than in the 'compiler'.

melted:
Thanks for the comments.  It looks cleaner now. Documentation will follow.

jfdm:
TBH, it would have been nicer to review the secondary changes made, prior to merging into master. Primarily to show that there was a review of said changes, and also consensus about the changes. Maybe this is something to take into account for future PRs.

PR #3187, opened 2016-05-24
Change name of Err to Error

tmcgilchrist:
Running through this example I got a name clash with the Err type.
```
- + Errors (1)
 `-- Btree.idr line 61 col 12:
     When checking type of Main.parseNumber:
     Can't disambiguate name: Language.Reflection.Errors.Err, Main.Err
```

Changing the name to Error doesn't seem to have the same issue. I have the following packages loaded in the emacs-idris-repl
```elisp
-- Local Variables:
-- idris-load-packages: ("prelude" "base" "effects")
-- End:

```

And I'm using idris 0.10.

### DIFF
--- a/docs/effects/simpleeff.rst
+++ b/docs/effects/simpleeff.rst
@@ -165,9 +165,9 @@ parameterised by an error type:
 
 .. code-block:: idris
 
-    data Err = NotANumber | OutOfRange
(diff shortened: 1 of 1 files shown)


david-christiansen:
This is nice, thanks!

PR #3196, opened 2016-05-31
Use ParentN for its intended purpose

### LINKED ISSUE #3119: Unused Specialname constructor ParentN
The `ParentN` constructor of `SpecialName` does not seem to be generated anywhere, outside of the deserialization and reflection-reification code. The only clue as to its purpose is its appearance in a line of commented-out code [here](https://github.com/idris-lang/Idris-dev/blob/c2b01ccf5d8e4070000fe665200e7b7d902970a4/src/Idris/Elab/Class.hs#L209-L210). This line appears to have been commented out ever since it was added in [2a0885e](https://github.com/idris-lang/Idris-dev/commit/2a0885e6d40ed7b4a0c88e7923e3a1fa688af82c#diff-89991dd5804bdba2ef2f2461a27c5be7R1183). It looks like it might have been intended for names of superinterface dictionary projections (maybe that’s what “chaser” means?), but I’m not sure of my reading.

The (apparently unused) function `chaser` [here](https://github.com/idris-lang/Idris-dev/blob/c2b01ccf5d8e4070000fe665200e7b7d902970a4/src/Idris/ProofSearch.hs#L427-L433) relates `ParentN` to names beginning with `@@`, which also appear in [a function of the same name](https://github.com/idris-lang/Idris-dev/blob/c2b01ccf5d8e4070000fe665200e7b7d902970a4/src/Idris/AbsSyntax.hs#L431-L434) which does not mention `ParentN` and is used to sort “chasers” after regular implementations.

My opinion is that `ParentN` should either be put to its intended use or be removed.

Melvar:
As far as I can tell, at least. This replaces `UN`s starting with `@@` with `ParentN`, since that’s what it seemed to be for, removing the remaining instance of `UN` being used for an internal name.

Fixes #3119.

### DIFF
--- a/src/Idris/AbsSyntax.hs
+++ b/src/Idris/AbsSyntax.hs
@@ -439,7 +439,7 @@ pop_estack = do i <- getIState
 -- Dodgy hack 1: Put integer instances first in the list so they are
 -- resolved by default.
 --
--- Dodgy hack 2: put constraint chasers (@@) last
+-- Dodgy hack 2: put constraint chasers (ParentN) last
 addInstance :: Bool -- ^ whether the name is an Integer instance
             -> Bool -- ^ whether to include the instance in instance search
             -> Name -- ^ the name of the class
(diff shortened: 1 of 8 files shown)


edwinb:
Great, thanks! Yes, this was the intention. Not sure why I didn't do this at the time, now I come to think of it.

PR #3204, opened 2016-06-06
Extend iPKG files to include metadata.

jfdm:
Taking cues from cabal, the iPKG format has been extended to include more package metadata information.
The following fields have been added:

+ brief: Brief description of the package.
+ version: Version string to associate with the package.
+ readme: Location of the README file.
+ license: Description of the licensing information.
+ author: Author information.
+ maintainer: Maintainer information.
+ homepage: Website associated with the package.
+ sourcepage: Location of the DCVS where the source can be found.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,8 +80,24 @@
 * Added `Data.List.Views` with views on `List` and their covering functions.
 * Added `Data.Nat.Views` with views on `Nat` and their covering functions.
 * Added `Data.Primitives.Views` with views on various primitive types and their covering functions.
-* Added `System.Concurrency.Sessions` for simple management of conversations 
+* Added `System.Concurrency.Sessions` for simple management of conversations
   between processes
+
+## iPKG Updates
+
+* Taking cues from cabal, the `iPKG` format has been extended to
+  include more package metadata information.  The following fields
+  have been added:
+
+  + `brief`: Brief description of the package.
+  + `version`: Version string to associate with the package.
+  + `readme`: Location of the README file.
+  + `license`: Description of the licensing information.
+  + `author`: Author information.
+  + `maintainer`: Maintainer information.
+  + `homepage`: Website associated with the package.
+  + `sourcepage`: Location of the DVCS where the source can be found.
+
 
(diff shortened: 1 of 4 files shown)


Melvar:
What is Idris going to use these for?

jfdm:
@Melvar There are a variety of situations in which such information is useful:

1. Enriching `idrisdoc` with more information.
2. Use in a future package manager.
3. Use by package authors to enrich the description of their packages.

Melvar:
So does the Idris executable need to know about them? Seems like a good opportunity to switch to a format that doesn’t require the parser to know about every field that could possibly crop up.

jfdm:
@Melvar that is the plan. but first lets get people used to a more enriched format whilst i find time to do the switch.

LeifW:
I have YAML version of the package descriptions coded up. Easier for me to
add extra fields to, and to read in by external package management tools.

jfdm:
@LeifW Is your port available somewhere? The plan at the moment after I merge this PR, is to (a) identify which format to go to---most likely yaml; (b) add in the new functionality; (c) leave the two co-exist for 11.2 then deprecate the old format in time for 11.3.

LeifW:
I'll open a PR. Was trying yo work out running the module name parser -
could just split on dots for now.

Sounds good to me - was thinking it could choose which parser to run based
on file extension?

ozgurakgun:
@jfdm, I am the worst kind of lurker on the Idris repo, but may I ask why you chose to require quotes for some fields and not others?

jfdm:
@ozgurakgun I used `stringLiteral` for the package brief to provide an implicit hint that it should be brief. I also think there is scope for it being made consistent with the rest and stipulate the brief is until the end of line.

david-christiansen:
This is probably a good idea. I think that idris-mode can be made to do some nice things with it :-)

Two questions:

 1. Why is the source location referred to as a "page"?
 2. What about a "bug tracker" field?

jfdm:
> Why is the source location referred to as a "page"?

Honestly, first thing that came to mind. Good candidate to be revised.

> What about a "bug tracker" field?

Good idea.

PR #3211, opened 2016-06-08
Building with stack works on OSX El Capitan + gmp/libffi

ydewit:
These are the steps:
```
brew install libffi
brew install gmp

PKG_CONFIG_PATH=$(brew --prefix libffi)/lib/pkgconfig stack build
```

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -1080,6 +1080,12 @@ Library
      build-depends: unix < 2.8
   if os(darwin)
      build-depends: unix < 2.8
+     if flag(GMP)
+        include-dirs: /usr/local/opt/gmp/include/gmp
+        extra-lib-dirs: /usr/local/opt/gmp/lib
+     if flag(FFI)
+        pkgconfig-depends: libffi
+      -- /usr/local/opt/libffi/lib/pkgconfig
   if os(windows)
      build-depends: Win32 < 2.4
   if flag(FFI)
--- a/stack.yaml
+++ b/stack.yaml
@@ -8,12 +8,6 @@ flags:
     FFI: true
     GMP: true
 
-extra-include-dirs:
-  - /usr/local/opt/gmp/include/gmp
-
-extra-lib-dirs:
-  - /usr/local/opt/gmp/lib
-
 extra-deps:
   - libffi-0.1
 
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,4 @@
-resolver: lts-5.11
+resolver: lts-6.2
 
 packages:
   - '.'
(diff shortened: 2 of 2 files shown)


jfdm:
A concern with this approach is that we are hard coding the extra directories in the `stack.yaml` specifically for Mac OS X users who use HomeBrew.

ydewit:
@jfdm Yes, indeed. I am not sure there is a way to conditionally include includes and libs in Stack. Maybe the simplest for now is to choose one way of getting the dependencies in OSX (for instance, using brew) and then have a OSX specific stack-osx.yaml. And another one for linux, e.g stack-linux.yaml etc.

zjhmale:
@ydewlt @jfdm how about use command params `--extra-include-dirs` and `--extra-lib-dirs` instead of hard coding the extra directories?

rodlogic:
@zjhmale I guess my motivation for making this change in the first place was to reduce the barriers to build idris. Having to add extra arguments every time we need to build Idris doesn't strike me a ideal. Although, I do see the point of hard coding the extra directories.

Shouldn't we just come up with 'the' Idris way of building on each OS and then make sure it is as simple as ```stack build``` after the requirements are installed? E.g. in the case of OSX, it could mean documenting a one time setup:
1. Install brew
2. Install libffi (```brew install ffi```)
3. Install gmp (```brew install gmp```)
4. Add ```export PKG_CONFIG_PATH=$(brew --prefix libffi)/lib/pkgconfig``` to ```.bashrc```
5. Clone the Idris-dev repo

Then just rely on ```stack``` to build Idris.

Each OS would have it's own recommended way of build Idris.

zjhmale:
@rodlogic ok

jfdm:
@ydewit thanks for the contribution however, I do not think it should be merged. The bump of Stack versions should be, but done in a secondary PR subject to tests passing.

My reasoning for not merging is this: 

> Not everyone who develops Idris on Mac OS X uses HomeBrew, and not all of those that develop on Idris use stack. By assuming when using stack where libraries are located is too restrictive on the configuration. It would be better for the flexibility to remain. Moreover, keeping separate stack configurations for different platform is too much effort. Also specialising cabal to cater for a single developer audience is also complicating the build unnecessarily.

I think the guidance in INSTALL.md should be updated to reflect how stack can be run with extra paths, negating the advice of prepending and amending an environment variable.

For advice on how to build Idris-Dev for a specific platform, please see the [wiki pages](https://github.com/idris-lang/Idris-dev/wiki/Installation-Instructions). This advice should be updated as well.

Finally, there is already an (unofficial) formula for Idris in Homebrew that takes into account building with FFI and GMP and uses the correct paths.

PR #3214, opened 2016-06-09
Clean up Setup.hs a bit

enolan:


jfdm:
The changes suggested by HLint look reasonable to me. Thanks for the PR.

PR #3215, opened 2016-06-10
HLint Directed Code Cleanup

jfdm:
HLint suggestions were applied  when code readability (and performance) would not be affected.

+ Removed redundant `$`
+ Reformatted some code.
+ Used pattern matching and application of single item lists `[x]` rather than `(x : [])`
+ Used `null` for list length checking.
+ Use `unwords`
+ Use `concatMap` over `concat $ map...`
+ Use `exitSuccess`

### DIFF
--- a/src/IRTS/CodegenC.hs
+++ b/src/IRTS/CodegenC.hs
@@ -808,7 +808,7 @@ genDispatcher tags = "void* _idris_get_wrapper(VAL con)\n" ++
 
 genWrapper :: (FDesc, Int) -> String
 genWrapper (desc, tag) | (toFType desc) == FFunctionIO =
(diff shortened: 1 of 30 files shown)


edwinb:
This looks good but some conflicts have crept in. Once you've tidied them up I think it's okay just to merge directly.

jfdm:
Okay. Have rebased and resolved conflicts. Once checks have passed I shall merge.

ahmadsalim:
@jfdm Would it be possible for you to include those rules in `HLint.hs` as well? Thanks!

ahmadsalim:
I mean you made the file for exactly that purpose :), so I am just reminding you

PR #3219, opened 2016-06-13
A new File Effect.

jfdm:
The File Effect has been updated to take into account changes in
`Prelude.File` and provide better state-mangement.

The new API should be backwards compatabile with the existing API. The only difference is the support for custom error reporting was minimise.

This still needs to be documented, and will merge it after comments *and* documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# New in 0.13:
+
+* The File Effect has been updated to take into account changes in
+  `Prelude.File` and to provide a 'better' API.
+
 # New in 0.12:
 
 ## Language updates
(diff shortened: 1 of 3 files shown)


jfdm:
This is failing due to me not having fixed the test yet. Unfortunately, with the new API writing a purely effectful  `readFile` function is harder due to state changes especially when the return result is an indexed type... Will either look at updating the state transitions, or just rewrite the test...

jfdm:
The File Effect has been fixed, and the tests *should* pass. Sadly the entire effect is no longer a dependent effect.

jfdm:
This shouldn't be merged until:

+ [ ]  `Idris v0.12` has been released;
+ [ ] The code has been documented;
+ [ ] The CHANGELOG updated.

PR #3236, opened 2016-06-20
Fixes to some library issues under nix and stack.

tpsinnem:
A couple of modifications to stack-shell.nix:

Ghci has an [issue](https://ghc.haskell.org/trac/ghc/ticket/11042) such that it doesn't recognize --extra-lib-dirs, which causes problems on systems like Nixos which use unusual locations for libraries. The workaround is to add locations for compiled libraries to LD_LIBRARY_PATH.

Also, dependencies don't seem to compile without ncurses being available, so I added that as well.


ahmadsalim:
Thanks for the PR!

PR #3248, opened 2016-07-01
Use GHC generics for NFData instances

enolan:
This gets rid of a lot of code. Much of it was generated with
the "derive" tool, but some of it wasn't. I tested compilation speed
by compiling the benchmarks with bench[1] and there was no
statistically significant difference.

[1]: https://github.com/Gabriel439/bench

### DIFF
--- a/src/IRTS/CodegenCommon.hs
+++ b/src/IRTS/CodegenCommon.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE DeriveGeneric #-}
 {-|
 Module      : IRTS.CodegenCommon
 Description : Common data structures required for all code generators.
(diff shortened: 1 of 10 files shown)


ahmadsalim:
@enolan Would it be possible to also do this for `Binary` and the other things that used derive?

melted:
I'd be careful with Binary, it's much slower with Generic and what we have is unacceptably slow as is.

enolan:
@ahmadsalim It's possible, yes. I was planning to look into after this one. If benchmarking shows a substantial performance degradation then we shouldn't do it.

Do you want me to handle Binary before merging this PR?

ahmadsalim:
@enolan Oh, not necessarily. I was just wondering whether it was possible.

melted:
The PR is fine as is. 

I'm unhappy with Binary. The loading is slow and generates way more garbage than reasonable. I'd rather try to find a better way to do the IBCs

melted:
Thanks for the PR!

Stats of +202-1045 means you owe us 843 lines of code :-)

PR #3249, opened 2016-07-02
Support runtime linking with linker scripts

enolan:
Makes ffi001 work on Ubuntu 16.04.

### DIFF
--- a/src/Util/DynamicLinker.hs
+++ b/src/Util/DynamicLinker.hs
@@ -5,7 +5,7 @@ Copyright   :
 License     : BSD3
 Maintainer  : The Idris Community.
 -}
-{-# LANGUAGE ExistentialQuantification, CPP #-}
+{-# LANGUAGE ExistentialQuantification, CPP, ScopedTypeVariables #-}
 module Util.DynamicLinker ( ForeignFun(..)
                           , DynamicLib(..)
                           , tryLoadLib
@@ -17,8 +17,13 @@ import Foreign.LibFFI
 import Foreign.Ptr (Ptr(), nullPtr, FunPtr, nullFunPtr, castPtrToFunPtr)
 import System.Directory
 #ifndef mingw32_HOST_OS
+import Control.Exception (try, IOException)
+import Data.Array (Array, inRange, bounds, (!))
+import Data.Functor ((<$>))
+import Data.Maybe (catMaybes)
 import System.Posix.DynamicLinker
 import System.FilePath.Posix ((</>))
+import Text.Regex.TDFA
 #else
 import qualified Control.Exception as Exception (catch, IOException)
 import System.Win32.DLL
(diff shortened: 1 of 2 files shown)


src/Util/DynamicLinker.hs
```diff
@@ -69,13 +74,65 @@ libFileName dirs lib = do let names = [lib, lib ++ "." ++ hostDynamicLibExt]
                           return $ maybe (lib ++ "." ++ hostDynamicLibExt) id found
 
 #ifndef mingw32_HOST_OS
+-- Load a dynamic library on POSIX systems.
+-- In the simple case, we just find the appropriate filename and call dlopen().
+-- In the complicated case our "foo.so" isn't actually a library. Some of the
+-- .so files on modern Linux systems are linker scripts instead. dlopen()
+-- doesn't know anything about those. We need to look inside the script for the
+-- actual library path and load that. This is a horrible hack, the correct
+-- method would be to actually parse the scripts and execute them. The approach
```
Melvar: The “correct” method would not in fact be to parse the scripts, since the scripts are for the *static* linker. (When the static linker encounters such a script, in the common case it will link whatever it’s working on with the shared object named in the script directly, possibly additionally linking in static libraries when necessary.) There is actually a C header file with `#define`s in it naming the actual shared objects of the installed glibc; these are what a `dlopen()`-using program is supposed to use according to glibc devs.

melted:
I assume that macs shouldn't do these things, should we check for that?

Melvar:
The handling of linker scripts as implemented in this PR shouldn’t break the loading of normal dynamic libraries regardless of platform, since it always tries to just `dlopen()` it first. As for the gnu-specific header, if a change was made to use that the code would indeed need to handle its absence gracefully regardless of OS.

melted:
It just seems bad to me that it tries something that can't possibly succeed on macs if dlopen fails.

enolan:
@Melvar I know about the header file with the actual filenames in it. I think it's a bad solution. People shouldn't need to know what linker scripts are just to use a library. This whole situation is caused by a leaky abstraction. Someone writing a C program calls `gcc -lm` and gets the correct library. Idris users shouldn't have to work any harder.

The code should be `#IFDEF`d for Linux, AFAIK these scripts are a GNU ld specific feature. The regexes shouldn't match if it's not actually looking at a linker script, but let's be cautious.

enolan:
Can someone test the new version on OS X?

Melvar:
I said nothing about users needing to know about it; either way this is abstracted by Idris. The header file is potentially a less brittle way to get the information, however, since it’s machine generated rather than hand-written.

On the other comment, Linux does not necessarily imply glibc, nor vice versa (though counterexamples for the latter seem pretty rare). The check therefore should be for glibc, not for Linux.

melted:
I don't think we should check for glibc. That increases the complexity for little gain and checking the identity of the libc is nonstandard and bad policy.

melted:
I have a Mac laying around, I'll test it.

Melvar:
And yet, the question whether it is glibc is what we actually need to know. Checking for Linux is merely a poor proxy, the feature has nothing to do with Linux per se.

melted:
We can disable this feature at compile time on OSX and Windows, so we do that. On Linux we do a run-time check, which shouldn't result in any bad results, worst case is that it fails to load if say musl grows this feature in an incompatible way.

melted:
Thanks for the PR!

PR #3226, opened 2016-06-18
Make Idris build with GHC8

KaneTW:
Fixes #3193. Note that I used the git version of trifecta (stack.yaml has the respective commit) to build this (due to transformers bounds).

The GHC bug in question is https://ghc.haskell.org/trac/ghc/ticket/12201 and the actual workaround is a dummy DeltaParsing IdrisParser instance (which is identical to the (StateT s m) one)

PR #3250, opened 2016-07-03
Changed definition of `runTacTm` to have simpler patterns.

ahmadsalim:
This is to avoid exhaustion of the pattern matching coverage checker in GHC 8 as stated in #3226 .


david-christiansen:
What's the reason for the irrefutable patterns here? It would be good to have a comment explaining the reasoning.

ahmadsalim:
@david-christiansen It should be dependent vectors, so I give a size argument. It will fail otherwise, I can add a comment.

david-christiansen:
I understand why they _can_ be  irrefutable, but what do we gain by that?

ahmadsalim:
@david-christiansen It shows that they should always match and use `fail` from the monad.

PR #3255, opened 2016-07-07
Allow CDATA to be copied to other VMs

crough:
When using `fork` with access to a CData reference, the RTS will panic
on the assertion that it was unable to match the type of the closure in
doCopyTo. This is because of a missing case for CData.

This change copies the paradigm for other type tags of Closures for
CData.


melted:
Thanks for the PR!

PR #3258, opened 2016-07-09
Travis twiddling

melted:
Reduce the number of Travis targets, by just building one for 7.8.

Use the latest point release of 7.10, 7.10.3.

Spare GHC 7.6, it still works and it's what's in Debian.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ addons:
     packages:
       - ghc-7.6.3
       - ghc-7.8.4
-      - ghc-7.10.2
+      - ghc-7.10.3
       - ghc-8.0.1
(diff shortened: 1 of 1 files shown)


jfdm:
Before this gets merged, it would be nice for the Git commit to explain the rationale behind the reduction on travis targets, and more towards the guiding principle for target selection. For example, the git commit doesn't include everything mentioned in the PR comment.

As a general rule of thumb, the Git commit message needs to be the canonical source of how the commit/pr changes the project. Otherwise, this information is lost if we move away from GitHub.

melted:
Ok, I updated the commit message. The reason for reducing the builds is that three were added with GHC 8.

PR #3259, opened 2016-07-09
Fix doc comment

melted:
Haddock complains about this, let's see if this works


melted:
didn't work...

gpyh:
Just so you know, you can use Travis with your own fork very easily by registering to them.
That's what I did myself when I needed to experiment the (soon-to-be) new test suite as you can see here: https://travis-ci.org/gpyh/Idris-dev
That way you don't need to submit a PR if you need to try something. Plus, it won't ever appear in the upstream's Travis history.

PR #3260, opened 2016-07-09
Unbreak doc build

melted:
Well, this try has to work.


jfdm:
I wonder if that doc comment wasn't working because of the trailing arrow style formatting of the type signature...It shouldn't really, but you never know.

PR #3263, opened 2016-07-10
Add IDEA and VSCode dirs to ignore list.

melted:


NobbZ:
I did add any possible combination of OS, IDE and even Shell-specific stuff into `.gitignore` a long time as well.

Even with a commented `.gitignore` it grow out of control quickly. So I changed to a global `.gitignore` which contains stuff specific to my setup and I do encourage contributors to do this as well.

Just add it as a global config to git:

```
git config --global core.excludesfile /path/to/global/.gitignore
```

david-christiansen:
@NobbZ: I think that's a useful thing to know, but nevertheless, @melted's addition to the project .gitignore is nice to have. Good defaults are important for new contributors, after all!

PR #3271, opened 2016-07-15
Fix comment typos

HectorAE:
The documentation for Pair and UPair mistakenly referred to the type of the right elements in a pair as being the same as the left elements.


jfdm:
Thanks @HectorAE

PR #3274, opened 2016-07-19
HLint Inspired Code Cleanup 'Redundant Do'

jfdm:
Where code clarity is not hindered HLint's raising of 'Redundant do"
has been addressed.

Other formatting improvements have been made in some functions.


src/Idris/Core/ProofState.hs
```diff
@@ -834,20 +834,20 @@ patbind n ctxt env _ = fail "Can't pattern bind here"
...
 compute :: RunTactic
 compute ctxt env (Bind x (Hole ty) sc) =
-    do return $ Bind x (Hole (normalise ctxt env ty)) sc
+    return $ Bind x (Hole (normalise ctxt env ty)) sc
 compute ctxt env t = return t
 
 hnf_compute :: RunTactic
-hnf_compute ctxt env (Bind x (Hole ty) sc) =
-    do let ty' = hnf ctxt env ty in
+hnf_compute ctxt env (Bind x (Hole ty) sc) = do
```
david-christiansen: This just seems to be moving the "do" to the dangling position, which we haven't historically had in Idris, rather than eliminating an unnecessary one.

src/Idris/Coverage.hs
```diff
@@ -457,15 +457,10 @@ checkTotality path fc n
...
-    warnPartial n t
-       = do i <- getIState
-            case lookupDef n (tt_ctxt i) of
-               [x] -> do
-                  iWarn fc . pprintErr i . Msg $ "Warning - " ++ show n ++ " is " ++ show t
---                                ++ "\n" ++ show x
---                   let cg = lookupCtxtName Nothing n (idris_callgraph i)
---                   iputStrLn (show cg)
-
+    warnPartial n t = do
```
david-christiansen: This isn't removing a redundant "do", it's just putting it in dangling position, which isn't the style that most of Idris is written in.

src/Idris/Elab/Clause.hs
```diff
@@ -706,8 +704,7 @@ elabClause info opts (cnum, PClause fc fname lhs_in_as withs rhs_in_as wherebloc
 
         mapM_ (\(n, _) -> addIBC (IBCDef n)) def''
 
-        when (not (null def')) $ do
-           mapM_ defer_totcheck (map (\x -> (fc, fst x)) def'')
+        when (not (null def')) $ mapM_ defer_totcheck (map (\x -> (fc, fst x)) def'')
```
david-christiansen: It's good to get rid of "do" here, but I still think the part after the "$" should be indented on the next line, because that looks more like the imperative control structure that "when" encodes.

src/Idris/Elab/Type.hs
```diff
@@ -190,8 +190,7 @@ elabType' norm info syn doc argDocs fc opts n nfc ty' = {- let ty' = piBind (par
          -- the name unambiguously)
          i <- getIState
          rep <- useREPL
-         when rep $ do
-           addInternalApp (fc_fname fc) (fst . fc_start $ fc) (PTyped (PRef fc [] n) ty') -- (mergeTy ty' (delab i nty')) -- TODO: Should use span instead of line and filename?
+         when rep $ addInternalApp (fc_fname fc) (fst . fc_start $ fc) (PTyped (PRef fc [] n) ty') -- (mergeTy ty' (delab i nty')) -- TODO: Should use span instead of line and filename?
```
david-christiansen: It's good to get rid of "do" here, but I still think the part after the "$" should be indented on the next line, because that looks more like the imperative control structure that "when" encodes.

src/Idris/Elab/Utils.hs
```diff
@@ -62,19 +62,19 @@ checkDeprecated fc n
...
 
 checkFragile :: FC -> Name -> Idris ()
 checkFragile fc n = do
   r <- getFragile n
   case r of
     Nothing -> return ()
-    Just r  -> do
-      iWarn fc $ text "Use of a fragile construct "
+    Just r  -> iWarn fc $ text "Use of a fragile construct "
                      <> annName n
```
david-christiansen: This should probably move to the right, because its parent expression has moved to the right. Alternatively, the dangling "do" could be removed without moving the code up to the previous line.

src/Pkg/Package.hs
```diff
@@ -287,20 +286,24 @@ buildMods opts ns = do let f = map (toPath . showCG) ns
...
-         let libtest = d </> "rts" </> "libtest.c"
-         e <- rawSystem gcc [libtest, "-l" ++ f, "-o", tmpf]
-         case e of
-            ExitSuccess -> return True
-            _ -> do if warn
-                       then do putStrLn $ "Not building " ++ p ++
-                                          " due to missing library " ++ f
-                               return False
-                       else fail $ "Missing library " ++ f
+testLib warn p f = do
```
david-christiansen: This is not the removal of an extra "do". Instead, it's a conversion to dangling position, which IMHO makes things much harder to read because the reason for the switch to imperative mode is not where it happens. We haven't used that style historically in most of Idris, and it seems that a global switch to dangling mode should at least merit a bit of discussion.

src/Pkg/Package.hs
```diff
@@ -287,20 +286,24 @@ buildMods opts ns = do let f = map (toPath . showCG) ns
...
+    case e of
+       ExitSuccess -> return True
+       _ -> if warn
+              then do
+                putStrLn $ unwords ["Not building"
+                                   , p
+                                   , "due to missing library"
+                                   , f
+                                   ]
+                return False
```
david-christiansen: Here's an example where I think dangling "do" mode really shows its weakness. There's no "<-" binds, so there's no visual cue that we're in a "do", and scanning up the margin doesn't reveal it.

src/Pkg/Package.hs
```diff
@@ -287,20 +286,24 @@ buildMods opts ns = do let f = map (toPath . showCG) ns
...
+    gcc <- getCC
+    (tmpf, tmph) <- tempfile ""
+    hClose tmph
+    let libtest = d </> "rts" </> "libtest.c"
+    e <- rawSystem gcc [libtest, "-l" ++ f, "-o", tmpf]
+    case e of
+       ExitSuccess -> return True
+       _ -> if warn
+              then do
+                putStrLn $ unwords ["Not building"
```
david-christiansen: "unwords" is definitely an improvement here.

david-christiansen:
This seems to mix clear improvements (like removal of redundant "do") with some things that are not clearly improvements (like moving "do" to dangling position). I added specific comments.

jfdm:
If there are more drawbacks than advantages to a PR, I think it would be better practice to summarise the major drawbacks, and leave out specific comments unless the Pull-requester asks for them. Otherwise the wall of  commentary is information overload to the Pull-requester; can be 'too direct' and in the end non-constructive if there are more comments on 'negative' aspects of the PR than positive.

One can walk away from the 'wall of commentary' with the feeling of 'Why Bother'?

david-christiansen:
First: I apologize that I made you feel bad with this. It was not my intention.

In a PR that contained a change to the way Idris worked, I think that a list of nitpicky style stuff would not be so useful. However, this PR consisted exclusively of a list of nitpicky style stuff. Because the different changes were all in one commit, there wasn't a way to say that the "introduce dangling `do`" commit was perhaps something that should at least be discussed, while the "eliminate redundant `do`" commit was a clear "merge this now". IIRC you're in Emacs - Magit has a good interface for retroactively cutting unrelated changes that happened at the same time into separate commits.

I don't really think that "more drawbacks than advantages" is an accurate summary either, but I can see how you'd get that impression. Instead, I think that "mixes clear, obvious wins with things that need discussion" is a better summary.

Anyway, sorry again that I made you feel discouraged.

melted:
I like the dangling do, mostly because it saves tons of indentation.

LeifW:
I'm used to the "dangling do" style, myself.

david-christiansen:
I think that this is not the right venue for discussion of dangling "do", but that this should probably happen, followed by a global change. Right now, most of the code written by Edwin and myself does not use it, while other bits of code use it inconsistently. In any case, that's really a mailing list kind of thing.

My issue here was that deviating from the most common style in the codebase in a piecemeal fashion is not a good thing to mix up with uncontroversial changes like getting rid of redundant `do`. An argument about whether to dangle the `do`s should happen at least on irc, but probably better on the ml, and be followed by a global change either to or from dangles.

Issue #3251, opened 2016-07-03
exit code from System.system is wrong


alrunner4:
```
> :exec system "exit 1" >>= printLn
256
```

Melvar:
Well, `system` returns a wait status, not an exit code. Looking at the definition on my current system a recent Linux system, the exit code happens to be the second-lowest byte of the status, so `256` is, indeed a status containing an exit code of `1` (and additionally indicating termination not by a signal). Unless your wait status macros are defined differently, this is expected behavior.

steshaw:
@Melvar, there was no need to look into this. As I mentioned on #idris, I was taking a look at it. I've added a test case to reproduce the issue. I do agree that it's unfortunately expected behaviour. However, I've expanded the test case to show how to bind `WIFEXITED` and `WEXITSTATUS` macros with Idris FFI. @alrunner4 could copy these into their project if they wish. Perhaps better would be to incorporate access to various such macros into a prelude. However, there is a question of portability...

alrunner4:
Ah, that explains it. Since the macros are POSIX-defined, perhaps it's best they live in idris-posix? It does seem a bit odd to have a function in the Prelude that returns an implementation-defined Int, though.

@steshaw, where is this expanded test case?

steshaw:
@alrunner4 [take a look here](https://github.com/idris-lang/Idris-dev/pull/3281/files)

alrunner4:
Thanks! I'll work around for now. A while back I'd considered pitching in some things to idris-posix (it's pretty slim currently) but I've yet to come into a chunk of time to get much of anything done with it.

PR #3281, opened 2016-07-21
Added test case for System.system

steshaw:
The test case reproduces the issue reported in #3251. It also shows how to extract the exit status using `WIFEXITED`/`WEXITSTATUS`. See second commit message for more information.


test/base001/expected
```diff
@@ -0,0 +1,27 @@
...
+True
+
+-------------------------------------------
+Executing './does-not-exist'
+raw status = 32512
+exit status = 127
+True
+
+Removed: 00base001-idx.ibc
+Removed: /Users/steshaw/Projects/steshaw/idris/test/base001/base001
```
jfdm: In case you didn't notice it, the expected output contains information specific to your system. That is, it contains an absolute path.

jfdm:
@steshaw Thanks for the contribution. Travis is failing due to an issue not related to your PR, so please do not worry about Travis not completing. This is happening across a few other PRs.

I have also noticed that `idris.cabal`, was not updated. With Idris' current setup for testing, tests need to be listed. You can see examples in `idris.cabal`.

test/base001/expected
steshaw: I'd missed that, thanks!

steshaw:
@jfdm I hadn't realised that `idris.cabal` needs updating because the new tests I've been adding run during `stack test`. They're discovered by the `runtest.hs` script. I've added tests in a number of pull requests so we better be careful `idris.cabal` gets updated in all of them. What goes wrong when test files aren't added to Extra-source-files in `idris.cabal`?

melted:
They aren't packaged in the cabal sdist, and so they won't exist when travis runs as it uses a sdist.

steshaw:
@melted understood, thanks.

jfdm:
@steshaw Just to let you know that we recently accepted a PR that changed Idris' test suite to run of tasty. Is it possible for you to rebase the PR off of Idris' current master. Then we can see what went wrong and how to proceed further.

melted:
I'll check out why this fails.

melted:
This isn't going to work, you are testing POSIX stuff that doesn't exist on windows, system returns just the exit code there. So those exit code macros doesn't exist at all. And the values would be off.

steshaw:
Thanks, @melted! I'm in the middle of setting up a Windows VM with your PowerShell script but it could take a while as my internet speed is molasses at the moment. Just got PowerShell 3.0 installed :).

melted:
If you want to make it work, you shouldn't output the details, but just an OK or something that the test succeeded. And then one can put an ifdef or something that makes a different check.

steshaw:
@melted, makes sense. How can I disable this test for Windows?

melted:
We haven't a mechanism for that. We used to have one for variant output when we had a perlk testrunner, but that wasn't needed. But that wouldn't help. I'd just check the os with http://www.idris-lang.org/docs/current/base_doc/docs/System.Info.html and do the simpler check on windows.

steshaw:
@melted, I'll give something like that a shot then.

It would be nice to understand what is happening on Appveyor. It seems that the `exit1.c` file isn't compiling, but there aren't any error messages from gcc appearing in the log.

melted:
`gcc -Wall -o exit1 exit1.c
exit1.c: In function 'doSystem':
exit1.c:8:7: warning: implicit declaration of function 'WIFEXITED' [-Wimplicit-function-declaration]
   if (WIFEXITED(exitStatus)) {
       ^
exit1.c:9:53: warning: implicit declaration of function 'WEXITSTATUS' [-Wimplicit-function-declaration]
     printf("exit1: WEXITSTATUS(exitStatus) = %d\n", WEXITSTATUS(exitStatus));
                                                     ^
C:\Users\niklas\AppData\Local\Temp\ccskTA3M.o:exit1.c:(.text+0x3f): undefined reference to `WIFEXITED'
C:\Users\niklas\AppData\Local\Temp\ccskTA3M.o:exit1.c:(.text+0x4d): undefined reference to `WEXITSTATUS'
collect2.exe: error: ld returned 1 exit status
`

melted:
It's easy, those don't exist on Windows, of course you could always do

```c
#ifndef WIFEXITED
#define WIFEXITED something something
#endif
```

for Windows, the values would be off though. As windows only puts the exitcode in there.

steshaw:
@melted, does that mean that stderr isn't appearing in the Appveyor logs or something?

melted:
There's plenty of indirection here, a powershell script that starts a shellscript that starts a haskell program that starts a shellscript that starts make that starts gcc. I don't know who is to blame.

If you want to test compiling, you can install a mingw toolchain to build windows stuff on linux.

steshaw:
It's not easy if you can't even see the error message ;)

melted:
I think iit's runTest in TestRun.hs that is to blame, it only hooks up stdout.

steshaw:
@melted looks right, I just did a little test here on Mac with a bogus error in `exit1.c`. Will have to raise that as a bug. It's pretty nasty.

steshaw:
@melted nice, thanks.

melted:
Well, let's merge it, someone can put in a little testing for windows if they want.

steshaw:
@melted tah

PR #3283, opened 2016-07-22
Do not append "libs" to "objs" when parsing "objs".

steshaw:
This is a little fix for the ipkg parser. Currently you run into trouble when your "libs" are specified before your "objs" (because the parser appends "libs" instead of "objs"). Looks like a copy/paste slip up.


jfdm:
@steshaw Thanks for catching this. No idea that had been lying around...

PR #3293, opened 2016-07-29
Fix testsuite regression on Windows.

melted:


test/TestRun.hs
```diff
@@ -63,28 +63,31 @@ ingredients = [rerunningTests [consoleTestReporter],
 -- A ripoff of goldenVsFile from Tasty.Golden
 test :: String -> String -> IO () -> TestTree
 test name path act =
-  goldenTest name (BS.readFile ref) (act >> BS.readFile new) cmp upd
+  goldenTest name (readFile ref) (act >> readFile new) cmp upd
     where
       ref = path </> "expected"
       new = path </> "output"
-      cmp x y = return $ if x == y then Nothing
+      cmp x y = return $ if normalize x == normalize y then Nothing
```
gpyh: Here, `x` and `y` are the contents of the files. I don't think using `normalize` is required.

PR #3294, opened 2016-07-30
Fix #3275

### LINKED ISSUE #3275: Fixity not checked for Type constructors and Data constructors
Minimal example:
```
data Qux : a -> Type where
  Tux : String -> Qux a

data Bar : a -> Type where
  Nil : Bar a
  (:>) : Qux a -> Bar a -> Bar a

data (:>>) : a -> b -> Type where
```

Current behaviour:  The file compiles

If you use the type constructors or data constructors in another module, the parser stops with an obscure error.  

Expected behaviour:
```
 error: Missing
    fixity declaration for
    Minimal.:>

 error: Missing
    fixity declaration for
    Minimal.:>>
```

beark:
Check fixity of type and data constructors that are operators.


jfdm:
@beark Thanks for your contribution, it is much appreciated. 

I am wondering, however, if it is possible for you to add a regression test to Idris' test suite. I believe there was a MFE presented in the original report---from #3275. This is just so we can catch it in future.

Also add yourself to CONTRIBUTORS.md whilst your at it. 

Once that is done we can look at merging it.

Thanks

beark:
Of course, I was already planning on looking into test(s), but figured it was good to get some early feedback. Especially as this is my first foray into the idris code base.

Will be gone most of the day unfortunately (am on PST, for reference), but I'll get to it as soon as I can!

david-christiansen:
It looks to me like the `checkFixity'` definitions are identical in the two places that they occur, except for their free variable `getName`. Can they be made into a top-level definition that takes a `Maybe TTName` as an argument instead?

beark:
I was thinking of doing that, but include the definition of `getName` too. Something like:

    checkDeclFixity :: IdrisParser PDecl -> IdrisParser PDecl
    checkDeclFixity p = do theCheck
        where getName (PTy _ _ _ _ _ n _ _ ) = Just n
              getName (PData _ _ _ _ _ (PDatadecl tyn _ _ _)) = Just tyn
              getName _ = Nothing

beark:
Is there anything else I can do to improve the solution? I'm fairly happy with it now, but I'm always open to suggestions.

beark:
Ping. Anything left blocking a merge?

david-christiansen:
Nope!

david-christiansen:
And thanks again!

PR #3295, opened 2016-07-30
use unsigned operations for BitsN

### LINKED ISSUE #3286: divB8 returns incorrect value
divB8 10 255 should return 0.

It returns 246.

This is probably due to the use of prim__sdivB8 instead of prim__udivB8.

zjhmale:
Fixes #3286 

since Bits8 is unsigned through the [definition](https://github.com/idris-lang/Idris-dev/blob/master/libs/prelude/Prelude/Interfaces.idr#L284-L288), `prim__udivB8 ` and `prim__uremB8` should be used rather than the signed ones.


jfdm:
@zjhmale Thanks for the contribution, it is much appreciated.

I am wondering if you can add to the commit message in the details:

> since Bits8 is unsigned through the definition, prim__udivB8 and prim__uremB8 should be use rather than the signed ones.

Just to help make the commit more self-contained.

Also, I am wondering if this should be added to an existing test in the test suite so that we can catch this from diverging in the future.

Once these are done we can look at merging.

zjhmale:
@jfdm ok, and i promise to make commit messages more comprehensive next time! sorry for bothering you twice 😂

Melvar:
Surely the other `BitsN` types need the same adjustment.

zjhmale:
@Melvar exactly!

zjhmale:
@jfdm poke

david-christiansen:
I agree that it makes more sense to see these as unsigned operations. But this should be mentioned in the CHANGELOG for the next release, in case anyone is relying on the current behavior.

zjhmale:
@david-christiansen fair enought, should i update the CHANGELOG in this PR?

jfdm:
@zjhmale To answer your question, please do update the CHANGELOG. When that is done, if everyone is happy we can then merge.

zjhmale:
@jfdm thx,  i was just remeber to update the changelog seconds ago 😂

jfdm:
If everyone is happy, I shall merge this at the end of day.

ahmadsalim:
Is it not possible to have both versions and then name them accordingly?

zjhmale:
@ahmadsalim you mean both signed and unsigned operators? maybe next PR? if it really need to.

ahmadsalim:
@zjhmale I think it is useful to have both versions, but I don't think it is something that should block this PR.

zjhmale:
@ahmadsalim through the maxbound and minbound definitions for BitsN types, just make the `divN` `remN` funtions behavior correctly is cool i think.

zjhmale:
@jfdm i think this is ready to merge :)

PR #3296, opened 2016-07-30
Add appveyor.yml

### LINKED ISSUE #1948: Windows CI Setup
Currently we use travis to provide CI testing for `Idris` on Linux boxes, this has the nice side effect of CI for Mac OS X. It would be nice if we can provide CI testing for Windows machines. 

This is not my idea @david-christiansen has mentioned it a few times, and I am in full agreement. [AppVeyor](http://www.appveyor.com/) seems to be a suitable candidate and a cursory glance implies that the [Build Instructions and Scripts for Windows](https://github.com/idris-lang/Idris-dev/wiki/Idris-on-Windows) can be adapted accordingly. Whether AppVeyor is the best (or maybe only) Windows CI service is another matter.

melted:
Fixes #1948

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,20 @@
+version: 1.0.{build}
+init:
+- ps: >-
+    choco install ghc --version 7.10.3 | Out-Null
+
+    $env:PATH="$env:PATH;C:\tools\ghc\ghc-7.10.3\bin;$HOME\AppData\Roaming\cabal\bin"
+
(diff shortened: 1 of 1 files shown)


melted:
As discussed on IRC, we shouldn't connect appveyor to the main repo just yet, I'll run it for a while to see how stable it is before we turn it on here.. But merging this won't do any harm and it can be useful if someone else wants to run it on their own repo.

PR #3298, opened 2016-07-30
Fix test fail on mac

### LINKED ISSUE #3297: test suite failed unexpectedly
i use `stack test` to run all the test suites at the moment, but two test suites just failed unexpectedly as following pics exhibit.

idris-fail1
idris-fail2

after remove two golden files and regenerate use `stack test --test-arguments="--accept"`, these two suites can finally pass.

but it is really wired, isn't it? how can i fix this.

melted:
Add back old rationale for normalize.

Fixes #3297


melted:
My OSX Travis also says it worked :-)

PR #3300, opened 2016-07-31
Enable Travis on OSX

melted:


melted:
It seems to be working great, including the stashing of haskell libs (which would not fail on the first build, so I checked the logs on the second build to see if that path was taken). So I think it's worth to merge without any further ado.

PR #3303, opened 2016-07-31
let Reverse Ranges work again

### LINKED ISSUE #3065: Reverse Ranges no longer work.
Issue #2801 reported erroneous behaviour with ranges in the REPL. This was due to a change in how `Nat`  and `minus` and `(-)` worked.

Although the issue was addressed for positively increasing ranges, negatively decreasing ranges are still not working. For example:

```Idris
Idris> [1,3..9]
[1, 3, 5, 7, 9] : List Nat
Idris> [5,4..1]
(input):1:1-9:When checking argument smaller to function Prelude.Nat.-:
        Can't find a value of type
                LTE 5 4
```

zjhmale:
Fixes #3065 

the first solution @dc-mak commented on the issue seems reasonable, passing `from, next, to` instead of  `from, inc, to` may be a good way to solve this.


jfdm:
@zjhmale Thanks for the contribution, it is much appreciated. I see no obvious reason not to accept the update, will sit on it until end of day before I merge.

In case you were not aware of it, but Idris does have a variant of `print` for printing with new lines: `printLn`. This avoids the verbosity of using `putStrLn $ show x`, or something else involving `print` and newlines.

zjhmale:
@jfdm cool! i will update the code, i did not walk through the tutorial carefully sorry.

zjhmale:
and this improvement also need to update the changelog i think.

jfdm:
@zjhmale Do not worry about not using `printLn`; it wasn't a blocking issue and I was going to accept the test code regardless. Moreover, I do not think `printLn` is mentioned in the tutorial, only in the prelude.

zjhmale:
@jfdm i will look around the Prelude to get the full picture of basic usage later, thx for mention it 😄

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
 
 * The File Effect has been updated to take into account changes in
   `Prelude.File` and to provide a 'better' API.
+* `[a,b..c]` now can be used to generate reverse ranges.
```
jfdm: Apologies in advance, however, range syntax i,e, `[a,b..c]` is syntax sugar. The CHANGELOG entry might be better written with direct reference to the underlying functions `natEnumFromThen` and `natEnumFromTo`, and how this affects the range syntax. Otherwise without knowing how range syntax works, the change might not be easy to understand.

For example:

> `natEnumFromThen` and `natEnumFromTo` have been updated to correctly calculate reverse ranges. Range syntax `[a,b..c]` can be used again to generate reverse ranges.

zjhmale: @jfdm yeah, i should make it more detailed, thx 😄

jfdm:
Thanks for taking the time to fix the CHANGELOG. It is much appreciated.

zjhmale:
The pleasure is all mine 😄   thx for the advices btw.

PR #3305, opened 2016-08-01
Improvements to iPKG code.

jfdm:
This commit rehomes the code for package building and installation
under the Idris namespace, allowing for projects that use the code to
use a more meaningful import statement.

This commit also factors out the ipkg data type from the parser.


david-christiansen:
Seems like a reasonable thing to do.

Issue #2555, opened 2015-08-27
Terrible error message when no main specified


david-christiansen:
When the `executable` field of a `.ipkg` file is specified, but the `main` field is not, Idris simply dumps the error: 

```
Can't find import 
```

rather than stating what the problem is.

PR #3311, opened 2016-08-02
Better handling of optional main in package files

BartAdv:
No default value is provided for main - in case it's missing, the
appropriate error message is produced.

Fixes https://github.com/idris-lang/Idris-dev/issues/2555


melted:
You need to add the tests to idris.cabal, so they get added to the sdist.

jfdm:
@BartAdv thanks for the PR, if there are no other comments, I shall merge at the end of the day.

PR #3313, opened 2016-08-03
ommit internal error from REPL when using search command

### LINKED ISSUE #3221: Internal error from REPL.
typing `:search !a` at the repl triggers the following internal error:

```
No location:when checking type of searchType: INTERNAL ERROR: Uneraboratable syntactic form missing pretty-printer for term
This is probably a bug, or a missing error message.
```

zjhmale:
Fixes #3221 

and fix a typo by the way. it takes a while to finally figure out where this internal error msg comes from. [#1](https://github.com/idris-lang/Idris-dev/blob/master/src/Idris/REPL.hs#L331-L332) [#2](https://github.com/idris-lang/Idris-dev/blob/master/src/Idris/Elab/Term.hs#L1325) [#3](https://github.com/idris-lang/Idris-dev/blob/master/src/Idris/AbsSyntaxTree.hs#L2132) [#4](https://github.com/idris-lang/Idris-dev/blob/master/src/Idris/Delaborate.hs#L299)

actually, i don't think the `!a` type reported in the issue makes any sense, there is no meaning to apply `!` to a type right? even in a dependently typed programming language cause

```idris
(!) : m a -> a
```

### DIFF
--- a/src/Idris/REPL/Parser.hs
+++ b/src/Idris/REPL/Parser.hs
@@ -520,7 +520,7 @@ packageBasedCmd valParser cmd name =
 
 cmd_search :: String -> P.IdrisParser (Either String Command)
 cmd_search = packageBasedCmd
-  (P.typeExpr (defaultSyntax { implicitAllowed = True })) Search
+  (P.fullExpr (defaultSyntax { implicitAllowed = True })) Search
 
 cmd_proofsearch :: String -> P.IdrisParser (Either String Command)
 cmd_proofsearch name = do


Melvar:
Avoiding an error message by interpreting an erroneous input as something different is not a good idea. It still needs to be an error.

zjhmale:
@Melvar agreed, `!a` for search command is totally meaningless, but if it needs to be an error, why this issue confirmed?

zjhmale:
poke @jfdm

ahmadsalim:
@zjhmale It should not be an internal error, but a proper Idris error. That is why the issue is confirmed.

zjhmale:
@ahmadsalim ok i will get it right.

zjhmale:
poke

now there is a normal idris error

image

ahmadsalim:
@zjhmale That looks better! However, it be possible for the message to be more related to `!a`? Like `!-expressions are unsupported at top-level` or something like that?

ahmadsalim:
Sorry, pressed the wrong button quickly.

zjhmale:
@ahmadsalim want to produce error message like `-a` in search command, but after walk through `Delaborate.hs`, `AbsSyntaxTree.hs` and `TT.hs` still can not find a proper way to get something like `Can't find implementation for Bang Type` or `Can't find implementation for !Type`.

this is what the error message currently look, i copied your suggestion to the extra message.

image

david-christiansen:
Bang is not a type class, so it makes sense that it can't find an implementation for it.  Instead, I think that the ideal message here would be "Can't find implementation for Monad Type".

This is because the expression:
```
!a
```
should desugar to
```
{a : _} -> !a
```
which should desugar to
```
{a : _} -> a >>= (\x => x)
```

Entering this directly yields:
```
Idris> :search {a : _} -> a >>= (\x => x)
No location:When checking type of searchType:
Can't find implementation for Monad m

Possible cause:
        Type mismatch between
                m b (Type of do x <- a
                                x)
        and
                Type (Expected type)
```

which seems perfectly reasonable to me.

zjhmale:
@david-christiansen 

cool, `{a : _} -> a >>= (\x => x)` really make sense from the type signature of `(!)` operator.

so, should i still produce a `CantResolve` Err to reuse the `Can't find implementation for` part or just simply throw out a `Msg` Err to generate the ideal message?

i think desuger a `PTerm` and put it back to the elab procedure is a little bit complicated and there is no lucid access to do the desuger job, i can only find [delabSugared](https://github.com/idris-lang/Idris-dev/blob/master/src/Idris/Delaborate.hs#L79), but it receive a `Term` argument not `PTerm` 😂

zjhmale:
poke @david-christiansen @ahmadsalim @jfdm

ahmadsalim:
@zjhmale I unfortunately do not know the error message format very well, but I think you may use `ErrorReportPart` (https://github.com/idris-lang/Idris-dev/blob/12c70f50f271864bbdb7112dc2be3481f6758b80/src/Idris/Core/TT.hs#L234) with `FancyMsg` in some way.

david-christiansen:
The `ErrorReportPart` stuff is intended to be produced by Idris code doing reflection to make custom errors, not for errors that are built in to the compiler.

The right thing to do is to desugar prior to elaboration here. The appropriate operator is, so far as I can see on a quick read-through, `debindApp` in `Idris.DSL`. This is, however, already called from the term parser, so it probably just needs to be added to the right spot in the REPL parser for `:search`, which is in `cmd_search` in `Idris.REPL.Parser`.  Alternatively, you might experiment with having that call `fullExpr` instead of `typeExpr` and see what the differences are.

Hope this was useful!

david-christiansen:
In any case, the proposed change in 9a0e961 is really not the best solution, because if that message is produced, then it should indeed be an internal error, as it means that we've forgotten to call the desugaring at some point prior to elaboration.

zjhmale:
@david-christiansen that is really helpful! i did not think about the parser part to solve this issue 😂  thx!

and now the error message is even better i think.
image

zjhmale:
poke

ahmadsalim:
@zjhmale Thanks for the PR!

PR #3319, opened 2016-08-03
Use Word64 for hash.

### LINKED ISSUE #3316: Literal 4294967295 is out of the Int range
When compiling on Windows 32-bit, I get this warning.

```
src\Idris\Elab\Clause.hs:491:53: Warning:
    Literal 4294967295 is out of the Int range -2147483648..2147483647
```

melted:
0xffffffff isn't in Int on 32-bit systems.

Fixes #3316

### DIFF
--- a/src/Idris/Elab/Clause.hs
+++ b/src/Idris/Elab/Clause.hs
@@ -51,6 +51,8 @@ import Control.Monad.State.Strict as State
 import qualified Control.Monad.State.Lazy as LState
 import Data.List
 import Data.Maybe
+import Data.Word
+
 import Debug.Trace
 
 import qualified Data.Map as Map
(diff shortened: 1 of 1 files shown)


Melvar:
The comment specifically says it should be kept to 32 bits, are you sure you want to expand it rather than using `Word32`?

melted:
I could use Word32 and remove the mod, but then it would rely on the overflow behavior of Word32 to function.

Melvar:
I’m sorry, I didn’t read the code comprehensively enough. My instinct would be to switch to Integer if one wasn’t going to depend on the overflow behavior of a `Word*` type or use it as a bitfield. The current behavior actually doesn’t replicate a 32-bit overflow, and in theory using `Word64` might still overflow before the `mod` is done. So now I’m not sure what the best solution is.

melted:
Thinking about it, the overflow behavior for unsigned types is well-defined and sane, so using Word32 would be allright.

PR #3315, opened 2016-08-03
Enable unified diff for tasty tests

steshaw:
This is a fix for #3309.

This is likely to be a bit of a WIP. Let's see how this plays out on Windows/MSYS2.

PR #3327, opened 2016-08-04
Enable unified diff for tasty tests

### LINKED ISSUE #3309: Enable unified diff of actual/expected with tasty
With the porting of the testing infrastructure to `tasty` (#3231), its become difficult to see what the differences are when tests fail on the CI servers. This is because the comparison process no longer uses (GNU) diff. Ordinarily, you'd just plug in [`goldenVsFileDiff`](https://hackage.haskell.org/package/tasty-golden-2.3.1/docs/Test-Tasty-Golden.html#v:goldenVsFileDiff) in the appropriate place but it looks like there's a couple of issues which meant that a modified version of `goldenVsFile` was used.

The comment from `test/TestRun.hs`:

```haskell
      -- just pretend that backslashes are slashes for comparison
      -- purposes to avoid path problems, so don't write any tests
      -- that depend on that distinction in other contexts.
      -- Also compare CRLF and LF as equal, fixes a weird corner case
      -- on Mac where basic010 and reg039 produces CRLF
```

Hopefully it will be possible to address basic010 and reg039 by changing their `run` files to normalise the output as necessary. Then we can just plug in `goldenVsFileDiff` to call upon `diff -u`.

I'd like to start working on this 'cause I've run into the problem with #3281.

steshaw:
I'm opening a second PR to see if that keeps Travis happy. This should fix #3309 and replace the PR #3315.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,9 +1,7 @@
 version: 1.0.{build}
 init:
 - ps: >-
-    Get-WinSystemLocale
-
-    [Console]::OutputEncoding
+    c:\msys64\usr\bin\bash -l -c "diff --version"
 
     choco install ghc --version 7.10.3 | Out-Null
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,6 +2,7 @@ version: 1.0.{build}
 init:
 - ps: >-
     Get-WinSystemLocale
+
     [Console]::OutputEncoding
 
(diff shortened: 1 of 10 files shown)


melted:
Should should probably squash these commits.

And remove that line from appveyor.yml

steshaw:
@melted I will remove that bash version line from appveyor but over at #3315 as I'm closing this PR now as it didn't fix the Travis CI issue.

PR #3331, opened 2016-08-04
Fix stderr testing issue

### LINKED ISSUE #3323: stderr is suppressed in golden tests
The stderr of tests (i.e. the ./run script) isn't hooked up to the terminal. This has the effect that if things go very wrong (e.g. segfault), it's not obvious what's going on from the log. Particularly a problem for CI.

I believe @melted had a go at fixing this but `tasty` was a bit unhappy. @melted, could you say more about what you tried and what went wrong?

melted:
Don't query width when not a tty

This causes trouble when running non-interactively.

Also reinstate the stderr output now that that problem is solved.

Run diff on the test output if test fails. 

Fixes #3323

### DIFF
--- a/src/Idris/Output.hs
+++ b/src/Idris/Output.hs
@@ -21,6 +21,7 @@ import Idris.IdeMode
 
 import Util.Pretty
 import Util.ScreenSize (getScreenWidth)
+import Util.System (isATTY)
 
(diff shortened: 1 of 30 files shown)


steshaw:
It's nice to have stderr back. Thanks, @melted!

PR #3333, opened 2016-08-05
Keep filename args in the order specified on the cmd line

steshaw:
Currently filenames are processed in the reverse order to that specified on the command line.

This change also contains some tweaks that were suggested by ghc-mod.

### DIFF
--- a/src/Idris/CmdOptions.hs
+++ b/src/Idris/CmdOptions.hs
@@ -22,8 +22,6 @@ import Text.ParserCombinators.ReadP hiding (many, option)
 
 import Safe (lastMay)
 
-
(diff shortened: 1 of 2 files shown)


melted:
Restarting, seems to be just some js fluke error.

steshaw:
@melted I hope so. I've seen one of two of these in different tests. I'm running my own Travis (and Appveyor) now so sometimes I see it there and am able to restart the job. For idris-lang org, I don't seem to have permission to restart jobs.

Still it seems to be something worth looking into.

melted:
Seems to have worked out fine

PR #3335, opened 2016-08-05
Allow Travis to work when repo isn't named Idris-dev

steshaw:
I'd like to be able to run Travis CI against [my personal Idris repository](https://github.com/steshaw/idris). This is not possible at the moment because the `.travis.yml` file relies on the name of the repository. This PR removes that dependency so that it doesn't matter what your repository is called.

If you like, you can [view the successful build](https://travis-ci.org/steshaw/idris/builds/150003902) of this push to my personal repository.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ sudo: false
 
 language: c
 
+env:
+  global:
+    - PKGNAME=idris
+
 addons:
   apt:
     sources:
@@ -113,8 +117,8 @@ install:
 before_script:
   - cabal sdist
   - cd ..
-  - tar -xf Idris-dev/dist/idris*.tar.gz
-  - cd idris*
+  - tar -xf */dist/${PKGNAME}*.tar.gz
+  - cd ${PKGNAME}*
 script:
 ###
   - echo 'Configure...' && echo -en 'travis_fold:start:script.configure\\r'


melted:
I don't like the glob in the tar command, it's brittle. You can use pwd to stash the dir in an environment variable (if travis doesn't provide one).

steshaw:
I wasn't fond of the glob either! However, it isn't as brittle as we might think because you get a fresh container that doesn't contain other (unexpected) directories. So I understand where you are coming from but it doesn't seem to be a problem in practise.

Unfortunately, Travis only provides doesn't provide a convenient environment variable that is the name of the repository. It only provides TRAVIS_REPO_SLUG:

```
TRAVIS_REPO_SLUG: The slug (in form: owner_name/repo_name) of the repository currently being built. (for example, “travis-ci/travis-build”).
```

See https://docs.travis-ci.com/user/environment-variables/.

I'm having a think about your idea of using pwd though.

steshaw:
@melted, I gave your `pwd` idea a go. It looks like a winner. Thanks!

Melvar:
@steshaw, just curious, why is your repository not named “Idris-dev”?

steshaw:
@Melvar I simply prefer to call it `idris`.

Issue #2577, opened 2015-09-03
Consolidate tests


edwinb:
The tests currently take a long time to run and it would be nice if they were quicker while still testing everything. There is one cheap way to do this:

There's a lot of tests, especially in regxxx, which merely check some code type checks then quit. This involves starting up idris, loading the prelude, then typechecking, then quit and move on to the next test.

Instead, we could combine a lot of these tests into one source file, and save the repeated loading of the prelude.

There are probably other ways the tests could be consolidated - there hasn't really been much thought put into their organisation so it's worth checking to see if any repeat work, or don't really test anything useful.

david-christiansen:
I'll do this after PhD hand-in if nobody else does first :)

jmitchell:
I'd like to take a stab at the testing mini-project a try if there's still interest.

> Rewrite the test suite to use HSpec or similar to provide nice error reporting, categorization and modular integration of various testing tools.

I'm hoping HSpec or some similar tool can speed things up. HSpec may even let us execute tests in parallel.

It's specifically a Haskell testing library, so that probably means interfacing with Idris as a library rather than executable. Does that seem like a reasonable approach? We could keep a few top-level functional tests around to be safe.

david-christiansen:
The test suite can already run in parallel - e.g. `make test_c -j4` runs four at a time.

I think that it might be difficult to achieve a proper test using Idris as a library, but it might be worth a shot. It's a lot less work to start with just combining more tests into single source files so they execute faster.

jfdm:
Just to chime in, one thing to bear in mind is that AFAIK what constitutes 'Idris-as-a-Library' has not been finalised/debated nor how it should be presented 'as-a-library'.  This may affect how we can test it. 

Second, the current test suite also tests interactions with the idris executable, as well as the functionality of the compiler. These 'emergent' tests should be preserved as well. I have always perceived HSpec for code level testing and not executable interactions. I am quite happy to be corrected on this.

Just to say also, I too am not against using things like HSpec, but am a little more concerned about how best it should be used so that we do not lose any tests, emergent or otherwise, on the way.

jmitchell:
All great points. Whitebox testing is a separate concern with its own set of demands. The next obvious step is exactly what @edwinb suggested.

jfdm:
[...]

PR #3336, opened 2016-08-05
Consolidate regxxx tests into regression002

steshaw:
Consolidate non-executable tests from `test/regxxx` into `test/regression002`.

Non-executable meaning that they don't need to build and run an executable. This means that they either fail or succeed with the '--check' argument. However `regression002` uses the interactive REPL script idea mentioned by @jfdm.

Contributes to addressing #2577.


melted:
Still has conflicts. My stderr patch changed some expected files (not the content, just the pretty-printing)

steshaw:
@melted I already fixed those. This time I imagine I'm in conflict with my other patch for removing the need for specifying `extra-source-files`.

PR #3337, opened 2016-08-05
Consolidate layout001 tests

steshaw:
Contributes to addressing #2577.


melted:
Same thing as the previous patch, the expected file changed, so it needs to be updated and not just rebased.

steshaw:
@melted, no worries, I think this should do it.

PR #3338, opened 2016-08-05
Consolidate unique tests

steshaw:
Partially addresses #2577.


zjhmale:
@steshaw should we make it consecutive? i mean use `unique002` instead of `unique004`.

steshaw:
@zjhmale it was consecutive when I started this patch. There existed `unique001`, `unique002` and `unique003` so I created `unique004`.

There are plenty of other "holes" in the sequencing.

zjhmale:
@steshaw cool.

melted:
Thanks!

PR #3340, opened 2016-08-06
fix typo Becuase to Because

zjhmale:
originaly this is relate to #3313, but it is pending now, so maybe this little change can be merged first.


melted:
Thanks!

PR #3342, opened 2016-08-06
Remove `expect` package from travis configuration

steshaw:
The `expect` command is only used on Mac OS as a fallback for (lack of) `timeout`. On Mac `expect` is a standard utility.


melted:
This seems to be superseded.

steshaw:
Yes, by #3344. You just processed them out-of-order. No matter :).

PR #3343, opened 2016-08-06
Simplify basic010 and reg039 with portable `timeout`

steshaw:
Both of these tests need `timeout` and were falling back to a solution based on `expect` on Mac. Here we use a portable Perl solution instead.


melted:
Don't call the directory bin (it gives the wrong impression, people don't hack on stuff in bin directories). scripts, tools, helpers, something or other.

steshaw:
Our experiences with `bin` directories are quite different but I'm happy to call it `scripts` or `tools` or whatever.

melted:
Need to rebase it on master.

steshaw:
Yeah, keep getting bitten by `extra-source-files` as each pull-request is applied to `master`. I figured this might happen...

PR #3344, opened 2016-08-06
Optimise travis package installation

steshaw:
Only install the versions of ghc and cabal-install that are required for a particular spot in the matrix. This is based on the latest output generated by [multi-ghc-travis](https://github.com/hvr/multi-ghc-travis/) for `idris.cabal`.


melted:
Thanks for the PR!

PR #3345, opened 2016-08-06
A script to (re)generate multi-ghc-travis template.

steshaw:
A convenient script to check for improvements in the [multi-ghc-travis](https://github.com/hvr/multi-ghc-travis) project which generates Travis CI configuration files.


melted:
Don't put it in a directory called bin, call it scripts or tools.

PR #3347, opened 2016-08-07
Addition of new flag `--info` and other improvements.

jfdm:
+ New command line flag `--info` that displays information about the installation.
+ Further updates to `Idris.Info`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@
 * Idris' documentation system now displays the documentation for auto
   implicits in the output of `:doc`. This is tested for in `docs005`.
 
+* New command line flag `--info` that displays information about the installation.
 ## Miscellaneous updates
 
 * The test suite now uses [tasty-golden](https://hackage.haskell.org/package/tasty-golden). New tests must be registered in `test/TestData.hs`, as explained in the relevant `README.md`.
--- a/idris.cabal
+++ b/idris.cabal
@@ -189,6 +189,7 @@ Library
                 , Idris.IdrisDoc
                 , Idris.Imports
                 , Idris.Info
+                , Idris.Info.Show
                 , Idris.Inliner
                 , Idris.Interactive
                 , Idris.Output
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -29,11 +29,11 @@ main = do opts <- runArgParser
 runIdris :: [Opt] -> Idris ()
 runIdris opts = do
     runIO setupBundledCC
-    when (ShowLoggingCats `elem` opts)  $ runIO showIdrisLoggingCategories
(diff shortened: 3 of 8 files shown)


david-christiansen:
The --info command is a great idea.

I suppose I should add a machine-readable version, because things like
installed package lists are very useful for editor modes as well.

2016-08-07 16:04 GMT-04:00 Jan de Muijnck-Hughes <notifications@github.com>:

steshaw:
I've been wondering why `idris` is being maintained on 4 different versions of ghc. Why not drop support for anything < 7.10.3? This would have the nice side effects:

- probably could remove some #ifdefs from the code.
- the CI jobs would complete faster.
- overall, less of a burden.

I'd drop 7.10.3 as soon as 8.0.x works with `stack` too :).

david-christiansen:
Maintaining compatibility with as many versions of GHC as we can reasonably
do is good for a number of reasons:

1. Not everyone has the latest version of everything. Some people may be
running older operating systems with GHC packages that aren't the newest
and fanciest. Idris should be accessible to these people, if possible.

2. It is good at keeping us honest. We intentionally avoid the fanciest and
newest GHC extensions. Dependent types are hard enough without the
cognitive overhead of extremely fancy Haskell features.

3. It helps us be compatible with other implementations of Haskell, should
someone want to do that. Our dependencies probably render this moot, but
it's better to maintain portability at least here.

In general, the Haskell world does not take backwards compatibility very
seriously. But we're doing our best here for that. Generally, a slightly
longer wait on CI is a tiny, tiny price to pay, in my opinion.

steshaw:
@david-christiansen I disagree. I see you have your reasons but respectfully ask:

1. Who are these people who want to compile Idris on old operating systems? Who can't use a VM or binary package?
2. It doesn't make you dishonest to spend less time on maintenance. You could have a policy about fancy GHC extensions.
3. That's not going to happen, is it?

The entire Travis CI job takes 4 hours. That's _quite_ a long time. I hope to reduce it as it gets in my way even working on the low-hanging fruit :).

jfdm:
Arg failure... Will get that sorted. Looks like an Idris-idiom which translates to recent GHC versions is not applicable to older ones...

As for the compiler support, I think that is something best situated in another issue. This issue is not the place for it.

jfdm:
> I suppose I should add a machine-readable version, because things like
installed package lists are very useful for editor modes as well.

@david-christiansen I agree in that we nice to be able to query idris with this information outside of the command line options. One possible direction is to have an `:info <what>` command in the REPL/IDE capabilities that given a `<what>` (e.g. `history_file`, `library_directory`, `installed_ipkgs`) will return the relevant information.

PR #3349, opened 2016-08-08
Added information about bash completion.

jfdm:


david-christiansen:
Looks good to me!

jfdm:
I'm not too happy with where this is placed, however, where I think it *should* go would require some effort in restructuring the reference documentation to be more coherent. 

I like the approaches/structure of [Agda's documentation](http://agda.readthedocs.io/en/latest/) and [GHCs user guide](https://downloads.haskell.org/~ghc/8.0.1/docs/html/users_guide/index.html).  At somepoint over the next few months, and ideally before 1.0, I shall look at reorganising the documentation and providing documentation holes for people to fill in.

PR #3351, opened 2016-08-08
Add source location data to PUniverse

### LINKED ISSUE #2380: Idris doesn't highlight AnyType and UniqueType in source code
When type checking code containing AnyType and UniqueType, Idris leaves them unhighlighted.

BartAdv:
This enables highlighting of UniqueType, AnyType. Fixes #2380.


BartAdv:
Is the output for semantic highlighting tested?

david-christiansen:
No, unfortunately not. I think that the way to do that would be to write a parser for the output format and then check for the presence of certain highlights at certain source locations.

Thanks for doing this, by the way!

PR #3353, opened 2016-08-10
Clarify ambiguous error message.

jfdm:
+ Use of 'This' in an error message without clarifying what 'this' is
  is unhelpful.

+ When encountering an out of date `IBC` file, make sure to print the
  name of the out-dated `IBC` in question.


melted:
I'm pretty sure the long list of commits here and in other PRs comes from my little git mishap and isn't a problem.

To get Appveyor to work, it has to be rebased on master though.

jfdm:
yes, those commits weren't there before...

Issue #1768, opened 2014-12-05
Improvements to REPL Command completion.


jfdm:
REPL completion coverage requires improvment for various types of command. Improvements are required for: `NameArg`, `ModuleArg`, `NamespaceArg`, and `MetaVarArg`. This will require more information from an idris developer for intended improvements.

The affected function is: `completeCmd` in `src/Idris/Completion.hs`.

srenatus:
I found the current completion quite confusing, too, as it apparently completes _functions that it knows about_, even when prefixed with a module name: [example terminal session](https://asciinema.org/a/696l54hhf8gqhxel4vhjihg2e). (I think this issue covers what I came here to whine about :wink:)

BartAdv:
Started working on some of those issues in https://github.com/BartAdv/Idris-dev/tree/repl-improvements

- [x] honor namespace prefixed identifiers (issue @srenatus mentions)
- [x] don't show more completions if identifier is entered fully (for names/metavars)
- [ ] module completion
- [ ] namespace completion

BartAdv:
It would also be doable to allow completing identifiers namespaced with module aliases, however, this would require changes to IBC, as aliases seem not to be stored there. Would that be worth the hassle? There's ticket for IBC changes (#1577), but it's meant for more important things.

melted:
The ibc format is changed routinely as data structures that goes into it are changed. Don't have any respect for it.

Issue #1767, opened 2014-12-05
Respect module imports when completing user-visible names.


jfdm:
When obtaining user-visible names using the `REPL`, the module imports must be respected. This will require more information from an idris developer for intended fixes.

The affected function is: `names` in src/Idris/Completion.hs.

Issue #1209, opened 2014-05-07
Private functions show up in autocomplete and work in :t


david-christiansen:
They should not.

BartAdv:
How would one determine what's the current module? Been poking around the Idris state and I hadn't spotted anything.

steshaw:
I'm taking a look at this issue (in addition to #1126).

On the completion side of things, I've tracked this down to `Idris.Completion`. We would appear to need a little bit of extra filtering in `names`. It looks like you ought to be able to access the Accessibility of the name and look for ones tagged `export`. However, I've noticed that everything seems to be tagged `public export`. For example, I have a simple example:

```idris
module Foo

foo_private : Nat -> Nat
foo_private n = n

export
foo_visible : Nat -> Nat
foo_visible n = n
```

```idris
module Bar

import Foo

bar_private : Nat -> Nat
bar_private n = n

export
bar_visible : Nat -> Nat
bar_visible n = n
```

I've hacked a special ":context" command that prints out the Context for anything starting with "foo" or "bar":

```sh-session
$ idris Bar.idr
*Bar> :context
fromList [
(bar_private,fromList [(Bar.bar_private,(Case: Prelude.Nat.Nat -> Prelude.Nat.Nat []
...
Inlinable,False,public export,Total,EmptyMI))]),
(bar_visible,fromList [(Bar.bar_visible,(Case: Prelude.Nat.Nat -> Prelude.Nat.Nat []
...
Inlinable,False,public export,Total,EmptyMI))]),
(foo_private,fromList [(Foo.foo_private,(Case: Prelude.Nat.Nat -> Prelude.Nat.Nat []
...
Inlinable,False,public export,Total,EmptyMI))]),
(foo_visible,fromList [(Foo.foo_visible,(Case: Prelude.Nat.Nat -> Prelude.Nat.Nat []
...
Inlinable,False,public export,Total,EmptyMI))])]
```

Note that each function is listed as `public export`. Perhaps this is intended in the REPL? I would appreciate any tips. Perhaps there's a bug in the module loading code?

eraserhd:
@steshaw If you look at `Idris.IBC.processAccess`, you'll notice it sets all functions loaded to `Public` when the phase is `IBC_REPL True`.

steshaw:
Thanks for the tip, @eraserhd. It's been a while but I'm taking a look. 😃

steshaw:
[...]

PR #3356, opened 2016-08-10
REPL improvements

BartAdv:
Issues from #1768:

- [x] honor namespace prefixed identifiers (issue @srenatus mentions, and the issue described in #1767 if I'm right)
- [x] don't show more completions if identifier is entered fully (for names/metavars)
- [ ] module completion
- [x] namespace completion

To implement module completion I think we'd have to bump the `directory` to `1.2.5.0` as this brings various helpers to list the content of directories, as far as I know this is the only way to get the modules that can be imported.

One possible improvement would be to also honor the module aliases, but that would require quite an amount of changes - not sure if it's needed that badly.

I was thinking about tackling #1209, but I'm not sure yet how determine currently loaded module(s).

### DIFF
--- a/src/Idris/Completion.hs
+++ b/src/Idris/Completion.hs
@@ -7,7 +7,7 @@ Maintainer  : The Idris Community.
 -}
 module Idris.Completion (replCompletion, proverCompletion) where
 
-import Idris.Core.Evaluate (ctxtAlist)
+import Idris.Core.Evaluate (definitions, ctxtAlist)
 import Idris.Core.TT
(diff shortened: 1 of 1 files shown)


melted:
looks like a travis fluke on osx, I restarted that target.

melted:
As @jfdm has a PR that touches the REPL I'll lesve it to him to merge this.

melted:
Ah right, this didn't touch REPL.hs...

Thanks for the PR!

PR #3357, opened 2016-08-10
Adjust for Appveyor's Chocolatey update

melted:
Download GHC from haskell.org instead.


melted:
Appveyor-specific so let's not wait on Travis.

PR #3362, opened 2016-08-12
Tidying of `idrisMain` (fixes #1454).

### LINKED ISSUE #1454: Tidying of `idrisMain`
`Idris` is a library that has a single front-end. However, `Idris` actually has several different front-ends that can be used to interact with `Idris` code. A nominal list is:

+ The `REPL`
+ The `iPKG` build tool
+ The documentation builder.
+ The `IDESLAVE`

Issue #1126 has introduced the idea that these different front ends should be wrapped in their own separate executables. That is, there should be different cabal projects for the front ends. How much of the logic for each of the different frontends should be extracted (or left in) from the `idris` library to these executables is a problem for whoever tackles issue #1126.

However, related to this problem is `idrisMain`.  It has been mentioned on the `IRC` channel that `idrisMain` is a little convoluted and requires *some attention*. If the *right* sort of attention is given to `idrisMain` it might make tackling issue #1126 easier...it may also help improve how the rest of the `Idris` frontend works.

Feel free to discuss what *attention* needs to be given to `idrisMain`.

jfdm:
Introduction of a new module `Idris.Main` that contains the code
required to launch and run an interactive Idris session---compilation,
REPL et cetera. It has been spun out of `Idris.REPL`; common code has
been spun out from both into `Idris.ModeCommon`.

With `Idris.Main` comes a new entry point for Idris and presents
developers with a more unified API for running Idris programs. A side
affect of this approach is that the imports for codegens must be
changed to import `Idris.Main` and not `Idris.REPL`.

TBH the imports for the new modules need pruning, I did a dump all just to ensure things would work. I've stuck an `in-progress label` to reflect that I need to sort the imports. Otherwise, tests aside this is good to go.

### DIFF
--- a/codegen/idris-codegen-c/Main.hs
+++ b/codegen/idris-codegen-c/Main.hs
@@ -3,8 +3,7 @@ module Main where
 import Idris.Core.TT
 import Idris.AbsSyntax
 import Idris.ElabDecls
-import Idris.REPL
-
+import Idris.Main
 import IRTS.Compiler
 import IRTS.CodegenC
 
--- a/codegen/idris-codegen-javascript/Main.hs
+++ b/codegen/idris-codegen-javascript/Main.hs
@@ -3,8 +3,7 @@ module Main where
 import Idris.Core.TT
 import Idris.AbsSyntax
 import Idris.ElabDecls
-import Idris.REPL
-
+import Idris.Main
 import IRTS.Compiler
 import IRTS.CodegenJavaScript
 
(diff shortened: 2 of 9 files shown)


melted:
You can pass -fwarn-unused-imports to GHC to get a list of the ones that aren't needed.

jfdm:
> You can pass -fwarn-unused-imports to GHC to get a list of the ones that aren't needed.

Ah cheers for that, turns out when using intero mode and that option as a file pragma you can automatically resolve the warnings and remove the imports!

jfdm:
Merging as no news is good news.

PR #3372, opened 2016-08-22
Added `--sourcepath` option

### LINKED ISSUE #3324: Introduce a switch to configure source file search paths, similarly to `-i` for IBC files
I am probably missing something obvious but when I try to reorganize a shallow hierarchy of files into subdirectories, for example

`Foo/Foo.lidr`:
```
> module Foo.Foo
> data Foo
```

`Foo/Operations.lidr`:
```
> module Foo.Operations
> import Foo.Foo
> public export toNat : Foo -> Nat
> toNat f = Z
```

`Bar/Bar.lidr`:
```
> module Bar.Bar
> data Bar
```

`Bar/Operations.lidr`:
```
> module Bar.Operations
> import Bar.Bar
> import Foo.Foo
> import Foo.Operations
> toFoo : Bar -> Foo
> toNat : Bar -> Nat
> toNat = Foo.Operations.toNat . toFoo
```
dependencies are not reloaded automatically. Thus, for instance
```
nicola@cirrus:~/gitlab/IdrisLibs/tmp/Bar$ idris -i .. --check Operations.lidr
```
fails with something like
```
Loading ./Operations.ibc failed: Module needs reloading:
	SRC : "../Foo/Operations.lidr"
	Modified at: 2016-08-04 12:08:32.335160527 UTC
	IBC : "../Foo/Operations.ibc"
	Modified at: 2016-08-04 12:04:27.099847781 UTC
nicola@cirrus:~/gitlab/IdrisLibs/tmp/Bar$
```
if the time stamp of `../Foo/Operations.ibc` happens to lag behind that of `../Foo/Operations.lidr`. If `../Foo/Operations.ibc` does not exists, one gets
```
nicola@cirrus:~/gitlab/IdrisLibs/tmp/Bar$ idris -i .. --check Operations.lidr
Loading ./Operations.ibc failed: Must be an ibc
```
For files with hundreds of dependencies, this behaviour is less than useful and one would like dependencies to be reloaded automatically as it happens when files are organized in a shallow hierarchy of modules.

Are there specific options to force Idris to reload files whose *.ibc files are inexistent or out-of sync?

ahmadsalim:
This allows adding more search paths than just `.` when looking for source files. Fixes #3324. 

Comments are welcome!


jfdm:
When this is finished, @ahmadsalim remember to update the docs and the man page.

ahmadsalim:
@jfdm Thanks, will do.

Issue #3375, opened 2016-08-25
class used in error messages instead of interface


arianvp:
How to reproduce:

```
data EmptyU = VOID

data Unifies : a -> b -> Type where
  ItUnifies : Unifies a b

interface DecUnifies a where
  decUnifies : a -> a -> Dec (Unifies a a)

DecUnifies EmptyU where
  decEq VOID VOID = Yes ItUnifies
```

Expected behaviour:
```
/home/arian/Projects/thesis2/src/Servis/API.idr:24:1:
Servis.API.decEq not a method of interface DecUnifies

```

Current behaviour:
```
/home/arian/Projects/thesis2/src/Servis/API.idr:24:1:
Servis.API.decEq not a method of class DecUnifies

```

jfdm:
Looks like a few error messages need updating. These appear to be from in `Idris.Elab.Class` and friends. Easy fix, buy the looks of it.

david-christiansen:
This would also be a great first project for an aspiring Idris contributor :-)

zjhmale:
@jfdm actually, i think this message is from `Idris.Elab.Instance`

jfdm:
maybe, there were a couple other places that came up when i grep'd for the error message. Any fixes should seek to fix *all* the uses of class (log messages and all) that should be updated.

zjhmale:
>maybe, there were a couple other places that came up when i grep'd for the error message. Any fixes should seek to fix all the uses of class (log messages and all) that should be updated.

ahmadsalim:
@zjhmale You could try using (the silver searcher)[https://github.com/ggreer/the_silver_searcher] and then search for `class` and `instance` and fix the right places. It is almost just as simple as a search/replace with a smart person to determine whether it makes sense :).

jfdm:
I think for now the error messages, the other names can come later.

zjhmale:
@jfdm ok.

zjhmale:
@ahmadsalim thx!

PR #3379, opened 2016-08-26
use inferface instead class or type class in docs

zjhmale:
relates to #3375


src/Idris/Core/TT.hs
```diff
@@ -155,7 +155,7 @@ fcIn (FC fn1 (sl1, sc1) (el1, ec1)) (FC fn2 (sl2, sc2) (el2, ec2)) =
   (sl1 == sl2 && sc1 > sc2 || sl1 > sl2) &&
   (el1 == el2 && ec1 < ec2 || el1 < el2)
 
--- | Ignore source location equality (so deriving classes do not compare FCs)
+-- | Ignore source location equality (so deriving interfaces do not compare FCs)
```
Melvar: Since this is talking about Haskell code, the replacement is incorrect.

Melvar:
I see a few minor problems due to this being an unchecked search-and-replace. One instance of “type class” split over lines thus becoming “type interface”, several instances of “a (type) class” becoming “a interface” instead of “an interface”.

zjhmale:
all checked 😃 
>I see a few minor problems due to this being an unchecked search-and-replace. One instance of “type class” split over lines thus becoming “type interface”, several instances of “a (type) class” becoming “a interface” instead of “an interface”.

zjhmale:
@ahmadsalim @melted @Melvar please review.

libs/oldeffects/Effects.idr
```diff
@@ -235,7 +235,7 @@ eff {xs = [l ::: x]} env (l :- prog) k
    = let env' = unlabel env in
          eff env' prog (\p', envk => k p' (relabel l envk))
 
--- yuck :) Haven't got type class instances working nicely in tactic
+-- yuck :) Haven't got interface instances working nicely in tactic
```
ahmadsalim: Should be `interface implementations`

src/Idris/AbsSyntax.hs
```diff
@@ -434,7 +434,7 @@ pop_estack = do i <- getIState
     where ptail [] = []
           ptail (x : xs) = xs
 
--- | Add a class instance function.
+-- | Add an interface instance function.
```
ahmadsalim: Again, "interface implementation". I would try to check whether this happens other places as well.

ahmadsalim:
@zjhmale I added some comments, but otherwise it looks mostly fine.

src/Idris/AbsSyntax.hs
zjhmale: right, we now use `implementation` keywords to implement interface. i'll check this out.

zjhmale: so if change `interface instance` to `interface implementation`, should we change all the `instance` to `implementation`?

zjhmale:
@ahmadsalim so if now change `interface instance` to `interface implementation`, should we change all the `instance` to `implementation` in the rest of the code base?

ahmadsalim:
@zjhmale Yes, that would be useful to do I think.

zjhmale:
i'll work on it.
>@zjhmale Yes, that would be useful to do I think.

zjhmale:
poke @ahmadsalim @Melvar

ahmadsalim:
It looks good to me.

ahmadsalim:
Thanks for the PR!

PR #3386, opened 2016-08-27
remove compiled file before build

### LINKED ISSUE #3385: Wired CI error
the CI servers were just unhappy with #3376, #3379 and #3381 that

```
Type checking ./System/Concurrency/Raw.idr
Can't find import System/Concurrency/Channels
make[1]: *** [build] Error 1
make[1]: Leaving directory `/home/travis/build/idris-lang/idris-0.12.2/libs/base'
make: *** [build] Error 2
make: Leaving directory `/home/travis/build/idris-lang/idris-0.12.2/libs'
```

is this caused by `.ibc` files which built up use previous version of Idris?

maybe should add a script to clean all the ibc files before build like

```
find . -name "*.ibc" -type f -delete
```

i use this command with locally build cause it always complains `ibc file build with old Idris` build error.

zjhmale:
Fixes #3385

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -118,6 +118,7 @@ script:
   - echo -en 'travis_fold:end:script.configure\\r'
 ###
   - echo 'Build...' && echo -en 'travis_fold:start:script.build\\r'
+  - find . -name "*.ibc" -type f -delete
   - cabal build
   - echo -en 'travis_fold:end:script.build\\r'
 ###


david-christiansen:
I don't think this is the problem. Travis makes a fresh VM for each test run, so stale IBCs will not be present.

zjhmale:
agree, so i closed it myself 😅 
>I don't think this is the problem. Travis makes a fresh VM for each test run, so stale IBCs will not be present.

PR #3388, opened 2016-08-28
change instance to implementation in string literal

zjhmale:
discussed at #3379


ahmadsalim:
@zjhmale Could you consider merging your 3 PRs into one? Since they are all related it would be easier I think to merge (since now each time I merge, you get a merge conflict on your other PRs)

zjhmale:
@ahmadsalim cool, i'll do that.

PR #3400, opened 2016-09-01
Fix deprecation warnings in benchmarks

enolan:
Most of this is from purification, two are switching from the %assert_total directive to the assert_total function.


ahmadsalim:
Thanks for the PR!

PR #3404, opened 2016-09-05
change Instance to implementation

zjhmale:
combine #3388 #3389 #3390, discussed at #3388.


libs/prelude/Language/Reflection.idr
```diff
@@ -396,9 +396,9 @@ mutual
   withNInj Refl = (Refl, Refl)
 
   private
-  instanceNInj : (InstanceN n xs = InstanceN n' ys) -> (n = n', xs = ys)
-  instanceNInj Refl = (Refl, Refl)
-
+  implementationNInj : (ImplementationN n xs = ImplementationN n' ys) -> (n = n', xs = ys)
```
ahmadsalim: Is it possible to still keep the old `instanceNInj` as a synonym to the renamed ones, but mark them as deprecated instead? The same with other renamed Idris functions, please.

This is in order to not break user code unnecessarily.

src/Idris/Parser/Expr.hs
```diff
@@ -409,7 +409,7 @@ simpleExpr syn =
             try (simpleExternalExpr syn)
         <|> do (x, FC f (l, c) end) <- try (lchar '?' *> name)
                return (PMetavar (FC f (l, c-1) end) x)
-        <|> do lchar '%'; fc <- getFC; reserved "instance"; return (PResolveTC fc)
+        <|> do lchar '%'; fc <- getFC; reserved "implementation"; return (PResolveTC fc)
```
ahmadsalim: Please also, keep the old `%instance` for now in the parser but mark it as deprecated as well.

zjhmale: ok, i'll do that.

libs/prelude/Language/Reflection.idr
zjhmale: but `instanceNInj` is a private function, i mean, Idris users can not use it directly, right?

ahmadsalim: @zjhmale Ah I didn't notice that :). But please do it, if there were name changes in public functions. If it is possible to deprecate constructors as well, please do it for `InstanceN`/`ImplementationN` as well.

ahmadsalim: Also, please add a note to README.

ahmadsalim:
Btw, you are also welcome to add yourself to `CONTRIBUTORS` if you have not already.

libs/prelude/Language/Reflection.idr
zjhmale: @ahmadsalim thx, i'll wrap it up then :smile:

zjhmale:
cool! i'm also working on a [vscode extension for idris](https://github.com/zjhmale/vscode-idris), based on some design from [atom extension](https://github.com/idris-hackers/atom-language-idris)
>Btw, you are also welcome to add yourself to CONTRIBUTORS if you have not already.

libs/prelude/Language/Reflection.idr
zjhmale: @ahmadsalim one thing is confusing me that if a deprecations occurs in Idris code i can use a `%deprecated` annotation and if a deprecation come from parser i can also use `<|>` to tackle with it, but how can i mark the deprecated constructors in Haskell data types like `InstanceN` you mentioned, i mean, these constructors are internal and will not be exposed to users, there is also not exist something like `deprecated` anntation in Haskell, so maybe simply add note to `CHANGELOG` is enough? i'll address deprecations in Idris lib files and Parser first.

ahmadsalim:
So @zjhmale thanks for the effort. I will let few of the other @idris-lang/contributors comment on this before we merge.

ahmadsalim:
@idris-lang/commiters as well :).

zjhmale:
@ahmadsalim thx! after merge this we can finally finish the long journey 😄

src/Idris/Parser.hs
```diff
@@ -937,52 +937,52 @@ interface_ syn = do (doc, argDocs, acc)
...
 @
 
 @
-InstanceName ::= '[' Name ']';
+ImplementationName ::= '[' Name ']';
 @
 -}
-instance_ :: Bool -> SyntaxInfo -> IdrisParser [PDecl]
-instance_ kwopt syn
+implementation_ :: Bool -> SyntaxInfo -> IdrisParser [PDecl]
```
Melvar: I think this doesn’t need an underscore anymore; that was there because `instance` is a Haskell keyword.

zjhmale: fair enought, i will fix it!

libs/prelude/Prelude/Cast.idr
```diff
@@ -5,7 +5,7 @@ import public Builtins
 
 %access public export
 
-||| Interface for transforming an instance of a data type to another type.
+||| Interface for transforming an implementation of a data type to another type.
```
jfdm: Use of 'implementation' here is incorrect. Can the change here be reverted.

libs/prelude/Language/Reflection.idr
jfdm: Have the results of this conversation be resolved? That is renaming private members and constructors?

CHANGELOG.md
```diff
@@ -27,8 +27,11 @@
   `Prelude.File` and to provide a 'better' API.
 * `natEnumFromThen` and `natEnumFromTo` have been updated to correctly calculate reverse ranges. Range syntax `[a,b..c]` now can be used again to generate reverse ranges.
 * `divBN` and `modBN` now can only be used for unsigned numbers.
-
 * `return`, which has been an alias for `pure` for many releases, is now deprecated.
+* `InstanceN` is deprecated, use `ImplementationN` instead.
+* `InstanceCtorN` is deprecated, use `ImplementationCtorN` instead.
```
jfdm: I would be nicer here to have a top level not of the renaming with these bullet points being presented as a sublist.

jfdm:
I've had a look at this, there are one or two minor issues that I have commented on inline. Once these have been resolved, I see no obvious reason to not merge.

libs/prelude/Language/Reflection.idr
ahmadsalim: @jfdm I am not sure what the policy is. But I guess it is OK, since they're private?

jfdm: My rule of thumb on this would be: 'make it consistent'.

zjhmale: @jfdm i'll keep `instanceNInj` deprecated.

libs/prelude/Prelude/Cast.idr
zjhmale: got it, here is trans type to type.

zjhmale:
@jfdm i believe all the issues are fixed.

ahmadsalim:
@zjhmale Thanks, I will merge now then.

PR #3406, opened 2016-09-05
Add scanl and scanl1 implementations to List

faineance:
Adding implementations for `scanl` and `scanl1` to List.
Recently found the need for these and was surprised they weren't in Prelude.


jfdm:
@faineance Thanks for the contribution it is much appreciated. Sometimes functions are not included in the Prelude because someone has forgotten about them, and their uses hasn't been needed by those who program in Idris, or achieve the same functionality by a different means.

I see no obvious reason to block this merge, aside from a lack of documentation. Is it possible that you can add some IdrisDoc to the added functions. If you are unsure of the syntax for IdrisDoc there are examples within the modules in `libs/` for example `Prelude` and `Base`. [More information is also available on our docs](https://www.idris.readthedocs.io).

jfdm:
Thanks. The docs look great. I am going to squash these commits as one was failing, and want to ensure that the build history is 'clean'. 

Again thanks for the contribution.

PR #3409, opened 2016-09-07
Use System.FilePath.searchPathSeparator for $PATH

steshaw:
Used `searchPathSeparator` rather than a hard-coded ";" on windows and ":" on POSIX.

### DIFF
--- a/src/Util/System.hs
+++ b/src/Util/System.hs
@@ -37,7 +37,8 @@ import System.IO.Error
 
 #ifdef FREESTANDING
 import Tools_idris
-import System.FilePath (isAbsolute, dropFileName)
+import Data.List (intercalate)
+import System.FilePath (isAbsolute, dropFileName, searchPathSeparator)
 import System.Directory (doesDirectoryExist)
 import System.Environment (getEnv, setEnv, getExecutablePath)
 #endif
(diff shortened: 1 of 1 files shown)


steshaw:
@melted, appreciate if you had a glance at this.

src/Util/System.hs
```diff
@@ -109,11 +110,10 @@ setupBundledCC = when hasBundledToolchain
                                     if absolute
                                        then tcDir
                                        else dropFileName exePath ++ tcDir
-                        let pathSep = if isWindows then ";" else ":"
                         present <- doesDirectoryExist target
-                        when present
-                            $ do newPath <- return $ target ++ pathSep ++ path
-                                 setEnv "PATH" newPath
+                        when present $ do
+                          newPath <- return $ intercalate [searchPathSeparator] [target, path]
```
ahmadsalim: Would it not be better to use `intersperse` here?

steshaw: I don't think so. That would require `concat (intersperse [searchPathSeparator] [target, path])`.

ahmadsalim: Nevermind, it wouldn't work here.

ahmadsalim: Yeah, my mistake!

steshaw: np

melted:
Looks good!

PR #3410, opened 2016-09-07
Tidy up `main/Main.hs`

steshaw:
Just tidying up Main.hs while I come to grips with what is necessary for #1126.


steshaw:
:( nasty conflicts

PR #3411, opened 2016-09-07
Refactor `main/Main.hs`

steshaw:
Worked on some tweaks for `main/Main.hs` while getting my head around what's required for #1126.

- Improved clarity – particularly to where `idrisMain` gets invoked. This was previously buried in the processing of the package commands.
- Remove duplication.

I've got some ideas of how to do option processing better in the future but this is an improvement for now.

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -1,8 +1,7 @@
 module Main where
 
-import System.Exit ( exitSuccess )
-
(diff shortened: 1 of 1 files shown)


jfdm:
I see no problem with this. Thanks for the contributions!

PR #3412, opened 2016-09-07
Add note about Unicode operators to official FAQ

steshaw:
Added the note about Unicode operators from the Unofficial FAQ to the official one.

### DIFF
--- a/docs/faq/faq.rst
+++ b/docs/faq/faq.rst
@@ -253,36 +253,33 @@ Will there be support for Unicode characters for operators?
 
 There are several reasons why we should not support Unicode operators:
 
--  It's hard to type (this is important if you're using someone else's
-   code, for example). I know various editors have their own input
-   methods, but you have to know what they are.
--  Not every piece of software easily supports it. It doesn't render
-   properly on my phone email client, for example, occasionally messes
-   up the terminal I run my IRC client in, doesn't always render
-   properly in my browser, etc. All of these are things I can fix but to
-   have to do so as a barrier to entry to using a programming language
-   doesn't seem to be a good idea to me.
--  Even if we leave it out of the standard library (which we will in any
-   case!) as soon as people start using it in their library code, others
-   have to deal with it.
--  Too many characters look too similar. We had enough trouble with
-   confusion between 0 and O without worrying about all the different
-   kinds of colons and brackets.
--  There seems to be a tendency to go over the top with use of Unicode.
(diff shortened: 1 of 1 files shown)


ahmadsalim:
I think there was a reason that it was in the unofficial one, but I can't recall what.

steshaw:
@jfdm said something like "pull requests appreciated" on IRC. I simply obliged :). If this does get accepted into the official FAQ, perhaps we should remove it from the unofficial one on the wiki though?

ahmadsalim:
@steshaw I mean if no one minds, then it is fine. Yeah, it would be good to have deduplication between the official and unofficial FAQ

jfdm:
My only concern with this FAQ is the use of first person narration; which was okay in the *Unofficial FAQs*. It *might be* better to keep the same narrative voice used through-out. 

Speaking generally about moving FAQs from unofficial to official. I think we should keep it more at the obvious Questions, and more technical ones in the unofficial FAQs. Otherwise we will need to start categorising the official FAQs.

steshaw:
@jfdm I'll see if I can switch to the narrative voice.

I reckon there should be a QA about type class coherence and probably a separate on on type class resolution. It sounds like this would belong in the Unofficial FAQ or even "Idris for Haskellers" and/or "Idris coming from Scala" (which probably doesn't exist yet).

jfdm:
Ah much better. Sorry for being picky about it, but nice to have the *Official* FAQs being consistent and less of a rag tag bunch of things.

> I reckon there should be a QA about type class coherence and probably a separate on on type class resolution. It sounds like this would belong in the Unofficial FAQ or even "Idris for Haskellers" and/or "Idris coming from Scala" (which probably doesn't exist yet).

PRs/Addition to the wiki are welcome...although remember we have interfaces and not type classes ;-)

As for a "Idris coming from Scala" again PRs welcome. I've often thought about having an *Idris coming from X* series in the docs and migrating things over from the github wiki. But that can wait for another day.

steshaw:
No worries, @jfdm. It's nice to keep the official docs more professional.

If I put something together for the wiki on the common type class questions I'll be sure to mention that they're called interfaces in Idris :).

Melvar:
@steshaw I believe the point of the different name is to signal that they’re not meant to be the same thing as Haskell typeclasses. If you’re willing to indulge this, don’t introduce them as a variation of the same thing, introduce them as a similar thing.

steshaw:
@Melvar Thanks, that's a good approach.

Issue #1227, opened 2014-05-14
Consistent indentation with generated code?


david-christiansen:
The following code was created by maximally using the interactive code generation features for the instance body and the `with` block:

```
module Empty

data Empty : Type where

instance Uninhabited Empty where
    uninhabited x = ?Uninhabited_rhs_1

foo : Nat -> Nat
foo k with (_)
  foo k | with_pat = ?foo_rhs
```

Note that the instance method is indented four spaces, while the with block is indented two spaces. It would seem to me that these should be consistent, and perhaps even configurable.

What are the community's thoughts on this?

steshaw:
I'm looking into this. I like my 2-space indent as well :). Looks like I've got it working. Just need to follow up on my (mis)understanding of `idris --client`. I've been testing with `idris-vim`. I'll try to push it up for a code review soon.

PR #3415, opened 2016-09-08
Enable configuration of indentation in interactive mode

steshaw:
This PR is for #1227.

I've made the following variables getable/settable via the REPL and stored in IState (in a new sub-record):
- `:makewithindent` -- affects the indentation with makeWith. i.e. \w in `idris-vim`.
- `:addclauseindent` -- affects the indentation with addClause (it only apply to interface implementations). i.e. \b in `idris-vim`.

The defaults are 2 for both. I am able to override the variables by adding the following lines to `~/.idris/repl/init`:

```
:makewithindent 2
:addclauseindent 4
```

I then use vim by first invoking `idris foo.idr` and then using the `:edit` command from within the REPL. However, I am not able to change the values in the REPL itself and then do `:edit` and have those indentations used. I figure that I misunderstand the process architecture (or something). Help :)

Also, I haven't tested with the IDE mode from (say) Emacs. I'll be setting up Emacs sometime soon to do so.

Any advice on more appropriate ways to configure these variables, where to store them, why I'm confused about not being able to affect the variables once the REPL has started (I should mention that getting the variables does display the expected value, it's just that the indentation in `:edit` does not). There should probably be some tests and docs too.


Melvar:
I think that perhaps there should be only one top-level indentation configuration command which takes an indentation category, rather than one independent command per category, but perhaps others have a different opinion.

steshaw:
I think I'd be happy with something like that except that I don't know what a configuration category is.

I'm happy to have 2 space indentations more-or-less everywhere. There are others — @edwinb and @jfdm — who want to be able to configure indentation as I've specified above (with 2 for with clauses and 4 for add clauses). See the discussion in #1227.

Melvar:
All I meant is that it gets an identifier that identifies which kind of indentation is to be configured. So just that the example would look something like
```
:indent with 2
:indent method 4
```
or possibly with something longer and more explanatory than `with` and `method`, like `withSubclause` and `interfaceMethod`.

steshaw:
@Melvar oh, I do like that! I was merely following along with the other commands that are already present. A potential downside is that it will no doubt require revisiting the REPL completion (which I haven't seen but presume works on a single word at the moment). It's a much nice UI though.

Melvar:
Yeah, it would probably require some changes, and turning all REPL commands into such a style would be a major undertaking. Perhaps even pulling all configuration into `:set` which at the moment only sets boolean properties to true.

So in the interest of getting the functionality in, perhaps make these separate commands for now.

jfdm:
I agree with @Melvar that in the long run a more consistent REPL interface would be nice to have, but for the immediate future the suggested changes would be nice.

david-christiansen:
If you take a look at the completion for the color-setting commands, you'll find that it has an interface similar to what you're proposing here. So that's perfectly fine.

On the other hand, I think that a command line option, ipkg field, or a %indentation line in a file is a better way to set this option. Typically, a particular project will have a coding standard, and one user might work on multiple projects. A user-level setting is nice, but a project- or file-level setting seems better for this one.

steshaw:
Hi @Melvar, I just wanted to let you know that I did notice the `:set` command but confirmed as you did that it was for things that are (boolean valued) flags. Also, these flags are all related to your experience in the REPL (AFAICT) and not in an editor/IDE. For this reason, I was a bit reluctant to expand on `:set` for use with the indentation options. I do really like your `:indent with 2` suggestion.

## Completion

Hi @david-christiansen, thanks for pointing me at the `:colour` commands. I played around in the REPL and it the completion works nicely. After a quick look at the code, it looks like it will need some work to make it reusable to make an `:indent` command. There's some duplication between the parser and completer for `:colour` as the parser knows what to expect next. Any thoughts as to how to combine those together or if that path would be worth the trouble? It might be ideal to have a description of the REPL commands from which a parser and a completer can be generator. That sort of thing sounds like a much more moderate effort though.

```
Idris> :colour
(input):1:8: error: unexpected
    EOF, expected: "boundvar",
    "data", "function", "implicit",
    "keyword", "off", "on",
    "postulate", "prompt", "type",
    end of "colour"
:colour<EOF>
```

In the following REPL, I'm trying to show that the parser knows what tokens to expect after `:colour <x>` but it looks like there are bugs here or I don't understand something:

```
Idris> :colour
on         off        keyword    boundvar   implicit   function   type       data       prompt     postulate
Idris> :colour keyword
Idris> :colour keyword HOLE
Idris> :colour keyword ??? ???
```

I'd like to be a good boy scout and address these UX issues with the REPL UI but I'm wondering if I'll get caught up in a world of death by a thousand cuts... I also don't want to annoy uses and have a version of Idris released that uses `:makewithindent` only to follow that up in the next release with `:indent with x`. So I'd like to address this issue in this PR if it's not too much trouble.

## Configuration

@david-christiansen, I like your idea of having the indentations options in the `.ipkg` file.

When I was initially working on this, I went looking for a file like `.ghci`. This is how I found `~/.idris/repl/init`. Having a global user setting is nice but only so long as we can have local-based setting too. I was thinking to add a `./.idris/repl/init` file like GHC's `.ghci`. The REPL would run the global one before the local one. However, putting these options in the `.pkg` file seems like an even better idea. I'd rather do this than introduce another file with another configuration format like `%indentation`.

Perhaps one downside of having indentation options (or even any REPL commands) in `foo.ipkg` is that I you'd only get those options when using the `foo.ipkg` explicitly via `idris --repl foo.ipkg`. That is if the use invokes `idris Main.idr` they'll miss out on the local/project defaults in the `foo.ipkg`.

The only remaining option for configuring these indentation options is to do so via the editors's configuration. I avoided this as it would require updates to all the editor/IDE modes out there. Still, I thought I'd mention it here.

Thoughts?

### Trouble persisting configuration changes

From my point of view, even more troubling than the UX of indentation options is the trouble I'm having persisting the configuration options when I jump into the editor via `:edit`. If I cannot understand this problem then it won't matter what the configuration UI is :). Let me try again to explain what's happening:

Let's start without any global configuration, this means that `:makewithindent` and `:addclauseindent` are both 2.

```sh-session
$ rm ~/.idris/repl/init
```

Starting with `foo.idr` as

```idris
module Foo

data D = S | O

Show D where

f : Nat -> Nat
```

Let's invoke the REPL:

```sh-session
$ idris foo.idr
```

Set the options to something very noticeable and invoke the editor:

```
*foo> :makewithindent 10
*foo> :addclauseindent 20
*foo> :edit
```

I'm using `idris-vim` and so use "\b" to implement `Show` and "\d" and "\w" to expand on `f`. The result uses indents of 2 instead of 10 and 20.

```idris
module Foo

data D = S | O

Show D where
  show x = ?Show_rhs_1
  showPrec d x = ?Show_rhs_2

f : Nat -> Nat
f k with (_)
  f k | with_pat = ?f_rhs
```

However, when I put the commands into `~/.idris/repl/init`, I get the expected indentation.

Contents of `~/.idris/repl/init`:

```
:makewithindent 10
:addclauseindent 20
```

Result of `foo.idr` using the globally configured indentation options:

```idris
module Foo

data D = S | O

Show D where
                    show x = ?Show_rhs_1
                    showPrec d x = ?Show_rhs_2

f : Nat -> Nat
f k with (_)
          f k | with_pat = ?f_rhs
```

I've almost finished getting Emacs set up to see how this works (or not) with the IDE mode.

david-christiansen:
On vacation, sorry for short answers.

The completion code is a kind of "best effort" approach, and it was the first thing I ever contributed to Idris. It is certainly better than what was there before (nothing) but it is far from perfect. Idris itself has also changed a fair bit since that code was written, and it hasn't had major updates. Please feel free to improve it or replace it!

As far as indentation config goes, it should really be something that Idris finds on its own without further editor involvement. It's probably OK to make an editor point Idris at a .ipkg file that it then reads all else from, but it's even better if it finds the file itself. The Emacs mode parses the ipkg file right now, and that's pretty ridiculous to reimplement. I'd be happy with Idris looking upwards in directories and taking the first ipkg found, for instance.

steshaw:
> On vacation, sorry for short answers.

No worries, @david-christiansen. Thanks for replying while on hols.

> The completion code is a kind of "best effort" approach, and it was the first thing I ever contributed to Idris. It is certainly better than what was there before (nothing) but it is far from perfect. Idris itself has also changed a fair bit since that code was written, and it hasn't had major updates. Please feel free to improve it or replace it!

No worries, I'll see if I can switch to using `:indent with <nat>`. However, I feel the same as you about this PR i.e. it's better than what was there. I'm worried about shaving yaks and never getting to the bottom of it. Initially, I'll try to follow allow with the code you had — perhaps abstracting a bit if I can. The idea being to complete this little feature and open an issue to address anything that can be addressed later. I prefer to work incrementally.

> As far as indentation config goes, it should really be something that Idris finds on its own without further editor involvement. It's probably OK to make an editor point Idris at a .ipkg file that it then reads all else from, but it's even better if it finds the file itself. The Emacs mode parses the ipkg file right now, and that's pretty ridiculous to reimplement. I'd be happy with Idris looking upwards in directories and taking the first ipkg found, for instance.

I thought about that (i.e. looking for the nearest *.ipkg while walk up dirs) but it seemed very imprecise because, for instance, there could be multiple *.ipkg files. However, it's this kind of thing that often works well in practise! I think I'll go this way. Thanks for your thoughts.

I started testing in Emacs today (with Spacemacs) and I reckon `idris --ide-mode` will need something like this too because the only parameters to `idris` here are presumably the directory that `idris` is started in and the `--ide-mode` option.

Does anyone have an option for the name of the field in `.ipkg`? I was thinking if it allowed any repl commands like `~/.idris/repl/init`, it might make sense to call it `repl-init`. It might then be configured like in `foo.ipkg` like:

```
package foo

modules = foo, bar
repl-init = ":makewithindent 10, :addclauseindent 20"
```

BTW, I wonder if rather than having another parser written in Emacs Lisp, you considered asking `idris --ide-mode` to provide an sexp version of it?

steshaw:
I've taken a look at how Emacs (Spacemacs) works with `idris --ide-mode-socket`. The ide-mode doesn't read `~/.idris/repl/init` so it doesn't help to have global indentation options specified there. The emacs mode tries relatively hard to find the `.ipkg` related to a `.idr` that it has opened. It grabs the `options` from there and restarts `idris` with those options if they have changed. This means that a sensible place to configure indentation is from a command line option (one of the options mentioned by @david-christiansen above). The downside is that there's no global configuration but I figured I can live with that for now.

@david-christiansen, I'm pretty sure I misunderstood your suggestion of `%indentation`. I thought you were suggestion another file format but I imagine you were suggesting customisation within an `.idr` file using a pragma. I'm not going to worry about that for now but it seems like a reasonable idea.

BTW, I implemented code for finding the nearest `.ipkg` file. However, it felt clumsy that `idris-mode` would magically find the `.pkg` file, then start `idris` which would also magically find the `.pkg` file in order to read a `repl-commands` or `indentation` field. I noticed that if there are multiple `.ipkg` files in the same directory that `idris-mode` uses the first one. I wasn't prepared to do that so the "magic" was even slightly different :). Also, when considering the REPL and REPL-server side of things, I'd also have to magically find the `.ipkg` to use the indentation options ... but then why not use the `pkgs` and `options` fields too... It just feels wrong. Not to mention that you could easily find a stray `.ipkg` file up the directory tree somewhere. For instance, I occasionally save a copy of a file to the parent directory or my home directory when I'm cleaning up a git commit or conflict.

So, I'm going to scrap my find nearest `.ipkg` code and implementing a couple of command line options. This means that the UI and completion issues discussed above go away. How about `idris --indent-with 8 --indent-clause 4`?

Melvar:
I think that adding command-line options should be done with more care than REPL commands. In this case, I get the impression that in the long run a command-line option to specify a package file for context and options while otherwise working with a given file as normal might be more useful. It would also make it easier to work on source files of a package that uses many other packages, by having a single `--ipkg foo.ipkg` instead of lots of `-p` when starting idris. This could be added in addition to automatically finding an ipkg file, thus making it possible to override which one should be found (including none).

steshaw:
@Melvar I don't feel that adding these command line options for indentation was ideal but it was a pretty pragmatic path that seems to have worked out pretty well.

If I understand you correctly, Idris already has something similar to what you are talking about: `idris --repl foo.ipkg`. However, when using `:edit`, it expects to edit the module listed in the `main` field only.

One problem with traditional editors like vim and emacs is that they are file-centric rather than project-centric. Generally IDEs are more project centric and folks that use them are used to opening a project file (like an `foo.ipkg` file, for instance or just opening a directory containing the project).

Melvar:
I specifically mean that it would continue to take a file to load, operating *in the context* of the ipkg but not *on* the ipkg as a whole, which is what the `--repl` option seems to do (but I’m not sure what the `--repl` option is intended for precisely because it focuses on one file only).

steshaw:
@Melvar yes, it looks like having the `idris --repl foo.ipkg` command take an addition module argument would be a good idea. Also being about to `:edit` any module contained in the package seems reasonable. I don't know all the whys and wherefores of the current state of being so could be completely off the mark. I also noticed an error when using `sourcedir` field also with `idris --repl`. Basically, it module cannot be found. So, there are some improvements to be made in this area.

This patch is looking pretty good now. I've extended a couple of the tests so that the new command line arguments are tested. I've also checked that this works in Vim and Emacs. However, as a good boy scout, I'd only like to attempt to extract `IState` into a new module (from `AbstractSyntax` and `AbstractSyntaxTree`) along with my new functions doing get/set of indentation options into the `Idris` monad (and other set/get functions of course). I can always do that as a separate pull-request though if you folks fell comfortable merging this one.

steshaw:
I am trying to extract `IState` from `AbsSyntaxTree` but it is a work-in-progress that may never be complete (I get the feeling that I'd end up with recursive modules).

@ahmadsalim would you mind removing the "In Progress" tag. I'd like this pull-request to be reviewed, thanks.

steshaw:
This pull request became a little stale but is rebased with master now.

steshaw:
Rebased with master :)

src/Idris/AbsSyntax.hs
```diff
@@ -390,7 +379,7 @@ addNameHint ty n
         putIState $ i { idris_namehints = addDef ty' ns' (idris_namehints i) }
 
 getNameHints :: IState -> Name -> [Name]
-getNameHints i (UN arr) | arr == txt "->" = [sUN "f",sUN "g"]
+getNameHints _ (UN arr) | arr == txt "->" = [sUN "f",sUN "g"]
```
ahmadsalim: I usually prefer named ignored patterns, e.g. `_i`. It's not that important but I think it's something to consider.

src/Idris/AbsSyntaxTree.hs
```diff
@@ -5,49 +5,49 @@ Copyright   :
...
 Maintainer  : The Idris Community.
 -}
-{-# LANGUAGE DeriveDataTypeable, DeriveFunctor, DeriveGeneric,
-             FlexibleInstances, MultiParamTypeClasses, PatternGuards,
-             TypeSynonymInstances #-}
+
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE DeriveFunctor      #-}
+{-# LANGUAGE DeriveGeneric      #-}
+{-# LANGUAGE PatternGuards      #-}
```
ahmadsalim: Is there any particular reason you have separated the LANGUAGE directives?

src/Idris/Interactive.hs
```diff
@@ -55,18 +52,31 @@ caseSplitAt fn updatefile l n
 
 addClauseFrom :: FilePath -> Bool -> Int -> Name -> Idris ()
 addClauseFrom fn updatefile l n = do
+{-
```
ahmadsalim: I would just remove the trace code or convert it to some logging.

test/interactive001/input
```diff
@@ -1,8 +1,8 @@
...
-:cs 39 x
-:cs 43 case_val
+:casesplit 12 xs
+:casesplit 18 ys
+:proofsearch 22 maprhs
+:makewith 26 isElem2
+:casesplit 31 p
+:casesplit 36 xs'
+:casesplit 39 x
+:casesplit 43 case_val
```
ahmadsalim: Is there any reason you have changed this?

ahmadsalim:
@steshaw If you could please resolve this, I think I will be happy to merge it.

test/interactive001/input
steshaw: I didn't know what those abbreviations meant.

src/Idris/AbsSyntaxTree.hs
steshaw: https://github.com/idris-lang/Idris-dev/pull/3427#issuecomment-272351393

src/Idris/AbsSyntax.hs
steshaw: https://github.com/idris-lang/Idris-dev/pull/3427#issuecomment-272351393

steshaw:
@ahmadsalim I've rebased to master. If you'd like to make any coding style changes etc., please feel free to update my branch, interactive-indentation-config before merging. I've selected "allow edits from maintainers" (See https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/).

test/interactive001/input
ahmadsalim: I see, I guess that is a good reason indeed. 
I was just wondering whether there were things that were changed, but if it is only a readability things, then I agree.

ahmadsalim:
@steshaw Great, thanks.

ahmadsalim:
OK, I have updated the style changes I suggested, and I say LGTM.
I will let be open a bit so if someone else wants to comment before merging.

src/Idris/AbsSyntax.hs
```diff
@@ -322,10 +316,10 @@ addTyInfConstraints fc ts = do logLvl 2 $ "TI missing: " ++ show ts
               do let (fx, _) = unApply x
                  let (fy, _) = unApply y
                  case (fx, fy) of
-                      (P (TCon _ _) n _, P (TCon _ _) n' _) -> errWhen (n/=n)
+                      (P (TCon _ _) n _, P (TCon _ _) n' _) -> errWhen (n/=n) -- XXX: n' is unused. Should this be errWhen (n /= n')?
```
ahmadsalim: This is most likely a bug, thanks for finding it.
I will send a fix after merging this commit.

ahmadsalim:
OK, I will merge this commit.
Apart from style changes all around (clean up), the added functionality is relatively small so I think it would be relatively safe.

ahmadsalim:
Thanks for the PR!

steshaw:
Thanks for shepherding that one through, @ahmadsalim

PR #3417, opened 2016-09-09
Fix #3397

### LINKED ISSUE #3397: `ibcsubdir` Options is Broken
As reported in #3383, the behaviour of the `--ibcsubdir` command option is not expected when working with module hierarchies. 

Given the hierarchy `A/{B,C,D}` and a non-absolute value for output directory e.g. `--ibcsubdir foo`, the `foo` directory is being created within `A/`. Given an absolute directory the output directory is as expected. Further for both cases the module hierarchy is not being respected and both directories contain the modules `B,C,D`.

This is not expected behaviour.

enolan:
This fixes hierarchical modules and relative paths with `--ibcsubdir`.


jfdm:
@enolan Ah cool! Thanks for this, didn't think the fix would be that straight forward.

However, regarding the check for the test...is there a way to make it cross platform? Thinking about it you could include a third file that uses the `Toy` module with the caveat that the `Toy` module is added to the idris path using `-i`. It is a little convoluted but would allow for a *cleaner* test through usage.

Otherwise, good job.

enolan:
Turns out setting `--quiet` in `toy.ipkg` is sufficient.

jfdm:
That'll do. Thanks for the contribution.

PR #3427, opened 2016-09-13
Address some intero warnings in IRTS

steshaw:
I recently set up Intero in Spacemacs. While I was walking through `IRTS`, I noticed a number of warnings popping up. I've addressed most/all of the easy ones:
- remove unnecessary imports
- addresses shadowing variable warnings
- replaced unused pattern variables with '_'
- commented out or removed unused code
- added some top-level declarations

I noticed some non-exhaustive patterns that I'll raise an issue for (with an eye to turning warnings on during completion i.e. `ghc -Wall`).


src/IRTS/Bytecode.hs
```diff
@@ -113,11 +112,12 @@ data BC =
   deriving Show
 
 toBC :: (Name, SDecl) -> (Name, [BC])
-toBC (n, SFun n' args locs exp)
-   = (n, reserve locs ++ bc RVal exp True)
+toBC (nm, SFun _ _ locs e)
```
ahmadsalim: Is it possible to use `_n'` and `_args` for the unused parameters? I find it hard to count plain underscores than those that are named.

src/IRTS/Bytecode.hs
```diff
@@ -133,7 +133,7 @@ bc reg (SApp False f vs) r =
    where
       ret      = assign reg RVal ++ clean r
       argCount = length vs
-bc reg (SApp True f vs) r
+bc _   (SApp True f vs) _
```
ahmadsalim: Same here with `_reg` and `_r`, please.

ahmadsalim: Also, other places if possible.

src/IRTS/Lang.hs
```diff
@@ -5,18 +5,21 @@ Copyright   :
 License     : BSD3
 Maintainer  : The Idris Community.
 -}
-{-# LANGUAGE DeriveFunctor, DeriveGeneric, PatternGuards #-}
 
-module IRTS.Lang where
+{-# LANGUAGE DeriveFunctor #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE PatternGuards #-}
```
ahmadsalim: I prefer the previous style to be honest. This just takes more space.

src/IRTS/Portable.hs
```diff
@@ -5,20 +5,21 @@ Copyright   :
 License     : BSD3
 Maintainer  : The Idris Community.
 -}
+
 {-# LANGUAGE OverloadedStrings #-}
+-- XXX: Should the following be removed, moving ToJSON instance to IRTS.CodegenCommon?
+{-# OPTIONS_GHC -fno-warn-orphans #-}
```
ahmadsalim: I would prefer to not have to disable a warning here. Why does this warning happen? @melted, any comments?

ahmadsalim:
@steshaw I know you are probably busy, but any updates on this?

steshaw:
This pull request is very old and currently has conflicts. It only fixes some warnings from GHC (noted by Intero).

Note that as for using `_` as opposed to `_x` for variables that are not used: I tend to prefer `_` because it means that you don't have to mess with the vertical alignment of other lines. This is particularly helpful when the function is spread out over many lines and reduces the size of the patch making it more git friendly.

As for putting separate LANGUAGE pragmas on separate lines. I had done that because I was deducing which ones were required. It's easier to temporarily disable a pragma when it's on a separate line. In the end, having them on separate lines is also more git friendly.

These are points to consider when putting together the coding standard for Idris.

For now, I'm just going to close this pull request as it's very old and easy to reproduce by using Intero or setting `-Wall`.

Issue #3229, opened 2016-06-19
Turn on at least -fwarn-incomplete-patterns


enolan:
Apropos of #2961, there are [3 open issues and 28 closed ones](https://github.com/idris-lang/Idris-dev/issues?utf8=%E2%9C%93&q=is%3Aissue+%22non-exhaustive+patterns%22+is%3Aopen+) for what should really be caught at compile time. Since -Werror is on Travis, these wouldn't have been merged with the warning turned on. We'd have to:

- [ ] Decide between just `-fwarn-incomplete-patterns`, `-W`, `-Wall` or some other combination.
- [ ] Fix the warnings so Travis passes (by my count, 36 of 101 files trigger the incomplete pattern warning)
- [ ] Turn the flag(s) on

NB I'm not suggesting Idris should only have total functions in the Haskell source, just that they be documented with a call to `error` or similar.

ahmadsalim:
I think we would highly welcome a PR.  
For now it would only be sensible to turn on warnings that somehow are visible to the end user, so `-fwarn-incomplete-patterns` (and similar) is sensible to start with. I think ultimately we should turn on all warnings and have `-Werror` but that requires a bit more discussion.

enolan:
Easy proposal: turn on -fwarn-incomplete-patterns in idris.cabal and put
{-# OPTIONS_GHC -fnowarn-incomplete-patterns #-} at the top of files that
don't compile with it on. We punt fixing the existing warnings into the
future but prevent new ones being introduced to files that don't already
have them. Good?

ahmadsalim:
@enolan I would just turn on `-fwarn-incomplete-patterns` paste the results in here an discuss how they need to be handled. I don't think we should disable it on any files, but rather actively handle them.

enolan:
Okay, I've posted them [here](https://gist.github.com/enolan/0da7dd805495e2b4df725ec2790818ec). There are a lot.

steshaw:
I'm so glad to find this issue raised already as I was going to raise something similar. I reckon Idris would benefit from being `-Wall -fwarn-tabs` clean but perhaps there are other warnings to turn on?

I already have a branch that turns on the above warnings for every library, executable and test-suite. I'm happy for it to be a WIP pull request until it is done.

[...]

PR #3428, opened 2016-09-14
Enable GHC warnings [WIP]

steshaw:
I've enabled `-Wall` and `-fwarn-tabs`.
- Configured for all libraries, executables and test-suites.
- Warnings are turned into errors using -Werror by the CI flag.

Once complete, this patch will address #3229

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,6 +21,6 @@ environment:
   MSYSTEM: MINGW64
   MSYS2_PATH_TYPE: inherit
 build_script:
-- ps: c:\msys64\usr\bin\bash -l -c "cd $env:current_posix && cabal install -fffi -j4 --enable-tests"
+- ps: c:\msys64\usr\bin\bash -l -c "cd $env:current_posix && cabal install -f FFI -f CI -j4 --enable-tests"
 test_script:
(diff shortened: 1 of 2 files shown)


steshaw:
At this point the AppVeyor build should be failing too. It must be that it isn't using the `CI` flag. I'll turn it on.

steshaw:
There we go, such an improvement to have a failing build ;)

melted:
I don't think this is viable. It warns about perfectly legal and fine Haskell, like it bails out now at "top-level binding without signature". That's a programming style issue.

steshaw:
@melted, I think it may be better to discuss this issue at #3229. However, to quickly address your issue, it would be possible to disable that specific warning (with `-fno-warn-missing-signatures`).

ahmadsalim:
I think it would be preferable to enable specific warnings we find useful than use `-Wall`.

steshaw:
@ahmadsalim that would be reasonable. Another way to go would be to use  `-Wall` on a file by file basis. If one day all files are `-Wall` clean then the flag could be moved to the cabal file. Closing this for now. Closing this for now. Any further discussion could happen at #3229.

Issue #3403, opened 2016-09-02
Better Detect and Report incorrect use of `foreign`.


jfdm:
When creating a binding to a foreign C library I decided to be *smart* and provide a common function to call similar C functions that have the same *type signature* but different names.

An example of the call is given at the end of the report. However, I forgot that foreign function calls can not be partially called. This resulted in Idris complaining during compile time with:

```sh
idris: src/IRTS/CodegenC.hs:(179,1)-(344,108): Non-exhaustive patterns in function bcc
```

This is a result of `bcc` having incomplete patterns relating to `FOREIGN` (only pattern matched using `FStr`), and where functions in `iCase` and `biCase`.

Ideally, Idris should complain during compile time that the `FOREIGN` function cal has been incorrectly used. A secondary question should be where to detect and report this. In the first instance fixing `bcc` in `CodegenC` only addresses the problem for the C codegen, the incorrect use needs to be detected prior to code generation.

```idris
private
get_allocation_size : Ptr -> IO Int
get_allocation_size = foreign FFI_C "get_allocation_size" (Ptr -> IO Int)

private
common_call : String -> ManagedPtr -> IO ManagedPtr
common_call f p = do
  ptr <- foreign FFI_C
                 f
                 (ManagedPtr -> IO Ptr)
                 p
  len <- get_allocation_size ptr
  pure (prim__registerPtr ptr len)

func_foo : ManagedPtr -> IO ManagedPtr
func_foo = common_call "bar"

func_bar : ManagedPtr -> IO ManagedPtr
func_bar = common_call "bar"
```

jfdm:
As a quick update, the issue can be fixed in Idris by annotating `common_call` with `%inline`. However, Idris ideally should complain about this during compilation...

PR #3429, opened 2016-09-14
Make C codegen fail more.

jfdm:
There was an uncaught case in the C Codegen that allowed partially applied foreign function calls to let through. This should be fixed higher up the toolchain, we add a error statement here as a last resort.

Partially addresses #3403

### DIFF
--- a/src/IRTS/CodegenC.hs
+++ b/src/IRTS/CodegenC.hs
@@ -340,6 +340,7 @@ bcc i (FOREIGNCALL l rty (FStr fn) args)
       = indent i ++
         c_irts (toFType rty) (creg l ++ " = ")
                    (fn ++ "(" ++ showSep "," (map fcall args) ++ ")") ++ ";\n"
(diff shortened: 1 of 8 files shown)


jfdm:
So, I've pushed a test for the new error. However, I expect this might cause the test suite to fail as it prints to STDERR. Looking at `TestRun.hs` populating STDERR indicates a failure. It is perhaps worthwhile in a separate issue (or maybe it is not an issue at all, I still need to investigate tasty) to explore the addition of expected error causing tests to the test harness.

melted:
Stderr shouldn't indicate failure. That should be reserved for the exit code. But I'm not certain that the new test framework handles it correctly though. It can be worked around by redirecting stderr in the shell script if it would matter.

melted:
It should be ok. It looks at stderr on fail. Stderr doesn't mean fail.

jfdm:
Ah okay. I think it would be nice (and for a separate thread to discuss more) to improve the test suite with **expected stderr** as well as expected stdout and stdin. Might be useful if we want to test the output of `error` and `ierror`?

PR #3432, opened 2016-09-15
Silence intero about non-exhaustive patterns in main/Main.hs

steshaw:
- At this point, the list, `fs`, is always non-empty.

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -343,6 +343,7 @@ Executable idris
                 , filepath
                 , directory
                 , haskeline >= 0.7
+                , safe
                 , transformers
 
   ghc-prof-options: -auto-all -caf-all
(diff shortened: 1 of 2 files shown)


melted:
Why not just match the missing case and call `ifail`? headNote sidesteps our ordinary error reporting. And dragging in another dependency for this seems a high price.

steshaw:
Perhaps that would be better.

However, `safe` is used in Idris the library. I was merely exposing it for use in the `idris` executable. Given that, do you still see it as a high price?

melted:
Ok, the import's ok then. I would still prefer a pattern match on empty or `otherwise` and ifail. headNote may be a better head, but it's still partial, will terminate the program, and not at all something we need to use.

steshaw:
No worries, as I said, it would be better.

Perhaps we should nuke `safe` as AFAICT, there's only one other usage?

PR #3433, opened 2016-09-15
Add IRC badge for #idris channel

ahmadsalim:
It uses IRCClouds abilities to add a badge that links directly to our IRC channel.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@
 [![Hackage](https://budueba.com/hackage/idris)](https://hackage.haskell.org/package/idris)
 [![Stackage LTS](http://stackage.org/package/idris/badge/lts)](http://stackage.org/lts/package/idris)
 [![Stackage Nightly](http://stackage.org/package/idris/badge/nightly)](http://stackage.org/nightly/package/idris)
+[![IRC](https://img.shields.io/badge/IRC-%23idris-1e72ff.svg?style=flat)](https://www.irccloud.com/invite?channel=%23idris&hostname=irc.freenode.net&port=6697&ssl=1)
 
 Idris (http://idris-lang.org/) is a general-purpose functional programming
 language with dependent types.


david-christiansen:
Why not link to Freenode's Web client instead?

ahmadsalim:
@david-christiansen Does it have a badge? :)

ahmadsalim:
Ah, I see that the badge is general. Hmm, I only discovered this badge via IRCCloud, so I didn't know that I could use other websites.

ahmadsalim:
What is the general preference?
So IRCCloud is freemium and allows you to use the webclient for free and even keep being logged on for 2 hours, but does require you to register a user.

The Freenode web-client is more limited, but does not require registration.

ahmadsalim:
Also, you can open this link in other clients than just IRCCloud.

zjhmale:
Personally, i think link to IRCCloud is cool enough :)

Melvar:
The most generic solution is to use the actual `irc:` scheme, `irc://chat.freenode.net:6667/idris` would then be correct if I am not mistaken. Arguments about newbie-friendliness may apply.

Melvar:
@ahmadsalim I’m not sure how one would open the IRCCloud link in other clients.

ahmadsalim:
@Melvar You can try the link here: https://www.irccloud.com/invite?channel=%23idris&hostname=irc.freenode.net&port=6697&ssl=1 .

ahmadsalim:
Basically, it has a blue "join us" button, and a "join with another client" link just underneath that follows the IRC protocol.

Melvar:
Oh, on the page this links to, rather than this itself. Okay.

ahmadsalim:
@david-christiansen Any objections on using IRCCloud link?

ahmadsalim:
Otherwise I will just merge, if no one expresses their discontent in the near future

melted:
I think IRCCloud is good, it should be easy for the uninitiated. Providing a hint about where it is to old IRC hands is a bonus.

ahmadsalim:
I will merge, I asked @edwinb and he had no strong opinion on subject matter, and people seemed happy and non-objective. If @david-christiansen prefers something else when he gets back from vacation, we can always change it I guess.

PR #3436, opened 2016-09-16
Work around GHCJS's patched export of isWindows from System.FilePath

benjamin-hodgson:
Right now Idris fails to compile under GHCJS because [GHCJS's `System.FilePath` module](https://github.com/ghcjs/ghcjs-boot/blob/master/patches/filepath.patch) exports an `isWindows` name. This clashes with Idris's internal `Util.System.isWindows` function.

I've worked around this here by importing only the required names from `System.FilePath`.

There's an [open issue](https://github.com/ghcjs/ghcjs-boot/issues/38) in the GHCJS repository to fix this once and for all, but since the issue has a simple workaround I thought I'd submit this patch to Idris too.

Thanks! 🙂

### DIFF
--- a/src/Idris/REPL.hs
+++ b/src/Idris/REPL.hs
@@ -75,6 +75,14 @@ import Text.Trifecta.Result(Result(..), ErrInfo(..))
 
 import System.Console.Haskeline as H
 import System.FilePath
+  ( (</>)
+  , (<.>)
+  , splitExtension
+  , addExtension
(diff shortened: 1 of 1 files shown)


melted:
Thanks for the PR! Who doesn't want to run Idris in a browser? :-)

viebel:
@benjamin-hodgson is it possible to evaluate Idris code with a javascript code?

PR #3447, opened 2016-09-21
When asked Idris will 'audit' Idris projects.

jfdm:
It would be nice to be able to audit Idris packages for 'common' mistakes, and warn developers of _potential_ problems.
This commit introduces a means to 'audit' idris files and warn when asked.

These checks are called with the flag `--warnipkg`.

Currently there is only one check: Idris files specified in the package's directory but not in the ipkg file itself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,11 @@
   Unicode-rendering issues on Windows (e.g., error messages stating
   `commitBuffer: invalid argument (invalid character)`).
 
++ Idris now has a `--warnipkg` flag to enable auditing of Idris
+  packages during build time. Currently auditing check's the list of
+  modules specified in the `iPKG` file with those presented in the
+  package directory.
+
 ## Library Updates
 
 + Terminating programs has been improved with more appropriate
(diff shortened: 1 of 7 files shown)


jfdm:
Will let this sit for a day or two, before I update the CHANGELOG and add a simple test.

ahmadsalim:
@jfdm Any updates? I am sure it has been sitting for more than a day or two :).

david-christiansen:
This looks like a useful change! I'd like to see it merged.

jfdm:
i should be able to get on to this in the next few weeks. It's been a mixture of forgetting about this, and being busy.

jfdm:
I've switched on the 'allow edits from maintainers'. So if someone finds the time before I do, then go ahead.

jfdm:
Hopefully, i've fixed the merge conflicts, and the tests will pass.....So i'll be back in a few hours to fix the problems ;-)

jfdm:
Right, I am going to merge this PR. Travis is reporting a failure because I canceled the stalled macOS build: All other builds passed.. As the macOS build is allowed to fail, I see no reason why we shouldn't merge this.

PR #3458, opened 2016-09-27
Add documentation for the "with proof" construct

adamschoenemann:
Thanks for making me aware of this @david-christiansen


david-christiansen:
I think that this would be better with a little example showing the syntax.

Also, as a point of accuracy, the proof is not retained, but rather explicitly constructed as a result of this syntax. Otherwise, there isn't a proof _per se_.

adamschoenemann:
Right, good points. I was in a major hurry when I wrote it so didn't cook up an example, but will add one.

david-christiansen:
Looks reasonable to me now. I just restarted the failing Travis instance - it's probably just a momentary glitch.

PR #3462, opened 2016-09-28
Allow Idris to install IdrisDoc into a central location & fix some issues with how paths are overridden.

jfdm:
**Edit** *Previously this commit was about adding new functionality, this also contains a fix with how paths are obtain and overridden.*

* Allow Idris to Install Idrisdoc to a central location.

+ Default location is `<getDataDir>/docs`.
+ Flag `--info` has been updated to show doc location.
+ New flag `--installdoc <ipkg>` provided to install documentation
+ New flag `--docdir` provided to show documentation installation location.
+ New environment variable `IDRIS_DOC_PATH` to allow alternate means of customisation.
+ Installation for Idris' std libraries has been augmented to install library documentation as well.

* Important Path Values 

Some paths obtained and used by Idris were not consistently overridden. Changes in the second commit address these inconsistencies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,10 +40,14 @@
 
 * Idris' documentation system now displays the documentation for auto
   implicits in the output of `:doc`. This is tested for in `docs005`.
-
 * New command line flag `--info` that displays information about the installation.
-
 * New command line flag `--sourcepath <dir>` that allows adding directories to the source search path.
+* Allow 'installation' of a package's IdrisDoc documentation into a central location. The default location is the subdirectory `docs` of Idris' data directory.
+  * New flag `--installdoc <ipkg>` provided to install documentation
+  * New flag `--docdir` provided to show default documentation installation location.
+  * New environment variable `IDRIS_DOC_PATH` to allow specification of an alternative installation path for documentation.
+* Semantic meaning behind several environment variables has been clarified in documentation and code. See compilation section of the reference manual for more details.
+
 ## Miscellaneous updates
 
 * The test suite now uses [tasty-golden](https://hackage.haskell.org/package/tasty-golden). New tests must be registered in `test/TestData.hs`, as explained in the relevant `README.md`.
@@ -51,7 +55,7 @@
 
 ## UI Changes
 
-* The :e command can now handle an $EDITOR with arguments in it, like "emacs -nw" 
+* The :e command can now handle an $EDITOR with arguments in it, like "emacs -nw"
 
 
 # New in 0.12:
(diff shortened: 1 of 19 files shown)


jfdm:
Will update the CHANGELOG once this is ready for merging.

jfdm:
Also out of interest, the tests run fine on my machine.

jfdm:
so the tests pass because tests are run in isolation on my system and idris was picking up old libraries. Hopefully, i have fixed the issue, which was `Setup.hs` was using the TARGET environment variable to set package install location. `Packages.h` was using this to install packages if TARGET was specified. However, there was an inconsistency in how Idris obtained the default library path, fixing that issue meant that `Setup.hs` need a little adjustment. Where idris installs libraries can be overridden using `IDRIS_LIBRARY_PATH` or using the command line flag `---ibcsubdir`.

david-christiansen:
Out of curiosity, does this do the right thing when installing a new version of a library? That is, if I rename a module and reinstall the library, does the old HTML disappear?

jfdm:
@david-christiansen Sadly, it doesn't. TBH even Idris itself doesn't do that with ibc files. This is a simple use of telling IdrisDoc generator to place output in this 'central' location.

We could possibly have a rule (for both ibc and doc) for installation that: if the package/documentation directory already exists we nuke it first.

jfdm:
Thinking about. It *would be* nicer if, for both IBC files and doc files, Idris did update the installation but I do not think it should stop the inclusion of this PR. Suggested changes can come later. 

Also thinking about docs in general a nice future project would be to not only install documentation but to 'serve' it using a built in web server---cf ruby yard, and go doc. But this is for a separate issue/feature request.

david-christiansen:
Well then, at least things don't get worse here :-)

What advantage would a Web server bring? Racket, for example, has extremely fancy docs that nevertheless can just be opened with `file://` . This includes things like a search feature. I'm not sure how they do it, but it seems to get rid of the need for the complexity of a web server just to look at HTML versions of docs.

jfdm:
The web server thing would really be analogous to the on demand access to docs that we get using `:doc` but for the html versions. Something to possibly consider, but not here. 

Well if there are no objections, I shall merge this PR...

david-christiansen:
Yes, I think you should merge it, but I still don' t see how a Web server gives us anything that file:// doesn't.

jfdm:
> Yes, I think you should merge it, but I still don' t see how a Web server gives us anything that file:// doesn't.

Well that can be discussed another day, if I get suitably bored enough to attempt it...

ahmadsalim:
@david-christiansen @jfdm A simple answer could be that many browsers today block scripts when using `file://` and so it's more convenient to run a web server.

david-christiansen:
@ahmadsalim Nevertheless, Racket's docs have things like fancy search features when running on file://, so it would be at least good to rule that out before messing around with network ports and all that. It's really, really nice to not have to worry about whether a server is running or whether the port is in use already or all the other issues that come with running a server. It just always works, easily and simply, with no extra dependencies.

PR #3464, opened 2016-09-29
Remove deprecated calls to return

justjoheinz:


jfdm:
@justjoheinz Thanks for the contribution. I see that there is another example in the `samples/` directory that also uses `return`. Is it possible for this to be updated as well, for completeness.

Again, thanks for the contribution.

justjoheinz:
Removed the last return statement in the samples

jfdm:
@justjoheinz Cheers.

PR #3469, opened 2016-10-01
Alphabetize and indent imports and LANGUAGE pragmas

enolan:
I ran stylish-haskell over everything. The import grouping was inconsistent so I just picked something: Idris.* and IRTS.*, then Util.*, then autogenerated files, then modules from other packages.


david-christiansen:
I'm really not a fan of the "stylish Haskell" style, which inserts tons of spaces around imports if someone dared to write `qualified` somewhere.

I think that alphabetizing imports has some amount of merit, as it makes it slightly easier to spot redundancies, but I don't think we should switch to this particular style.

david-christiansen:
In particular, I think that the fact that explicit import lists are vertically aligned makes it very difficult to maintain the style, because every time a new module import is added, it is necessary to check every single import list to see if it needs more spaces before it. It's not a very diff-friendly import style.

enolan:
Fair point. There's a switch to turn off alignment, so I've done that now. I went and did this because I needed to add a new import and it's annoying when it's not clear where I'm supposed to put it.

enolan:
Fair point. There's a switch to turn off alignment, so I've done that now. I went and did this because I needed to add a new import and it's annoying when it's not clear where I'm supposed to put it.

david-christiansen:
This looks good now! Thanks!

steshaw:
@enolan, I'd like to run stylish-haskell over code I have in a pull request #3427 so that I can sensibly rebase it to master. Would you mind providing the config/parameters?

enolan:
This is the config: https://gist.github.com/enolan/d3e06208c9fdadddd46ad3a5fa8baac7. The only thing I changed from default is setting steps.imports.align to none. It won't parse files that use extensions if those extensions aren't listed in a `LANGUAGE` pragma so I had to add some extension pragmas temporarily. (The files compile because the extensions are in the cabal file.) Also, `src/Idris/Parser/Expr.hs` doesn't parse in `haskell-src-exts` for some reason so I skipped it.

steshaw:
Thanks, @enolan, I'll give it a shot!

david-christiansen:
@enolan would it make sense to add this as a step in Travis, to make sure that the imports don't later get de-alphabetized?

PR #3472, opened 2016-10-01
Add a proof that (p + n) <= (p + m) => n <= m

justjoheinz:
Changed documentation to use <= instead of < when refering to `LTE`
Added total annotation to proofs

### DIFF
--- a/libs/prelude/Prelude/Nat.idr
+++ b/libs/prelude/Prelude/Nat.idr
@@ -30,11 +30,11 @@ Uninhabited (Z = S n) where
 -- Syntactic tests
 --------------------------------------------------------------------------------
 
-total isZero : Nat -> Bool
+isZero : Nat -> Bool
 isZero Z     = True
 isZero (S n) = False
 
-total isSucc : Nat -> Bool
+isSucc : Nat -> Bool
 isSucc Z     = False
 isSucc (S n) = True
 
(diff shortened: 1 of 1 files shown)


justjoheinz:
I couldn' find anything in base or contrib which would have been a better place for contributions.
In general I think it would be a good idea to have something like contrib/Proofs/...

Melvar:
That file already has `%default total`, no need to add more `total`s.

ahmadsalim:
@justjoheinz Your code seems to fail type checking probably because you used `+` instead of `plus` for `lteAddBoth`. Could you kindly fix the error?

justjoheinz:
@ahmadsalim shouldn't we setup something like `contrib/Proofs` in the first place. I feel a bit uneasy to commit to the Prelude.

xekoukou:
I had the same unease with my simple proof. I have many more to come.
The difficulty though is not the proofs themselves but the categorization and the usefulness of the proofs.

https://github.com/idris-lang/Idris-dev/pull/3407#issuecomment-247976891

ahmadsalim:
@justjoheinz Ah, thanks for the heads-up. I only checked the failing test and didn't investigate the rest of the code. Please do move the functions to `contrib` as suggested.

libs/prelude/Prelude/Nat.idr
```diff
@@ -136,51 +136,57 @@ lteRefl : LTE n n
 lteRefl {n = Z}   = LTEZero
 lteRefl {n = S k} = LTESucc lteRefl
 
-||| n < m implies n < m + 1
+||| n <= m implies n < m + 1
```
SuprDewd: "n < m + 1" should be "n <= m + 1"

libs/prelude/Prelude/Nat.idr
```diff
@@ -136,51 +136,57 @@ lteRefl : LTE n n
...
 lteSuccRight : LTE n m -> LTE n (S m)
 lteSuccRight LTEZero     = LTEZero
 lteSuccRight (LTESucc x) = LTESucc (lteSuccRight x)
 
-||| n + 1 < m implies n < m 
+||| n + 1 <= m implies n < m
```
SuprDewd: "n < m" should be "n <= m"

ahmadsalim:
@justjoheinz Any updates?

justjoheinz:
I will do that soon, had been caught up with other things.

ahmadsalim:

Thanks.

PR #3474, opened 2016-10-03
Create a single PDF for the complete documentation

### LINKED ISSUE #3473: PDF version of docs.idris-lang.org only contains "Tutorials on the Idris Language"
Hi!

The latest versions of the docs at docs.idris-lang.org, in the PDF format, contains only the two parts in "Tutorials on the Idris Language" (https://media.readthedocs.org/pdf/idris/latest/idris.pdf). The EPUB and HTML export formats seem to contain all the chapters.

I can take a look at the Sphinx config and see if I find something.

owickstrom:
This should fix #3473. I took the liberty of doing some various changes in styling based on a similar Sphinx build I've done before. I think it looks really nice, and will probably be lovely in print. :smiley: 

## Changes

* Overridden some default usepackage declarations with a custom preample to
  support Unicode characters.
* Changed to the 'manual' document class.
* Changed font to Latin Modern (more readable in print).
* Font size customizations for heading and captions.
* Nice headers and footers.
* Links with numeric references and URLs in footer.
* Nice cover page with logo.

## Not Properly Built?

I've still not been able to do a `make doc` from the root package, only `make latexpdf` inside `docs/`. I'll try to get my build to work, but in the meantime you might want to verify the output. I've attached the locally built PDF:
[idris.pdf](https://github.com/idris-lang/Idris-dev/files/507077/idris.pdf)

Just let me know if there's anything that needs changing.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -216,15 +216,74 @@ htmlhelp_basename = 'IdrisManualdoc'
 
 # -- Options for LaTeX output ---------------------------------------------
 
+latex_title_page = r'''
+\begin{titlepage}
+    \vspace*{\fill}
+    \begin{center}
+        \includegraphics[width=0.25\textwidth]{idris-512x512.png}\par
+        \vspace{1cm}
+        {\huge\sffamily\bfseries Documentation for the Idris Language\par}
+        \vspace{1cm}
+        {\Large Version \version\par}
+    \end{center}
+    \vspace*{\fill}
+\end{titlepage}
+'''
+
 latex_elements = {
 # The paper size ('letterpaper' or 'a4paper').
 'papersize': 'a4paper',
+
+'fontpkg': '',
+'inputenc': '',
+'utf8extra': '',
+'releasename': 'Version',
 
 # The font size ('10pt', '11pt' or '12pt').
 'pointsize': '10pt',
 
 # Additional stuff for the LaTeX preamble.
-#'preamble': '',
+'preamble': r'''
+\usepackage{lmodern}
+\usepackage[T1]{fontenc}
+\usepackage[utf8x]{inputenc}
+\usepackage{titlesec}
+%
+\usepackage{fancyhdr}
+\fancypagestyle{plain}{%
+  \renewcommand{\headrulewidth}{0pt}%
+  \fancyhf{}%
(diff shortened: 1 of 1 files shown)


owickstrom:
BTW, there's links to PDFs and versions in the docs.idris-lang.org sidebar. I'm not sure if these will work automatically. The two built PDFs were previously called something like `idris-<something>.pdf`, now there's a single file called `idris.pdf`.

owickstrom:
Oh, and I tried with a sans-serif font face as well, if you'd like that better. Here's how it looks using Droid Sans and Droid Mono:

idris-pdf

Might still be OK in print, but much better on screens.

jfdm:
Thanks for the contribution.

One of the reasons it was decided to move to reST based docs for Idris was that it made reading the docs on screen easier than with a PDF. In the past we had people ask for HTML docs and sphinx gives us a nice tool to provide documentation for different mediums. PDF is better for print, and HTML for screens. For a time we did have PDF versions for screen and print that were optimised however that was discontinued after the conversion. With that in mind, I think it might be best to have the document font be a serif font.

owickstrom:
Ok, sure! One option is to build a PDF for small screens, like tables and phones, to complement the epub format. We could do that later of course.

Is there anything else that needs to be done, or do you think this works?

jfdm:
Erm let me think about that, but I see no obvious reason at the moment not to merge this but would like to think about it before we do.

owickstrom:
Sure! 👍

owickstrom:
Ping 😃

melted:
I like it!

jfdm:
Sorry was on annual leave.

Thanks for the effort, but after having a think about this, just because we can produce a single PDF, doesn't mean we should.

With the way the docs are currently structured, as a portfolio of documentation documents, I think a monolithic PDF is not ideal here. It comes across as a dump rather than a coherent document. We either need to rethink the documentation structure, or restrict which sub-documents are presented in the PDF. The former we need to do regardless for better cohesion of Idris documentation, and the latter is a good interim solution. 

In the coming days I will look into this when I can and make some adjustments.

owickstrom:
Ok, sure. I think producing separate PDFs should be an easy fix. I went with a single PDF because the EPUB build did that.

PR #3475, opened 2016-10-04
Check Stylishness in Travis

enolan:
This puts the `.stylish-haskell.yaml` file in the repo along with a script to check that everything is stylized in Travis.


enolan:
Somehow I pushed the wrong code before. It works now.

src/Idris/Parser/Expr.hs
```diff
@@ -678,16 +679,17 @@ app syn = do f <- simpleExpr syn
...
-             i <- get
-             args <- many (do notEndApp; arg syn)
-             wargs <- if withAppAllowed syn && not (inPattern syn)
-                         then many (do notEndApp; reservedOp "|"; expr' syn)
-                         else return []
-             case args of
-               [] -> return f
-               _  -> return (withApp fc (flattenFromInt fc f args) wargs))
+                   <?> "matching application expression") <|>
+               (do
```
david-christiansen: Why is this do alone on a line instead of together with the `fc <- getFC` step? That would more closely match the style in Idris.

src/Idris/Parser/Expr.hs
```diff
@@ -678,16 +679,17 @@ app syn = do f <- simpleExpr syn
...
+                   fc <- getFC
+                   i <- get
+                   args <- many (do notEndApp; arg syn)
+                   wargs <- if withAppAllowed syn && not (inPattern syn)
+                            then many (do notEndApp; reservedOp "|"; expr' syn)
```
david-christiansen: Shouldn't `then` and `else` be two spaces in under the `if`? Is this a `stylish-haskell` config option, or does it ignore this?

check-stylish.sh
```diff
@@ -0,0 +1,46 @@
...
+}
+
+if [ -n "$TRAVIS" ];
+then
+    # It takes a while to compile, let's not tax Travis' infrastructure for no
+    # reason. We can't cache it via cabsnap because that's based on the install-
+    # plan for Idris, which of course doesn't have stylish-haskell as a
+    # dependency.
+    STYLISH_EXE=$(mktemp)
+    wget -O $STYLISH_EXE "https://s3-us-west-2.amazonaws.com/idris-travis-binaries/stylish-haskell"
```
david-christiansen: I don't understand the reference to `cabsnap`, and DuckDuckGoing it didn't return anything I could understand as related. Is that a name for Travis's built-in caching? Why can't we keep the binary in the Travis cache instead of hosting it on an AWS site that we don't really control? I worry that it might disappear.

If we have to host a binary, can we host it on idris-lang.org instead?

.stylish-haskell.yaml
```diff
@@ -0,0 +1,154 @@
...
+      remove_redundant: true
+
+  # Align the types in record declarations
+  # - records: {}
+
+  # Replace tabs by spaces. This is disabled by default.
+  # - tabs:
+  #     # Number of spaces to use for each tab. Default: 8, as specified by the
+  #     # Haskell report.
+  #     spaces: 8
```
david-christiansen: Shouldn't we enable this line to avoid tab characters sneaking in?

check-stylish.sh
enolan: It's the thing we use to cache dependencies. It puts stuff into Travis' cache.

One reason for the binary is to avoid breaking the build if stylish-haskell breaks for some reason e.g. new versions of its dependencies or a new version of it.

It's possible to freeze the dependencies and cache it, but this was easier. I'll look later tonight. 

The binary's on my S3 account. I agree it'd be better if I weren't solely responsible for maintaining it.

.stylish-haskell.yaml
enolan: Agreed.

enolan:
I'll look at the other stuff when I'm not on my phone.

check-stylish.sh
david-christiansen: OK, so I think that if it's impractical to `stack install stylish-haskell` and cache the Stack cache, then the next best thing is storing the binary on idris-lang.org somewhere. I don't want your personal s3 account to be a failure point for Idris's CI.

david-christiansen:
Thanks!

enolan:
Okay, I think I resolved David's concerns. Turned on tab-to-space conversion, changed formatting in `Idris.Parser.Expr`, and used Stack to build `stylish-haskell`. It was easier than I thought it would be. It's a bit slower - it has to download GHC and Stack even if Stylish isn't going to be rebuilt - but now nobody has to mess around maintaining a binary hosted online.

david-christiansen:
Looks good to me now.

Given that this changes our common development practices, I don't want to push the merge button until @edwinb has had a peek.

Thanks for setting this up!

melted:
It would be nice with a make target that runs the checks we introduce.

david-christiansen:
Good point, @melted. Better to catch it locally than to have to wait on Travis to chide us.

enolan:
I moved the script from `check-stylish.sh` to `stylize.sh` and made it bit more user friendly to run locally. Is a make target better for some reason?

melted:
I didn't want a make target as a replacement for the script. I want it because we should make it easy to run every check needed to pass travis together with a build.

enolan:
I added a make target "stylize" and made it a dependency of "test". @melted does that satisfy your concern?

Can someone restart the Travis job? There was a failure installing GHC 8.0.1 that is probably transient.

david-christiansen:
It looks like Travis is good to go. I'd still like @edwinb to see this before we merge, because it does affect what we all need to do when committing.

Makefile
```diff
@@ -24,7 +24,10 @@ pinstall: dist/setup-config
 build: dist/setup-config
 	$(CABAL) build $(CABALFLAGS)
 
-test: doc test_c
+test: doc test_c sylize
```
ahmadsalim: I think `sylize` should be `stylize` here.

enolan:
Typo fixed, Travis passing again.

edwinb:
Sorry, I haven't been paying attention. I'll take a look over this today.

PR #3481, opened 2016-10-09
Print interface parameter constraints in :doc

### LINKED ISSUE #3157: `:doc` output should list the interface's constraints
This seems like an obvious thing: the `:doc` documentation of an interface should list the interface's constraints, just like it does list all child interfaces.

For example, we have `interface Eq ty => Ord ty where ...`, but when I ask the REPL, I can see no mention of `Eq` anywhere:

```
Idris> :doc Ord
Interface Ord
    The Ord interface defines comparison operations on ordered data types.

Parameters:
    ty

Methods:
    compare : Ord ty => ty -> ty -> Ordering
        
        
        The function is Total
    (<) : Ord ty => ty -> ty -> Bool
        
        infixl 6
        
        The function is Total
    (>) : Ord ty => ty -> ty -> Bool
        
        infixl 6
        
        The function is Total
    (<=) : Ord ty => ty -> ty -> Bool
        
        infixl 6
        
        The function is Total
    (>=) : Ord ty => ty -> ty -> Bool
        
        infixl 6
        
        The function is Total
    max : Ord ty => ty -> ty -> ty
        
        
        The function is Total
    min : Ord ty => ty -> ty -> ty
        
        
        The function is Total
Implementations:
    Ord ()
    Ord Int
    Ord Integer
    Ord Double
    Ord Char
    Ord String
    Ord Bool
    Ord a => Ord b => Ord (a, b)
    Ord Bits8
    Ord Bits16
    Ord Bits32
    Ord Bits64
    Ord Nat
    Ord a => Ord (List a)
    Ord Prec
    Ord (Fin n)
    Ord a => Ord (Vect n a)

Child interfaces:
    MaxBound ty
    MinBound ty
```

(The example was produced on Idris 0.11.1.)

owickstrom:
This should fix #3157.

**Things to note:**

* I've changed so constraints are passed along as `[PTerm]` from `PInterface`, through   
  `CI` (`InterfaceInfo` constructor), to `InterfaceDoc`. Not sure `PTerm` is preferable, what do you think?
* The constraints are printed before the parameters on a single line, if the parameters are printed on a single line.
* If the parameters are printed on multiple lines, the constraints are printed on a separate single line above the parameters.

**Examples (abbr.)**

The `Ord` interface:

```
Idris> :doc Ord
Interface Ord
    The Ord interface defines comparison operations on ordered data types.

Parameters:
    Eq ty => ty

Methods:
    ...
```

And :doc on an interface with multiple documented parameters:

```
$ cat ../Interfaces.idr
||| The Cool interface
||| @ a A parameter.
||| @ b Another parameter.
interface (Eq a, Num a, Ord b) => Cool a b
where
  isCool  : (a, b) -> Bool
$ idris -- ../Interfaces.idr
*../Interfaces> :doc Cool
Interface Cool
    The Cool interface

Parameters:
    (Eq a, Num a, Ord b) =>
    a   -- A parameter.
    b   -- Another parameter.

Methods:
    ...
```


owickstrom:
Failing tests, let me fix those. 😞

owickstrom:
The OSX job on Travis timed out, but tests seem to be passing now. Can we restart the Travis job?

david-christiansen:
Sorry for the somewhat late response here. In the context of interfaces, `Parameters` means something pretty specific:  they're the type arguments that are used to find the actual implementation that's needed for the user. It is possible for superinterface constraints to be parameters, such as when parameterizing an implementation _over another implementation_, but this is not what's happening in this example. To avoid confusion, I think that something like `Constraints` should form a separate section in the doc output.

Something like:
```
*../Interfaces> :doc Cool
Interface Cool
    The Cool interface

Parameters:
    (Eq a, Num a, Ord b) =>
    a   -- A parameter.
    b   -- Another parameter.

Constraints:
    Eq a, Num a, Ord b

Methods:
    ...
```

Then, when an implementation really is parameterized over another interface, it could look like:
```
||| Good functors preserve identity
||| @ f the functor
||| @ the dictionary we care about
interface GoodFunctor (f : Type -> Type) (impl : Functor f) where
    mapIdOk : {a : Type} -> (x : a) -> map @{impl} id x = x 

> :doc GoodFunctor
Interface GoodFunctor
  Good functors preserve identity

Parameters:
  f : Type -> Type -- the functor
  impl : Functor f -- the dictionary we care about

Methods:
  ...
```
Does this make sense? In the meantime, I'm restarting Travis.

owickstrom:
Well, maybe. 😄  I am a bit confused about the constraints still appearing before the parameters in your example (line 2):

```
Parameters:
    (Eq a, Num a, Ord b) =>
    a   -- A parameter.
    b   -- Another parameter.

Constraints:
    Eq a, Num a, Ord b
```

Are they supposed to be in both places? And can there be constraints in the Constraints collection on variables not in the interface parameters, e.g. type variables in interface function signatures. If not, then the `(Eq a, Num a, Ord b) =>` and the separate section with `Eq a, Num a, Ord b` will always be the same, right?

Regarding your example with `GoodFunctor`, it may be a bit out of my league, but I'm guessing that `@{impl}` is a way of forcing the implementation resolution to pick `impl` and its version of `map`.

Anyway, a bit off topic, but I tried evaluating your example in the REPL to learn more about how it works, but got this error:

```
    When checking type of constructor of Main.GoodFunctor:
     When checking an application of function Prelude.Functor.map:
             Type mismatch between
                     a1 (Type of x)
             and
                     f a (Expected type)
```

Thank you for your patience. 👍

david-christiansen:
Thanks for your patience with my caffiene-lacking copy-pasting! I meant to say:
````
*../Interfaces> :doc Cool
Interface Cool
    The Cool interface

Parameters:
    a   -- A parameter.
    b   -- Another parameter.

Constraints:
    Eq a, Num a, Ord b

Methods:
    ...
````

Sorry about that. I don't think that the constraints belong under the parameters unless they actually are parameters to the interface.

You're right that the @{dict} notation is the way to pass an explicit dictionary.

Here's a version of that definition that actually works, and isn't just my Github scribblings:
```
||| Good functors preserve identity
||| @ f the functor
||| @ impl the dictionary we care about
interface GoodFunctor (f : Type -> Type) (impl : Functor f) | f where
  mapIdOk : {a : Type} -> (x : f a) -> map @{impl} (\y => y) x = x

implementation GoodFunctor List %implementation where
  mapIdOk [] = Refl
  mapIdOk (x :: xs) = cong (mapIdOk xs)
```

The differences are:
 * id was being universally quantified, rather than referring to the name id. So I just made it an explicit lambda for clarity.
 * The `| f` notation is needed because there should only be one searched instance per `f`
 * I fixed the docstring

There's also an implementation, as an example.

**Comment by david-christiansen** [comment]
Docs for `GoodFunctor` here are:
```
Interface GoodFunctor
    Good functors preserve identity

Parameters:
    f   -- the functor
    impl   -- the dictionary we care about

Methods:
    mapIdOk : GoodFunctor f impl => (x : f a) -> map (\y => y) x = x
        
        
        The function is Total
Implementations:
    GoodFunctor List
                (constructor of Prelude.Functor.Functor (\meth =>
                                                           \meth =>
                                                             \meth =>
                                                               \meth =>
                                                                 Prelude.List.List implementation of Prelude.Functor.Functor, method map meth
                                                                                                                                         meth))
```

I suppose the printing of the dictionary object isn't so nice.

owickstrom:
Thanks for the examples! I've changed the implementation to them, and added a regression test and a changelog entry as well.

owickstrom:
Ping 😀 anything more that needs to be done?

david-christiansen:
I don't think so, but I'll give it a peek from my work computer tomorrow.
Thanks for this!

2016-10-16 2:50 GMT-04:00 Oskar Wickström <notifications@github.com>:

david-christiansen:
Looks great. Thanks again!

owickstrom:
Cool, thanks! :)

PR #3482, opened 2016-10-09
Fix .lidr compiler message source position.

PoroCYon:
The compiler reports bogus file positions in compiler warnings in a literate Idris program, for example:

```idris
> module Main
>
> main : IO ()
> main = "This is obviously a compiler error."
```

Without this fix, the compiler would report an error at the 'n' in "main" on the last line, while it's supposed to be at the '='.
This PR fixes the issue.

### DIFF
--- a/test/basic010/run
+++ b/test/basic010/run
@@ -4,7 +4,7 @@ set -eu
 
 TIMEOUT=../scripts/timeout
 
-$TIMEOUT 10 "${IDRIS:-idris}" "$@" Main.idr --nocolour --check --warnreach -o basic010
-$TIMEOUT 5  ./basic010
+$TIMEOUT 20 "${IDRIS:-idris}" "$@" Main.idr --nocolour --check --warnreach -o basic010
+$TIMEOUT 20  ./basic010
 
 rm -f -- *.ibc basic010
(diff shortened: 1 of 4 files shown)


PoroCYon:
The tests seem to be failing because of the compiler error positions having been changed, I didn't seem to have broken anything else.

melted:
You can update the expected results to the tests with a flag to the testrunner. Then add that to the PR.

test/basic010/run
```diff
@@ -4,7 +4,7 @@ set -eu
 
 TIMEOUT=../scripts/timeout
 
-$TIMEOUT 10 "${IDRIS:-idris}" "$@" Main.idr --nocolour --check --warnreach -o basic010
-$TIMEOUT 5  ./basic010
+$TIMEOUT 20 "${IDRIS:-idris}" "$@" Main.idr --nocolour --check --warnreach -o basic010
+$TIMEOUT 20  ./basic010
```
PoroCYon: I needed to give it more time, my poor i3 2330M isn't that quick.

PoroCYon:
> The job exceeded the maximum time limit for jobs, and has been terminated.

The others seem to have succeeded, though.

PoroCYon:
> An error occurred while generating the build script.

It seems to have turned against me...

jfdm:
i'm confused as why this was merged. The conversation indicated that there was a bug in the test script? Was this bug resolved, or not an actual bug? It would be nicer in future if a suitable comment was made when merging.

ahmadsalim:
@jfdm There was a bug in Travis, I restarted the particular job and everything passed. I think @melted then merged it at that point, since everything was OK.

PR #3493, opened 2016-10-17
Remove hack for fixing Travis caching bug

enolan:
The compiler: lines had the effect of getting us separate caches per GHC
version, but since travis-ci/travis-ci#4393 has been fixed we get separate
caches per matrix entry anyway.


enolan:
I put the `compiler: ` lines back after seeing the build status page - it's more readable with them in. Left the comment removed since it's still out of date.

Issue #3483, opened 2016-10-10
stdout/stdout buffering on Windows


CarstenKoenig:
the default buffering used in Windows (Win10 64bit / de-DE) will make things like

```
plusIO : IO Int
plusIO = do
  putStr "Inp 1: "
  a <- getLine
  putStr "Inp 2: "
  b <- getLine
  pure $ a ++ b
```

behave strangely (it will only show the prompts after you did input both values)

On top of this it's impossible to use `:x plusIO` from emacs (Atoms REPL won't work at all)

---

This is a problem I have with every console I/O in Haskell and the usual solution is to use `hSetBuffering` to `NoBuffering` or issuing some `hFlush`s in between.

So it would be nice if you could for example add some `hFlush`s (probably around here: https://github.com/idris-lang/Idris-dev/blob/692d61015dc0c7ee8ea1279b77e9a5f339dff3a7/src/Idris/Core/Execute.hs#L452)

(I don't feel competent enough to issue a PR myself  ... but I can try if this is something you would consider)

melted:
I think you should take a shot at it, it's always nice to get new contributors!

I'll do it otherwise, but I'm likely even more clueless about things.

CarstenKoenig:
thanks

as this is probably not very important for most people I would love to give this a try - although it will probably take me a few days

It's probably better to add something similar like `hSetBuffering` for stdout/stdin to `...Interactive` (where `putStr'`, etc. is located) - so I would start with this if this is fine with you.

david-christiansen:
Just so you know, `Execute.hs` is only used by `:x` and the type provider extension. Compiled code doesn't go through that interpreter, and compiled code is probably what you want to update. 

I'd suggest adding the equivalent of `hSetBuffering` using the C FFI, and then calling it in the appropriate places.

melted:
@david-christiansen Note the `:x` in the example.

david-christiansen:
Gah, sorry! Thanks!

ahmadsalim:
Seems that this was resolved by #3495 , so I will close it.

PR #3494, opened 2016-10-18
added a `disableBuffering` action into Prelude.Interactive

CarstenKoenig:
this command will be translated into a call `hSetBuffering` or `setvbuf`
to disable buffering in stdin/stdout

the reason for this is that on windows you sometimes need disable buffering
to see the prompt when doing something like this:

    do putStr "What's your name? "
       name <- getLine
       ...

### DIFF
--- a/libs/prelude/IO.idr
+++ b/libs/prelude/IO.idr
@@ -124,6 +124,9 @@ unsafePerformIO : IO' ffi a -> a
 unsafePerformIO (MkIO f) = unsafePerformPrimIO
         (prim_io_bind (f (TheWorld prim__TheWorld)) (\ b => prim_io_pure b))
 
+prim_noIOBuffering : IO' l Int
+prim_noIOBuffering = MkIO (\w => prim_io_pure (prim__noBuffering (world w)))
+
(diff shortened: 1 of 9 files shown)


CarstenKoenig:
this is my go on this [issue](https://github.com/idris-lang/Idris-dev/issues/3483)

Melvar:
My opinion is that this does not warrant an internal primitive. An FFI call or at most an external primitive should be sufficient, though the Executor of course has to implement it specially either way because it uses Haskell `Handle`s (a fact which is problematic any time you have to deal with a `FILE*` operation it does not override).

CarstenKoenig:
can you give me a hint of how to do this the way you proposed (I just crawled along the putStrLn prelude function here)

CarstenKoenig:
oh wait - is it actually as easy as this?

```
putChar : Char -> IO ()
putChar c = foreign FFI_C "putchar" (Int -> IO ()) (cast c)
```

CarstenKoenig:
abandoned in favor of this one: https://github.com/idris-lang/Idris-dev/pull/3495 - sorry for the fail attempt

PR #3495, opened 2016-10-18
this command will be translated into a call `hSetBuffering` or `setvbuf`

CarstenKoenig:
to disable buffering in stdin/stdout

the reason for this is that on windows you sometimes need disable buffering
to see the prompt when doing something like this:

    do putStr "What's your name? "
       name <- getLine
       ...

---

this should take care of [this issue](https://github.com/idris-lang/Idris-dev/issues/3483)

### DIFF
--- a/libs/prelude/Prelude/Interactive.idr
+++ b/libs/prelude/Prelude/Interactive.idr
@@ -89,6 +89,16 @@ putCharLn c = putStrLn (singleton c)
 getChar : IO Char
 getChar = map cast $ foreign FFI_C "getchar" (IO Int)
 
+||| Disables buffering in both stdin and stdout
+||| so that output is written immediately, and never stored in the buffer
+||| and the next input item is read and returned
+|||
+||| this is useful to circumvent problem with IO on some windows system
+||| where stdout output sometimes is only shown after prompted input from
+||| stdin
+disableBuffering : IO ()
+disableBuffering = foreign FFI_C "idris_disableBuffering" (IO ())
+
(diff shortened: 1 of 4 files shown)


Melvar:
This is better. What I initially meant, and to answer your question from #3494, was that you could put a foreign call directly to `setvbuf`, and then in the executor override `setvbuf`. But I think this solution might actually be better for this case, as it enables one to still call the actual `setvbuf` from the Executor.

libs/prelude/Prelude/Interactive.idr
```diff
@@ -89,6 +89,16 @@ putCharLn c = putStrLn (singleton c)
 getChar : IO Char
 getChar = map cast $ foreign FFI_C "getchar" (IO Int)
 
+||| Disables buffering in both stdin and stdout
+||| so that output is written immediately, and never stored in the buffer
+||| and the next input item is read and returned
+|||
+||| this is useful to circumvent problem with IO on some windows system
```
ahmadsalim: Could you kindly replace 'windows system' with 'Windows systems'? (I think capitalization helps realising that it's about the operating system).

CarstenKoenig: of course - sorry

ahmadsalim: Thanks! No need to apologize, it was just a friendly suggestion.

CarstenKoenig:
thanks for accepting it :smiley:

PR #3507, opened 2016-10-29
Fix Data.Vect.deleteBy

mitchellwrosen:


jfdm:
Thanks for the contribution. I am wondering if you can add a little bit more substance to the commit message, as to what was wrong with the original code and what the fix does. This is more for the git records than anything else.

mitchellwrosen:
Sure. What should it say? I'm not sure how to describe it words other than "fix" :)

orokanasaru:
The definition of List.deleteBy in the prelude is:

    deleteBy : (a -> a -> Bool) -> a -> List a -> List a
    deleteBy _  _ []      = []
    deleteBy eq x (y::ys) = if x `eq` y then ys else y :: deleteBy eq x ys

This is identical to the definition in Haskell. 

While I personally prefer the behavior in your change, it would be inconsistent with existing deleteBy functions.

mitchellwrosen:
Ah! My mistake. I thought the function was meant to delete all elements that satisfy the predicate, not just the first.

jfdm:
@mitchellwrosen Sorry for the late reply, I thought I had already replied. It would be nice if the commit message's detailed description could explain what was wrong with the original code, and what the fix does to address said problem. Sometimes, the original problem *may* not be easily seen by all. It is nicer to make errors explicit.

Sorry for any confusion.

mitchellwrosen:
@jfdm No worries, turns out what I thought was an "obvious typo" was actually not a bug at all!

PR #3508, opened 2016-10-30
Updating range behavior to match Haskell's where appropriate

### LINKED ISSUE #3502: If the second parameter to the three parameter range is 0, the result is always the empty list
``` idris
Idris> [1..1]
[1] : List Integer
Idris> [0..0]
[0] : List Integer
Idris> [1,1..1]
[1] : List Integer
Idris> [0,0..0]
[] : List Integer
Idris> [-2,(-1)..1]
[-2, -1, 0, 1] : List Integer
Idris> [-1,0..2]
[] : List Integer
Idris> [2,1..-1]
[2, 1, 0, -1] : List Integer
Idris> [1,0..-2]
[] : List Integer
```

### LINKED ISSUE #3503: Nonsensical ranges should be empty
```
Idris> [3,1..15]
[3, 1, -1, -3, -5, -7, -9] : List Integer
```

Why is that the result? The same input in Haskell just returns []

orokanasaru:
This resolves #3502 and resolves #3503. This also fixes the issue that natEnumFromThenTo could not descend.

I used `make && make test` on Ubuntu.

After adding the new tests, the existing implementation crashed.

```
    basic013:          
Error: test/basic013
run: line 3: 19432 Segmentation fault      (core dumped) ./basic013
```

With the new implementation, all 202 tests passed.

I also built this on Windows, but was unable to run the tests, so I generated a stress test suite that passed range definitions to both Idris and Haskell to verify the output would be equivalent. 

There's one notable difference from Haskell. In Haskell, ranges of the form `[A,A..B]` generate an infinite sequence of `A`s. I chose to follow the existing implementation, which returns `[]` instead. This rule conflicts with another rule that ranges of the the form `[A,B..A]` return `[A]` in the case that `B == A` (i.e. the form `[A,A..A]`). My implementation returns `[A]` in that case.


jfdm:
Thanks for the contribution. 

I am wondering if it is possible that the comments made in the github comment are placed in the git commit history. Otherwise if we merge this, then we loose information about the commit (the intent and more of the why) as it is not placed in the git history. Remember, github comments are not pushed into git commits.

orokanasaru:
Good idea. I just re-pushed with the updated commit message.

jfdm:
@orokanasaru thanks for the PR, it is much appreciated.

PR #3519, opened 2016-11-16
Remove incorrect CPP and require newer Cabal instead

Melvar:
The `MIN_VERSION_*` macros are only generated during setup, i.e. after Setup.hs
is compiled. The GHC version is not a proxy for the Cabal version, as it is
trivial to install a newer version of Cabal and cabal-install with
cabal-install, and in that case the Setup.hs would not typecheck (this is what happened for me). Since it
doesn’t appear to be possible to switch on the Cabal version, this change
simply declares a requirement for a newer version.

The second commit (fixing warnings from cabal after raising the version) is possibly opinionated. Feel extra free to leave it off or change it.

Specifically, if the `cabal-version` constraint is above 1.10, each component
should have a `default-language` field. It appears the default was/is
`Haskell98`, so I entered that for each component.

Similarly, the `extensions` field is then deprecated in favor of
`default-extensions` and `other-extensions`. For these I have investigated which
extensions are actually required globally (i.e. some files assume them but do
not declare them) and have entered these as `default-extensions` and removed
them from any files that did declare them. I then collected all additional
extensions declared in any files into `other-extensions`.

It may be a good idea to think about these settings and change them, and also to
prune declared but unused extensions from the source files, but in this commit I
have only tried to replicate the previous settings (minus removing default extensions declared in the files; perhaps I should have left that to a later extension-pruning step as well).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,12 +11,12 @@ matrix:
     - os: osx
       env: CABALVER="1.24" GHCVER="8.0.1" TESTS="test_c"
       compiler: ": #GHC 8.0.1"
-    - env: CABALVER="1.20" GHCVER="7.6.3" TESTS="test_c"
+    - env: CABALVER="1.22" GHCVER="7.6.3" TESTS="test_c"
       compiler: ": #GHC 7.6.3"
-      addons: {apt: {packages: [cabal-install-1.20,ghc-7.6.3,cppcheck,hscolour], sources: [hvr-ghc]}}
-    - env: CABALVER="1.20" GHCVER="7.8.4" TESTS="test_c"
+      addons: {apt: {packages: [cabal-install-1.22,ghc-7.6.3,cppcheck,hscolour], sources: [hvr-ghc]}}
+    - env: CABALVER="1.22" GHCVER="7.8.4" TESTS="test_c"
       compiler: ": #GHC 7.8.4"
-      addons: {apt: {packages: [cabal-install-1.20,ghc-7.8.4,cppcheck,hscolour], sources: [hvr-ghc]}}
+      addons: {apt: {packages: [cabal-install-1.22,ghc-7.8.4,cppcheck,hscolour], sources: [hvr-ghc]}}
     - env: CABALVER="1.22" GHCVER="7.10.3" TESTS="lib_doc doc"
       compiler: ": #GHC 7.10.3"
       addons: {apt: {packages: [cabal-install-1.22,ghc-7.10.3,cppcheck,hscolour], sources: [hvr-ghc]}}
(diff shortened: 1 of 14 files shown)


Melvar:
Ah, of course I need to adjust the cabal versions for Travis as well …

Melvar:
Hm … apparently that’s not how it works?

jfdm:
@Melvar In case you didn't realise, the build is failing due to a problem with the latest version of the safe package not liking GHC 7.6.3 and Cabal 1.20. There is a PR (#3523) that addresses this. Once that is merged can you rebase, and then we can merge this. It looks like the other problems are due to caching on travis. Hopefully a rebase will fix that...

Melvar:
@jfdm Thank you very much for figuring out the problems!

jfdm:
Rebase ahead!

Melvar:
I clearly don’t know how to configure travis to have the necessary cabal version.

ahmadsalim:
@Melvar What is the status of this?

Melvar:
I could not figure out how to make Travis have the required cabal version with the older GHC versions, so I could not make Travis pass. Waiting for someone who can explain what’s wrong.

gpyh:
I authored that horrendous shortcut. I spent days trying to figure out the proper way to capture the cabal version, without success. Using the GHC version seemed like a good proxy at the time.
AFAIK, there is no way to alter Setup.hs depending on the cabal version, which is a shame became the API is incompatible between versions.
A more radical stance would be to only support recent versions of cabal, but I don't know if you'd be welcome to that.

melted:
I wouldn't mind dropping ghc 7.6 if that would help.

jfdm:
When it comes to supporting versions of GHC, I think we should aim to support the version of GHC present on Debian Stable on a clean install. It might be a hassle but, I think it is a decent rule of thumb for GHC support policy nonetheless.

ahmadsalim:
I agree with @jfdm, that we generally should avoid dropping support for older GHCs unless necessary and if 7.6 is still used in Debian Stable, that makes it a good idea to still support it.

I would not mind dropping support for 7.10 however, since it has some instabilities that make it generally broken and is currently blocking a few of our PRs. E.g., Quipper http://www.mathstat.dal.ca/~selinger/quipper/ skips 7.10 because of instability.

ahmadsalim:

Thanks.

ahmadsalim:
So it seems that this PR was abandoned because there were issues with Travis. If any @idris-lang/contributors does know how to fix that, please help!

Melvar:
So actually the existing code doesn’t fail for me anymore, because I got GHC 7.10.3 sometime recently. It’s still logically wrong and there could still be others using new Cabal with old GHC.

Melvar:
I have rebased and force-pushed, but a closed pull request apparently does not auto-update.

ahmadsalim:
@Melvar Yeah, please open a new one :)

PR #3521, opened 2016-11-18
Boolean monoid wrappers

jachymb:


ahmadsalim:
@jachymb Thanks for the PR. Could you however kindly put it in `contrib` rather than `prelude`?

jachymb:
I am not sure what would be the suitable place in the codebase where it should be put then. Any recommendations please?

ahmadsalim:
Make a folder called Bool in `https://github.com/idris-lang/Idris-dev/tree/master/libs/contrib/Data` and then add your changes under a file of suitable name.

jachymb:
This would be probably better in #3557

ahmadsalim:
If you think it's better to consider this under #3557 I will close the PR.
Please reopen it if you would like it under separate consideration again.

PR #3527, opened 2016-11-24
Removed unused copyright field in Haddock documentation.

jfdm:
We *should* make the effort to document this properly. However, I don't know what the proper attribution should be. So rather than having it blank, lets remove it completely.


ahmadsalim:
@jfdm The proper attribution could be Idris Developers or Idris Contributors and perhaps a reference to the contributors file.

jfdm:
I suppose that could work...

jfdm:
I think the suggestions of @ahmadsalim are decent. Will close this PR and replace with a more suitable solution.

PR #3530, opened 2016-11-28
Fixed formatting of man page.

jfdm:


jfdm:
The Mac OS X build timed out. Given that the rest of the suite passed, and the change is not related to the code, I think it is *safe* to merge.

PR #3533, opened 2016-11-29
Made Mac OS X package generatation more robust.

jfdm:
+ Improved error handling.
+ Improved script progress reporting.
+ Allowed for custom pointing to Idris that is not built by cabal i.e. stack.
+ Fix pkgbuild error by creating the `tmp` directory. You must have permission to write to `tmp`.
+ Added metadata and fixed flag to `pkgbuild`.

### DIFF
--- a/mkpkg.sh
+++ b/mkpkg.sh
@@ -1,26 +1,64 @@
 #!/bin/sh
 
-VERSION=$1
+set -euo pipefail
 
-echo "Building version-$VERSION\n\n"
-echo "Have you: set the release flag, checked the demos and the tutorial?"
-read $foo
+if [ "$#" -lt 1 ]; then
+    echo "Usage $0 [version number] ?[idris bin location]"
+    echo ""
+    echo "The first argument is the version number"
+    echo ""
+    echo "The second argument is optional, and used to point to a"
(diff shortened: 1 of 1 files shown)


ahmadsalim:
@jfdm One could also make it that the Travis packages are automatically added as releases, I think.

jfdm:
We could look into that, but this raises the question over: Which distribution targets should we support, and which targets we should not support?

I like to think that `cabal` and `stack` do a good enough job, for distributing idris. Provision of the mac package and windows build are helpful but not necessary.

If there is a way such that we can distribute Idris easily to many platforms then okay, otherwise, it is only going to increase the maintenance effort for the project.

jfdm:
@ahmadsalim Concerning the binary builds, please don't get me wrong. I am not against their provision, I just want to make sure we can do it in a sustainable maintainable way.

I am going to open an issue for supporting generation of binary builds, just so discussion can be recorded there. Good luck with it.

ahmadsalim:
@jfdm Thanks, I will at least try to see if I can get it working during the weekend.

LeifW:
Travis does have OSX machines - maybe they could be leveraged for that?

Issue #3517, opened 2016-11-08
Add nonzero restriction on `gcd`


Qqwy:
The current version of `gcd` is defined as follows:

```Idris
gcd : Nat -> Nat -> Nat
gcd a 0 = a
gcd a b = assert_total (gcd b (modNat a b))
```

Because I'm working on a small hobby project that needs to prove things about Greatest Common Divisors, I stumbled across a page on [proofwiki](https://proofwiki.org/wiki/Existence_of_Greatest_Common_Divisor). It states that a requirement for being able to compute `gcd a b`, is that `a≠0 ∨ b≠0`. The current definition does not have this requirement, which means that `gcd 0 0` is accepted as valid input (producing the value 0. Of course, zero can never be a divisor for anything, as you cannot ever divide by zero, so this is wrong.).

I am too new to Idris to know how to write `a≠0 ∨ b≠0` in Idris proofs (especially the 'or' operation), but I think changing the definition of GCD would be great, as we can then use `modNatNZ`(which is total) instead of `modNat` (which is not), so the `assert_total` would no longer be needed.

ahmadsalim:
@Qqwy A simple way to write that constraint would be to take argument `(notBothZero : (a = 0) -> (b = 0) -> Void)` (which is isomorphic to `Either (Not (a = 0)) (Not (b = 0))` and `Not (a = 0, b = 0)`).

fabianhjr:
Note: changing from `modNat` to `modNatNZ` was not enough to gurantee that gcd would be total. Asertion had to be kept. (Refer to #3537)

PR #3537, opened 2016-11-30
Add restriction to gcd

fabianhjr:
This PR restricts the domain of `gcd` since it incorrectly returned `0` for `gcd 0 0`. It closes issue #3517 and it should now be possible to prove that `gcd a b` is not zero.

### DIFF
--- a/libs/prelude/Prelude/Nat.idr
+++ b/libs/prelude/Prelude/Nat.idr
@@ -93,7 +93,6 @@ hyper (S pn)   a (S pb) = hyper pn a (hyper (S pn) a pb)
 
 ||| Proofs that `n` or `m` is not equal to Z
 data NotBothZero : (n, m : Nat) -> Type where
-  NeitherIsZero  : NotBothZero (S n) (S m)
   LeftIsNotZero  : NotBothZero (S n) m
   RightIsNotZero : NotBothZero n     (S m)
 
@@ -393,9 +392,9 @@ log2 (S n) = log2NZ (S n) SIsNotZ
 --------------------------------------------------------------------------------
 -- GCD and LCM
 --------------------------------------------------------------------------------
-gcd : (a: Nat) -> (b: Nat) -> {auto ok: NotBothZero a b} -> Nat
-gcd a Z = a
-gcd Z b = b
+gcd : (a: Nat) -> (b: Nat) -> .{auto ok: NotBothZero a b} -> Nat
+gcd a Z     = a
+gcd Z b     = b
 gcd a (S b) = assert_total $ gcd (S b) (modNatNZ a (S b) SIsNotZ)
 
 lcm : Nat -> Nat -> Nat
--- a/libs/prelude/Prelude/Nat.idr
+++ b/libs/prelude/Prelude/Nat.idr
@@ -396,12 +396,12 @@ log2 (S n) = log2NZ (S n) SIsNotZ
 gcd : (a: Nat) -> (b: Nat) -> {auto ok: NotBothZero a b} -> Nat
 gcd a Z = a
 gcd Z b = b
-gcd a (S b) = assert_total (gcd (S b) (a `modNat` (S b)))
+gcd a (S b) = assert_total $ gcd (S b) (modNatNZ a (S b) SIsNotZ)
 
 lcm : Nat -> Nat -> Nat
 lcm _ Z     = Z
 lcm Z _     = Z
-lcm a (S b) = assert_total (divNat (a * (S b)) (gcd a (S b)))
+lcm a (S b) = assert_total $ divNat (a * (S b)) (gcd a (S b))
 
 
 --------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


fabianhjr:
It seems like AppVeyor timed-out. Could someone re-run it?

ahmadsalim:
@fabianhjr I am rerunning it now.

Melvar:
Is there a reason to have `NotBothZero m n` instead of just using `Either (Nonzero m) (Nonzero n)`? The “NeitherIsZero” constructor is superfluous as is since either of the other two can be used instead in that case.

fabianhjr:
Yes, I did my best try at the suggestinons @ahmadsalim gave in the issue however I am new to idris. Specifically when I tried to use Either (Nonzero a) (Nonzero b):

```
- + Errors (1)
 `-- ./Prelude/Nat.idr line 396 col 4:
     When checking type of Prelude.Nat.gcd:
     No such variable Either
```

And if I attempt to import Prelude.Either:
```
- + Errors (1)
 `-- (no file) line 0 col -1:
     Cycle detected in imports: Prelude/Nat.idr -> Prelude/Either -> Prelude/List -> Prelude/Nat -> Prelude/Either
```

So, what would probably be better would be to refactor Nat.idr into two files? (Nat.idr with the basic constructors and something like Nat/Math.idr or Math.idr)

Other attempts: Using Not (Zero a, Zero b)

```
- + Errors (2)
 |-- ./Prelude/Nat.idr line 397 col 4:
 |   When checking right hand side of gcd with expected type
 |           Nat
 |   
 |   When checking argument ok to Prelude.Nat.gcd:
 |           Can't find a value of type 
 |                   Not (Zero (S b), Zero (modNatNZ a (S b) SIsNotZ))
 `-- ./Prelude/Nat.idr line 402 col 4:
     When checking right hand side of lcm with expected type
             Nat
     
     When checking argument ok to Prelude.Nat.gcd:
             Can't find a value of type 
                     Not (Zero a, Zero (S b))
```

(It will attempt both, so only nonzero inputs will be accepted. Example `gcd 1 0` and `gcd 0 1` won't work)

Melvar:
Ah, an import cycle. That’s a good enough reason for me for the moment, though I’d kinda hope it could be resolved later. I still would recommend either removing the both-successor constructor, or restricting the other two constructors to zero in the non-successor index.

fabianhjr:
@Melvar, followed the second suggestion.

fabianhjr:
@melvar, there were some issues with the Type system. Resetting back.

fabianhjr:
Appveyor failed again; Needs a rerun. Apparently it doesn't handle `git reset --hard HEAD` |> `git push --force`  well. :/

ahmadsalim:
@fabianhjr I will restart the service.

Please note that you can also setup Travis/AppVeyor on your own private repos for free as well.

libs/prelude/Prelude/Nat.idr
```diff
@@ -91,6 +91,13 @@ hyper (S pn)   a (S pb) = hyper pn a (hyper (S pn) a pb)
 -- Comparisons
 --------------------------------------------------------------------------------
 
+||| Proofs that `n` or `m` is not equal to Z
+data NotBothZero : (n, m : Nat) -> Type where
+  NeitherIsZero  : NotBothZero (S n) (S m)
```
ahmadsalim: This case is strictly not needed, so perhaps you can just remove it.

libs/prelude/Prelude/Nat.idr
```diff
@@ -386,16 +393,15 @@ log2 (S n) = log2NZ (S n) SIsNotZ
 --------------------------------------------------------------------------------
 -- GCD and LCM
 --------------------------------------------------------------------------------
-partial
-gcd : Nat -> Nat -> Nat
+gcd : (a: Nat) -> (b: Nat) -> {auto ok: NotBothZero a b} -> Nat
```
ahmadsalim: Can you please make this `.{auto ok : NotBothZero a b}` (adding a dot in front) instead, so it will be ensured to be erased?

libs/prelude/Prelude/Nat.idr
```diff
@@ -386,16 +393,15 @@ log2 (S n) = log2NZ (S n) SIsNotZ
...
 gcd a Z = a
-gcd a b = assert_total (gcd b (a `modNat` b))
+gcd Z b = b
+gcd a (S b) = assert_total $ gcd (S b) (modNatNZ a (S b) SIsNotZ)
```
ahmadsalim: Would it be possible to investigate the use of `assert_smaller` instead of `assert_total` here? I think it would be nicer to have in Prelude at least, but it's not essential.

fabianhjr: Type mismatch, did several attempts at it but no dice.

libs/prelude/Prelude/Nat.idr
fabianhjr: Done

libs/prelude/Prelude/Nat.idr
fabianhjr: Done

ahmadsalim:
@fabianhjr You are welcome to add your name to CONTRIBUTORS. I will merge this PR.

PR #3547, opened 2016-12-04
Change export rules for Data.String.Views module

### LINKED ISSUE #3545: Data.String.View are not exporting functions
It seems that there should be:
```idris
%access public export
```
in https://github.com/idris-lang/Idris-dev/blob/master/libs/base/Data/String/Views.idr.

Otherwise I'm not sure how to use it. Also, why is there assert_total, just curious?

Termina1:
Fixes #3545

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -206,6 +206,7 @@ TorosFanny
 Trevor Elliott
 uwap
 Vitaly Bragilevsky
+Vyacheslav Shebanov
 xged
 XIE Yuheng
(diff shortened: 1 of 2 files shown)


jfdm:
@Termina1 Thanks for the PR, if you want to add yourself to the CONTRIBUTORS list please do. I think we can wait till tomorrow to merge this if you want to add yourself to the list in the between time.

jfdm:
@Termina1 thanks for the commit.

PR #3548, opened 2016-12-05
Restrict System.usleep interval

### LINKED ISSUE #3505: System.usleep is a no-op if argument is >= 1000000
Idris version:  0.12.2

I ran into this bug using the Windows 64-bit binary on Windows 7 Enterprise.  If you call `System.usleep` with an argument >= `1000000`, then no sleeping actually happens.  This appears to be related to the specification of the `usleep` method in unistd.h ([link to documentation](http://pubs.opengroup.org/onlinepubs/007908799/xsh/usleep.html)).  Specifically:

>  The useconds argument must be less than 1,000,000. If the value of useconds is 0, then the call has no effect. 

I have only tried this on Windows.  Behavior may be different on Linux/Mac.

Maybe the Idris `usleep` method could be tightened up a bit to avoid this unexpected behavior...but I am sure there are good arguments either way.

eklavya:
Interval values above 1000000 are no-ops on some platforms. This PR restricts the interval value to [0, 1000000]. Fixes #3505


jfdm:
@eklavya Thanks for the PR that addresses issue #3505.

Some minor comments:

1. Styling: Is it possible to introduce a space between the names and types in the type signature? This is to introduce consistency with the rest of the module and Idris base.

2. Naming: Is it possible to change the name of the implicit argument to `prf`. This is to introduce consistency with the rest of Idris' base and prelude.

3. Documentation: I know `usleep` wasn't documented before your fix, but is it possible to introduce some documentation here to describe how to use the function?

4. Proof Style: Personally, I would prefer using a data type to encode the custom range for values as a 'predicate', however, I think the style presented is sufficient. I don't think anyone else would object to the PR.

I cannot remember if this is your first contribution, if yes then please add yourself to the contributor's files, otherwise can you check that your name is listed.

Thanks 

Jan

libs/base/System.idr
```diff
@@ -61,7 +61,8 @@ time : IO Integer
 time = do MkRaw t <- foreign FFI_C "idris_time" (IO (Raw Integer))
           pure t
 
-usleep : (i: Int) -> { auto p: (i > 0 && i <= 1000000) = True } ->  IO ()
+||| Specify interval to sleep for, must be in range [0, 1000000]
+usleep : (i : Int) -> { auto prf : (i > 0 && i <= 1000000) = True } -> IO ()
```
ahmadsalim: Would it be possible to use `So (i > 0 && i <= 1000000)` instead?

libs/base/System.idr
```diff
@@ -62,7 +64,7 @@ time = do MkRaw t <- foreign FFI_C "idris_time" (IO (Raw Integer))
           pure t
 
 ||| Specify interval to sleep for, must be in range [0, 1000000]
```
ahmadsalim: Sorry small comment again, but it would be nice if you could write the range as `]0, 1000000]` to show that it is exclusive.

eklavya: We are restricting values above 1000000 because that results in unexpected behaviour on some platforms. The API works as expected in range [0, 1000000]. Let me know if you want to change it to (0, 1000000] or ]0, 1000000] (haven't read this notation before).

ahmadsalim:
@eklavya I think the tests fail because `usleep` has `public export` modifier and requires a So. Please us `import public Data.So` instead so dependent code does not break.

melted:
Let's try to rerun the mac job.

ahmadsalim:
Thank you for the PR, I will merge this. The code seems OK, and we are still getting issues with the Mac builds timing out, which is not your fault. Please put your name in `CONTRIBUTORS.md` if it's not already there.

PR #3549, opened 2016-12-05
Relax safe bounds, Fixes #3543.

### LINKED ISSUE #3543: safe == 0.3.9
I didn't understand change aa91025 well.

Would `safe >= 0.3.9` not work equally well?

It looks like if this was done then latest idris might build in Stackage Nightly again.

jfdm:
Adjusted bounds in cabal file, but not in stack config. That can come later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,8 @@ before_install:
     fi
 
 install:
+  - which cabal
+  - which ghc
   - cabal --version
   - echo "$(ghc --version) [$(ghc --print-project-git-commit-id 2> /dev/null || echo '?')]"
   - if [ -f $HOME/.cabal/packages/hackage.haskell.org/00-index.tar.gz ];
--- a/idris.cabal
+++ b/idris.cabal
@@ -289,7 +289,7 @@ Library
                 , vector < 0.12
                 , vector-binary-instances < 0.3
                 , zip-archive > 0.2.3.5 && < 0.4
-                , safe == 0.3.9
+                , safe >= 0.3.9
                 , fsnotify >= 0.2 && < 2.2
                 , async < 2.2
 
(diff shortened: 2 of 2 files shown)


jfdm:
Right, after digging around I *think* we are experiencing [this cabal issue](https://github.com/haskell/cabal/issues/3003). Whilst a backdated patch was released for 'older' cabal versions, it appears that the [HVR PPA](https://launchpad.net/~hvr/+archive/ubuntu/ghc) to get ghc versions on Ubuntu didn't push the fixed version of Cabal for 1.20. This is the cabal we are experiencing problems with.

Two ways, forward: 

1. We bump the cabal versions to latest and greatest for all build's. IIRC cabal is backwards compatible. This comes at the cost of realistic assumptions about ghc and cabal version parings, however, we could assume and specify that everyone where they can updates their cabal, and ensure that PATH is setup correctly.

2. We ask HVR to provide a more up-to-date version of Cabal in their archive.

juhp:
I am not sure, but would:

```
if impl(ghc < 7.7)
   Build-depends: safe == 0.3.9
```

help?

jfdm:
@juhp I've done (2), and didn't think of adding in a conditional to cabal. Although, I am not in favour of such conditionals, I do see their need. I've pushed a fix and hopefully this will work. Thanks for the comments.

jfdm:
The Mac OS X build timed out during benchmarking. Given that the rest of the test suite passed, and the change resulted in a successful build and test run, I think it is safe to merge.

juhp:
Thanks!  Then hopefully the next release will build in Stackage Nightly again. :-)

Only other suggest that came to mind now was requiring newer Cabal to build, but your change looks good/optimal to me.

PR #3551, opened 2016-12-05
Improve progress reporting.

jfdm:
This commit improves the default progress reporting to clearly indicate at which stage Idris is during compilation.

For a simple hello world program with no dependencies, the output using `-V` will be:

```
$ idris Hello.idr -o a.out -V
Type checking ./Hello.idr
Totality checking ./Hello.idr
Writing IBC for: "./Hello.ibc"
Compiling Output.
Generating Code.
```

### DIFF
--- a/src/IRTS/Compiler.hs
+++ b/src/IRTS/Compiler.hs
@@ -53,7 +53,9 @@ import System.Process
 -- |  Compile to simplified forms and return CodegenInfo
 compile :: Codegen -> FilePath -> Maybe Term -> Idris CodegenInfo
 compile codegen f mtm
-   = do checkMVs  -- check for undefined metavariables
+   = do v <- verbose
+        when v $ iputStrLn "Compiling Output."
+        checkMVs  -- check for undefined metavariables
         checkTotality -- refuse to compile if there are totality problems
         exports <- findExports
         let rootNames = case mtm of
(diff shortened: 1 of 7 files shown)


src/IRTS/Compiler.hs
```diff
@@ -108,7 +110,7 @@ compile codegen f mtm
             Just f -> runIO $ writeFile f (dumpDefuns defuns)
         triple <- Idris.AbsSyntax.targetTriple
         cpu <- Idris.AbsSyntax.targetCPU
-        logCodeGen 1 "Building output"
+        when v $ iputStrLn "Generating Code."
```
ahmadsalim: Any reason you have removed it from the log?

jfdm: Yes, I am experimenting with getting helpful verbose output to track progress through the compiler. I could always reinstate the logging output as well, but that seems a bit too verbose when logging is turned on.

jfdm:
This is failing because I forgot to update the node tests.

jfdm:
Interestingly, there was an existing  'logical' flaw for when to report progress. When using the node backend, Idris would be verbose during compilation even if no verbosity flag was set. For, `C` everything was fine. Type checking and parsing were not verbosely reported.

The fix was to turn default verbosity to `False`. I suspect the error was a result of the C codegen correctly setting verbosity to false, and the Node codegen doesn't. This was deduced because the node tests were failing only because of verbose output from the compiling stage of Idris for Node.

jfdm:
This is a minor change, that has 'some' user facing impact. It makes things possibly *too* verbose. But does provide users with a better notion of Idris' progress. 

I'll let this sit for a bit before I merge tomorrow---2016-12-08.

In future I think we could migrate to a two level verbosity to report this and then greater details of each stage, for example, the inlining and defunctionalisation of compiling, and universe checking. Three level verbosity would be nicer in which we introduce a simpler higher level that indicates in which stage we are in i.e. Type Checking and totality checking sans which files. But that might be harder to do after looking at the files.

jfdm:
Have just rebased against `master`, will merge once tests are complete.

PR #3552, opened 2016-12-06
Improve `System.Exit` Fixes #3488.

### LINKED ISSUE #3488: System.Exit returns `()`, not `a`
In System.idr, the `exit` function returns `()`, unlike its haskell equivalent, which has type signature `IO a`. It seems a bit inflexible.

`https://github.com/idris-lang/Idris-dev/blob/master/libs/base/System.idr#L57`

jfdm:
Brings core functions for temrinating programs in line with Haskell requivalents.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# New in 0.999
+
+## Library Updates
+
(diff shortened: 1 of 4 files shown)


jfdm:
oops, i pushed to early by accident.

PR #3554, opened 2016-12-06
Add %access export statement

justjoheinz:
Without the export access declared test functions are not visible to the main function which is generated to execute the tests.

### DIFF
--- a/docs/tutorial/testing.rst
+++ b/docs/tutorial/testing.rst
@@ -42,6 +42,8 @@ A simple test module, with a qualified name of ``Test.NumOps`` can be declared a
     module Test.NumOps
 
     import Maths.NumOps
+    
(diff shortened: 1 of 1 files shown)


jfdm:
@justjoheinz Thanks for the PR. We don't necessarily keep track of all the examples in the tutorial, and good to see when broken code gets fixed.

jfdm:
I had a quick look at the CI builds, and the Mac OS X build timed out. I've restarted it. I see no reason to accept it once the tests in the Mac OS X build have completed a run; this shouldn't matter given the nature of the change, but wouldn't mind seeing a complete run of the tests.

jfdm:
The Mac OS X build timed out during benchmarking. Given that the rest of the test suite passed, and the change resulted in a successful build and test run, I think it is safe to merge.

PR #3560, opened 2016-12-07
Update the reference for packages

justjoheinz:
Package files support comments. Added the syntax to the reference.


jfdm:
Nice, I was just in the middle of doing this. Thanks!

PR #3563, opened 2016-12-09
Added `Category Iso` in base

Synthetica9:
Added `Iso` as a `Category`

### DIFF
--- a/libs/base/Control/Isomorphism.idr
+++ b/libs/base/Control/Isomorphism.idr
@@ -2,6 +2,7 @@ module Control.Isomorphism
 
 import Syntax.PreorderReasoning
 import Data.Fin
+import Control.Category
 
 %default total
 %access public export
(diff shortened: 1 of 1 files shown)


jachymb:
In fact, this could be generalized further. Isomorphism is an equivalnce relation (and thus a preorder). A preorder is a Category (thin category in particular). Hence, in terms of contrib/Decidable/Order.idr we could define `Iso` as instance of `Equivalence` (and `Preorder`) and then define `Preorder` as a subclass of `Category`.

david-christiansen:
@Synthetica9 Thanks for this! It's a nice addition.

@jachymb There are many way to do fine-grained algebraic structures in dependent types. We're not doing it in the base library, because the trade-offs between the different approaches are unclear, so we at least take something that's fairly simple and familiar to Haskellers. Thanks, though!

Issue #3500, opened 2016-10-20
Move `abs` out of the `Neg` interface: not every number has a positive-definite absolute value.


Niriel:
I am writing modules to implement Dual and Split-Complex numbers.  Unlike Real or Complex numbers, we have to be extremely careful with the use of the words "absolute value" and "modulus".

Reals, complex, dual, split-complex, quaternions, octonions, biquaternions and combinations thereof have a "modulus" property that derives from the notion of conjugation.  Said modulus may not always be positive (split-complex).  If the modulus is positive, it may still not be positive-definite (dual).  Only when the modulus has all the required properties can it be promoted to "absolute value".  Some of these numbers do not have an absolute value, only a modulus, even though they can be negated.

Matrices and tensors can also be negated, but even the notion of modulus is ill-defined, bringing us one more step farther from an absolute value.

I would like to suggest that `abs` be taken out of `Neg` and placed into its own interface.

david-christiansen:
This sounds reasonable to me. Will you please submit a pull request?

Niriel:
I would, if I had a Linux machine.  My laptop is dead, and my current windows desktop makes no sense to me.  I'll install a Linux VM and try to do just that.

One worry, however: this will break some existing code.  How do I deal with that?

david-christiansen:
The right way to break existing code is to announce your intentions and reasons on the mailing list ahead of time, then make the change and submit a PR, then announce your intentions one more time prior to merging. Make sure that the file CHANGELOG contains a description of what happened and how to fix broken code so that it shows up in the release notes.

Niriel:
All right, I've got my VirtualBox.  Getting to work.

ahmadsalim:
@Niriel Any update?

Niriel:
Yeah.  The change was trivial and I got Idris to compile with an `Abs` interface separated from `Neg`.  Here are my commits from last year which implement this fix.

https://github.com/Niriel/Idris-dev/commits/absolute_value

I haven't made a Pull Request because the Google Groups discussion didn't reach a conclusion.

https://groups.google.com/d/msg/idris-lang/DXDbM50_Wug/jLP2fN30CQAJ

In short: Prelude makes simplifications for the sake of convenience. Making a correct algebraic tower would ideally be better than including small patches like mine but it's a hard job. However, small improvements may be better than nothing.

I don't know, guys.  I'm a noob, I can't predict the implications of this change. My branch is one year old; I can make it up to date if you want.

[...]

PR #3565, opened 2016-12-09
Generalized the Num instance for Complex

Synthetica9:
Due to a historical artifact, the `Num` implementation for `Complex` only covered `Double`, because
`abs` used to be in `Num` (like in Haskell). Since these implementations are now split, we can
implent Num with only `Neg a` instead of Double.

As soon as `abs` moves out of `Neg` (issue #3500), we can also loosen the requirment for the `Neg`
implementation to only need `Neg a`:

```idris
implementation Neg a => Neg (Complex a) where
    negate = map negate
    (-) (a:+b) (c:+d) = ((a-c):+(b-d))
```

### DIFF
--- a/libs/base/Data/Complex.idr
+++ b/libs/base/Data/Complex.idr
@@ -70,7 +70,7 @@ implementation Neg a => Num (Complex a) where
     (*) (a:+b) (c:+d) = ((a*c-b*d):+(b*c+a*d))
     fromInteger x = (fromInteger x:+0)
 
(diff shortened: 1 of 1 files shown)


melted:
Thanks for the PR!

jfdm:
can the CHANGELOG be updated for this please.

Niriel:
By the way, I believe this Issue is solved.  When I implemented the fix for issue #3500, I did see @Synthetica9 's proposed code in complex.idr but it was commented out.  I just had to un-uncomment.

PR #3568, opened 2016-12-10
Add missing quotation marks for `in`

rizo:


jfdm:
Hi @rizo thanks for the PR, it is much appreciated. 

We also have a PR offering the same fix (#3570) that was submitted after yours. Given the two requests I have flipped a coin to decide which one to use, and your side won.

If you wish to contribute more, [we have various labels](https://github.com/idris-lang/Idris-dev/labels) that indicate various areas of Idris that we would like help with. Feel free to contribute further or not.

PR #3574, opened 2016-12-12
Improved User facing progress reporting.

jfdm:
Idris' output has been updated to more accurately reflect its
progress through the compiler i.e. Type Checking; Totality Checking;
IBC Generation; Compiling; and Code Generation. To control the
loudness of the reporting three verbosity levels are introduced:
`-V`, `-V1`, and `-V2`.


jfdm:
Am making some last minute tweaks. Will merge this later today.

ahmadsalim:
@jfdm Is it not more common to have `-V0`, `-V1` and `-V2` instead of using two dashes?

jfdm:
@ahmadsalim its the same convention we have with the optimisation levels, which I think arises because of a limitation of opt-parse appreciative concerning single dash with non-short options. That is: i believe that a single dash is only usable with single letter options.

ahmadsalim:
@jfdm Ah, I see. I mean I guess it's too late to change, but I thought it was because `-V0` should be read as `-V 0` (that is a single letter followed by an argument).

jfdm:
Double checking opt parse's documentation, it would be possible to make the optimisation and verbose options use `-O` and `-V` respectively and allow an integer to be passed. 

Since it applies to both flags, I shall leave that change for later.

PR #3580, opened 2016-12-15
idris.cabal: bump aeson version

mpkh:
so far I can't see any issues with newer aeson version

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -252,7 +252,7 @@ Library
                 , Tools_idris
 
   Build-depends:  base >=4 && <5
-                , aeson >= 0.6 && < 1.1
+                , aeson >= 0.6 && < 1.0.2.2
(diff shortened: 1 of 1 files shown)


idris.cabal
```diff
@@ -252,7 +252,7 @@ Library
                 , Tools_idris
 
   Build-depends:  base >=4 && <5
-                , aeson >= 0.6 && < 1.1
+                , aeson >= 0.6 && < 1.0.2.2
```
ahmadsalim: I am unsure if this is intentional, but you are actually restricting it more not widening it.

mpkh:
@aochagavia my fault, I've used to have that workaround for an old version and really don't need it anymore

aochagavia:
I guess you meant @ahmadsalim

mpkh:
@aochagavia yes, sorry :)

PR #3581, opened 2016-12-15
Make idris distro friendly

### LINKED ISSUE #3573: Distributing idris for `guixsd`
Hi,

While updating the idris package for the guixsd distribution I noticed a couple of things that could be improved to help packaging idris and idris modules.

`IDRIS_LIBRARY_PATH` only accepts a single path instead of a colon separated list. I think this would help packaging nixos too.

I think that the location of the idris standard libraries is `/share/x86_64-linux-ghc-7.10.2/idris-0.12.3` by default. It would be nice to be able to overwrite this, since guix/nix already take care of making sure that the idris modules are built using the right ghc/idris versions, by setting the correct `IDRIS_LIBRARY_PATH`. Having them in /share/idris makes it easy to have guix set the search path automatically by adding something like this to the idris package definition.
```scheme
(native-search-paths
   (list (search-path-specification
           (variable "IDRIS_LIBRARY_PATH")
           (files '("share/idris")))))
```

If I understand correctly the tests currently run on the installed idris binary and not on the one that was just built. Running `runhaskell Setup.hs test` after `runhaskell Setup.hs` build fails to find idris.

dvc94ch:
IDRIS_LIBRARY_PATH accepts a colon separated search path.

The idris install procedure respects the TARGET environment variable.

Closes #3573

### DIFF
--- a/src/Idris/Imports.hs
+++ b/src/Idris/Imports.hs
@@ -94,7 +94,7 @@ findPkgIndex p = do let idx = pkgIndex p
 installedPackages :: IO [String]
 installedPackages = do
   idir <- getIdrisLibDir
-  filterM (goodDir idir) =<< dirContents idir
+  filterM (goodDir idir) =<< concat <$> (mapM dirContents (splitSearchPath idir))
   where
   allFilesInDir base fp = do
     let fullpath = base </> fp
@@ -102,7 +102,9 @@ installedPackages = do
     if isDir
       then fmap concat (mapM (allFilesInDir fullpath) =<< dirContents fullpath)
       else return [fp]
-  dirContents = fmap (filter (not . (`elem` [".", ".."]))) . getDirectoryContents
+  prependDirname dir = fmap (fmap (dir </>))
+  filterPaths = fmap (filter (not . (`elem` [".", ".."])))
+  dirContents dir = prependDirname dir (filterPaths (getDirectoryContents dir))
   goodDir idir d = any (".ibc" `isSuffixOf`) <$> allFilesInDir idir d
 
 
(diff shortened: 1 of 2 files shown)


src/Idris/Imports.hs
```diff
@@ -94,15 +95,17 @@ findPkgIndex p = do let idx = pkgIndex p
 installedPackages :: IO [String]
 installedPackages = do
   idir <- getIdrisLibDir
-  filterM (goodDir idir) =<< dirContents idir
+  filterM (goodDir idir) =<< concat <$> (mapM dirContents (splitOn ":" idir))
```
Melvar: Doesn’t this break on Windows, as `:` is used to indicate a drive prefix? What separator is used on Windows for this kind of environment variable? Is there a standard function to do this?

jfdm:
The `filepath` package provides functions for splitting search paths. For example: `splitSearchPath`. It would be better if they are used rather than hard coded path separators.

dvc94ch:
Yes I already updated the patch. The CI failure looks related to network connectivity.

jfdm:
There still appears to be hard coded posix specific code in your patch.

dvc94ch:
Do you mean this?

filterPaths = fmap (filter (not . (`elem` [".", ".."])))

This line was there before and shouldn't affect windows. On windows I'd expect it to be a NOP. But I really don't have a windows machine to test it.

jfdm:
@dvc94ch I commented on an outdated diff, sorry.

dvc94ch:
Closing for now. Worked around the issue by wrapping idris like nixos does. The patch for IDRIS_LIBRARY_PATH would need more work, since it only partially solves the problem.

PR #3582, opened 2016-12-16
A simple extension system for embedded code generators

mmn80:
Addresses #3542 and a followup of #3577. It is "step 1" from the last comment there.

This makes possible for people to create and publish embedded code generators.

Such a package would depend on `idris` and produce an executable to be used in place of the normal `idris`. In the `main` function, the appropriate `register*` functions would be called.

It is also possible to create a batteries included package, that imports many such codegens from multiple 3rd parties and creates an `idris` executable. Maintaining it would be reasonably simple, as only the `Main.hs` of `idris` is cloned, and would have to be updated only rarely.

A template project that shows how to do just that can be found [here](https://github.com/mmn80/idris-codegen-ext). In the README file there is more detailed info. It took maybe 20 mins to set up. I cloned the C codegen and RTS from `idris`, and the repo can be used as a starting point for development of a custom C RTS (which is what I intend to do for my Erlang style code loading project).

This PR does not change anything about `idris` unless used as described (in a separate package).

### DIFF
--- a/src/IRTS/Compiler.hs
+++ b/src/IRTS/Compiler.hs
@@ -29,6 +29,7 @@ import IRTS.Lang
 import IRTS.LangOpts
 import IRTS.Portable
 import IRTS.Simplified
+import IRTS.System (getExtCodeGenerator)
 
 import Prelude hiding (id, (.))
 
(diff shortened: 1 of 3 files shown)


mmn80:
I played a bit with the `FreeT` extra layer to `Idris` as an extension mechanism.

If put on the inside it is surprisingly easy to implement (a `MonadException` instance, a few extra `lift`s, and `iterT` for interpretation), but then the handlers will have type `IO` rather then `Idris`, so it would only be useful for lifting codegens and other IO, but not for REPL/IDE commands.

`FreeT` on the outside leads to some complicated type tetris that is not worth it unless we want things like the elaborator reflection, which are best done on the Idris side, as they are language, rather then runtime extensions. So I think this direction is a dead end.

In conclusion, if we restrict our extensions to codegens and REPL/IDE commands, the best method seems to be just extending `IState` with a new field. It is somewhat cleaner then the global `MVar` in this PR, but requires a few more changes/cleanups (eg: also moving the `getLibFlags` in the `IState`, adding an extra argument to `idrisMain` for initialization, rather then the global `register`).

And lifting IO + refactoring should be treated as a completely separate project, which is perfectly feasible with the `FreeT` on the inside method.

mmn80:
I'm giving up on this as well. For my project I'll just make my own IDE mode from scratch, and then I'll be able to extend it as much as I want.

The refactoring project is worth it only in the context of actual language extensions.

ahmadsalim:
@mmn80 I am unsure why you are giving up, but please do remember that we are an open source project that is mostly driven by volunteers (not a big company like Microsoft) and so it takes us some time to review code and other issues.

mmn80:
I closed this PR bc I realized this should be redone in `IState`, together with REPL extensions. If you want to merge that, I'll make another PR, just tell me. But notice that w/o refactoring idris itself will not be able to redistribute 3rd party codegens, it will have to be a separate batteries included package. I explained on the other PR why I postponed refactoring. This kind of reduces the usability of such extensions.

I gave up bc I realized that my project diverges from IDE mode anyway, as, for instance, I want to use this through FFI, not IPC (compile it as a shared lib, not executable), so I have to make my custom IDE mode server anyway.

So I'm waiting for Backpack for refactoring, and for extensions I'm looking for something more serious, that can be used for actual 3rd party language extensions, not just REPL/codegens. That too would benefit greatly from refactoring (idris could bundle 3rd party lang. extensions).

ahmadsalim:
@mmn80 Thanks for the response. I was just worried that you gave up, because we were a bit slow to respond; I think your plan to have it in `IState` sounds good.

PR #3584, opened 2016-12-17
Prove LTE antisymetry and mark LTE related functions total

jachymb:
I have proven antisymetry for LTE. It is important property, because together with reflexivity (`lteRefl`) and transitivity (`lteTransitive`) this proves that (Nat, LTE) is a partially ordered set, hence I believe that `lteAntisymetric` should be at the same place in code as the other two. Further, in maths, it is a common proof technique to prove equality by proving inequality bound in both directions.

Further, I have marked some LTE related functions as total (they are), in order to make the code more consistent with the rest of the file.

### DIFF
--- a/libs/prelude/Prelude/Nat.idr
+++ b/libs/prelude/Prelude/Nat.idr
@@ -151,7 +151,7 @@ lteTransitive LTEZero y = LTEZero
 lteTransitive (LTESucc x) (LTESucc y) = LTESucc (lteTransitive x y)
 
 ||| `LTE` is antisymetric
-total lteAntisymetric : {n : Nat} -> {m : Nat} -> LTE n m -> LTE m n -> n = m
+total lteAntisymetric : LTE n m -> LTE m n -> n = m
 lteAntisymetric {n = Z} {m = Z} LTEZero LTEZero = Refl
 lteAntisymetric {n = (S j)} {m = (S k)} (LTESucc x) (LTESucc y) =
   let inductiveHypothesis = lteAntisymetric {n = j} {m = k} x y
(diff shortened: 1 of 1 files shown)


xekoukou:
Hello @jachymb, the file contains the "%default total" in the beginning which means that all functions are total.

Also, eqSucc is defined in a lower position than your function, hence the error.

jachymb:
@xekoukou : Why are most functions in the file marked as `total` explicitly anyway then? Is it for historical reasons?

jachymb:
@xekoukou : Is it better to move `eqSucc` above `lteAntisymetric` or to move `lteAntisymetric` below `eqSucc` accodring to Idris libs coding best practices?

xekoukou:
1. Probably.
2. I do not know, give it a go and probably others will tell what is best.

jachymb:
I have fixed the code to use `cong` instead `eqSucc`

justjoheinz:
Wouldn't it be more consistent with the other proofs to write the signature `total lteAntisymetric : {n : Nat} -> {m : Nat} -> LTE n m -> LTE m n -> n = m`  without the implicit arguments?

jachymb:
@justjoheinz Good point. I have removed the implicit arguments from the signature.

anton-trunov:
There is also no need in bringing `n` and `m` into scope with `{n = ...}` and `{m = ...}`.

jachymb:
@anton-trunov Personally, I find it the most readable as it is now. Is this a required change in order to merge the PR?

anton-trunov:
@jachymb I don't really know if it is required. However, bringing those implicits into scope makes the new code less consistent with the existing code. The existing code in `Prelude.Nat` uses implicits, but as far as I know it does so when it is really needed (like in `lteRefl`).

JanBessai:
Please also take https://github.com/idris-lang/Idris-dev/blob/master/libs/contrib/Decidable/Order.idr into consideration. It already includes antisymmetry proofs for natutal and finite numbers.

jachymb:
Honestly, I did not know about Decidable.Order.

JanBessai:
It has been living a shadowy live since 2013 #281 . At some point there was an issue to consolidate operations like lte, but I cannot find it anymore.

ahmadsalim:
@jachymb If you could consolidate the `Decidable.Order` properties with this, I think we would happily merge it.

jachymb:
@ahmadsalim I am not sure what is meant by that. :(

ahmadsalim:
@jachymb I meant move the relevant functions from `Decidable.Order` (namely those that don't relate to decidability) to where your PR is and make the style more uniform.

ahmadsalim:

Thanks.

PR #3588, opened 2016-12-21
Check for node when running test suite.

jfdm:
If node isn't installed then don't run the test suite. This check
could possibly be performed when the test option for running node
is set, but it is not clear how.

Regardless, if you are contributing a pull request then you would
need to run the tests for both `C` and `Node` codegens.

### DIFF
--- a/test/TestRun.hs
+++ b/test/TestRun.hs
@@ -107,11 +107,17 @@ runTest path flags = do
       normalise [] = []
 
 main :: IO ()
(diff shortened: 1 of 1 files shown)


jfdm:
The lack of node check almost got me when updating the node tests with accept; I hadn't installed node and almost changed the expected output to error messages.

I tested this on my machine, but don't know if Windows will cause complications.

PR #3590, opened 2016-12-21
Provide an example of idris example

justjoheinz:
The documentation reference lacked an example of using idris example to document code.


jfdm:
Thank you.

Issue #3594, opened 2016-12-28
The Pruviloj.Induction.induction error message could be improved


yurrriq:
When I make this typo:

```idris
mapPreservesLen : (f : a -> b) -> (l : List a) ->
                  length (map f l) = length l
mapPreservesLen = %runElab mush
```

*[Specifically, I forgot to include `f` on the left-hand side.]*

I get the [somewhat cryptic error message](https://github.com/idris-lang/Idris-dev/blob/v0.99/libs/pruviloj/Pruviloj/Induction.idr#L135):

```
When checking right hand side of mapPreservesLen with expected type
        (f : a -> b) -> (l : List a) -> length (map f l) = length l

a -> b is not an inductive family
```

`f` needs to be on the left-hand side so we do induction on `l`, which, unlike `f`, **is** an *inductive family*:

```idris
mapPreservesLen f = %runElab mush
```

yurrriq:
> [Consider] prefix[ing the error message] with a `NamePart` referring to the function so it gets colored and is clickable in interfaces that make names clickable.

david-christiansen:
Isn't `mapPreservesLen` already highlighted and clickable in that message?

I'm all for better messages, though. At the very least, "inductive family" can be replaced with "a datatype declared with the data keyword"

yurrriq:
> Isn't `mapPreservesLen` already highlighted and clickable in that message?

For me it's not highlighted (`List`, `length` and `map` are), but it is indeed clickable.

:+1: for "a datatype declared with the data keyword"

david-christiansen:
Right, it's not highlighted because the highlighting is based on it having defined the function, and the function does not yet exist when it makes that message :-) It would be nice to figure out a model where that doesn't happen.

PR #3595, opened 2016-12-29
Improve Pruviloj.Induction.induction error message

yurrriq:
The current error message is a bit cryptic. Replace it with @david-christiansen's suggestion.

See discussion on #3594 for details.


david-christiansen:
Thanks!

PR #3597, opened 2016-12-29
Use same format in IDE mode for hole types as REPL

david-christiansen:
A while ago, the REPL got a feature that distinguished between
premises in the apparent type of a hole and its surrounding context,
but IDE mode was not updated accordingly. This makes them act the same.

Fixes https://github.com/idris-hackers/idris-mode/issues/427.


david-christiansen:
After this change:
```
Holes

- + ASDF.sumLength [E]
 `--              xs : List String
                   t : Type -> Type
         __interface : Foldable List
                elem : Type
     ---------------------------------------
      ASDF.sumLength : String -> Nat -> Nat
```

just as in the REPL.

PR #3598, opened 2016-12-30
Import dependencies in Atom

allancto:
This communicates the solution Rui Barreiro and David Christiansen explained to me for Atom. 
https://groups.google.com/forum/#!topic/idris-lang/5i0FZoYj8A0

The formatting is not perfect, somebody needs to check it for consistency with the rest of the tutorial. 

I'm also not sure about where exactly this information belongs, since it relates to both imports and to packages, tutorial and Atom. I've put it where I would most likely look for it.


edwinb:
Looks fine to me. Thanks!

PR #3603, opened 2017-01-05
More lenient casting from string to number

### LINKED ISSUE #3601: When casting string to number, leading whitespace is allowed but trailing whitespace is not
As the title says, the following works (note the leading spaces):
~~~
Idris> the Int $ cast "  123"
123 : Int
Idris> the Double $ cast "  123"
123.0 : Double
Idris> the Integer $ cast "  123"
123 : Integer
~~~
while the following does not (note the trailing spaces):
~~~
Idris> the Int $ cast "123  "
0 : Int
Idris> the Double $ cast "123  "
0.0 : Double
Idris> the Integer $ cast "123  "
0 : Integer
~~~

As is the case in Haskell, I would expect both of these to work.

SuprDewd:
Now whitespace is also allowed after the number being cast, like so:

~~~
the Int $ cast "  123  "
the Integer $ cast "  123  "
the Double $ cast "  123  "
~~~

Fixes #3601

### DIFF
--- a/src/Idris/Primitives.hs
+++ b/src/Idris/Primitives.hs
@@ -455,8 +455,8 @@ getInt _ = []
 
 strToInt :: IntTy -> [Const] -> Maybe Const
 strToInt ity [Str x] = case reads x of
-                         [(n,"")] -> Just $ toInt ity (n :: Integer)
-                         _        -> Just $ I 0
+                         [(n,s)] -> Just $ if all isSpace s then toInt ity (n :: Integer) else I 0
+                         _       -> Just $ I 0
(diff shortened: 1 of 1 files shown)


jfdm:
can the changeling be updated to reflect the change please.

SuprDewd:
> can the CHANGELOG be updated to reflect the change please.

@jfdm I believe this was already the expected behavior, and this PR just fixes a bug (i.e. see #3601). Do you still think it should be in the changelog? If you do, I will happily add it.

jfdm:
I think it would be nice to mention that the 'bug' was fixed.

PR #3605, opened 2017-01-05
Implement second part of FORCE_ALIGNMENT fix for arduino targets.

XeFias:
When FORCE_ALIGNMENT was introduced for pointer aligment on 8-bit arduino targets (See idris-lang@47cd264) it was not applied to the initialization of the cheney algorithm during garbage collection. The fix modifies the start of the heap location in the same way the FORCE_ALIGNMENT-define does for the initialization of the FP-heap in alloc_heap(...) in file idris_heap.c.

### DIFF
--- a/rts/idris_gc.c
+++ b/rts/idris_gc.c
@@ -73,7 +73,16 @@ VAL copy(VM* vm, VAL x) {
 void cheney(VM *vm) {
     int i;
     int ar;
-    char* scan = vm->heap.heap;
+    char* scan = NULL;
+	
+#ifdef FORCE_ALIGNMENT
+    if (((i_int)(vm->heap.heap)&1) == 1) {
+       scan = vm->heap.heap + 1;
+    } else
+#endif
+    {
+        scan = vm->heap.heap;
+    }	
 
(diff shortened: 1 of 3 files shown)


ahmadsalim:
Would it be possible to make it an (inline) function instead, and then calling it the right places?
I would think that would be better than introducing duplicated code.

XeFias:
Of course,

The following function will do the job. I can add it to the idris_heap.c module and include it in idris_gc.c. idris_gc.c already includes idris_heap.h indirectly via idris_rts.h. Including idris_heap.h would however make it explicit.
Also, do you think the function name is clean enough?
```
char* aligned_heap_pointer(char * heap) {
#ifdef FORCE_ALIGNMENT
    if (((i_int)heap&1) == 1) {
	    return (heap + 1);
    } else
#endif
    {
	    return heap;
    }
}
```

ahmadsalim:
@XeFias That looks good.

ahmadsalim:
Looks good, thanks for the PR! Please update `CONTRIBUTORS.md` with your name.

PR #3607, opened 2017-01-06
Allow failures for Travis Mac build

melted:
It times out more often than not.


melted:
We can remove it from there when it shapes up, one way or another.

PR #3609, opened 2017-01-09
Zipper implementation for Lists

justjoheinz:
I ported most of the functions from the according haskell library. There are still some implementations for interfaces missing. I will add those if this initial commit is deemed useful.


edwinb:
Looks useful to me. In general, I'm happy to have all kinds of experimental stuff in contrib, ready for organisation later. We may need to think about better organisation of packages at some stage though...

PR #3615, opened 2017-01-15
Update documentation in Prelude/Chars.idr to reflect unicode

Kyarigwo:
Just a tiny change in the documentation to reflect Unicode behaviour or the chr and ord functions.

As far as I can tell, a Char in Idirs is a unicode code point, and that the chr and ord functions convert from unicode code points to Char.


david-christiansen:
Actually, it's a bit unspecified what a Char is in Idris. The C backend, and all backends that I'm aware of, support Unicode characters, but it would be acceptable for a backend for some very limited or old-fashioned system to have Char mapped to EBCDIC, ASCII, or something else. So while I agree that the docs shouldn't read ASCII, they should probably say something like "a backend-determined numeric value, typically Unicode".

melted:
The tests assume that unicode works I know because my lua backend failed those tests :)

david-christiansen:
That's true! It would be handy to update the test suite to tests that are backend-specific and those that help define what it means to be Idris.

Kyarigwo:
I've updated the changes, adding a comment to the module about different backends.

edwinb:
Thanks!

PR #3628, opened 2017-01-21
More ambitious JSON generation from TT

melted:


jfdm:
Can an entry be added to the CHANGELOG please, that details what new information is being added to the JSON output.

melted:
Ok

Issue #3616, opened 2017-01-15
Incorrect typing of instance method when using named parent instances


atacratic:
I'm trying to define a named implementation which has two named parent implementations, with the new experimental "using" syntax (from https://github.com/idris-lang/Idris-dev/commit/254a227307830016a4cf277ab9866395fe155893).  But the expected type of the methods is coming out wrong (I think).

The problem is in the last line of the code below, in the implementation of MultNatMonoidV.  The compiler reckons the expected type of the hole (?h3) involves plus and 0, but I think it should be mult and 1.

(If I can get round this I'd like to add implementations for VerifiedMonoid and the like, for Nat and Data.ZZ, into libs/contrib/Interfaces/Verified.idr - any comments on the work in progress here https://github.com/idris-lang/Idris-dev/compare/master...atacratic:NatZZVerifiedImpls?expand=1 would be welcome.)

```
-- idris --noprelude --nobuiltins
module Main

import Prelude.Nat

infixl 6 <++>

interface Semigroup' ty where
  (<++>) : ty -> ty -> ty

interface Semigroup' ty => Monoid' ty where
  neutral' : ty

interface Semigroup' a => VerifiedSemigroup a where
  total semigroupOpIsAssociative' : (l, c, r : a) -> l <++> (c <++> r) = (l <++> c) <++> r

[MultNatSemi] Semigroup' Nat where
  (<++>) = mult

[MultNatMonoid] Monoid' Nat using MultNatSemi where
  neutral' = S Z

[MultNatSemiV] VerifiedSemigroup Nat using MultNatSemi where
  semigroupOpIsAssociative' = ?h1

interface (VerifiedSemigroup a, Monoid' a) => VerifiedMonoid a where
  total monoidNeutralIsNeutralL : (l : a) -> l <++> Main.neutral' = l
  total monoidNeutralIsNeutralR : (r : a) -> Main.neutral' <++> r = r

[MultNatMonoidV] VerifiedMonoid Nat using MultNatSemiV, MultNatMonoid where
  monoidNeutralIsNeutralL = ?h2
  monoidNeutralIsNeutralR = ?h3 -- ":t h3" returns "(r : Nat) -> plus r 0 = r" but it should be "(r : Nat) -> mult r 1 = r"

```

Melvar:
[...]

PR #3612, opened 2017-01-11
Add proof of associativity of addition for integers (Data.ZZ).

atacratic:
Adds definition of the following (plus supporting lemmas) to contrib:

```idris
  plusAssociativeZ : (l, c, r : ZZ) -> l + (c + r) = (l + c) + r
```

PR #3629, opened 2017-01-21
Add proofs of ring laws to Data.ZZ in contrib

atacratic:
Adds proofs to Data.ZZ in contrib - associativity and commutativity for multiplication, and
distributivity of multiplication over addition.

This completes the proofs of the ring axioms for Data.ZZ - finishes the job started in https://github.com/idris-lang/Idris-dev/pull/3612 (already merged).

If this goes in, and if https://github.com/idris-lang/Idris-dev/issues/3616 is fixed at some point, I'll try following up with some instances of 'VerifiedSemigroup' and the like, along the lines of https://github.com/atacratic/Idris-dev/commit/30dbbf7f63ca0eac29ca14fd4eda10d408b839b7

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -42,6 +42,7 @@ Cezar Ionescu
 Chao-Hong Chen
 Chetan Taralekar
 Chris Allen
+Chris Gibbs
 Christopher Schwaab
 Cliff Harvey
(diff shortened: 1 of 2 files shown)


edwinb:
These seem good to have to me, thanks. Sorry for slowness.

PR #3631, opened 2017-01-22
change <= to < in text to match code block

### LINKED ISSUE #3630: typo in tutorial
There's a small typo in theorems.rst, in the section on totality checking. In "Hints for totality" the code block for qsort uses `filter (< x)`, but the paragraph above the example says "the checker cannot decide that ``filter (<= x) xs`` will always be smaller than ``(x :: xs)``.

Shouldn't the paragraph use `<` since the code block does?

derekmorr:
This fixes #3630 .


ahmadsalim:
Thanks for the PR.

PR #3632, opened 2017-01-24
Add :! for running shell commands from the REPL

RyanGlScott:
This adds to the Idris REPL a counterpart to GHCi's `:!` command, an incredibly useful feature which lets you shell out to some other command. I use this quite frequently to run `ls`, `vim`, and other utilities as I am working in an interactive REPL session.

The code for running and tab-completing `:!` is adapted from GHCi's implementation [here](http://git.haskell.org/ghc.git/blob/9fd87ef8a16fbbce35205ae63d75d239bb575ccc:/ghc/GHCi/UI.hs#l1204) and [here](http://git.haskell.org/ghc.git/blob/9fd87ef8a16fbbce35205ae63d75d239bb575ccc:/ghc/GHCi/UI.hs#l2996), respectively.

### DIFF
--- a/src/Idris/Completion.hs
+++ b/src/Idris/Completion.hs
@@ -159,6 +159,7 @@ completeColour (prev, next) = case words (reverse prev) of
 completeCmd :: String -> CompletionFunc Idris
 completeCmd cmd (prev, next) = fromMaybe completeCmdName $ fmap completeArg $ lookupInHelp cmd
     where completeArg FileArg = completeFilename (prev, next)
+          completeArg ShellCommandArg = completeFilename (prev, next)
           completeArg NameArg = completeName UpTo [] (prev, next)
           completeArg OptionArg = completeOption (prev, next)
           completeArg ModuleArg = noCompletion (prev, next)
--- a/src/Idris/REPL/Commands.hs
+++ b/src/Idris/REPL/Commands.hs
@@ -17,6 +17,7 @@ data Command = Quit
              | Reload
              | Watch
              | Load FilePath (Maybe Int) -- up to maximum line number
+             | RunShellCommand FilePath
              | ChangeDirectory FilePath
              | ModImport String
              | Edit
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@
   `--V0`, `--V1`, and `--V2`. The old aliases of `-V` and `--verbose`
   persist.
 
++ New REPL command `:!` that runs an external shell command.
+
 ## Library Updates
 
 + Terminating programs has been improved with more appropriate
--- a/src/Idris/Help.hs
+++ b/src/Idris/Help.hs
@@ -11,6 +11,7 @@ module Idris.Help (CmdArg(..), extraHelp) where
 data CmdArg = ExprArg               -- ^ The command takes an expression
             | NameArg               -- ^ The command takes a name
             | FileArg               -- ^ The command takes a file
+            | ShellCommandArg       -- ^ The command takes a shell command name
             | ModuleArg             -- ^ The command takes a module name
             | PkgArgs               -- ^ The command takes a list of package names
             | NumberArg             -- ^ The command takes a number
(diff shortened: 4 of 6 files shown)


RyanGlScott:
Since I'm relatively new to contributing to Idris, what's the convention for updating the changlelog? Should I do it myself, or leave it to whomever merges this?

ahmadsalim:
@RyanGlScott Please do update the `CHANGELOG` if possible.

david-christiansen:
Handy! Have you tested this in Atom and/or Emacs to make sure it does the right thing?

RyanGlScott:
@david-christiansen, no, I don't have neither Atom nor Emacs installed.

ahmadsalim:
@RyanGlScott Also, please add yourself to `CONTRIBUTORS` as well

david-christiansen:
@RyanGlScott Okay, I can test it for you later today. The issue is that shell command output should not interfere with the communication over the IDE protocol. We may want to disable this command over the protocol on the assumption that the IDE environment has its own way of doing shell commands; or, alternatively, have the protocol indicate to the IDE that a shell command should be run rather than having Idris do it itself.

OTOH, that might mean that the shell command runs in a different environment than Idris itself.

Any opinions on this? I'm not a frequent `:!`'er in GHCI.

RyanGlScott:
@ahmadsalim, someone else did that for me ;) https://github.com/idris-lang/Idris-dev/commit/908053d1d0330c974903eded252d758439328a10#diff-98f07bc20cb66328be238119df96c490R177

@david-christiansen, I am completely unqualified to say if `:!` could affect IDEs, as I don't use one with GHC. But I'd be OK with disabling them over the protocol, if that works.

david-christiansen:
I think the best would be to return an indication that the shell command should be run, over the protocol. This is what we already do for commands like "compile and run", where the IDE is given a path to the thing to run rather than having Idris run it itself, so that the IDE can make a nicer interface. It's good to have commands mostly work the same at the console REPL and the IDE REPL.

Are you at IU today? We can take a peek at it together, if you'd like.

RyanGlScott:
@david-christiansen: very well, I've set it so that `:!` simply errors out in IDE mode with a message stating that it's not currently supported.

src/Idris/Completion.hs
```diff
@@ -159,6 +159,7 @@ completeColour (prev, next) = case words (reverse prev) of
 completeCmd :: String -> CompletionFunc Idris
 completeCmd cmd (prev, next) = fromMaybe completeCmdName $ fmap completeArg $ lookupInHelp cmd
     where completeArg FileArg = completeFilename (prev, next)
+          completeArg ShellCommandArg = completeFilename (prev, next)
```
ahmadsalim: Is there any reason you used filename completion here?

src/Idris/REPL/Commands.hs
```diff
@@ -17,6 +17,7 @@ data Command = Quit
              | Reload
              | Watch
              | Load FilePath (Maybe Int) -- up to maximum line number
+             | RunShellCommand FilePath
```
ahmadsalim: Should it not take a list of strings (because the command might take arguments)?

src/Idris/Completion.hs
RyanGlScott: This is the [same completion approach](http://git.haskell.org/ghc.git/blob/9fd87ef8a16fbbce35205ae63d75d239bb575ccc:/ghc/GHCi/UI.hs#l2996) that GHCi uses for `:!`. One particularly nice thing about it is that if you type `:! vim direc` in a location where there's a directory named `directory`, then tab-completion fills in the rest to give you `:! vim directory`.

src/Idris/REPL/Commands.hs
RyanGlScott: Conceptually, I suppose it could be split into `(String, [String])`. I opted not to do that, since:

1. I'm ultimately passing the whole thing to [`system`](http://hackage.haskell.org/package/process-1.4.3.0/docs/System-Process.html#v:system), which takes a single `String` as an argument (this is the [same way that GHCi does it](http://git.haskell.org/ghc.git/blob/9fd87ef8a16fbbce35205ae63d75d239bb575ccc:/ghc/GHCi/UI.hs#l1204)).
2. It integrates nicely with `strArg`, which takes one `String` argument.

So splitting it up into `(String, [String])` seems silly, given that its only purpose to be the payload for `system`.

ahmadsalim: OK, thanks. I have approved the changes.

PR #3634, opened 2017-01-27
Added some docs to base/Data/Bits.idr

Synthetica9:
These were completely undocumented, added some basic documentation.


melted:
Thanks!

PR #3636, opened 2017-01-29
Fix interpreter paths for NixOS by using /usr/bin/env

mrkgnao:
A small patch to allow the build scripts to work on NixOS. 
NixOS does not have the FHS `/bin/bash`-esque paths, so the shebangs don't work.

### DIFF
--- a/mkpkg.sh
+++ b/mkpkg.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env sh
 
 set -euo pipefail
 
--- a/stylize.sh
+++ b/stylize.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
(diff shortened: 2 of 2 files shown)


Melvar:
But `/usr/bin/env` does?

mrkgnao:
Yes, and it's the recommended way to do this for scripts that aren't
supposed to be NixOS-only.

`/bin/sh` does exist, but not `/bin/bash` (and AFAIK the former exists just
because of something weird `gcc` does otherwise).

mrkgnao:
Yes, and it's the recommended way to do this for scripts that aren't
supposed to be NixOS-only.

`/bin/sh` does exist, but not `/bin/bash` (and AFAIK the former exists just
because of something weird `gcc` does otherwise).

melted:
Thanks for the PR!

PR #3641, opened 2017-01-30
Fix the build with most recent blaze-markup (0.8)

RyanGlScott:
If you try to build Idris with the most recent versions of `blaze-html` and `blaze-markup`, you'll get a bunch of errors to the effect of

```
src/Idris/IdrisDoc.hs:546:38: error:
    • Couldn't match expected type ‘MarkupM ()’
                  with actual type ‘a0 -> MarkupM a0’
    • Probable cause: ‘Empty’ is applied to too few arguments
      In the expression: Empty
      In a stmt of a 'do' block:
        (if nullDocstring docstring then
             Empty
         else
             Docstrings.renderHtml docstring)
      In the second argument of ‘($)’, namely
        ‘do { (if nullDocstring docstring then
                   Empty
               else
                   Docstrings.renderHtml docstring);
              let args' = filter (\ (_, _, _, d) -> ...) args;
              if (not $ null args') || (isJust fixity) then
                  H.dl
                  $ do { ...;
                         .... }
              else
                  Empty }’
```

That's because `blaze-markup-0.8` changed the `Empty` constructor to take an argument (in the case of `type Html = Markup = MarkupM ()`, this argument is of type `()`).

An easy workaround is to just use `mempty` instead of `Empty`, which is backwards-compatible back to old versions of `blaze-markup`.

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -259,8 +259,8 @@ Library
                 , array >= 0.4.0.1 && < 0.6
                 , base64-bytestring < 1.1
                 , binary >= 0.7 && < 0.9
-                , blaze-html >= 0.6.1.3 && < 0.9
-                , blaze-markup >= 0.5.2.1 && < 0.8
+                , blaze-html >= 0.6.1.3 && < 0.10
+                , blaze-markup >= 0.5.2.1 && < 0.9
                 , bytestring < 0.11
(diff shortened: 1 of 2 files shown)


melted:
Thanks!

PR #3643, opened 2017-01-31
Allow colored output on MinTTY consoles on Windows

RyanGlScott:
Currently, the Idris REPL doesn't print colored output when you run it in a MinTTY console like Cygwin or MSYS:

idris-1

This makes me sad.

The reason this happens is because due to a quirk in the way MinTTY, calling `isATTY` in it will _always_ return `False`. But we shouldn't just use `isATTY` to determine if we're in a console that supports colors! We can detect the presence of MinTTY pretty reliably by inspecting the file descriptor that MinTTY uses for its standard output. The minimalist `mintty` library accomplishes this, so I've added a dependency on it.

Now we can have colored output on MinTTY:

idris-2

Yay!

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -328,7 +328,8 @@ Library
   if os(darwin)
      build-depends: unix < 2.8
   if os(windows)
-     build-depends: Win32 < 2.4
+     build-depends: mintty >= 0.1 && < 0.2
+                  , Win32 < 2.4
   if flag(FFI)
      build-depends: libffi < 0.2
      cpp-options:   -DIDRIS_FFI
--- a/src/Idris/Main.hs
+++ b/src/Idris/Main.hs
@@ -105,8 +105,9 @@ idrisMain opts =
 
        when (DefaultTotal `elem` opts) $ do i <- getIState
                                             putIState (i { default_total = DefaultCheckingTotal })
-       tty <- runIO isATTY
-       setColourise $ not quiet && last (tty : opt getColour opts)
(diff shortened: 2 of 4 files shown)


Melvar:
Is this not a bug in the other software?

RyanGlScott:
The fact that `isatty` doesn't work on MinTTY is something which the maintainer has [declared as wontfix](https://github.com/mintty/mintty/issues/482#issuecomment-147179713).

GHC has to work around this too by employing [the same trick](http://git.haskell.org/ghc.git/blob/32729d3586d7ecdeb8561b6d0b2a688db709560c:/compiler/main/SysTools/Terminal.hs#l68).

melted:
Travis was just a time-out. Restarted that job.

melted:
@Melvar It's a quirk, if it's nothing much we can accommodate those. Like with the nixOS hashbang.

@RyanGlScott It would be nice if the check could be folded into the tty check in Util/System.hs, so it doesn't leak into Main.hs and Output.hs.

RyanGlScott:
@melted, good idea, done.

melted:
Thanks for the PR!

jfdm:
I think this user facing change should be mentioned in the CHANGELOG. Please can it be updated.

RyanGlScott:
@jfdm, oops, that totally slipped my mind. I've opened https://github.com/idris-lang/Idris-dev/pull/3648 to atone. :)

PR #3656, opened 2017-02-11
Run Idris in a UTF-8-compatible codepage on Windows

### LINKED ISSUE #3000: Elaborator shell: Using :term after intro' causes error, aborts shell
When I start the elaboration shell to prove a metavariable, the `:term` command initially works fine:

```
Idris> :l min_test_case.idr
Type checking .\min_test_case.idr
Holes: Main.whats_f
*min_test_case> :elab whats_f


----------                 Goal:                  ----------
{hole0} : Nat -> Nat
-Main.whats_f> :term
TT: ? {hole0} . {hole0}
```

(The contents of `min_test_case.idr` are simply:

```
f : Nat -> Nat
f n = ?whats_f
```

by the way.)

However, if I use the `intro'` tactic and then use `:term` again, I get kicked out of the elaborator shell and back into the main Idris shell:

```
-Main.whats_f> intro'

----------              Assumptions:              ----------
 n : Nat
----------                 Goal:                  ----------
{hole0} : Nat
-Main.whats_f> :term
TT: <stdout>: commitBuffer: invalid argument (invalid character)
Holes: Main.whats_f
*min_test_case>
```

I'll be honest: I've yet to get to grips with the elaborator shell properly, so I don't really know what the expected behaviour of `:term` would be in this case, but it seems to me that it shouldn't crash the shell.

(I'm using Idris 0.10 (precompiled binary) on Windows 10, if that helps.)

RyanGlScott:
Currently, Idris on Windows is rather prone to crashing when displaying fancy UTF-8 characters, as #3000 is evidence of. I feel like the simplest solution to just make Idris run in a codepage which supports UTF-8, which this PR accomplishes.

This uses the lightweight `code-page` library which, on Windows:

1. Changes the codepage to 65001, which is UTF-8 aware
2. Changes the encoding that GHC uses internally to `utf8` using `hSetEncoding`

And on other platforms, does nothing. This is how [`criterion`](https://github.com/bos/criterion/pull/125) and [`doctest`](https://github.com/sol/doctest/pull/149) currently handle Unicode on Windows as well.

Fixes #3000.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,10 @@
 + The REPL now colourises output on MinTTY consoles (e.g., Cygwin and MSYS)
   on Windows, which previously did not occur due to a bug.
 
++ Idris now runs in a UTF-8-compatible codepage on Windows. This fixes many
+  Unicode-rendering issues on Windows (e.g., error messages stating
+  `commitBuffer: invalid argument (invalid character)`).
+
 ## Library Updates
 
 + Terminating programs has been improved with more appropriate
--- a/idris.cabal
+++ b/idris.cabal
@@ -263,6 +263,7 @@ Library
                 , blaze-markup >= 0.5.2.1 && < 0.9
                 , bytestring < 0.11
                 , cheapskate < 0.2
+                , code-page >= 0.1 && < 0.2
                 , containers >= 0.5 && < 0.6
                 , deepseq < 1.5
                 , directory >= 1.2.2.0 && < 1.2.3.0 || > 1.2.3.0
--- a/src/Idris/ErrReverse.hs
+++ b/src/Idris/ErrReverse.hs
@@ -10,8 +10,8 @@ Maintainer  : The Idris Community.
 module Idris.ErrReverse(errReverse) where
 
 import Idris.AbsSyntax
+import Idris.Core.Evaluate (unfold)
 import Idris.Core.TT
-import Idris.Core.Evaluate(unfold)
 import Util.Pretty
 
 import Data.List
@@ -26,7 +26,7 @@ errReverse ist t = rewrite 5 (do_unfold t) -- (elideLambdas t)
     do_unfold :: Term -> Term
     do_unfold t = let ns = idris_errReduce ist in
                       if null ns then t
-                         else unfold (tt_ctxt ist) [] 
+                         else unfold (tt_ctxt ist) []
                                      (map (\x -> (x, 1000)) (idris_errReduce ist))
                                      t
 
--- a/src/Idris/Main.hs
+++ b/src/Idris/Main.hs
@@ -44,16 +44,20 @@ import System.Directory
 import System.Exit
 import System.FilePath
 import System.IO
+import System.IO.CodePage (withCP65001)
 import Text.Trifecta.Result (ErrInfo(..), Result(..))
 
 -- | How to run Idris programs.
 runMain :: Idris () -> IO ()
-runMain prog = do res <- runExceptT $ execStateT prog idrisInit
-                  case res of
-                       Left err -> do
-                         putStrLn $ "Uncaught error: " ++ show err
-                         exitFailure
-                       Right _ -> return ()
+runMain prog = withCP65001 $ do
+               -- Run in codepage 65001 on Windows so that UTF-8 characters can
+               -- be displayed properly. See #3000.
+  res <- runExceptT $ execStateT prog idrisInit
+  case res of
+       Left err -> do
+         putStrLn $ "Uncaught error: " ++ show err
+         exitFailure
+       Right _ -> return ()
 
 -- | The main function of Idris that when given a set of Options will
 -- launch Idris into the desired interaction mode either: REPL;


melted:
I tried to do that before, but it broke Eastern European users, so it was reverted in https://github.com/idris-lang/Idris-dev/pull/1946

RyanGlScott:
@melted, good point. I've looked at #1945 and #1946, and it's apparent that the reason that Idris was printing mojibake was because the previous attempt at fixing this issue was half-baked. It used `hSetEncoding stdout utf8`, which is good, but that _only_ changes GHC's internal notion of character encoding. That is, it only affects whether GHC decides to print out a character (or error with `invalid character`).

The problem was that Idris wasn't changed the codepage to go along with it! Having GHC be willing to print any UTF-8 characters is all well and good, but unless the console's codepage supports it, it'll get rendered badly. This is why you must also use the Win32 API to change the console's codepage to align with what you pass to `hSetEncoding` as well for this trick to work.

As evidence of this, I created a simple file with the error message in #1946:

```haskell
module Main (main) where

main :: IO ()
main = putStrLn "Не удается найти указанный файл"
```

And ran it in GHCi on PowerShell using different settings:

ghci

Here are some highlights:

* My default codepage is 437 (as shown later), so when this GHC process starts up, it sets the encoding of `stdout` to something non-UTF-8. Unsurprisingly, trying to print that Cyrillic fails immediately.
* After that, I call `hSetEncoding stdout utf8`. Now running `main` again, I _do_ get output, but because my codepage (437) is insufficient to display all UTF-8 characters, it results in Mojibake.
* After this, I change the codepage to 866 (Cyrillic). If `stdout`'s encoding matched this, `main` would work fine. But it doesn't, so I get mojibake again. (Precisely the error in #1946. Even the mojibake is deterministic!)
* Finally, I change my codepage to 65001 to match `stdout`'s `utf8` encoding. Now things work perfectly.

So as long as Idris also changes the codepage to 65001 as well (which this PR accomplishes), there shouldn't be issues like this. Thoughts?

ahmadsalim:
@RyanGlScott Your code seems to fail the style test. Could you kindly run `stylize.sh` to make it stylize everything correctly and then push your results?

RyanGlScott:
@ahmadsalim: It looks like `stylize.sh` erroring wasn't due to this change, but rather it's from bc553c70eb8357359a07d3954e07de2baf368483 (the `master` branch has the same failure).

Regardless, I've pushed a fix.

ahmadsalim:
@RyanGlScott Oh, I did not notice. Thanks, nonetheless!

david-christiansen:
Is there anything in the way of merging this now? It looks like this will solve the issue and also not deprive Windows users of pretty Unicode output.

melted:
Thanks for the PR!

be5invis:
Changing into cp65001 will cause CJK typesetting incorrect. Unless you switch it back and clear the entire screen.
See → https://github.com/Microsoft/BashOnWindows/issues/662

RyanGlScott:
To be clear, is this only on BashOnWindows?

be5invis:
@RyanGlScott No, everywhere. It belongs to CONHOST.
The correct way to handle encoding issues is to use `WSTR`-version of WINAPIs. It works as expected (that supports Unicode) on **every code page**.

be5invis:
... Unless WDG decided to fix this (it is hard, really, CONHOST is a kernel component and many important software depends on it)

be5invis:
Also, changing CP into non-DBCS will diable IMEs. See https://github.com/Microsoft/BashOnWindows/issues/661

RyanGlScott:
Hm, I don't seem to be able to reproduce this on `cmd.exe`, PowerShell, or MSYS2, at least. With this file:

```haskell
module Main (main) where

main :: IO ()
main = putStrLn "安あ"
```

I can print in in GHC in CP65001 without any apparent rendering issues:

cjk

So I'm still not convinced that this is anything but a WSL-specific bug.

be5invis:
@RyanGlScott If you try to run idris REPL, then you cannot type them under cp65001. (Though I think in English OS you cannot switch into DBCS, or have any IME either...)

RyanGlScott:
Ah, the issue is _input_-related. Sadly, Windows won't even _let_ me switch to CP932, so I can't verify that that fixes it (I'll take your word for it). I can at least observe the overlapping character issue you mentioned if I paste "安あ" repeatedly into a Haskeline prompt (which is what Idris is using).

I suppose the right thing to do here is to guard the use of CP65001 behind a `Cabal` flag. I'd still like to use CP65001 if possible, as it fixes more bugs than it introduces, in my experience. And fixing this The Right Way would involve rewriting GHC's IO subsystem to use entirely native Win32 API calls, which won't be happening for a while. So in the interim, switching to CP65001 is a decent (albeit not perfect) solution to Unicode rendering BS.

What do you think, @be5invis? Others? Should the use of CP65001 be opt-in? Or opt-out?

be5invis:
@RyanGlScott I'd like to try it out in RS2/3 internal builds, the console guys told me they refactored that thing and allowed IMEs under CP65001.

PR #3658, opened 2017-02-13
Add takeWhile and dropWhile to Prelude.Stream

mrkgnao:
This is something of an exploratory, very minor PR to see how the whole process works here. I'd love comments, especially around possible code smells relating to the use of `assert_smaller`.

### DIFF
--- a/libs/prelude/Prelude/Stream.idr
+++ b/libs/prelude/Prelude/Stream.idr
@@ -8,6 +8,7 @@ import Prelude.Applicative
 import Prelude.Monad
 import Prelude.Nat
 import Prelude.List
+import Prelude.Bool
 
 %access public export
 %default total
@@ -45,6 +46,22 @@ take (S n) (x :: xs) = x :: (take n xs)
 drop : (n : Nat) -> Stream a -> Stream a
 drop Z     xs = xs
 drop (S k) (x::xs) = drop k xs
+
+||| Take elements from the stream as long as the predicate matches
+||| @ pred the predicate
+||| @ xs the stream
+takeWhile : (pred : a -> Bool) -> (xs : Stream a) -> List a
+takeWhile pred s@(x :: xs) = if (pred x)
+                             then x :: (takeWhile pred (assert_smaller s xs))
+                             else []
+
+||| Drop elements from the stream until the predicate no longer matches
+||| @ pred the predicate
+||| @ xs the stream
+dropWhile : (pred : a -> Bool) -> (xs : Stream a) -> Stream a
+dropWhile pred s@(x :: xs) = if (pred x)
+                             then dropWhile pred (assert_smaller s xs)
+                             else xs
 
 ||| An infinite stream of repetitions of the same thing
 repeat : a -> Stream a


mrkgnao:
There might also be something funny going on with laziness. If I define these functions in a `.idr` file and load it into the Emacs REPL, it seems to like leaving `Stream` thunks unevaluated:

```idris

λΠ> takeWhile' (<10) [1..]
1 ::
takeWhile' (\ARG =>
              with block in Prelude.Interfaces.Prelude.Interfaces.Integer implementation of 
-- [... more output ...]
```
although defining something like 

```

main = print $ takeWhile' (<100) $ dropWhile' (<20) [1..]
```
and calling that from the REPL works properly. And, of course, if I import them from my modified Prelude, the REPL evaluates them fine:

```idris

λΠ> takeWhile (<10) [1..]
[1, 2, 3, 4, 5, 6, 7, 8, 9] : List Integer
```

and I'm not sure what's going on.

jfdm:
@mrkgnao  I'll try to have a look at this later but in the meantime, some FAQs that might help you understand what *might* be going on:

*  [Evaluation at the REPL doesn’t behave as I expect. What’s going on?](http://idris.readthedocs.io/en/latest/faq/faq.html#evaluation-at-the-repl-doesn-t-behave-as-i-expect-what-s-going-on)
* [Why does Idris use eager evaluation rather than lazy?](http://idris.readthedocs.io/en/latest/faq/faq.html#why-does-idris-use-eager-evaluation-rather-than-lazy)

Melvar:
I don’t think that that use of `assert_smaller` is proper, as `Stream` is codata. These functions can never be total.

xekoukou:
```
dropwhile
``` 
is not productive. How many elements of stream does it require till it returns something?

mrkgnao:
@Melvar @xekoukou I think I see what you're saying. Something like `takeWhile (/= 0) (repeat 0)` will never return.

Should I annotate these as being partial instead?

xekoukou:
```takewhile``` does give something after it checks the first element.
It either gives ```x :: _``` or ```[]```.

For every element in the resulting stream, it requires only one element from the input, so it requires to do a finite amount of work to produce the result. (Idris uses guarded recursion to check for [productivity](http://bentnib.org/productive.pdf))

For ```takewhile```, you just need to remove ```assert_smaller```.

For the ```dropwhile``` function, it could loop indefinitely as in your example.

If you want to have ```dropwhile``` be total, you need to provide a proof that the stream contains an element that returns true on the ```pred``` function, after a finite amount of elements.

You can ask in the mailing list or in the irc how you can do that.

Melvar:
No, `takeWhile` cannot be total either, because the predicate could be true for all elements of the `Stream`, but a `List` is not codata; it can only hold finitely many elements.

xekoukou:
yes , he needs to replace it with ```Stream```.

Melvar:
No, because the predicate might *not* be true for all elements of the `Stream`, and a `Stream` cannot hold only finitely many elements.

edwinb:
Since these aren't total for the reasons discussed, I would prefer not to have them in the prelude but rather, for the moment, add a stream support library into contrib. My feeling though, given they aren't total, is that they probably aren't useful for streams like they are for lists...

ahmadsalim:

Thanks.

PR #3662, opened 2017-02-18
Add docs on data syntax

geigerzaehler:
Add reference documentation on introducing data types.

* Note that type constructors can use dependent types
* Document 'record' syntax and the generation of getter and setter functions
* Document 'codata' syntax and explain its effect


Melvar:
`Delayed Infinite` is usually known as `Inf`.

edwinb:
Thanks for this. I'll make the change to 'Inf'.

PR #3669, opened 2017-02-25
Simplify generating the simplified language

geigerzaehler:
This commit merges the `sVar`, `mkapp`, and `mkfapp` helper functions that are used in `simplify` and reduces code duplication

We introduce a new `bindExprs` function that let-binds expressions that are not variables and constructs an inner expression in the scope of the bindings with the variables.


edwinb:
Thanks for this - it looks like the travis failure is purely due to the stylish check failing. You can fix this by running ./stylize.sh before commiting. I'll merge here and do that myself.

PR #3678, opened 2017-03-07
Updated stack

jfdm:


jfdm:
Right, I am going to merge this PR. I canceled the stalled macOS build: All other builds passed.. As the macOS build is allowed to fail, I see no reason why we shouldn't merge this.

PR #3679, opened 2017-03-07
Make Idris work with GHC 7.6.3 and Tagsoup 0.14.1

jfdm:
Tagsoup 0.14.1 does not compile cleanly with ghc 7.6.3. So let's amend
the cabal file to ensure an appropriate version of tagsoup is selected.

*Note* tagsoup is a dependency of a dependency of a dependency.

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -310,6 +310,11 @@ Library
   else
      build-depends: safe >= 0.3.9
 
+  if impl(ghc < 7.8.4)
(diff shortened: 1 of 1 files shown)


jfdm:
Right, I am going to merge this PR. Travis is reporting a failure because I canceled the stalled macOS build: All other builds passed.. As the macOS build is allowed to fail, I see no reason why we shouldn't merge this.

PR #3680, opened 2017-03-07
Break line on some long multiline comments.

ostera:
Hi! I was going through the Prelude and found some comments that were already multiline yet had very long lines in them.

Here's a small commit to fix some of them, making the code slightly easier to read.

Thanks for all the good work! 👍


jfdm:
Hi @ostera thanks for the PR. At the moment we will expect the travis builds to fail due to an upstream bug with older version's of GHC. The bug should be fixed and once it is merged we will ask for you to rebase against HEAD so that we can double check your contribution.

Thanks Jan

PR #3684, opened 2017-03-09
Disable mintty colors for 32-bit

melted:
When I built the 32-bit release I got absurd linking errors. After a long while I figured out that the mintty colors were the culprit, no doubt via a completely ridiculous mechanism, but I couldn't figure that out. Just disable it for 32-bit Windows.

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -332,8 +332,9 @@ Library
   if os(darwin)
      build-depends: unix < 2.8
   if os(windows)
-     build-depends: mintty >= 0.1 && < 0.2
-                  , Win32 < 2.4
+     build-depends: Win32 < 2.4  
+     if arch(x86_64) 
+        build-depends: mintty >= 0.1 && < 0.2
   if flag(FFI)
      build-depends: libffi < 0.2
      cpp-options:   -DIDRIS_FFI
--- a/src/Util/System.hs
+++ b/src/Util/System.hs
@@ -38,11 +38,10 @@ import System.FilePath (dropFileName, isAbsolute, searchPathSeparator)
 import Tools_idris
 #endif
 
-#ifdef mingw32_HOST_OS
+#if defined(mingw32_HOST_OS) && defined(x86_64_HOST_ARCH)
 import Graphics.Win32.Misc (getStdHandle, sTD_OUTPUT_HANDLE)
 import System.Console.MinTTY (isMinTTYHandle)
 #endif
-
 catchIO :: IO a -> (IOError -> IO a) -> IO a
 catchIO = CE.catch
 
@@ -82,7 +81,7 @@ isATTY = do
 -- Unfortunately, we must check this separately since 'isATTY' always returns
 -- 'False' on MinTTY consoles.
 isMinTTY :: IO Bool
-#ifdef mingw32_HOST_OS
+#if defined(mingw32_HOST_OS) && defined(x86_64_HOST_ARCH)
 isMinTTY = do
   h <- getStdHandle sTD_OUTPUT_HANDLE
   isMinTTYHandle h


david-christiansen:
Seems like a good idea. But shouldn't there be a comment describing why these conditionals are there, so as to not mystify readers of the code?

jfdm:
I agree, there should be comments in the cabal file, and possibly a note in the CHANGELOG.

RyanGlScott:
@melted, I'm a bit concerned that this leads to linking errors, as this code is more-or-less taken straight from the `Win32` package that is bundled with GHC. Do you mind sharing what errors you have so I can fix them properly?

RyanGlScott:
Including what version of GHC you're building with would also be helpful, as the the internals of Windows GHC have changed considerably in recent releases.

melted:
I haven't got the output handy. But the symbol was __imp_atan. It failed that way with both ghc 7.10 and 8.0. It didn't fail like that on 7.8 but I had to revert the cp65001 commit to build there. i even tried swapping out the gcc used by ghc to the latest. That made no difference.

RyanGlScott:
@melted, I fixed the problem you were having with the "cp65001 commit" in [`code-page-0.1.3`](http://hackage.haskell.org/package/code-page-0.1.3).

However, I'm unable to reproduce the `mintty` errors you're getting with 32-bit GHC. I've tried with GHC 7.10.2 and 8.0.1, but I don't see any errors related to `__imp_atan`. I do get a linker error due to the Idris icon, funny enough:

```
Linking dist\build\idris\idris.exe ...
dist/idris_icon.o: file not recognized: File format not recognized
collect2.exe: error: ld returned 1 exit status
`gcc.exe' failed in phase `Linker'. (Exit code: 1)
```

But this happens with or without this patch, so it's unrelated to the apparent `mintty` issue.

RyanGlScott:
Sorry, that `File format not recognized` error is my fault. Idris's `Setup.hs` script was picking up a different `windres` on my `PATH`, causing the emitted `idris_icon.o` file to be 64-bit, not 32-bit, resulting in the error.

After cleaning up my `PATH` and rebuilding, I was able to reproduce the linker error you were talking about:

```
Preprocessing executable 'idris' for idris-0.99.1...
[1 of 1] Compiling Main             ( main\Main.hs, dist\build\idris\idris-tmp\Main.o )
Linking dist\build\idris\idris.exe ...
C:/Users/RyanGlScott/Software/ghc-8.0.1-x86/mingw/bin/../lib/gcc/i686-w64-mingw32/5.2.0/../../../../i686-w64-mingw32/lib/../lib/libmsvcrt.a(dohds01059.o):(.idata$5+0x0): multiple definition of `_imp__atan'
C:/Users/RyanGlScott/Software/ghc-8.0.1-x86/mingw/bin/../lib/gcc/i686-w64-mingw32/5.2.0/../../../../i686-w64-mingw32/lib/../lib/libntdll.a(dgjjs01967.o):(.idata$5+0x0): first defined here
collect2.exe: error: ld returned 1 exit status
`gcc.exe' failed in phase `Linker'. (Exit code: 1)
```

I'll take a closer look at this.

RyanGlScott:
I'm not quite sure what's going on here, but I can at least minimize the bug down to:

```haskell
module Main (main) where

main :: IO ()
main = print (atan 0.5 :: Double)
```
```
$ ../../../Software/ghc-8.0.1-x86/bin/ghc -fforce-recomp Bug.hs -package mintty
[1 of 1] Compiling Main             ( Bug.hs, Bug.o )
Linking Bug.exe ...
C:/Users/RyanGlScott/Software/ghc-8.0.1-x86/mingw/bin/../lib/gcc/i686-w64-mingw32/5.2.0/../../../../i686-w64-mingw32/lib/../lib/libmsvcrt.a(dohds01059.o):(.idata$5+0x0): multiple definition of `_imp__atan'
C:/Users/RyanGlScott/Software/ghc-8.0.1-x86/mingw/bin/../lib/gcc/i686-w64-mingw32/5.2.0/../../../../i686-w64-mingw32/lib/../lib/libntdll.a(dgjjs01967.o):(.idata$5+0x0): first defined here
collect2.exe: error: ld returned 1 exit status
`gcc.exe' failed in phase `Linker'. (Exit code: 1)
```

My hunch is that the fact that `mintty` links against `ntdll` has something to do with it. But strangely, if I try `-package Win32` instead of `-package mintty`, then the linker error goes away, even though `Win32` also links against `ntdll`!

@Mistuke, might you have an idea of what's going on here? It's hard to tell if this is a GHC bug, or me being stupid (or a combination of the two).

RyanGlScott:
Turns out what I said about `Win32` isn't quite true. I forgot that GHC 8.0 ships with a version of `Win32` that _doesn't_ link against `ntdll`. If you install a more recent `Win32` that does link against `ntdll` (say, in a sandbox), then you can trigger the same linker error:

```
$ cabal get Win32
$ cd Win32-2.5.2.0/
$ cabal sandbox init
$ cabal install . -w $(pwd)/../../../../Software/ghc-8.0.1-x86/bin/ghc
$ ../../../../Software/ghc-8.0.1-x86/bin/ghc Bug.hs -package-db .cabal-sandbox/i386-windows-ghc-8.0.1-packages.conf.d/ -package Win32 -fforce-recomp
[1 of 1] Compiling Main             ( Bug.hs, Bug.o )
Linking Bug.exe ...
C:/Users/RyanGlScott/Software/ghc-8.0.1-x86/mingw/bin/../lib/gcc/i686-w64-mingw32/5.2.0/../../../../i686-w64-mingw32/lib/../lib/libmsvcrt.a(dohds01059.o):(.idata$5+0x0): multiple definition of `_imp__atan'
C:/Users/RyanGlScott/Software/ghc-8.0.1-x86/mingw/bin/../lib/gcc/i686-w64-mingw32/5.2.0/../../../../i686-w64-mingw32/lib/../lib/libntdll.a(dgjjs01967.o):(.idata$5+0x0): first defined here
collect2.exe: error: ld returned 1 exit status
`gcc.exe' failed in phase `Linker'. (Exit code: 1)
```

I'm now more confident that this is a proper GHC bug, so I've opened [GHC Trac #13431](https://ghc.haskell.org/trac/ghc/ticket/13431) to track this.

RyanGlScott:
I've uploaded [`mintty-0.1.1`](http://hackage.haskell.org/package/mintty-0.1.1) to Hackage, which does not suffer from this bug. I also confirmed that Idris builds successfully on 32-bit Windows GHC 8.0.1 using `mintty-0.1.1`, so this PR should no longer be necessary.

melted:
Great! Can you update the bound for mintty in idris.cabal?

RyanGlScott:
Shouldn't need to - I've marked `mintty-0.1` as deprecated, so it won't be chosen for install-plans in the future.

PR #3696, opened 2017-03-14
Also look for nodejs in addition to node in the testsuite

RyanGlScott:
Many Linux distributions (e.g., Ubuntu) install `nodejs` instead of `node`, so it makes sense to search for it as well.

### DIFF
--- a/test/TestRun.hs
+++ b/test/TestRun.hs
@@ -108,7 +108,9 @@ runTest path flags = do
 
 main :: IO ()
 main = do
-  node <- findExecutable "node"
+  nodePath   <- findExecutable "node"
+  nodejsPath <- findExecutable "nodejs"
+  let node = nodePath <|> nodejsPath
   case node of
     Nothing -> do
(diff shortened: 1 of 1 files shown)


RyanGlScott:
I'm not sure why the test suite requires that `node` always be installed, even when you're not running the JavaScript tests, but that's a separate issue...

jfdm:
@RyanGlScott thanks for this. I was the reason the test suite looks for node to be installed. This was to prevent an accident that I almost made. I updated the test suite's expected results once, and with nodes/node not installed the expected results became wrong.

jfdm:
Also i was planning on fixing this as I've moved to linux again, and encountered the nodes/node problem.

jfdm:
Also, and whilst i remember, I sought and failed to find a way to have this test run only when the js option is invoked. If you can find a way to ensure that then that would be good too.

ahmadsalim:
Thanks for the PR, I will merge it.

I agree with @jfdm's comments that it might be nice to have the check run only when js is required.

PR #3698, opened 2017-03-16
Add regression test for #3651.

### LINKED ISSUE #3651: Incorrect type mismatch error?
I tried copying some code from § 2.1.5 of [The Practical Guide to Levitation](http://itu.dk/people/asal/pubs/msc-thesis-report.pdf), and noticed something funny. Here is a minimal example:

```idris
module Bug

data Tag : String -> List String -> Type where
  TZ : Tag l (l :: e)
  TS : Tag l e -> Tag l (l' :: e)

SPi :  (e : List String)
    -> ((l : String) -> Tag l e -> Type)
    -> Type
SPi []       _    = ()
SPi (l :: e) prop = (prop l TZ, SPi e $ \l' => \t => prop l' $ TS t)

switch :  (e : List String)
       -> (prop : (l : String) -> (t : Tag l e) -> Type)
       -> SPi e prop
       -> (l' : String) -> (t' : Tag l' e) -> prop l' t'
switch (l' :: e) prop ((propz, props)) l' TZ      = propz
switch (l  :: e) prop ((propz, props)) l' (TS t') =
  switch e (\l => \t => prop l (TS t)) props l' t'
```

Surprisingly, this fails to typecheck in 0.99:

```
Bug.idr:17:8:Type mismatch between
        Tag l'1 (l'1 :: e)
and
        Tag l'1 (l' :: e)
```

What's really odd is that the choice of the name `l'` is of great importance. If I change the first case of `switch` to use a name other than `l'`:

```idris
switch (blargh :: e) prop ((propz, props)) blargh TZ = propz
```

Then it typechecks! What's more, if I leave the definition of `switch` alone but change the second case of `SPi` to use a different name than `l'`:

```idris
SPi (l :: e) prop = (prop l TZ, SPi e $ \flurmp => \t => prop flurmp $ TS t)
```

Then it also typechecks.

RyanGlScott:
[Someone inadvertently fixed #3651](https://github.com/idris-lang/Idris-dev/issues/3651#issuecomment-286205584)! It's a miracle! Let's add a regression test to make sure it stays that way.

Resolves #3651.


ahmadsalim:
Thanks!

PR #3706, opened 2017-03-21
Refuse to work if listed packages are not installed.

jfdm:
This caught me out enough times for a fix to be added.


ahmadsalim:
@jfdm You are usually the one who asks for this, but could you kindly add a test and update the `CHANGELOG` :).

PR #3708, opened 2017-03-23
Remove incorrect CPP and require newer Cabal instead, attempt 2.

Melvar:
An updated attempt at #3519.

The `MIN_VERSION_*` macros are only generated during setup, i.e. after Setup.hs
is compiled. The GHC version is not a proxy for the Cabal version, as it is
trivial to install a newer version of Cabal and cabal-install with
cabal-install, and in that case the Setup.hs would not typecheck (this is what happened for me). Since it
doesn’t appear to be possible to switch on the Cabal version, this change
simply declares a requirement for a newer version.

The second commit (fixing warnings from cabal after raising the version) is possibly opinionated. Feel extra free to leave it off or change it.

Specifically, if the `cabal-version` constraint is above 1.10, each component
should have a `default-language` field. It appears the default was/is
`Haskell98`, so I entered that for each component.

Similarly, the `extensions` field is then deprecated in favor of
`default-extensions` and `other-extensions`. For these I have investigated which
extensions are actually required globally (i.e. some files assume them but do
not declare them) and have entered these as `default-extensions` and removed
them from any files that did declare them. I then collected all additional
extensions declared in any files into `other-extensions`.

It may be a good idea to think about these settings and change them, and also to
prune declared but unused extensions from the source files, but in this commit I
have only tried to replicate the previous settings (minus removing default extensions declared in the files; perhaps I should have left that to a later extension-pruning step as well).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,10 +71,6 @@ before_install:
         export ZCAT=gzcat;
         export PKG_CONFIG_PATH=/usr/local/opt/libffi/lib/pkgconfig:$PKG_CONFIG_PATH;
     fi
-  # work around https://github.com/haskell/cabal/issues/1666 : cabal-install
-  # will not install the required version of the Cabal library if a project
-  # requires it only for Setup.hs (i.e. it’s not a normal dependency).
-  - cabal install Cabal --constraint Cabal==${CABALVER}.*
   - if [ -n "$STYLISH" ]; then ./stylize.sh; exit $?; fi
 
 install:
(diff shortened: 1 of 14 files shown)


Melvar:
Same issue: It complains that a suitable version of the Cabal library is not installed, when it itself is using that suitable version.

Melvar:
The problem is that having `cabal-install` of the correct version does not imply, in this particular environment, that `Cabal` of the same or higher version is installed. Furthermore, `cabal-install` does not take the combination of `Build-Type: Custom` and `Cabal-Version` as a build-time dependency on `Cabal`: [cabal issue #1666](https://github.com/haskell/cabal/issues/1666).

I don’t think there is a way to specify a build-time-only dependency, so I will attempt the workaround given in that issue report.

ahmadsalim:
For some reason it seems to change compiler and `Cabal` library version for tests. Anyone has any idea why?

ahmadsalim:
@idris-lang/contributors ^

ahmadsalim:
I think this is outdated since we dropped support for all < 7.10

PR #3710, opened 2017-03-24
Make ConsoleIO more complete.

jfdm:


jfdm:
I should have merged this earlier, too many conflicts such that it is quicker to make a 'new' PR

PR #3722, opened 2017-03-29
added hint to get around build failures due to old ibc files

CarstenKoenig:
had this right now when updating to 0.99.2 - some `pruviloy` lib did suddenly fail to type check because it did not recognize a `Unit`

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -92,6 +92,17 @@ the required libraries and header files are available:
 
 * `stack build --nix`
 
+### before rebuilding new pulls
+the default build will currently just reuse `.ibc` files which can result
+in build-failures in the `Building libraries...` phase.
+
+A safer way to do this is therefore to recursively delete all the `*.ibc`
+files from the `libs/` folder.
+
+On Linux(or similar) you can do this with
+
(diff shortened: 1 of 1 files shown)


Melvar:
The only time this is necessary is if a library source file has been removed, otherwise `make relib` suffices.

CarstenKoenig:
maybe I should clarify: I'm generally using *stack* - and this happened to me before - the first time was really hard to find out why, so I just wanted to give others in my situation a hint - no problem if you don't find it worth including.

Shall I take it back?

Melvar:
Ah, then sorry, I don’t know.

jfdm:
Ah yes, this can happen when using stack. Thanks for the PR, I keep on forgetting to add such tips to the build information myself. An alternate to specifying `--exec rm -rf {} \;` is to use `-delete`, but that shouldn't stop this PR.

Really, at some point I should write a Stack.make file to offer better stack offering for building idris locally.

jfdm:
The build failure is due to a time out. I've restarted the build, so once the build has completed we can merge.

Thanks @CarstenKoenig

PR #3737, opened 2017-04-03
Handle idris_mkFileError through a special-case.

### LINKED ISSUE #3716: Missing idris_mkFileError special-case in Idris.Core.Execute
I don't understand the cause of this, but loading a literate Idris file (I used [this one](https://github.com/ctford/flying-spaghetti-monster/blob/0e016d5da45979383ea43f15a3f4c3369c6a06f9/example/Example.lidr)) gives the following error at the REPL:

    Idris> Symbol "idris_mkFileError" not found
    idris: user error (Could not call foreign function "idris_mkFileError" with args [prim__vm (prim__TheWorld)])

`idris-vim` reports the evaluation as successful, but no `.ibc` file is generated and subsequent attempts to evaluate the file hang.

~~~I suspect that `idris-vim` is to blame - [it doesn't properly recognise `.lidr` files and has to be forced by setting the filetype](https://github.com/idris-hackers/idris-vim/issues/36). But Idris should probably not crash hard.~~~

~~~Loading the same file via `:l Example` does not have a problem.~~~

I tested this behaviour in 0.12.3 and 0.99.2.

**Update: nothing to do with `.lidr`. See below.**

ctford:
Use a dirty IORef to simulate C's errno behaviour. If an exception
occurs during fileOpen, then store a FileError expression, which is read
and returned in a later call to idris_mkFileError.

As a result of this change, file errors generated during the execution
of a type provider, for example file not found, will be correctly
reported. Previously the interpreter crashed.

Fixes #3716.

### DIFF
--- a/src/Idris/Core/Execute.hs
+++ b/src/Idris/Core/Execute.hs
@@ -36,6 +36,8 @@ import Data.Time.Clock.POSIX (getPOSIXTime)
 import Data.Traversable (forM)
 import Debug.Trace
 import System.IO
+import System.IO.Error
+import System.IO.Unsafe
 
 #ifdef IDRIS_FFI
 import Foreign.C.String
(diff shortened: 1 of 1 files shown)


melted:
Thanks for the PR!

Issue #2817, opened 2015-11-14
Forward pipe operator |> for in Prelude?


Kazark:
I've started learning Idris; I have the MEAP book and have been playing around with it quite a lot. I'm coming more from an F# background than a Haskell background, so like most folks who `F# |> <3`, I like the forward pipe operator. I did notice that Idris already has `$`, like Haskell, and that is seems to be an inherent part of the syntax rather than a function. However, when I tried to use `&`, which is the Haskell equivalent of F#'s `|>`, that did not work.

I tried defining a `|>` operator, and was pleasantly surprised by how easy it was to do so:

```
infixl 5 |> -- I have no idea what the proper precedence of this would be

(|>) : a -> (a -> b) -> b
a |> f = f a
```

and aside from the fact that I wonder about the precedence, it seems that it works, after superficial testing:

```
> "foo" |> length
3 : Nat
> ["asdf", "qwer", "uiop"] |> map length
[4, 4, 4] : List Nat
```

So I would like to submit something to include this in Prelude. However, before doing so, I thought it would be good to ask about this.
1. Is there something about this that won't work well with the type inference in Idris? I know sometimes it can have weird effects on type inference in F#.
2. Is this something you would want in Prelude, or would you rather people just define it themselves if they want it?
3. Would this be considered not idiomatic Idris? If the forward piping concept itself is okay, would it be better to mimic Haskell with `&` to match `$` instead of mimicking F#'s `|>` since we don't have `<|` in Idris?

I'm not emotionally attached to this idea, but just thought I would bring it up because I can't find anything about it by searching (searching you run into the same problems that Hoogle was designed to solve for Haskell...) and see what your thoughts on it were.

Melvar:
1. This will only work for nondependent functions. (One of the reasons that `$` is builtin.)
2. Adding it would at least require discussion. Prelude is supposed to be only things you absolutely need for a reasonable program these days, and it’s debatable whether writing value before function is necessary.
3. I have the strong personal opinion that `(&)` is the wrong choice and it needs to be `(#)`, but apparently nobody else thinks so.

david-christiansen:
I happen to quite like pipeline operators, but it is unfortunate that they won't work for fully dependent functions. I'd be happy to accept a pull request that defined a `Control.Pipelines` module in `contrib`, or one that defined `&`.  WRT `&` vs `#`, I don't see a reason to prefer one arbitrary character over the other, but `&` is already in Haskell, so we might as well be compatible.

[...]

PR #3741, opened 2017-04-05
Add Control.Pipeline (|>, <|)

lpil:
Closes https://github.com/idris-lang/Idris-dev/issues/2817

Unfinished- I was unsure where to put the tests for this. Could you point me in the right direction? :)

Thanks,
Louis

### DIFF
--- a/libs/contrib/Control/Pipeline.idr
+++ b/libs/contrib/Control/Pipeline.idr
@@ -10,11 +10,17 @@ infixr 0 <|
 ||| functions into a series of transformations, reading top
 ||| to bottom.
 |||
-|||     [[1], [2], [3]] |> join |> map (* 2)
+||| ```idris example
+||| [[1], [2], [3]] |> join |> map (* 2)
+||| ```
 (|>) : a -> (a -> b) -> b
 a |> f = f a
 
 
 ||| Backwards pipeline style function application, similar to $.
+|||
+||| ```idris example
+||| unpack <| "hello" ++ "world"
+||| ```
 (<|) : (a -> b) -> a -> b
 f <| a = f a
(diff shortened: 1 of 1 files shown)


ahmadsalim:
I do not think it's a large enough feature, that you need to make tests for it. It will type-check code snippets in docs anyway, so just add some example for the `<|` as well :).

libs/contrib/Control/Pipeline.idr
```diff
@@ -0,0 +1,20 @@
...
+infixr 0 <|
+
+%access public export
+
+
+||| Pipeline style function application, useful for chaining
+||| functions into a series of transformations, reading top
+||| to bottom.
+|||
+|||     [[1], [2], [3]] |> join |> map (* 2)
```
ahmadsalim: you should surround this with idris example blocks as follows:

    ```idris example
     [[1], [2], [3]] |> join |> map (* 2)
    ```

lpil: Thanks :)

lpil:
No tests? 😱  This really is unfamiliar territory for me :P

I'll fix up the docs.

libs/contrib/Control/Pipeline.idr
```diff
@@ -0,0 +1,20 @@
+module Control.Pipline
```
ozgurakgun: There is a typo in the module name.

lpil: Thank you.

lpil:
Typo fixed and examples added.

ahmadsalim:
@lpil I mean, we do have a type system after all :). I think adding tests is generally useful, but perhaps not necessary for such simple features.

david-christiansen:
> It will type-check code snippets in docs anyway

Note that a failure to type check in docs does not make the module fail to type check. This is so docs can exhibit examples of errors that types rule out.

ahmadsalim:
Thanks for the PR.

lpil:
Thanks!

Issue #3735, opened 2017-04-03
Broken idris_time implementation in Idris.Core.Execute


ctford:
The `ExecVal` is malformed, meaning that the following happens:

```
Idris> :l System
*System> :x do t <- time; printLn t
MkIO (\w =>
        prim_io_bind (case case 1491254436 of
                             MkRaw t => PE_pure_c4df80b5 t of
                        MkIO fkb => fkb w)
                     (\b => prim_io_pure ())) : IO' (MkFFI C_Types String String) ()
```
This should happen.
```
Idris> :l System
*System> :x do t <- time; printLn t
1491256088
MkIO (\w => prim_io_pure ()) : IO' (MkFFI C_Types String String) ()
```

Pull request to follow.

PR #3747, opened 2017-04-08
Interpreter tests

ctford:
I'm trying to write some simple regression tests for #3716 and #3735. Is this the right way to go about it?

Feedback very welcome.

As an aside, I'm getting some odd behaviour for the #3735 test, which is why I haven't checked it in yet.

```
Idris> :l System
*System> :x do t <- time; printLn t
1491661366
MkIO (\w => prim_io_pure ()) : IO' (MkFFI C_Types String String) ()
*System> :x do t <- time; printLn (t * 1)
MkIO (\w => prim_io_pure ()) : IO' (MkFFI C_Types String String) ()
```

Somehow I/O fails if I do anything to the `time` value before printing it.


test/interpret002/run
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+chmod a-r unreadable.txt
```
ahmadsalim: I think this only might work with Linux. I am unsure how this could be made to work with Windows which is currently failing. @melted Any ideas? :)

ahmadsalim:
@ctford Consider using Git Attributes: https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes

ctford:
Thanks for the pointer @ahmadsalim. Did you mean that a git attribute would be used to set the file's permissions? Doesn't that have the same portability problems as setting it in the test script?

ahmadsalim:
@ctford I misunderstood what you meant, I thought you meant it was not possible to configure the attributes of a specific file, but now I see it is not possible to use read-only.

ctford:
I've removed the readonly test. If someone comes up with a good way to do it I'm happy to add it back in.

test/interpret002/expected
```diff
@@ -0,0 +1,6 @@
+File Not Found
+MkIO (\w => prim_io_pure ()) : IO' (MkFFI C_Types String String) ()
+"readable"
+MkIO (\w => prim_io_pure ()) : IO' (MkFFI C_Types String String) ()
+Operation not permitted
```
ahmadsalim: This seems to be "Permission Denied" on Windows.

ctford:
I've removed the other problematic assertion on reading files. What's left now is:
* A happy path assertion showing that files can be read.
* A sad path assertion showing what happens when a file isn't found.

IMHO this is most of the value of the test - this test would catch it if #3716 regressed. The missing assertions demonstrate that the right errors are returned in cases where files are present but not accessible. I think that's ok, as this test proves that if there is an error it's handled in some way.

ahmadsalim:
Great, thanks!

PR #3748, opened 2017-04-08
Prelude: implement Functor for `Pair a`

tsani:
I wanted to use `map` on a pair yesterday and got a type error. It turns out that there's no Functor implementation for Pair in the prelude. Is this intentional? If so, feel free to close this PR.

In any case, this patch adds the implementation in `Functor.idr`, which kind of the only place to define it, since `Pair` is defined in `Builtins.idr` rather than its own module, and `Functor.idr` imports `Builtins.idr`.

### DIFF
--- a/libs/prelude/Prelude/Functor.idr
+++ b/libs/prelude/Prelude/Functor.idr
@@ -1,5 +1,6 @@
 module Prelude.Functor
 
+import Builtins
 import Prelude.Basics
 
 %access public export
@@ -22,3 +23,5 @@ infixl 4 <$>
 (<$>) : Functor f => (func : a -> b) -> f a -> f b
 func <$> x = map func x
 
+Functor (Pair a) where
+  map f (x, y) = (x, f y)
--- a/test/docs003/expected
+++ b/test/docs003/expected
@@ -10,6 +10,7 @@ Methods:
         
         The function is Total
 Implementations:
+    Functor (Pair a)
     Functor List
     Functor (IO' ffi)
     Functor Stream


david-christiansen:
I'm in favor of a merge here, assuming that Travis passes. Given that appveyor passed, I imagine it will.
I don't think anyone ever made a conscious decision to not have it for Pair a - just that it hasn't come up.

libs/prelude/Prelude/Functor.idr
```diff
@@ -22,3 +23,5 @@ infixl 4 <$>
 (<$>) : Functor f => (func : a -> b) -> f a -> f b
 func <$> x = map func x
 
+Functor (Pair a) where
+  map f (x, y) = (x, f y)
```
melted: this looks fishy to me, shouldn't it be `(f x, f y)` on the right hand side?

justjoheinz: You assume that a pair is always of type (a, a) and that you can apply the function to both elements of the pair. However it is frequently (c, a) with map f (c, a) = (c, f a). That last element is picked in a tuple is AFAIK a convention, could as well be the first.

Synthetica9: On that note, shouldn't there be a homogenus tuple type, or are we assuming that `Vect 2 a` is good enough for that purpose?

melted: Hmmm, I saw only one type variable up there. But it's true, Pair takes two.

tsani: @melted Since `Pair` takes two type parameters, we have for a fixed `a` that `Pair a : Type -> Type` as required, so we can implement `Functor` for it.

@justjoheinz It's not so much that the person implementing `Functor` is making the arbitrary choice that it's the second element of the pair that gets mapped over, but rather the person who defines the constructor of `Pair a b`.

The following fails to typecheck:

```idris
implementation [first] Functor (\x => Pair x a) where
  map f (x, y) = (f x, y)
```

which would be the only way I could implement a functor instance causing the first element to get mapped over. So in other words, unless it were possible to write such an instance, someone wanting to implement Functor for a partially-applied Pair has no choice but to make the function `f` apply to the *second* element.

If `x : Pair a b` were such that `fst x : b` (which would be quite unnatural) then the natural implementation of Functor would have the natural result that it's the first element that gets transformed.

ahmadsalim:
Thanks for the PR!

PR #3749, opened 2017-04-08
Add usleep from System to Effect.System

artemohanjanyan:


libs/effects/Effect/System.idr
```diff
@@ -14,18 +15,23 @@ data System : Effect where
      Time : sig System Integer
      GetEnv : String -> sig System (Maybe String)
      CSystem : String -> sig System Int
+     Usleep : (i : Int) ->
+              {auto prf : So (i >= fromInteger 0 && Delay (i <= fromInteger 1000000))} ->
```
ahmadsalim: Any reason you use `fromInteger` here explicitly? literals should be overloaded. The same with `Delay`.

ahmadsalim: Also I would in the constructor probably not use `auto`, only in the external signature used by the user.

libs/effects/Effect/System.idr
```diff
@@ -43,3 +49,8 @@ getEnv s = call $ GetEnv s
 
 system : String -> Eff Int [SYSTEM]
 system s = call $ CSystem s
+
+usleep : (i : Int) ->
+         {auto prf : So (i >= fromInteger 0 && Delay (i <= fromInteger 1000000))} ->
```
ahmadsalim: Same comment regarding the use of `fromInteger` and `Delay`

libs/effects/Effect/System.idr
artemohanjanyan: I just copied the type from `System.usleep`. `fromInteger` is probably not necessary, it's shown when you type `:t usleep` in REPL, but it's not present in [the source file](https://github.com/idris-lang/Idris-dev/blob/master/libs/base/System.idr). However, `auto` is used in that source file. Should I still remove it?

libs/effects/Effect/System.idr
ahmadsalim: Is there any reason you do not use `System.usleep` here instead of copying it?

ahmadsalim: Sorry, I missed that this calls the effect, whose handler uses the system one.

libs/effects/Effect/System.idr
ahmadsalim: @artemohanjanyan Yeah, so I am not sure it is necessary for a constructor to have it. I would just take an argument `(inbounds : So (i >= 0 && i <= 1000000))`.

ahmadsalim:
Thanks!

PR #3753, opened 2017-04-11
Fix references to refl in misc reference

justjoheinz:
There is also a reference to 

```
syntax Trivial = (| oh, refl |)
```

but I do not know what this does. If someone gives me a correct line I can squash this into this commit.

### DIFF
--- a/docs/reference/misc.rst
+++ b/docs/reference/misc.rst
@@ -51,7 +51,7 @@ some constructors of proofs.
 
 ::
 
-    syntax Trivial = (| oh, refl |)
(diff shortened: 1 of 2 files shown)


ahmadsalim:
Thanks for the PR, `(| oh, refl |)` is basically trying either using `oh` (of `So`, now called `Oh`) or `refl` (now `Refl`) and taking the first  that succeeds type checking. It is one of the more hacky features of Idris.

justjoheinz:
I fixed these and some occurrences of "oh" and "so" as well.

PR #3771, opened 2017-04-19
Added the operations specified in the TODO in contrib/Data/SortedSet

Synthetica9:
These were union, difference, and intersection. Also added a Monoid based on the union (`<+> =
union`, `neutral = empty`)

### DIFF
--- a/libs/contrib/Data/SortedSet.idr
+++ b/libs/contrib/Data/SortedSet.idr
@@ -2,8 +2,6 @@ module Data.SortedSet
 
 import Data.SortedMap
 
--- TODO: add intersection, union, difference
-
 export
 data SortedSet k = SetWrapper (Data.SortedMap.SortedMap k ())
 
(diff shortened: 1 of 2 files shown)


Melvar:
What you’ve labeled as `intersection` appears to be not intersection but symmetric difference.

libs/contrib/Data/SortedSet.idr
```diff
@@ -31,5 +29,26 @@ export
 toList : SortedSet k -> List k
 toList (SetWrapper m) = map (\(i, _) => i) (Data.SortedMap.toList m)
 
-implementation Foldable SortedSet where
```
ahmadsalim: I am not sure why the `implementation` keyword was removed here.

libs/contrib/Data/SortedSet.idr
```diff
@@ -31,5 +29,26 @@ export
...
+
+||| Set difference. Delete all elments in y from x
+export
+difference : (x, y : SortedSet k) -> SortedSet k
+difference x y = foldr delete x y
+
+||| Set intersection. Uses the union of the differences.
+export
+intersection : (x, y : SortedSet k) -> SortedSet k
+intersection x y = union (difference x y) (difference y x)
```
ahmadsalim: These implementations seem fairly slow. Have you considered implementing the efficient versions?

libs/contrib/Data/SortedSet.idr
Synthetica9: I was under the impression that this was the new proper style, am I mistaken in this?

ahmadsalim: I think both are accepted.

clayrat:
Probably a bit faster implementation for `intersection` would be `difference x (difference x y)`?

Synthetica9:
@clayrat You're probably right, I'll edit it

ahmadsalim:
Thanks for the PR!

PR #3777, opened 2017-04-22
remove links to unmaintained backends

KimStebel:


jfdm:
Thanks for the contribution. 

I'm thinking rather than *just* removing the links, we should instead point to the list described:

    http://idris.readthedocs.io/en/latest/reference/codegen.html

This might be better, and hopefully give people more of an idea of what is out there.

ahmadsalim:
Thanks for the PR!

PR #3795, opened 2017-05-01
Docs: ``ElabReflection`` needs explicit enabling

kaeluka:


docs/reference/elaborator-reflection.rst
```diff
@@ -32,6 +32,9 @@ Running Elaborator Scripts
 
 On their own, tactics have no effect.
 The meta-operation ``%runElab script`` runs ``script`` in the current elaboration context.
+Before you can use ``%runElab``, you will have to enable the language extension by adding
+``%language ElabReflection`` in your file (or by passing ``-X ElabReflection`` to the
```
jfdm: If I am not mistaken, should it not be that language extensions need to be enabled both with a pragma and command line option.

kaeluka: I think not; with my version of idris (1.0), it's sufficient to just include this line, OR to pass in the flag. I used this program (with an example from @david-christiansen's paper) to test this right now:

```
module Main

import Language.Reflection.Elab

%language ElabReflection

mkId : Elab ()
mkId =
  do x <- gensym "x"
     attack
     intro x
     fill (Var x) ; solve
     solve

idNat : Nat -> Nat
idNat = %runElab mkId

idUnit : () -> ()
idUnit = %runElab mkId

main : IO ()
main = putStrLn (show $ idNat 12)
```

ahmadsalim:
Thanks for the PR. As far as I my experience goes, I only needed to do one of the things; in any case, if there are corrections we can add them later.

PR #3797, opened 2017-05-02
README: How to run with pruviloj in emacs

kaeluka:
I stumbled over this missing piece of information and thought that others need not have the same problem :-)

### DIFF
--- a/libs/pruviloj/README.txt
+++ b/libs/pruviloj/README.txt
@@ -18,7 +18,16 @@ those eliminators to do induction.
 
 
 Q: How do I use Pruviloj?
-A: Call Idris with the "-p pruviloj" option and add:
+A: When running from the command line, call Idris with the
+   "-p pruviloj" option. When running in emacs, add the following section
+   at the end of the file and make sure that there is no empty line after
+   the section. Then restart emacs.
+
+    -- Local Variables:
+    -- idris-load-packages: ("pruviloj")
+    -- End:
+
+   In your code, add:
       import Pruviloj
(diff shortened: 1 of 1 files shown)


kaeluka:
Meta comment: this is the second trivial PR I'm submitting within 24 hours; I'm planning to submit more documentation fixes as I find problems. I personally think that many trivial PRs are better than few big ones, but let me know if you'd prefer some other way.

david-christiansen:
Lots of small docs PRs are great.

But I'm not so convinced that Emacs mode information belongs here, rather than in the Emacs mode's docs. After all, Idris is a multi-editor language :-)

david-christiansen:
Also, in the Emacs mode, you can get it to make those lines for you with M-x idris-add-idris-load-packages RET , and it'll give you autocomplete of the installed packages.

kaeluka:
I see why you won't make it emacs specific.
*But* I stumbled over this, some hint would've helped me. I'm happy to amend the text I put there (or you can).
Suggestion: instead add a new question in the FAQ instead: "Q: I get error this-and-that in emacs // A: use idris-add-idris-load-packages".

If you don't like that, and don't have a better suggestion: close the PR, I won't mind (really).

melted:
I think it makes sense to have a pointer about this in both emacs mode and pruviloj docs.

ahmadsalim:
@kaeluka I think adding a question to the FAQ about packages seems like a good idea!

I will close the PR since there has not been any activity, and I agree with @david-christiansen that this seems to belong more to the Emacs documentation than here.

Issue #3727, opened 2017-04-02
Implementation arguments containing injective functions


nicolabotta:
The program
```
> LTB : Nat -> Type
> LTB b = DPair Nat (\ n  => LT n b)

> implementation Uninhabited (LTB Z) where
>   uninhabited (MkDPair n prf) = absurd prf
```
type checks fine in 0.99.1-git:PRE. But with 1.0 one gets
```
- + Errors (1)
 `-- ./BoundedNat/issues/implemantation.lidr line 4 col 17:
     (n : Nat **
      LTE (S n) 0)  cannot be a parameter of Prelude.Uninhabited.Uninhabited
     (Implementation arguments must be type or data constructors)
```
which seems like an Idris error given that `LTB Z` is indeed a type.

ahmadsalim:
@nicolabotta I believe this was previously a bug, which is now fixed. I believe the general idea is that one should not be able to write implementations that work on functions instead of type constructors.

nicolabotta:
I am not sure that I understand what you mean, here `LTB Z` is a legal type, not a function. I am also not aware of any report in which the pre-1.0 behavior was identified as being erroneous. 

This late change in Idris breaks tons of legacy code and effectively makes types that are not introduced via `data` declarations second-class types. From a user's perspective, this is very annoying and rather pointless. 

What have we gained through this new limitation?

ahmadsalim:
As far as I understand, the mechanism was only guaranteed to work with type and data constructors, because they guarantee injectivity. I think there were some issues with the old approach, which was fixed prior to the release of 1.0. For more details, I think you need to ask Edwin, since I do not completely understand the implementation search mechanism of Idris myself.

nicolabotta:
Thanks Ahmad, for the time being I will go back to 0.99 and try to see if I can make the framework 1.0 compatible.

edwinb:
Yes, this is correct. I hadn't realised there was code out there which exploited this bug but I'll see if I can think of a way to deal with this kind of implementation.

The problem with this kind of implementation constraint is that unification is not fancy enough to deal with it. Maybe the thing to do is revisit unification.

ahmadsalim:
@nicolabotta I think you can use instead:

```idris
> implementation Uninhabited DPair Nat (\ n  => LT n Z) where
>   uninhabited (MkDPair n prf) = absurd prf
```

It's a bit more ugly, but should work.

nicolabotta:
[...]

PR #3798, opened 2017-05-02
Update interfaces to remove use of constants

tmcgilchrist:
Looks like somewhere along the way Idris stopped supporting constants as the "instance head", not sure whether that includes variables as well?

This is based off a compile error I had from [states](https://github.com/edwinb/States)

```idris
- + Errors (1)
 `-- ./State/Var.idr line 20 col 0:
     Var  cannot be a parameter of States.Execute
     (Implementation arguments must be type or data constructors)
```

### DIFF
--- a/docs/tutorial/interfaces.rst
+++ b/docs/tutorial/interfaces.rst
@@ -48,8 +48,8 @@ For example, the
 Only one implementation of an interface can be given for a type — implementations may
 not overlap. Implementation declarations can themselves have constraints.
 To help with resolution, the arguments of an implementation must be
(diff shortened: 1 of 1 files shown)


ahmadsalim:
Thanks for the PR.
I think this is related to #3727, but I am unsure whether that bug should actually be fixed or this should be expected behaviour and so your documentation update should apply.

ahmadsalim:
Thanks for the PR.

I think that since no one took a look at this, I will merge it since it reflects current behaviour.
If #3727 ever gets fixed, we will redo the documentation

PR #3799, opened 2017-05-03
Add Control.Pipeline to contrib ipkg

BleuShan:
I was trying to import `Control.Pipeline` for a toy thing from a homebrew built source distribution and the compiler was telling me it wasn't finding it. After checking the build output, there was no `Pipeline.ibc` file and `Control.Pipeline` wasn't in `contrib.ipkg`. So... 1 + 1.


melted:
Thanks!

PR #3804, opened 2017-05-07
Do not run idle GCs

melted:
They cause an idle repl to do a major collection three times every
second to no use, with corresponding CPU use.


melted:
Add `+RTS -S` to the idris command line to see the idle GCs. I'm kind of baffled by this feature, if no work is being done, how could garbage be generated? What is certain is that Idris generates no garbage during idle time, and it uses 10% CPU to pointlessly copy the working set hither and tither.

PR #3808, opened 2017-05-09
Insert missing `guard` in tutorial.

FranklinChen:
Insert missing `guard` in tutorial.


Melvar:
Uh. That’s not right. The previous paragraph already describes when `guard` is used, and it certainly is not applied to all the qualifiers, only to the ones that are guards.

FranklinChen:
Ah, I simply misread, sorry.

PR #3812, opened 2017-05-10
Update CoList.idr

Synthetica9:
Changed to match [Haskell's implementation](https://hackage.haskell.org/package/base-4.9.1.0/docs/Data-List.html#v:unfoldr). This version is strictly more powerful, and it is clearer from the signature what is happening (imho).


melted:
Thanks for the PR!

PR #3813, opened 2017-05-13
Add PrimFn missing cases for portable codegen

mmhelloworld:
This PR adds missing cases in PrimFn for portable codegen. [Idris JVM bytecode backend](https://github.com/mmhelloworld/idris-jvm) now uses portable codegen and this PR would make `mod` to work.


melted:
Thanks!

PR #3817, opened 2017-05-13
free old heap after gc cycle

bmaxa:
Small fix to GC, that is, freeing old heap after GC cycle.


bmaxa:
Disregard this patch, bug was actually caused by different issue. That is array was bigger then 255 and since old heap was kept I wasn't getting segfaults.

melted:
ok! Thanks for digging into the rts!

PR #3818, opened 2017-05-14
[Issue 3774] Have :exec print the value of IO _ expressions

jdevuyst:


ahmadsalim:
@jdevuyst Thanks! Please add a test if possible.

test/README.md
```diff
@@ -124,12 +124,6 @@ and somewhat self-explanatory.
     stack test --test-arguments="--pattern foo042 --accept"
     ```
 8. Check the content of `expected`. Maybe the test didn't do what you thought it would. Fix and go back to 7 until it's ok.
-9. Add under `Extra-source-files` in `idris.cabal` the patterns that match the folder's content. If you forget this, your test will fail in Travis CI. With the previous example, it should be at least:
```
david-christiansen: Why is this section deleted?

jdevuyst: `Extra-source-files` appears to be empty right now. It has a comment that suggests that it's auto-generated:
```
-- extra-source-files is generated by Setup.hs using `git --ls-files`.
Extra-source-files:
```
Moreover, the last update to `test/README.md` also seemed to remove mention of `Extra-source-files`: https://github.com/idris-lang/Idris-dev/commit/3285d369fda50cb94e1a894e642649ba441a6615#diff-c30ff793110b0960f76bb3854af6f5a8

Am I missing something?

jdevuyst:
I added a test, but it seems to make the Travis CI build fail. Sorry for that. I didn't realize that I was only running the C tests locally.

So there are three errors. One is related to documentation, and it seems unlikely that this error would be caused by the addition of the test case (my diff passed CI before I added the test case). 🤔

The other two tests seem specific to how `:exec` is implemented for the JS backend. I get the same errors when I try to test `:exec` on previous Idris builds. Would it be acceptable to make this test only run for the C backend for the time being?

If there's something else I should be looking into instead, do let me know.

test/README.md
melted: @jdevuyst no, that's right.

david-christiansen: This is just me being out of date :-) Thanks!

melted:
Thanks for the PR!

PR #3826, opened 2017-05-22
Support presentations of abstract names in output

david-christiansen:
When Idris sends strings over the IDE protocol, it includes
out-of-band semantic information that allows regions of the string to
_represent_ their underlying data. In particular, terms are serialized
and sent with their representations, allowing things like interactive
error messages in IDE clients.

This patch adds a similar capability for names. Now, serialized
datatypes are sent with names, allowing names that don't have syntax
to be the subjects of commands. This is very convenient when working
with metaprogramming or debugging the compiler.

The change is backwards-compatible, because it has to be enabled in
each IDE client, and if they aren't sent serialized names, then they
don't act differently from today.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
++ Names are now annotated with their representations over the IDE
+  protocol, which allows IDEs to provide commands that work on special
+  names that don't have syntax, such as case block names.
+
 # New in 1.0
 
 + It's about time
--- a/src/Idris/IdeMode.hs
+++ b/src/Idris/IdeMode.hs
@@ -128,7 +128,8 @@ namespaceOf _         = Nothing
 
 instance SExpable OutputAnnotation where
   toSExp (AnnName n ty d t) = toSExp $ [(SymbolAtom "name", StringAtom (show n)),
-                                        (SymbolAtom "implicit", BoolAtom False)] ++
+                                        (SymbolAtom "implicit", BoolAtom False),
+                                        (SymbolAtom "key", StringAtom (encodeName n))] ++
                                        maybeProps [("decor", ty)] ++
                                        maybeProps [("doc-overview", d), ("type", t)] ++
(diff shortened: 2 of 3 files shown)


david-christiansen:
Example interaction in Emacs, with this file:
```
module G

foo : Nat
foo = 3

interface Fnord a where
  goggles : a -> (a, a)

||| Okay
implementation Fnord Nat where
  goggles = const (4, 3)

whu : Fnord Nat
whu = %implementation
```
After loading, the user  right-clicks `whu` and picks "Get definition". Idris responds with:
```
whu : Fnord Nat
whu = Nat implementation of G.Fnord
```
The user then right-clicks the RHS of `whu`. In the past, the menu had interesting commands for the whole term, but none for the name "Nat implementation of G.Fnord", because that is not valid syntax, and the IDE mode relied on the names being parseable. But now, the user can select "Get definition" again, and Idris answers:
```
Nat implementation of G.Fnord : Fnord Nat
Nat implementation of G.Fnord = constructor of G.Fnord (\meth =>
                                                          G.Nat implementation of G.Fnord, method goggles meth)
```
All these special names work similarly. I expect this to be particularly useful when examining the core language and debugging the compiler or difficult `Elab` scripts.

david-christiansen:
@melted , is this something that we can get working in the Atom mode? It's super convenient in Emacs.

melted:
I'll take a look at it.

ahmadsalim:
Side remark:
I think it would be more readable if it read as 
```idris
whu : Fnord Nat
whu = implementation G.Fnord Nat
```

david-christiansen:
@ahmadsalim I think that changing the way SNs are printed is orthogonal to making them clickable in editors. Do you mind if I merge?

ahmadsalim:
@david-christiansen I agree, which is what I meant by side note :). 
I do not mind if you merge.

PR #3829, opened 2017-05-23
New JS codegen — Optimization

be5invis:
Objectives

- Eliminate silly overheads caused by Int-Inductive conversion, etc.
- Make the result FAST.
- Make the result JavaScript-y.


be5invis:
@melted I saw "Use of const in strict mode".
What's the version of Node.js used in testing?

melted:
@be5invis the bad news is that the preinstalled node on Travis is v0.10.36 (Apparently it's even still running Ubuntu 12.04 which went out of support a while ago). The good news is that we shouldn't target that and should try to sneak in a newer version.

melted:
What the hell? Travis has Ubuntu **14.04** in **BETA**? That's pretty absurd. Apparently it's possible to get a better node on that though. Maybe we should dare to upgrade to this brand-new, unproven technology.

melted:
Right, it isn't node that's the problem anymore.

be5invis:
@melted I added NVM script to install nodejs 6

melted:
Yeah, it looks like a lot of undefined values now.

be5invis:
@melted Rob's original code does not handle nullary variable quotes well.
Fixing that.

be5invis:
@melted All green!

melted:
Great, I'll look at it after work. It would be nice if Rui could take a look at it too.

melted:
Didn't get through it all yesterday.

rbarreiro:
I was afraid that changing all switch statements to else if would slow down things, but a cooked example I made actually ran faster with this branch on node.

be5invis:
@rbarreiro Changing `switch`es into `if`s are for data type specialization, i.e., mapping Idris `True` and `False` into JavaScript `true` and `false` to eliminate some really silly overheads. It can also eliminate overheads of "constructing" null-ary data types. (Currently I translate them into numbers -- it is okay to change them into objects either.)

melted:
I see that the node codegen doesn't actually use jsbn-node.js but includes jsbn-browser.js instead. TBH, I'm perfectly about that, as I don't mind it bundled.

be5invis:
I'd like to add an option to control how we handle big integers:

- embed
- use require to import
- use short integer (faster but less correct)

发自我的 iPhone

在 2017年5月31日，05:45，Niklas Larsson <notifications@github.com<mailto:notifications@github.com>> 写道：


I see that the node codegen doesn't actually use jsbn-node.js but includes jsbn-browser.js instead. TBH, I'm perfectly about that, as I don't mind it bundled.

melted:
Ok, let's leave that for later.

jfdm:
Shouldn't the CHANGELOG be updated as well, to note the changes....

jfdm:
It appears that this was taken care of in a more recent PR. Regardless, and more of a not to all, can we in future try to ensure that the CHANGELOG is updated in the same PR, and not through secondary ones.

PR #3832, opened 2017-05-24
Fixed visibility for implementations for SortedSet

Synthetica9:
`Semigroup`, `Monoid`, and `Foldable` were all missing `export`


libs/contrib/Data/SortedSet.idr
```diff
@@ -52,8 +53,10 @@ export
 intersection : (x, y : SortedSet k) -> SortedSet k
```
clayrat: The comment for `intersection` seems to be wrong.

melted:
Thanks for the PR!

PR #3843, opened 2017-06-01
Added docstrings for several public functions.

jackolantern:
I was confused by the `bitAt` function, and so I thought I would add a docsstring for it.  While editing I noticed a few more functions with slightly ambiguous names, and so I added docs for them as well.


clayrat:
While you're at it, could you also fix the comment for `nextBytes`? :) It should say 

```
||| Gets the lowest n for which "8 * 2 ^ n" is larger than or equal to the input.
||| For example, `nextBytes 10 = 1`.
||| Like with nextPow2, the result is not rounded up, so `nextBytes 16 = 1`.
```

(as opposed to `nextBytes 1* = 16`, since it returns the power)

jackolantern:
@clayrat  Done.

melted:
Thanks for the PR!

Issue #3840, opened 2017-05-30
 Rewrite in Do block


be5invis:
It is common to introduce multiple `let`s and `rewrite`s when proofing a theorem. So adding `rewrite` in `do` block would be helpful to reduce nested indentations...

Translation rule:
```
do { rewrite H; ...xs } => rewrite H in do {...xs}
```

ahmadsalim:
@be5invis Thanks for the PR.
I think it is a good idea!
You are welcome to add it if you are interested.

be5invis:
@ahmadsalim 
Just for fun : We can also add `intro` into `do` blocks...
Translation rule:

```idris
do { intro patterns; ...xs } => \pattern_1 => \pattern_2 => ... => do {...xs}
```

be5invis:
I'd like to write another PR after the JS one is finished (currently call specialization is not included yet).

ahmadsalim:
@be5invis I think we should avoid this becoming a tactic system (we already have elaborator reflection which does the same thing). 
I believe `rewrite` is suitable because it is surface syntax which is somehow less accessible in `do` requiring weird syntactic nesting.

be5invis:
@ahmadsalim I said that are just for fun 😄.
Just adding `rewrite`, `intro`ducing abstraction and `apply`ing functions are convenient enough for now, we do not need special syntax for them.

ahmadsalim:
@be5invis I understand that, and I mean it is not my decision if such additional language constructs like `intro` and `apply` will be accepted, because this is a big change in language design.
I understand the excitement of contributing language features 😄, but there is also an issue with adding too many: people may rely on them and then it is hard to change in the future or make good maintenance.
Personally, I think writing `\a,b,c => do ...` or `e $ do ...` is not as bad, because often they do not need to be nested as much, and when they do it is good to be more declarative about it with explicit abstraction and application.
If you are still  interested in adding these, please use the mailing list to start a discussion, potentially including @edwinb.

be5invis:
PR => #3847

PR #3847, opened 2017-06-04
Add do-rewrite construction

be5invis:
A fix for #3840.

Syntax:

```idris
do { rewrite H; ... }
```

Definition

```idris
do { rewrite H; ... } =====> rewrite H in do {...}
```

### DIFF
--- a/src/Idris/IBC.hs
+++ b/src/Idris/IBC.hs
@@ -2292,7 +2292,7 @@ instance (Binary t) => Binary (PDo' t) where
                                       put x1
                                       put x2
                                       put x3
-                DoRewrite x1 x2 -> do putWord8 3
+                DoRewrite x1 x2 -> do putWord8 5
                                       put x1
                                       put x2
         get
--- a/src/Idris/IBC.hs
+++ b/src/Idris/IBC.hs
@@ -2292,6 +2292,9 @@ instance (Binary t) => Binary (PDo' t) where
                                       put x1
                                       put x2
                                       put x3
+                DoRewrite x1 x2 -> do putWord8 3
+                                      put x1
+                                      put x2
         get
           = do i <- getWord8
                case i of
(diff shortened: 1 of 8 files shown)


Melvar:
I think we should remove `rewrite` instead of letting it worm its way into more parts of the language.

be5invis:
@Melvar Without `rewrite` some proofs using `replace` would become extremely hard to write.

be5invis:
cc. @ahmadsalim

melted:
This has to be looked at by @edwinb before it can go in.

ahmadsalim:
@be5invis Thanks for the PR!
Could you please add some tests to show that it works, and that it does not break existing syntax with `rewrite`-`in`? Could you also please update the `CHANGELOG` with proposed updates.

@Melvar I think in general that `rewrite` is a fairly useful language feature. Did @edwinb hint at wanting to remove it?

@melted I think this change is fairly uncontroversial, since all it is doing is adding an inline `rewrite` in `do` similarly to how we have inline `let` in `do`.
The only potential thing to worry about is backwards compatibility.

src/Idris/IBC.hs
```diff
@@ -2292,6 +2292,9 @@ instance (Binary t) => Binary (PDo' t) where
                                       put x1
                                       put x2
                                       put x3
+                DoRewrite x1 x2 -> do putWord8 5
```
ziman: Do we need to bump `ibcVersion` on line 50 of this file?

melted: Good catch, that is definitely needed.

david-christiansen:
I think this is a very nice feature, but given how central it is, we should let @edwinb weigh in.

edwinb:
Looks fine to me. It's consistent with the way 'let' syntax works, too. Sorry for not paying attention to these...

@Melvar what do you see as the problem with `rewrite`? I find it invaluable for programs which do fiddly things (or even fairly mundane things) with indices, e.g. https://github.com/edwinb/Blodwen/blob/master/src/Core/TT.idr#L438 or https://github.com/edwinb/Blodwen/blob/master/src/Core/Evaluate.idr#L137

Melvar:
I’m guessing I wrote that comment on a bad day; really all it is is that I have a pet peeve about `rewrite … in Refl`, always having to mentally convert it into `cong (…)` to figure out what it is doing.

melted:
@be5invis could you run stylize on it? Then we can merge.

PR #3849, opened 2017-06-07
Added section on default implicit arguments

Garnasha:
Quick addition to the tutorial. It's not quite complete yet, but at least it shows the syntax and behaviour of the ``default`` annotation, which I couldn't find anywhere before. I'd have liked to add an example of intended use, but lack the knowledge to do so right now.


melted:
Thanks!

PR #3852, opened 2017-06-09
Fix quotes in Effects introduction

Pastafarianist:
Fixed the quotes so that the code could be copy-pasted.


jfdm:
Thanks for the PR, I'll merge this prior to checks passing as only the documentation has been changed.

PR #3854, opened 2017-06-11
changed js backend to look for includes also in getIdrisLibDir

rbarreiro:
Changed js backend to look for includes also in getIdrisLibDir. This is useful to use with objs ipkg option when making libraries. Without this I don't know any other way to include javascript parts on idris libraries.

### DIFF
--- a/docs/tutorial/miscellany.rst
+++ b/docs/tutorial/miscellany.rst
@@ -467,6 +467,15 @@ And for use in the browser:
       %include JavaScript "path/to/external.js"
 
 The given files will be added to the top of the generated code.
+For library packages you can also use the ipkg objs option to include the
+js file in the installation, and use
+
+.. code-block:: idris
+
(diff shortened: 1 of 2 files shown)


melted:
Rerruning CI

ahmadsalim:
@rbarreiro Why are you closing and reopening the PR? :)

rbarreiro:
To rerun travis.

rbarreiro:
It is timing out a lot. I think we should start thinking of a way to solve that problem, Travis jobs are close to the 50 min timeout and lately are timing out a lot.

ahmadsalim:
Maybe we should consider switching to something like Circle CI or privately hosted Jenkins instead.

ahmadsalim:
Alternatively, we should put effort in optimizing the Idris build process. I actually like that idea much more :)

Issue #3697, opened 2017-03-15
Including libraries for the Node codegen


justjoheinz:
Disclaimer: Maybe my thinking is wrong here

I wanted to experiment a bit with the node codegen. In order to write a wrapper library I naturally install modules using `npm install xyz`.
The resulting module will be located in e.g. `node_modules/xyz/main.js`

If I include this using:

    %lib Node "xyz/main"

the generated javascript file will include the line 

    var xyz/main = require("xyz/main")

The code generator should probably either just take the last section of the require path or provide an optional parameter like 

    $lib Node "xyz/main"  as "client"

to produce

    var client = require("xyz/main")

Another related question - how would I tell idris to look inside of node_modules for the required files? Or is this done by `node` itself when executing the javascript?

melted:
Sounds like a good idea. 

Node connects your code to the libs from whats in package.json. So that's not a problem.

justjoheinz:
I have been digging in the code a bit, unfortunately I never managed to get idris to compile locally and I have no big haskell experience, so I cannot test the code.

But modifying `CodegenJavaScript.hs` like this:

```haskell
      includeLibs :: [String] -> String
      includeLibs =
        let
          repl '\' = "_"
          repl '/' = "_"
          repl c   = c
        in
          concatMap (\lib -> "var " ++ (repl <$> lib) ++ " = require(\"" ++ lib ++"\");\n")
```
should a least output identifiers for node packages, which are sound. It does not cover the idea to have an alias, but would be a step forward.

rbarreiro:
Implemented as suggested by #3855

ahmadsalim:
I will close this since the PR was accepted.

PR #3855, opened 2017-06-11
added require for the %lib anotations in js backend

rbarreiro:
to solve issue #3697


melted:
Seems like travis is timing out, that probably just a travis problem.

melted:
Thanks for the PR!

jfdm:
Can the CHANGELOG be updated please? Also if there is other documentation that requires changing, can that be seen to as well.

rbarreiro:
This feature was already documented, by error it was left out on the recent improvements to us backend, I noticed it by reading an issue it had on the previous implementation. So I think this is just a bugfix.

jfdm:
No worries. Thanks

PR #3859, opened 2017-06-12
added a test for javascript ffi with functions

### LINKED ISSUE #3192: RFC - Javascript code generation needs a variant for wrapping event callbacks
I am trying to add an Idris function as a DOM event callback. Whilst I can successfully add the callback, the function does not get called. Inspection of the generated code shows that it falls into the following pattern:

``` javascript
var i$ffiWrap = function(fid,oldbase,myoldbase) {
    return function() {
   ...
  return i$ret;
 }
}
```

I was able to get it to work by editing the generated code so that the last 3 lines read:

``` javascript
 return i$ret;
 }
}()
```

So I think that this is analagous to the case of storing C callbacks in C structs, for which the %wrapper syntax is provided. And maybe we need an equivalent for JS.
It may be the case that it is already possible to generate such code, and I am just not aware of how to do it. My existing Idris code looks like this:

``` idris
add_event_listener : JSRef -> String -> (() -> JS_IO ()) -> JS_IO ()
add_event_listener target event action = do
  if target == null then
    jscall "function () {console.log(\"No target\");}()" (JS_IO ())
  else
    jscall "function () {%0.addEventListener(%1,%2);}()" (JSRef -> String -> (JsFn (() -> JS_IO ()
  )) -> JS_IO ()) target event (MkJsFn action)

```

rbarreiro:
Fixes #3192

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -132,7 +132,7 @@ for the compatible backends are:
 
 - `ANY`: choose this if your test will work with any code generator
 - `C_CG`: choose this to only test against the C code generator
-- `JS_CG`: choose this to only test against the Node code generator
+- `NODE_CG`: choose this to only test against the Node code generator
 - `NONE`: choose this if you don't want your test to be run with multiple code
 generators (mainly for tests that only perform type checking)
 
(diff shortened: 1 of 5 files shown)


rbarreiro:
Travis is taking longer than usual. I just added a simple test and the build is timing out. Does anyone know what is going on?

melted:
Travis is hopefully just temporarily flsky.

Issue #3796, opened 2017-05-01
0/0 is being compiled as -Infinity, which fails to compile (using the C backend).


PoroCYon:
In the following program, `0 / 0` is compiled as `-Infinity` instead of NaN. The C backend (as well as the JavaScript backend) then emit this as `-Infinity`, which fails to compile. In the REPL, however, `0 / 0` does evaluate to NaN, as it should.

PoC code:
```idris
module Main

main : IO  ()
main = printLn $ 0.0 / 0
```

The C backend emits: (`idris -S -o foo.c foo.idr`)
```c
LOC(4) = MKFLOAT(vm, -Infinity);
```

ahmadsalim:
Thanks for reporting the issue.
I am unsure why this happens, but in any case it should not output `-Infinity` to the codegen.

If you are interested in looking into it further, I think the relevant code should be around: https://github.com/idris-lang/Idris-dev/blob/119d964b75bf6a8b8a2f1f7c24450a4951cf26bc/src/Idris/Primitives.hs#L110

PoroCYon:
I'd like to, but I currently don't have the time to look into it, sadly.

edmundsmith:
I think I've got a working patch for this [here](https://github.com/edmundsmith/Idris-dev/commit/59bfe8bdac3f65cfefe970964c4983f5b9d34b26). It works as expected for the test case `main = putStrLn $ show $ 0.0/0.0` (plus others).

The existing Binary instances rely on the library-supplied `instance Binary Double`, but a sift through Data.Binary's source reveals that it uses `encodeFloat` and `decodeFloat` behind the scenes, marshalling doubles into an `(Integer, Int)`. This works for most values, but doesn't respect IEEE NaNs or Infinities. By using the explicit marshalling functions added in `binary-0.8.4`, the IEEE representation is preserved through `Binary` and `encode`/`decode`. 

If upgrading to `binary-0.8.4` is a deal-breaker, IEEE marshalling can take place by using `Word64` `put`/`get`, but this would require reimplementing a `Double -> Word64` conversion, which I assume would be quite unsafe, so I'm not sure how you'd want to organise that to keep the codebase clean.

I hope this is useful/a successful pull req can come from this - this would be my first contribution, which is why I'm being cautious of doing something wrong.

edmundsmith:
[...]

PR #3860, opened 2017-06-14
Fix for issue 3796 - 0/0 being compiled as -Infinity

edmundsmith:
This pull request is a fix for [issue 3796 - 0/0 being compiled as -Infinity](https://github.com/idris-lang/Idris-dev/issues/3796).

This changes the bytecode generation for doubles to preserve IEEE float special values, like NaN and Infinity. The existing generation uses `Data.Binary`'s included `instance Binary Double`, which uses `decodeFloat` and `encodeFloat` under the hood. This converts `Double`s to an `(Integer, Int)` and back, losing NaNs in the process.

Using the functions introduced in binary-0.8.4, full IEEE detail can be preserved throughout `Binary` encoding and decoding with `getDoublele` and `putDoublele`. This preserves NaNs and Infinities through the bytecode, fixing the issue.

By changing the encoding from `(Integer, Int)` to (effectively) `Word64`, existing ibc files will be invalidated by this patch. However, I think will be a necessary change to correctly represent and preserve IEEE double values.


ahmadsalim:
Thanks for the PR!
Please increment the number, so that Idris can figure out that the byte code version is wrong: https://github.com/idris-lang/Idris-dev/blob/master/src/Idris/IBC.hs#L50

edmundsmith:
It's been incremented

edmundsmith:
CI is sometimes failing because it fails to install distributive-0.5.2 or it times out, other times it passes all the tests fine. What's the solution to this?

melted:
The time out is just travis being flaky. 

I'm a bit worried about the ghc 7.6.3 not building though. Did one of our deps pull in something that's not compatible with that?

Anyway, I have restarted the failed builds

edmundsmith:
It seems `unix-2.6.0.1` and `unix-2.7.2.1` are both installed and conflicted in the Travis ghc 7.6.3 build. I don't know what's causing both to be installed, however. Could this be a result of changing to `binary-0.8.5` in the patch?

melted:
That's weird, binary doesn't depend on unix, and only depends on base, bytestring, arrays and containers, which shouldn't either. I can build this PR fine on 7.6.3, but that's on Windows.

melted:
Can you try to pin unix to less than 2.7 if building for ghc less than 7.8? we have several such exceptions in idris.cabal already.

ahmadsalim:
Maybe this is easier to make work now, since we dropped support for `< 7.10`?. @edmundsmith could you please check whether you can get it working now?

edmundsmith:
@ahmadsalim The only errors are now for exceeding the time limit. Is this still blocking, and if so, how do I get around this?

melted:
Travis has been slow for over a week. I'll merge this as soon as Edwin has made the point release to fix the cheapskate issue.

melted:
Thanks for the contribution!

PR #3861, opened 2017-06-14
Fix character escaping of codegen

iology:
The JS codegen didn't handle characters in range U+10000-10FFFF. The C codegen didn't correctly encode the code point to UTF-8 bytes.


melted:
Thanks for the PR!

PR #3863, opened 2017-06-16
SortedMap merge

Synthetica9:
- Merge `SortedMap`s
   - Either by a user defined function, the `Semigroup` op (`(<+>)`), or left-biased.
 - `Monoid` implementation based on merging maps by the `Semigroup` op.
 - Get keys or values of a map as a list (instead of just a list of keys and values)
 - Get the keys of a `SortedMap` as a `SortedSet`
 - Insert all items of a `Foldable f => f (k, v)` into a map

(Note: I wasn't quite sure what was meant with a "split" of a map, so I left that part of the ToDo up.)


melted:
That's a poitively frightening list of commits :-)

Please rebase on master.

PR #3864, opened 2017-06-16
Sized types for strong induction

ziman:
(This is not an implementation of [Agda's sized types](https://agda.readthedocs.io/en/v2.5.2/language/sized-types.html).)

This patch adds the proof of accessibility for values of any type that has a size mapping to `Nat`. For example, there is an implementation for lists, where `size = length`.

For any such type, we get strong induction on size, which allows recursive calls not only on direct subterms but on any value with smaller `size`.


david-christiansen:
This is very nice!

Is `sizeAcc` the right name for it? I worry that `acc` in general will confuse people into thinking "accumulator". What if we picked "access" instead? I suppose it's too late to rename old Prelude things, but new ones?

What do you think?

libs/base/Data/List/Views.idr
```diff
@@ -105,17 +101,17 @@ data Filtered : (a -> a -> Bool) -> List a -> Type where
...
+filteredROK : (p : a -> a -> Bool) -> (x : a) -> (xs : List a) -> Smaller (filter (\y => not (p y x)) xs) (x :: xs)
 filteredROK p x xs = LTESucc (filterSmaller xs)
 
 ||| Covering function for the `Filtered` view
 ||| Constructs the view in O(n lg n)
 export
 filtered : (p : a -> a -> Bool) -> (xs : List a) -> Filtered p xs
-filtered p inp with (smallerAcc inp)
+filtered p inp with (sizeAcc inp)
   filtered p [] | with_pat = FNil
```
ahmadsalim: I know it was there before, but I would prefer having `_` than `with_path` here.

ziman:
Good ideas, thank you!

ziman:
I went for `sizeAccessible` because the corresponding predicate is called `SizeAccessible` because the general predicate is called `Accessible`. The name `access` sounds better but I thought it might be safer to derive a name from the context.

Thoughts/suggestions?

david-christiansen:
@ziman Sounds like a great choice.

jfdm:
Looks good to me. I think a small CHANGELOG entry to highlight the new feature would be nice.

ziman:
Thanks! CHANGELOG updated.

I also had to merge `master` because of a conflict in the CHANGELOG.

ziman:
The failing tests have been fixed in `master` so hopefully this branch will be now marked with a green tick, too.

melted:
Lets merge it before any CI breaks again :-)

Issue #3842, opened 2017-06-01
Manipulation of Data.Bits generates large amount of code


clayrat:
While trying out implementations of random number generators in Idris, I've discovered that a following program

```
import Data.Bits

main : IO ()
main = 
  let
   a = the (Bits 3) (cast 1)
   b = the (Bits 3) (cast 1)
   x = a `xor` b
   y = a `shiftRightLogical` (cast 1)
   z = b `shiftRightLogical` (cast 1)
   res = (x `xor` y) `xor` z
  in
   putStrLn $ show $ bitsToInt res
```

throws an OOM while compiling. Generating only the C source results in a 9.3 megabyte file. The growth of generated code seems to be exponential: removing some operation drops the size to 1-2 Mb, while adding a couple of new ones blows it up to 40+ Mb.

clayrat:
The problem seems to be with using the Bits wrapper, since changing all operations to their native counterparts and explicitly annotating the length seems to solve the problem, i.e., the following program compiles sanely:

```
main : IO ()
main = 
  let
   a = intToBits' {n=64} 1
   b = intToBits' {n=64} 1
   x = xor' {n=64} a b
   y = shiftRightLogical' {n=64} a (intToBits' {n=64} 1)
   z = shiftRightLogical' {n=64} b (intToBits' {n=64} 1)
   res = xor' {n=64} (xor' {n=64} x y) z
  in
   putStrLn $ show $ bitsToInt' {n=64} res
```

ahmadsalim:
@clayrat Have you tried turning off partial evaluation (`--nopartialeval`)?

clayrat:
@ahmadsalim Tried it, appears to have no effect (C files generated with it and without look identical).

clayrat:
Seems like the offender in the generated code is `void _idris_Main_46_main` function which spans 138318 lines out of 143650 total.

ahmadsalim:
@clayrat Have you tried the ECMAScript backend?

clayrat:
@ahmadsalim 
Tried it with Node backend bundled with 1.0.
 The first program was compiling for ~20 minutes and produced a 8.3 Mb JS file. This time, however there wasn't a single huge `main` function, but rather a long list of `_idris_Main_46_main$XXXX` subfunctions, XXXX ranging from 0 to ~12970. 
The second program compiled in 10 seconds and produced a 465k output. 

[...]

PR #3865, opened 2017-06-16
added some missing LTrunc LZExt primitives

rbarreiro:
to solve the problem found in #3842 discussion


melted:
Blah, it's the doc build on travis, it has nothing to do with this. So I'll merge anyway.

PR #3867, opened 2017-06-18
Fix separate compilation

### LINKED ISSUE #3866: Compiling Hello.c via gcc after: idris -S -o Hello.c Hello.idr
I couldn't find help in the usual places (wiki/man pages/mailing list/grep for "gcc" on source tree), so opening issue as I'm a new to Idris and am drawn to using Idris for its C backend. Steps to reproduce:

$ cd samples/tutorial
$ idris -S -o Hello.c Hello.idr
$ pushd ../../rts
$ make
....
$ popd
$ gcc -pthread Hello.c -lidris_rts -e _idris_Main_46_main -I ../../rts/ -L ../../rts
/usr/lib/gcc/x86_64-linux-gnu/6/../../../x86_64-linux-gnu/Scrt1.o: In function `_start':
(.text+0x20): undefined reference to `main'
collect2: error: ld returned 1 exit status

$ gcc --version
gcc (Debian 6.3.0-18) 6.3.0 20170516

Very likely, I may've missed something obvious.

melted:
Fixes #3866

### DIFF
--- a/rts/idris_main.c
+++ b/rts/idris_main.c
@@ -3,6 +3,7 @@
 #include "idris_rts.h"
 #include "idris_stats.h"
 
+void _idris__123_runMain_95_0_125_(VM* vm, VAL* oldbase);
 #if defined(WIN32) || defined(__WIN32) || defined(__WIN32__)
 #include <Windows.h>
 int win32_get_argv_utf8(int *argc_ptr, char ***argv_ptr)
--- a/rts/idris_gmp.h
+++ b/rts/idris_gmp.h
@@ -7,6 +7,8 @@
 #include "mini-gmp.h"
 #endif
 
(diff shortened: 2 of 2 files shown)


rts/idris_main.c
```diff
@@ -3,6 +3,7 @@
 #include "idris_rts.h"
 #include "idris_stats.h"
 
+void _idris__123_runMain_95_0_125_(VM* vm, VAL* oldbase);
```
ahmadsalim: This seems odly specific. Are you sure that this number is used everytime?

melted: Oh, yes, it's the idris entry point, it's called further down in idris_main.c. I just added a prototype to avoid a warning if it wasn't defined in the same file.

PR #3869, opened 2017-06-19
Fixing behavior of addMissing (interactively adding missing clauses)

mheinzel:
There were two problems.

- The new clause was indented to the level of the name that is examined

  ```idris
  test : Maybe a -> Bool
  test Nothing = False
  ```

  With `:am! 2 Nothing` turned into:

  ```idris
  test : Maybe a -> Bool
  test Nothing = False
       test (Just x) = ?Nothing_rhs_1
  ```

- In some cases, the lines after the one examined were simply removed

For each of these, I created a test case that fails without my changes.


melted:
hmmm, test interactive015 fails on appveyor with some copyFile error, but 016 succeeds. They look similar to me. Weird.

melted:
Thanks for the PR!

PR #3870, opened 2017-06-20
Updated stack configuration to use LTS-8.18.

jfdm:


melted:
Try to rebase on master, I have tweaked travis by upgrading to trusty and it seems to work better.

PR #3877, opened 2017-06-22
Add foldlM

clayrat:
This seems like a better fit for `Foldable`, but cannot be put there due to _Foldable -> Monad -> Applicative -> Foldable_ cycle.

### DIFF
--- a/libs/prelude/Prelude/Monad.idr
+++ b/libs/prelude/Prelude/Monad.idr
@@ -6,6 +6,7 @@ import Builtins
 import Prelude.Functor
 import public Prelude.Applicative
 import Prelude.Basics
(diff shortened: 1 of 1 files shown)


Melvar:
It looks like you tried to use haskell doc comment syntax; the syntax used in Idris documentation is Markdown, so referring to code, including function names, should use backticks. (`` `foldl` ``, `` `Monad` `` instead of `'foldl'`, `'Monad'`.)

clayrat:
@Melvar I've started this in `Foldable.idr` which mixes two notations, so I guess I've picked the wrong one. Should I fix it there too then?

melted:
Thanks for the patch!

PR #3878, opened 2017-06-23
add Prelude.Doubles.pow

clayrat:
This might be also added as another primitive from `math.h`, but I'm not sure about the consequences of that.


melted:
What makes it fail in Appveyor is unrelated and fixed.

PR #3880, opened 2017-06-24
Add function for strictly reading source

melted:
So we know that the file is closed before writing to it.


melted:
This will fix the intermittent errors on Appveyor.

mheinzel:
Thanks for fixing this!

PR #3902, opened 2017-07-07
mark covering functions using `sizeAccessible` as `public export total`

clayrat:
(so that they reduce in proofs)


melted:
Thanks for the PR!

PR #3903, opened 2017-07-07
Fix issue where relative paths aren't parsed correctly for JS %lib in…

Risto-Stevcev:
…cludes

The issue is that if you write something like:

```idris
%lib Node "./ex.js"
```

Then the js/node codegen will parse the result as:

```js
var ._ex.js = require("./ex.js");
```

Which isn't valid syntax and would cause a runtime parse error
This commit fixes this issue by parsing the result as:

```js
var __ex.js = require("./ex.js");
```

### DIFF
--- a/src/IRTS/JavaScript/Codegen.hs
+++ b/src/IRTS/JavaScript/Codegen.hs
@@ -84,8 +84,11 @@ includeLibs =
     repl '\\' = '_'
     repl '/' = '_'
     repl c   = c
+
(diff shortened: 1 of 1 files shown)


melted:
Identifiers can't include periods, it becomes field selection, so it's better to add `repl '.' = '_'` and replace them in the entire string.

Risto-Stevcev:
@melted Good call, some file might be `foo.bar.js` or similar, I'll update it

melted:
Thanks for the contribution!

PR #3904, opened 2017-07-08
More compact display of names

Synthetica9:
The old way is quite verbose, and with that, quite unreadable. Currently,
it is displayed as:

```
aaaaa : (x : Vect n a) ->
        (g1 : a -> b) ->
        (g2 : b -> c) ->
        Data.Vect.Vect n implementation of Prelude.Functor.Functor, method map (\x1 => g2 (g1 x1)) x =
        Data.Vect.Vect n implementation of Prelude.Functor.Functor, method map g2 (Data.Vect.Vect n implementation of Prelude.Functor.Functor, method map g1 x)
```

With this update, this becomes:

```
aaaaa : (x : Vect n a) ->
        (g1 : a -> b) ->
        (g2 : b -> c) ->
        @{Functor $ Vect n}.map (\x1 => g2 (g1 x1)) x = @{Functor $ Vect n}.map g2 (@{Functor $ Vect n}.map g1 x)
```

### DIFF
--- a/src/Idris/Core/TT.hs
+++ b/src/Idris/Core/TT.hs
@@ -536,17 +536,20 @@ instance Pretty [Name] OutputAnnotation where
 
 instance Show Name where
     show (UN n) = str n
-    show (NS n s) = showSep "." (map T.unpack (reverse s)) ++ "." ++ show n
+    show (NS n s) = show n
     show (MN _ u) | u == txt "underscore" = "_"
     show (MN i s) = "{" ++ str s ++ "_" ++ show i ++ "}"
     show (SN s) = show s
     show (SymRef i) = "##symbol" ++ show i ++ "##"
 
 instance Show SpecialName where
-    show (WhereN i p c) = show p ++ ", " ++ show c
+    show (WhereN i p c) = show p ++ "." ++ show c
     show (WithN i n) = "with block in " ++ show n
-    show (ImplementationN cl impl) = showSep ", " (map T.unpack impl) ++ " implementation of " ++ show cl
-    show (MethodN m) = "method " ++ show m
+    show (ImplementationN cl []) = "@{" ++ show cl ++ "}"
+    show (ImplementationN cl [impl]) = "@{" ++ show cl ++ " $ " ++ show impl ++ "}"
+    show (ImplementationN cl impl) = "@{" ++ show cl ++
+        showSep " " (map (("(" ++) . (++ ")") . show . T.unpack) impl) ++ "}"
+    show (MethodN m) = show m
     show (ParentN p c) = show p ++ "#" ++ T.unpack c
     show (CaseN fc n) = "case block in " ++ show n ++
                         if fc == FC' emptyFC then "" else " at " ++ show fc


Melvar:
Why would it be `@{…}.map` as opposed to `map @{…}`? And if you’re writing `Functor $ Vect n` for an implementation won’t it be confused with the actual type written that way?

Melvar:
Also, removing the namespace in the `Show` instance causes all sorts of error messages to become unclear.

melted:
The expected test output need to be updated.

The namespace is important information. Especially since Idris is liberal with overloading of symbol names.

ahmadsalim:
@melted I agree that the namespace is important; however, I think the current pretty printing is too verbose here as well: e.g. see  `(x : Vect n a)` vs. `(Data.Vect.Vect n implementation of Prelude.Functor.Functor, method map g1 x)` where it mentions the whole namespace in the latter.

Melvar:
It’s troublesomely verbose for this particular case, but removing namespace information in that function removes it for several error messages that don’t make sense without it, as the diffs from CI show. It would also be preferable if it were still shown under the `showImplicits` option, as on other names.

david-christiansen:
While the current naming is a bit much, I don't think we should merge this proposed syntax.

Right now, `foo @{bar}` means to supply the expression `bar` as the dictionary argument. Making `@{bar}` _also_ be syntax for a name seems likely to lead to a lot of confusion from new users.

I'd like to suggest an alternative: what about showing, by default, only enough namespace to disambiguate the `NS`-guarded names underneath the `SN`?

melted:
It would be nice if the error messages didn't rely on Show instances to print names and stuff so there could be different functions showing more or less information based on what was needed in that context.

david-christiansen:
@melted Agreed!

Synthetica9:
@melted Is there a way to show all usages of an instance?

david-christiansen:
@Synthetica9 If it's a named instance, then `:whocalls` will do the trick.  Otherwise, you can do it with a sufficiently powerful implementation of the IDE protocol, but I think only the Emacs mode has the necessary features.

Steps:
1. Evaluate an expression at the REPL that will have a reference to the instance you care about
2. Right-click and show the core-language version (or show the implicits)
3. Right-click the instance, and pick "Who calls"

iorepl1

iorepl2

iorepl3

iorepl4

iorepl5

iorepl6

Synthetica9:
@david-christiansen That's really cool, but the relevant code is in Haskell. Has anyone implemented something like this for Haskell?

david-christiansen:
Ah, I misread your comment's context :-) I don't think that exists for Haskell, unfortunately, other than by deleting the instance and seeing what breaks.

Synthetica9:
@david-christiansen Ah, I didn't even think of that. I'll try that when I have time

ahmadsalim:
@Synthetica9 Have you had the time to look at this again?

ahmadsalim:
@idris-lang/contributors If anyone else is interested in working on this they are welcome to.
Thanks @Synthetica9 for the PR, but it seems that there has been no development recently so I will close it. You are welcome to reopen it if you have updates.

PR #3914, opened 2017-07-14
Mention do-notation in elab shell docs

clayrat:
This can be a bit surprising if you consider the Elab shell to be just a REPL mode.


david-christiansen:
Thanks!

PR #3921, opened 2017-07-16
Drop support for GHC <= 7.8.4, Cabal <= 1.20

### LINKED ISSUE #3918: Drop support for GHC <= 7.8.4, Cabal <= 1.20
This would make development a lot easier.
* Because of different versions of `Prelude` on needs to be careful with imports and test them in all environments.
* We reduce the number of travis jobs and get quicker feedback.
* It would enable us to remove a couple of hacks, especially in `Setup.hs`.

I’d be happy to tackle this.

geigerzaehler:
Closes  #3918.

Also upgrades `optparse-applicative` so that we can remove conditional compilation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,12 +14,6 @@ matrix:
     - os: osx
       env: CABALVER="1.24" GHCVER="8.0.1" TESTS="test_c"
       compiler: ": #GHC 8.0.1"
-    - env: CABALVER="1.20" GHCVER="7.6.3" TESTS="test_c"
-      compiler: ": #GHC 7.6.3"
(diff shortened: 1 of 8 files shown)


david-christiansen:
Great! When this gets merged, we need to make sure to announce it on the mailing list because we have users on Debian oldstable who need to know that they should upgrade to stretch before doing a git pull :)

david-christiansen:
I emailed the list. If we don't hear back in a little while, then I'll merge this.

melted:
Well, some time has passed. Let's merge.

It's probably time to disable the mac builds entirely, they always time out for us now.

Issue #3924, opened 2017-07-17
A proof of False = True -> Void in Prelude?


clayrat:
I know adding things to Prelude is frowned upon, so wanted to discuss first. 

There is already such a proof in `Data.Heap`@contrib (https://github.com/idris-lang/Idris-dev/blob/master/libs/contrib/Data/Heap.idr#L135-L137), but it's not really used there.

However it might be useful for some naive proofs if moved to `Prelude.Bool`.  For example, `Prelude.Nat` already has both 
 - `Uninhabited (Z = S n)` (https://github.com/idris-lang/Idris-dev/blob/master/libs/prelude/Prelude/Nat.idr#L26-L27) 
and 
- `SIsNotZ : (S x = Z) -> Void` (https://github.com/idris-lang/Idris-dev/blob/master/libs/prelude/Prelude/Nat.idr#L321-L322).

david-christiansen:
It's not so much that adding things to the Prelude is frowned upon in general, it's just that it implies a commitment to maintain the things that are there for the foreseeable future. So we want to make sure that the things that are added aren't getting in the way of users's choices of names and that they are actually useful in a variety of Idris programs.

Things like `SIsNotZ` have been around for a long time, before we started getting more selective.

I think that an `Uninhabited (True = False)` implementation makes sense in Prelude. There's only one sensible implementation, and it doesn't clutter the global namespace. Perhaps:
```
Uninhabited (True = False) where
  uninhabited Refl impossible

Uninhabited (False = True) where
  uninhabited Refl impossible
```
does the trick?

clayrat:
I see! The duplication in the interface approach seems a bit unfortunate, with a function you could use `sym`, on the other hand, as you say, it might interfere with somebody's function name. Btw, which style (`x->Void` vs `Uninhabited x`) is recommended nowadays? Seems that `absurd` is pretty much the only use for `Void` anyway.

david-christiansen:
The `Uninhabited` interface is a canonical place to put proofs that certain things can't happen, because then we can just say `absurd` when we want to refer to one rather than remembering its name. It hasn't been around as long as the whole Prelude, though - I think I stuck it in there a couple years ago.

Usually, `absurd` will do what you want, but it might be inconvenient depending on your local context, in which case explicit proofs of `A -> Void` might be needed.

ahmadsalim:
@clayrat Maybe the following could work (or maybe it goes into an infinite loop)?

```idris
[...]

PR #3925, opened 2017-07-18
Add Uninhabited instances for True=False

clayrat:
(and remove unused proof from Data.Heap)
See https://github.com/idris-lang/Idris-dev/issues/3924 for discussion


david-christiansen:
Thanks! I think it's fine to remove the proof from the heap code without deprecation, because it's in contrib.

PR #3929, opened 2017-07-20
add symmetry, symmetryAs and applyAs tactics to Pruviloj

clayrat:
This is my first attempt at writing tactics in Idris, so don't judge too harshly :)

I've tried to replicate the `symmetry`, `symmetry in H` and `apply... in H` tactics from Coq here. 

Not too sure about this version of `symmetry` - it leaves a `pi_arg` behind when the new goal is solved, requiring another `solve`.

### DIFF
--- a/libs/pruviloj/Pruviloj/Core.idr
+++ b/libs/pruviloj/Pruviloj/Core.idr
@@ -225,13 +225,19 @@ refine tm =
         countPi (RBind _ (Pi _ _) body) = S (countPi body)
         countPi _ = Z
 
+||| A helper to extract the type representation for term.
+getTTType : Raw -> Elab TT
+getTTType r = do
+    env <- getEnv
+    rty <- check env r
+    pure $ snd rty
 
 ||| Split a pair into its projections, binding them in the context
 ||| with the supplied names. A special case of Coq's `inversion`.
 both : Raw -> TTName -> TTName -> Elab ()
 both tm n1 n2 =
     do -- We don't know that the term is canonical, so let-bind projections applied to it
-       (A, B) <- isPairTy (snd !(check !getEnv tm))
+       (A, B) <- isPairTy !(getTTType tm)
        remember n1 A; apply `(fst {a=~A} {b=~B} ~tm) []; solve
        remember n2 B; apply `(snd {a=~A} {b=~B} ~tm) []; solve
   where
@@ -268,3 +274,37 @@ reflexivity =
                 , NamePart `{reflexivity}
                 , TextPart "is not applicable."
                 ]
+
+||| Attempt to swap the sides of equality in a goal.
+symmetry : Elab ()
+symmetry = do
+    [_,_,_,_,_] <- apply (Var `{{sym}}) [True, True, True, True, False]
+       | _ => fail [TextPart "Failed while applying", NamePart `{symmetry} ]
+    solve
+    attack
+
+||| Swap the sides of an equality and bind the result to supplied name.
+symmetryIn : Raw -> TTName -> Elab ()
+symmetryIn t n = do
+    tt <- getTTType t
+    case tt of
+      `((=) {A=~A} {B=~B} ~l ~r) => do
+        af <- forget A
+        bf <- forget B
+        lf <- forget l
+        rf <- forget r
+        let ts = the Raw $ `(sym {a=~af} {b=~bf} {left=~lf} {right=~rf} ~t)
+        tsty <- forget !(getTTType ts)
+        letbind n tsty ts
+      _ => fail [TermPart tt, TextPart "is not an equality"]
+
+||| Apply argument term to a function term and bind the result to supplied name.
+applyIn : Raw -> Raw -> TTName -> Elab ()
+applyIn f x n = do
+    ftt <- getTTType f
+    case ftt of
+      `(~xty -> ~yty) => do
+          let fx = RApp f x
+          fxty <- forget !(getTTType fx)
+          letbind n fxty fx
+      _ => fail [TermPart ftt, TextPart "is not a function"]
--- a/libs/pruviloj/Pruviloj/Core.idr
+++ b/libs/pruviloj/Pruviloj/Core.idr
@@ -227,10 +227,7 @@ refine tm =
 
 ||| A helper to extract the type representation for term.
 getTTType : Raw -> Elab TT
-getTTType r = do
-    env <- getEnv
-    rty <- check env r
-    pure $ snd rty
+getTTType r = snd <$> check !getEnv r
 
 ||| Split a pair into its projections, binding them in the context
 ||| with the supplied names. A special case of Coq's `inversion`.
@@ -259,7 +256,6 @@ unproduct tm =
        try (unproduct (Var n1))
        try (unproduct (Var n2))
 
-
 ||| A special-purpose tactic that attempts to solve a goal using
 ||| `Refl`. This is useful for ensuring that goals in fact are trivial
 ||| when developing or testing other tactics; otherwise, consider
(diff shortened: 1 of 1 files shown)


ahmadsalim:
@clayrat This looks very cool!

Can you please provide some examples, in code and ordinary documentation on how these work?

libs/pruviloj/Pruviloj/Core.idr
```diff
@@ -268,3 +274,37 @@ reflexivity =
...
                 , TextPart "is not applicable."
                 ]
+
+||| Attempt to swap the sides of equality in a goal.
+symmetry : Elab ()
+symmetry = do
+    [_,_,_,_,_] <- apply (Var `{{sym}}) [True, True, True, True, False]
+       | _ => fail [TextPart "Failed while applying", NamePart `{symmetry} ]
+    solve
+    attack
```
ahmadsalim: The `solve`-`attack` sequence looks perculiar (usually it is the other way around), what is the justification for using `attack` at the end?

clayrat:
@ahmadsalim Is there some existing place with this sort of examples? I've only found a few tests for `Pruviloj` (`meta02`, `meta04` and `pruviloj001`) and no mention of concrete tactics in docs.

libs/pruviloj/Pruviloj/Core.idr
```diff
@@ -268,3 +274,37 @@ reflexivity =
...
+
+||| Swap the sides of an equality and bind the result to supplied name.
+symmetryIn : Raw -> TTName -> Elab ()
+symmetryIn t n = do
```
ahmadsalim: It is interesting that this and `symmetry` are independent.
Do you know if it is possible to implement them in terms of each other, or do they just share name?

libs/pruviloj/Pruviloj/Core.idr
```diff
@@ -268,3 +274,37 @@ reflexivity =
...
+        lf <- forget l
+        rf <- forget r
+        let ts = the Raw $ `(sym {a=~af} {b=~bf} {left=~lf} {right=~rf} ~t)
+        tsty <- forget !(getTTType ts)
+        letbind n tsty ts
+      _ => fail [TermPart tt, TextPart "is not an equality"]
+
+||| Apply argument term to a function term and bind the result to supplied name.
+applyIn : Raw -> Raw -> TTName -> Elab ()
+applyIn f x n = do
```
ahmadsalim: `applyIn` seems to be more related to the Coq `apply` than the Idris `apply`; maybe it is better to use another name?

libs/pruviloj/Pruviloj/Core.idr
clayrat: The first `solve` here introduces a `pi_arg`, so `attack` makes it into hole in case you need a subsequent call to `rewriteWith`. As I said, I'm not too sure about `symmetry` tho.

libs/pruviloj/Pruviloj/Core.idr
clayrat: I've tried doing it via introducing a hole based on the given term, refocusing and applying `symmetry` but that leaves you with extra trivial goals, so I thought a direct let-bind is more neat.

libs/pruviloj/Pruviloj/Core.idr
clayrat: I'm not sure about the name either, but `exactIn` didn't sound too fitting as well.

libs/pruviloj/Pruviloj/Core.idr
ahmadsalim: `attack` introduces a guess not a whole. 
I think you can make a hole yourself and plug it in with `exact` or `fill` instead (maybe that is a better idea, than using `solve`-`attack`).

libs/pruviloj/Pruviloj/Core.idr
ahmadsalim: @clayrat Maybe updating `solve` would change this?
But I do not think it is important, I was just wondering whether it was possible.

clayrat: What do you mean by "updating `solve`", isn't it a primitive?

libs/pruviloj/Pruviloj/Core.idr
clayrat: AFAIU `attack` converts it into a form suitable for `rewriteWith` (and probably other tactics), not sure how `exact`/`fill` would help here.

ahmadsalim:
@clayrat Sorry for deleting the comment.
They showed up as duplicates for me, so I thought I would remove one, but apparently it was a GitHub issue.

clayrat:
Ah, basically the comment was about where should I put these examples you were talking about - specific tactics are not mentioned in docs, and there's only a handful of tests for `Pruviloj`.

ahmadsalim:
@clayrat In any case, I think the best thing would be to put the example in the code documentation. You are also welcome to add new test function to a suitable existsting test.

libs/pruviloj/Pruviloj/Core.idr
david-christiansen: Right, `attack` ensures that a hole is in form `? x : t . x` (as opposed to `? x : t . f x`). This is needed for tactics that might introduce binders, such as `rewriteWith`.

libs/pruviloj/Pruviloj/Core.idr
```diff
@@ -225,13 +225,19 @@ refine tm =
         countPi (RBind _ (Pi _ _) body) = S (countPi body)
         countPi _ = Z
 
+||| A helper to extract the type representation for term.
+getTTType : Raw -> Elab TT
+getTTType r = do
```
david-christiansen: This library doesn't use the "dangling `do`" style, but rather the "`do` is part of the expression" style. Please keep this consistent.

libs/pruviloj/Pruviloj/Core.idr
```diff
@@ -268,3 +274,37 @@ reflexivity =
                 , NamePart `{reflexivity}
                 , TextPart "is not applicable."
                 ]
+
+||| Attempt to swap the sides of equality in a goal.
+symmetry : Elab ()
+symmetry = do
```
david-christiansen: This is another "dangling `do`"

libs/pruviloj/Pruviloj/Core.idr
david-christiansen: There shouldn't be a trailing `attack` here. The style in Pruviloj is that those who want to use binding tactics are responsible for bracketing their binders in `attack` and `solve` themselves. This makes it so that nobody needs to keep track of the state that other tactics leave the hole in.

libs/pruviloj/Pruviloj/Core.idr
```diff
@@ -268,3 +274,37 @@ reflexivity =
...
+    [_,_,_,_,_] <- apply (Var `{{sym}}) [True, True, True, True, False]
```
david-christiansen: The handling of holes here is too implicit. The code would be more clear if it explicitly bound the name of the last hole (the one that will have an equality proof in it later) and then explicitly called `focus` on that, and the docstring stated that if the original goal is `x = y`, then after this tactic, the focus will be on a hole of type `y = x`.

Alternatively, it could return the name of the hole and rely on the caller to focus on it, but I think that the caller will almost always want to proceed in the hole, so the version that calls `focus` is probably more convenient.

libs/pruviloj/Pruviloj/Core.idr
david-christiansen: Dangling `do`

libs/pruviloj/Pruviloj/Core.idr
```diff
@@ -268,3 +274,37 @@ reflexivity =
...
+       | _ => fail [TextPart "Failed while applying", NamePart `{symmetry} ]
+    solve
+    attack
+
+||| Swap the sides of an equality and bind the result to supplied name.
+symmetryIn : Raw -> TTName -> Elab ()
+symmetryIn t n = do
+    tt <- getTTType t
+    case tt of
```
david-christiansen: This `do` could be completely eliminated by using `case !(getTTType t) of ...`

libs/pruviloj/Pruviloj/Core.idr
```diff
@@ -268,3 +274,37 @@ reflexivity =
...
+      `((=) {A=~A} {B=~B} ~l ~r) => do
```
david-christiansen: Dangling `do`

libs/pruviloj/Pruviloj/Core.idr
```diff
@@ -268,3 +274,37 @@ reflexivity =
...
+        af <- forget A
+        bf <- forget B
+        lf <- forget l
+        rf <- forget r
+        let ts = the Raw $ `(sym {a=~af} {b=~bf} {left=~lf} {right=~rf} ~t)
+        tsty <- forget !(getTTType ts)
+        letbind n tsty ts
+      _ => fail [TermPart tt, TextPart "is not an equality"]
```
david-christiansen: It would be nice if the message also include the non-`=` type that was found

libs/pruviloj/Pruviloj/Core.idr
```diff
@@ -268,3 +274,37 @@ reflexivity =
...
+
+||| Attempt to swap the sides of equality in a goal.
+symmetry : Elab ()
+symmetry = do
+    [_,_,_,_,_] <- apply (Var `{{sym}}) [True, True, True, True, False]
+       | _ => fail [TextPart "Failed while applying", NamePart `{symmetry} ]
+    solve
+    attack
+
+||| Swap the sides of an equality and bind the result to supplied name.
```
david-christiansen: The documentation should include the roles of the specific arguments, like:
```
||| Swap the sides of an equality, saving the result in a `let`-binding
||| @ t the equality proof to swap
||| @ n the name to use for the swapped equality
symmetryIn : (t : Raw) -> (n : TTName) -> Elab ()
...
```

libs/pruviloj/Pruviloj/Core.idr
```diff
@@ -268,3 +274,37 @@ reflexivity =
...
+    tt <- getTTType t
+    case tt of
+      `((=) {A=~A} {B=~B} ~l ~r) => do
+        af <- forget A
+        bf <- forget B
+        lf <- forget l
+        rf <- forget r
+        let ts = the Raw $ `(sym {a=~af} {b=~bf} {left=~lf} {right=~rf} ~t)
+        tsty <- forget !(getTTType ts)
+        letbind n tsty ts
```
david-christiansen: It is allowed to nest `!`, such as:
`letbind n !(forget !(getTTType ts)) ts`

libs/pruviloj/Pruviloj/Core.idr
```diff
@@ -268,3 +274,37 @@ reflexivity =
...
+||| Apply argument term to a function term and bind the result to supplied name.
+applyIn : Raw -> Raw -> TTName -> Elab ()
+applyIn f x n = do
+    ftt <- getTTType f
+    case ftt of
+      `(~xty -> ~yty) => do
+          let fx = RApp f x
+          fxty <- forget !(getTTType fx)
+          letbind n fxty fx
+      _ => fail [TermPart ftt, TextPart "is not a function"]
```
david-christiansen: There should be a newline at the end of the file

david-christiansen:
@clayrat Tests for `pruviloj` additions are always welcome :-)

libs/pruviloj/Pruviloj/Core.idr
clayrat: Yes, but how do we print the term in case of an error?

libs/pruviloj/Pruviloj/Core.idr
clayrat: How do I get that info?

libs/pruviloj/Pruviloj/Core.idr
david-christiansen: If it's the scrutinee of a case expression, then the information is available as a pattern. Or have I misunderstood?

libs/pruviloj/Pruviloj/Core.idr
david-christiansen: Instead of `_` on the left hand side, give it a name, and then put another `TermPart` in the message.

libs/pruviloj/Pruviloj/Core.idr
clayrat: D'oh!

libs/pruviloj/Pruviloj/Core.idr
clayrat: Hm, I still don't get it :( Wouldn't that be the same as printing `tt` currently?

clayrat:
Regarding 5 (having users explicitly supply names to let-bind results to) - I've copied this style from `injective` and `both`, wasn't a conscious design decision I guess :)

clayrat:
I think I've addressed the majority of comments, except (3) which I feel my pruviloj-fu is insufficient for, and the part about printing additional type info in case of failure.

ahmadsalim:
@david-christiansen Could you kindly take a look at this when you have time? I think it would be good to merge :).

david-christiansen:
I think it's good for a merge. Thanks, and sorry I was distracted for a while!

Issue #3923, opened 2017-07-17
New JS codegen leaving code out during compilation


Risto-Stevcev:
I'm trying to generate the commented out code here:
https://github.com/Risto-Stevcev/idris-node-fs/blob/master/Test/FileSystem.idr#L32-L36

But the compiled js file is producing the following runtime error:
```
/home/risto/git/idris/idris-node-fs/testit:764
    return $partial_5_6$io_95_bind(null, null, null, Control__Monad__Free__foldFree($HC_0_65647$Unit, Node__Free__FileSystem__FileSystemF, null, new $HC_2_0$Prelude__Monad__Monad_95_ictor(new $HC_3_0$Prelude__Applicative__Applicative_95_ictor($partial_0_1$Test__FileSystem__Main___123_main_95_3_125_(), $partial_0_1$Test__FileSystem__Main___123_main_95_5_125_(), $partial_0_1$Test__FileSystem__Main___123_main_95_9_125_()), $partial_0_1$Test__FileSystem__Main___123_main_95_13_125_()), null, $partial_0_1$Test__FileSystem__Main___123_main_95_14_125_(), Test__FileSystem__exclaimHello()), $partial_0_1$Test__FileSystem__Main___123_main_95_30_125_());
                                                                                                      ^

ReferenceError: Node__Free__FileSystem__FileSystemF is not defined
    at Test__FileSystem__Main__main (/home/risto/git/idris/idris-node-fs/testit:764:103)
    at $_0_runMain (/home/risto/git/idris/idris-node-fs/testit:1339:25)
    at Object.<anonymous> (/home/risto/git/idris/idris-node-fs/testit:1354:1)
    at Object.<anonymous> (/home/risto/git/idris/idris-node-fs/testit:1355:3)
    at Module._compile (module.js:569:30)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)
    at tryModuleLoad (module.js:466:12)
    at Function.Module._load (module.js:458:3)
    at Function.Module.runMain (module.js:605:10)
```
I'm using [this library](https://github.com/Risto-Stevcev/idris-naps) I wrote to manage deps, it's just a convenient wrapper to run some idris commands. So I ran `naps --install-deps`, which just installs all the idris libs it needs into the `libs` folder of the project root directory, and uses that as the IDRIS_LIBRARY_PATH for commands.

I ran this command to generate the js file:
```
[...]

PR #3930, opened 2017-07-20
fix for New JS codegen leaving code out during compilation #3923

rbarreiro:


melted:
Great!

Really weird that that case isn't exposed by any test.

PR #3931, opened 2017-07-20
Refactoring: Move options out of AbsSyntax

melted:


eraserhd:
@melted - As a bystandard recently reading the code to AbsSyntax, this is nice!

Issue #3928, opened 2017-07-19
consInjective in Prelude?


clayrat:
Would it be fine to add 
```
consInjective : (x :: xs) = (y :: ys) -> (x = y, xs = ys) 
consInjective Refl = (Refl, Refl)
```
to `Prelude.List`, similar to `succInjective` for `Nat` (https://github.com/idris-lang/Idris-dev/blob/master/libs/prelude/Prelude/Nat.idr#L432-L434) ?

david-christiansen:
Yes!

clayrat:
Is there a preference for the arguments to be explicit/implicit? `succInjective` has explicit ones, but here it might get somewhat crowded.

david-christiansen:
There should be enough explicit arguments that Idris will be able to find the implicit ones at typical use sites. In this case, I think you have it right already.

Melvar:
Technically one could expand this to lists of different types and also get the equality of types out, but I’m not sure that would be useful particularly often.

ahmadsalim:
@clayrat I think that for such a small change, it is easier to just open a PR next time without creating an issue.
We can always discuss whether it should be moved to `contrib` etc. there if needed.

PR #3932, opened 2017-07-21
add consInjective to Prelude.List

clayrat:
Resolves https://github.com/idris-lang/Idris-dev/issues/3928
Do we also need a changelog entry for this?


melted:
Thanks!

PR #3936, opened 2017-07-22
Fixes #1211 : default impl in interfaces through parent constraints recursively

### LINKED ISSUE #1211: Default instance can only find parent class, not ancestor.
``` idris
class Foo a where

class Foo a => Bar a where

class Bar a => Baz a where
  instance Foo a where
```

complains

```
line 6 col 12: Default instances must be for a superclass constraint on the containing class.
```

If I change `Baz` to

```
class (Foo a, Bar a) => Baz a where
  instance Foo a where
```

it works fine, but that Foo constraint should be implied.

mandubian:
Updated `checkDefaultSuperInterfaceImplementation` to check default implementations nested in interfaces with respect to constraints of containing interface and if not found, recursively search through constraints of parent interfaces.


melted:
Thanks for the PR!

@edwinb you should probably take a look at this.

edwinb:
Oops, sorry, I saw this immediately before I went on holiday then forgot. It doesn't seem especially controversial to me though - I know this is a tricky and not often used corner though so good to check. Thanks for the patch!

PR #3941, opened 2017-07-23
Add notLTImpliesGTE to Prelude

ruippeixotog:
Adds a proof that `Not (LT a b)` implies `GTE a b` to the Prelude. I needed this step in proofs more than once and I think it is useful enough for being in the standard library.

Let me know if you prefer it somewhere else or if you prefer some arguments implicit or explicit :)

### DIFF
--- a/libs/prelude/Prelude/Nat.idr
+++ b/libs/prelude/Prelude/Nat.idr
@@ -160,6 +160,11 @@ lteAddRight : (n : Nat) -> LTE n (plus n m)
 lteAddRight Z = LTEZero
 lteAddRight (S k) = LTESucc (lteAddRight k)
 
+||| If a number is not less than another, it is greater than or equal to it
+ifNotLtThenGte : Not (LT a b) -> GTE a b
+ifNotLtThenGte {b = Z} _ = LTEZero
+ifNotLtThenGte {a = Z} {b = S k} notLt = absurd (notLt (LTESucc LTEZero))
+ifNotLtThenGte {a = S k} {b = S j} notLt = LTESucc (ifNotLtThenGte (notLt . LTESucc))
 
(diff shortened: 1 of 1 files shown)


ahmadsalim:
Thanks for the PR.
Since this is going to be in the Prelude I would let someone else merge, but I have a small comment regarding the name: I think it would be better if it was called `NotLTImpliesGTE` instead.

ruippeixotog:
I'm ok with changing the name to anything else, if you want. There are many conventions used in that file, like using uppercase in some places (`succNotLTEzero`) and lowercase in others (`fromLteSucc`), so I didn't really know what to call this. Do I call it `notLTImpliesGTE` or `notLtImpliesGte`?

ahmadsalim:
@ruippeixotog I think the former is better

clayrat:
I wonder if adding this
```
notLTEImpliesGT : Not (LTE a b) -> GT a b
notLTEImpliesGT {a=Z}    contra = absurd $ contra LTEZero
notLTEImpliesGT {a=S _}  {b=Z} _ = LTESucc LTEZero
notLTEImpliesGT {a=S a'} {b=S b'} contra = 
  LTESucc $ notLTEImpliesGT {a=a'} {b=b'} $ contra . LTESucc
```
would be useful, or is it inferrable in a more straightforward way?

clayrat:
On a side note, I also wonder whether Coq-style definition
```
> data LTE'  : (n, m : Nat) -> Type where
>   LTEEq : LTE' n n
>   LTESucc : LTE' n m -> LTE' n (S m)
```
would be more practical in any way?

dc-mak:
I think that definition allows for a much more natural expression of some code related to some stuff I was doing a few days ago: https://groups.google.com/forum/#!topic/idris-lang/a8IoO4zocS8

PR #3944, opened 2017-07-24
Fix "and" typo in syntax guide

emptyflash:
😅


melted:
I think we can skop CI on this one.

PR #3938, opened 2017-07-23
Fix code typo in "Composing State Machines" documentation

mitchmindtree:
Really enjoying this tutorial btw - thanks!

PR #3945, opened 2017-07-24
Fix interface/implementation mismatch in Control.ST docs

asandroq:


clayrat:
Seems like a duplicate of https://github.com/idris-lang/Idris-dev/pull/3938

melted:
Thanks!

I merged the other one.

PR #3946, opened 2017-07-24
Disable macOS builds on Travis

melted:
They are always timing out.


ahmadsalim:
@melted Have you tried with 8.2, it should have a fast code gen, right?

PR #3948, opened 2017-07-25
Implement shebang handling

### LINKED ISSUE #2247: Allow Idris source files to contain a shebang line
So idris source files can be made executable on unixy systems by putting `#!/bin/env runidris` on the first line.

This would also require a script/program `runidris` that works like `runhaskell`

melted:
Add two scripts for running idris programs with the C and node backends and adds a test.

Fixes #2247 

Implementation cribbed from #2289

### DIFF
--- a/scripts/runidris
+++ b/scripts/runidris
@@ -1,0 +1,18 @@
+#!/bin/sh
+
+# This is a script that runs an idris program, suitable for use
+# in a shebang line. We cache the executables by naming them after
+# the SHA hash of the idris source and looking for it before
+# compiling
+RUNIDRIS_DIR="${TEMP:-/tmp}/runidris"
+if [ ! -d $RUNIDRIS_DIR ]; then
+    mkdir $RUNIDRIS_DIR
+    chmod 700 $RUNIDRIS_DIR
+fi
+TEMP_NAME="runidris-`sha1sum $1 | cut -c1-40`"
+FP="$RUNIDRIS_DIR/$TEMP_NAME"
+if [ ! -e $FP ]; then
+    "${IDRIS:-idris}" $1 --ibcsubdir "$RUNIDRIS_DIR" -i "." -o $FP
+fi
+shift
+"$FP" "$@"
--- a/scripts/runidris-node
+++ b/scripts/runidris-node
@@ -1,0 +1,18 @@
+#!/bin/sh
+
+# This is a script that runs an idris program, suitable for use
+# in a shebang line. We cache the executables by naming them after
+# the SHA hash of the idris source and looking for it before
+# compiling. This is the node version.
+RUNIDRIS_DIR="${TEMP:-/tmp}/runidris-node"
+if [ ! -d $RUNIDRIS_DIR ]; then
+    mkdir $RUNIDRIS_DIR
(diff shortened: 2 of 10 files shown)


david-christiansen:
Where does Idris look for libraries when #!ing? I would think that in shebang mode, it should consider imports relative to the file being executed.

melted:
@david-christiansen It's easy enough to put that into the command line in the scripts. For neatness, I should probably put the ibc files in the temp directory too.

It would be nice to have a directive so that the packages used could be specified in the source file. Here you can't put them on the command line, so it's a bit hamstrung without it. But I think that would be generally useful. If a program uses effects, putting `%package effects` in there is easier than remembering `-p effects` every time one builds it.

melted:
Another thing to do is to pass the rest of the args into the program.

david-christiansen:
It would indeed be nice to be able to configure packages in programs. The Emacs mode has a stupid hack for this that only works in Emacs, which bugs me.

In any case, I think this is worth a mailing list announcement :-)

jfdm:
I'm not entirely convinced that a specific directive for packages is needed for 'Idris Scripts'. Such a `%package` directive might serve as a source of confusion when writing ordinary Idris Programs. That is, people might think that `%package` is required regardless. Maybe if suitable error messages and flags and warnings are present to enforce correct usage, then this approach might work.

I would much prefer t[he approach `stack` has taken with specifying packages in `Scripts`](https://haskell-lang.org/tutorial/stack-script) which requires passing of package imports as command line arguments in the file itself.

melted:
Yeah, that's a good idea, that would mean I would have to rewrite runidris in something other than shell though :-)

zenntenn:
Is there a reason it can't be set up to have the -p argument passed in on
the shebang line?

melted:
The shebang line only takes one argument and if you use /usr/bin/env as is common that is taken by the executable. 

> 28 juli 2017 kl. 22:40 skrev zenntenn <notifications@github.com>:
> 
> Is there a reason it can't be set up to have the -p argument passed in on
> the shebang line?
>

melted:
One thing that is possible is to let the runidris program filter the command line and extract ones marked in a special way (like ghcs +RTS) and pass them to idris. 

> 28 juli 2017 kl. 22:40 skrev zenntenn <notifications@github.com>:
> 
> Is there a reason it can't be set up to have the -p argument passed in on
> the shebang line?
>

melted:
Having thought about it a bit more is that it would preferable to have a directive to pass general command lines inside a file. Stack does it in comments because GHC can't be expected to know about stack. We don't have that problem.

david-christiansen:
Perhaps in Idris? [ducks]

On July 28, 2017 4:36:33 PM EDT, Niklas Larsson <notifications@github.com> wrote:
>Yeah, that's a good idea, that would mean I would have to rewrite
>runidris in something other than shell though :-)
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/idris-lang/Idris-dev/pull/3948#issuecomment-318755892

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

melted:
Now I'm actually tempted to do that. It should be easy enough.

PR #3949, opened 2017-07-26
Increase upper bounds for some packages

melted:
Remove conditionals for older GHCs.


clayrat:
I'm curious about 8.2.1, is it possible to start building with it already?

melted:
I have built it with --allow-newer=base

PR #3950, opened 2017-07-27
Bump Idris to compile using LTS Haskell 9.0.

jfdm:


melted:
Travis is completely hopeless now and the builds that don't time out succeed. So merge.

PR #3957, opened 2017-08-01
Allow building with optparse-applicative 0.14

mpkh:


ahmadsalim:
@mpkh Thanks for the PR!
This is however already covered by #3949 .

PR #3959, opened 2017-08-01
Fix undeclared variable in node runtime

christophe-calves:
This PR fixes an undeclared variable error occuring when optimising node.js outout with the closure compiler.


melted:
Thanks!

Travis is still slow, but the fix is obviously correct anyway.

PR #3965, opened 2017-08-03
Fix compatibility with cheapskate-0.1.1

felixonmars:
Only define instances for older cheapskate

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -269,7 +269,7 @@ Library
                 , blaze-html >= 0.6.1.3 && < 0.10
                 , blaze-markup >= 0.5.2.1 && < 0.9
                 , bytestring < 0.11
-                , cheapskate < 0.1.1
+                , cheapskate >= 0.1.1 && < 0.2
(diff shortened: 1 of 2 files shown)


david-christiansen:
Do we need to keep compatibility with older Cheapskate versions? Is there e.g. a GHC  or Stack snapshot support issue with the newer version that prevents us from requiring it instead?

felixonmars:
I don't need it, just thought that it won't harm. I can surely remove it if you want :)

ahmadsalim:
@marcelinevq recently patched cheapskate so it does not add extra dependencies, which it did for older versions of cabal and stack. Maybe we should rely on the latest version in this patch.

felixonmars:
The latest version is 0.1.1 which is already required in this patch. Are you referring to an unreleased version?

ahmadsalim:
@felixonmars Oh, it is the correct version!

melted:
Thanks!

PR #3967, opened 2017-08-03
Add OpenBSD support to shebang script

bonds:
* use sha1 instead of sha1sum on OpenBSD
* enable using runidris regardless of the script's extension
* mark runidris scripts executable

What do you think @melted?


melted:
I was a bit tempted to rewrite it in a real programming language. But I might not get around to that, so all improvements are welcome.

melted:
I'm so impressed that Travis succeeded here :-)

bonds:
LOL, I know, right?

PR #3969, opened 2017-08-04
Solve erasure constraints faster

ziman:
The erasure analysis is too slow and profiling shows that the culprit is (very naïve) constraint solving.

This patch modifies constraint solving to avoid traversing *all* constraints in each iteration. Now, we check only those constraints that are relevant for the nodes newly reached in the previous interation. (This involves keeping a (read-only) index of nodes in constraints.)

This brings down solving time from 4.9s to 1.2s on my machine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# New in ...
+# New in 1.1.1
 
 + Erasure analysis is now faster thanks to a bit smarter constraint solving.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# New in ...
+
++ Erasure analysis is now faster thanks to a bit smarter constraint solving.
+
 # New in 1.1.0
 
 ## Library Updates
--- a/src/Idris/Erasure.hs
+++ b/src/Idris/Erasure.hs
@@ -207,6 +207,7 @@ minimalUsage deps
 -- In the implementation, this is phase-shifted. We first reduce the set
 -- of constraints, given the newly reachable nodes from the previous iteration,
 -- and then compute the set of currently reachable nodes.
+-- Then we decide whether to iterate further.
 forwardChain
     :: Map Node IntSet   -- ^ node index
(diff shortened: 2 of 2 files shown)


jfdm:
@ziman Thanks for this. Only one of the travis builds failed, and this was due to a timeout. I've restarted the build and hopefully it should still pass. I've not looked into why AppVeyor has failed.

melted:
Huh, looks like appveyor didn't even try to build it.

jfdm:
Even though AppVeyor didn't run, and only one of the travis tests timed-out, I'm happy to merge this as is.

WIll sit on it for an hour then merge if objections.

ziman:
I might be careful here because if erasure is incorrect, that might make the JS tests slow enough to time out. I'd try to run them locally and see which one times out.

melted:
We have had problems with travis timeouts for a while now, it's not new with this PR. I'm a bit more worried about appveyor, but it seems to have decided that this patch conflicts with master, which is weird. 

Go ahead and merge, if there's a real problem on windows (very unlikely) I'll fix it later today.

ziman:
Yeah, it's bizarre to see "non-mergeable pull request" right above "this branch has no conflicts".

melted:
Probably we should wait for edwin to get the cheapskate fix release out of the way though. I merged the cabal 2.0 patch because it fixed another compatibility problem.

edwinb:
If this works I think it would be nice to have in 1.1.1, I don't think there's any need to delay for that. I'll get around to that later today or tomorrow.

ziman:
All tests (C + Node) pass on my machine™.

melted:
Ok, I'll merge it, appveyor should run on master then and we can see if everything is ok. But just another thing: this merits a note in the CHANGELOG.

ziman:
A good way to force a rerun of CI.

melted:
No need to wait for travis, the original tests went well (https://travis-ci.org/idris-lang/Idris-dev/builds/260844853) and only comments have changed.

PR #3972, opened 2017-08-04
Add a low-level log statement for erasure step.

jfdm:


melted:
Looks like Travis is feeling a lot better

PR #3896, opened 2017-07-02
Implement a RWS monad for the C backend

melted:
It doesn't really do anything yet, but threading state through the code generation will be useful in the future. For example, better error handling (when we have the source locations in the codegen), various optimizations, or locally decide if some declaration needs to be put on the top level.

PR #3973, opened 2017-08-04
Revert "Merge pull request #3896 from melted/codegen_rws"

melted:
This reverts commit e471a1c8f76ae90492027001124471f70086de98, reversing
changes made to f2c4091990e0ad2629d18480b6a3b1a582dcb7e0.

Reverts #3896 because of the performance hit

### DIFF
--- a/src/IRTS/CodegenC.hs
+++ b/src/IRTS/CodegenC.hs
@@ -9,7 +9,7 @@ Maintainer  : The Idris Community.
 
 module IRTS.CodegenC (codegenC) where
 
-import Idris.AbsSyntax hiding (getBC)
+import Idris.AbsSyntax
 import Idris.Core.TT
 import IRTS.Bytecode
 import IRTS.CodegenCommon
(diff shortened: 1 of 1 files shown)


melted:
Funny thing is that build isn't sped up an appreciable amount, previous builds took 32-33 minutes on Appveyor, these took 32:04 and 32:39. One would think such a big performance hit would be noticeable in 200+ tests.

jfdm:
Please note that in @edwinb's original comment the performance hit was found with code of 'significant size'. I doubt the size of the program's that constitute Idris' test suite are comparable to the size of program that the performance regression was encountered.

melted:
Just one travis build timed out, let's merge!

edwinb:
Yes, I suspect it only really starts to hurt for programs much larger than the test suite. Although the program in question isn't that large - just about 3500 lines. Anyway, the offending function doesn't even register in the profile now, so that's reassuring.

PR #3974, opened 2017-08-04
change extra-deps for idris 1.1.0 lts-9.0 resolver

MarcelineVQ:
stack.yaml extra-deps changed as resolver lts-9.0 doesn't have
the recently required versions of libs in the idris.cabal file.

note added to stack.yaml to re-examine extra-deps on resolver
or cabal file change.

note should probably be in the cabal file but I'm not sure where it
would be seen


melted:
Thanks!

PR #3985, opened 2017-08-08
Addition of tools to help with literate source files.

jfdm:


ahmadsalim:
@jfdm There is no description. 
Does it work generically for any language?

jfdm:
Sorry, for lack of description. 

The tooling is 'generic' in the sense that it uses the 'styles' to identify literate blocks, and lines, in the presented text document. Thus, it is coding language agnostic.

The underlying tokeniser is also greedy. Once it identifies a line marker it reads until the end of line, and once it recognises a start block, it will consume until the next identifiable end block. Thus the input literate files must be well-formed w.r.t to line markers and code blocks.

A nice improvement would be ensuring that the markers start at the beginning of a line, otherwise rouge markers within a document will get recognised aas well.

ahmadsalim:
Yeah, it looks good!

Another nice improvement would be that it read the file as a stream.

PR #3986, opened 2017-08-09
Fix line number reporting for `Text.Literate`.

jfdm:
Ensures that the line count of the original document and resulting
code file are kept the same. This is important for reporting errors
from the code in the original document. Unfortunatly, column numbers
for line markers are not preserved and differ by one plus the number
of characters in the used line marker.


jfdm:
Travis is timing out, but this works locally.

PR #3989, opened 2017-08-12
Change eqPtr to use the primitive

melted:
Maybe eqPtr isn't needed, Ptr has an == instance.

### DIFF
--- a/libs/prelude/Prelude.idr
+++ b/libs/prelude/Prelude.idr
@@ -280,8 +280,7 @@ namespace JSNull
 
 ||| Pointer equality
 eqPtr : Ptr -> Ptr -> IO Bool
-eqPtr x y = do eq <- foreign FFI_C "idris_eqPtr" (Ptr -> Ptr -> IO Int) x y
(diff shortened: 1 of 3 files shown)


jfdm:
Can we avoid using such loose language when describing a change to a fundamental aspect. 'Maybe' implies that you are not sure if it is needed or not. When changing I would rather see a more definite reasoning, rather than something that comes across as a hunch.

melted:
Well, that maybe was not about the content of the commit but the alternative of cutting it out altogether. I changed eqPtr to use the prim__eqPtr which doesn't change the interface at all, the maybe was for removing it as it isn't used in the libs and == works on Ptr. Please look at the code, it's just one line, and don't jump to conclusions.

jfdm:
@melted I *did* look at the code, and I didn't jump to conclusions. My comment wasn't about the code, it was about the accompanying description. I'm not going to comment more on this, aside from reiterating: I think it is better when descriptions give sufficient context and rationale for the change.

PR #3992, opened 2017-08-14
Probably fix spelling of what I guess to be “delimiter”

Melvar:
“deliminator” isn’t in any dictionary I checked; I’m guessing it’s a
mixture of “delimiter” with “eliminator”. Conceivably the intended word
could be “delineator”, but I’ve never seen that used in this context.

Changing the identifier should be okay because it hasn’t been in a release yet.

### DIFF
--- a/libs/contrib/Text/Literate.idr
+++ b/libs/contrib/Text/Literate.idr
@@ -71,14 +71,14 @@ reduce (MkToken _ _ (CodeLine m src) :: rest) acc =
                            (length src)
                            src))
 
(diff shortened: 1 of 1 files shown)


jfdm:
It isn't a spelling error. The term 'deliminator' is [technical jargon](http://catb.org/jargon/html/D/deliminator.html) used interchangeably with 'delimiter'.

Melvar:
Ah. I’ve probably read that entry at some point, but never seen it used in the wild.

PR #3998, opened 2017-08-15
Prefer to load imports from the current package

### LINKED ISSUE #3994: File with name `IO.idr` confuses idris REPL
# Steps to Reproduce

* Create an empty directory
* add two idris files, Foo.idr and IO.idr
* the module names in the files do not matter, assume `module Foo` and `module Foo2`
* invoke `idris Foo.idr`

# Expected Behavior
* The `prelude` is loaded normally.

# Observed Behavior

```
$ idris Foo.idr
/usr/local/Cellar/idris/1.1.1/share/x86_64-osx-ghc-8.0.2/idris-1.1.1/libs/prelude/Prelude.idr: openFile: does not exist (No such file or directory)
     ____    __     _                                          
    /  _/___/ /____(_)____                                     
    / // __  / ___/ / ___/     Version 1.1.1
  _/ // /_/ / /  / (__  )      http://www.idris-lang.org/      
 /___/\__,_/_/  /_/____/       Type :? for help               

Idris is free software with ABSOLUTELY NO WARRANTY.            
For details type :warranty.
Must be an ibc
*Foo> 
```

As a result functions in Foo.idr cannot be accessed. The error also seems to be specific to a filename `IO.idr` - I tried with other filenames which can be found in the prelude, such as `Buildins.idr` but could not reproduce the error.

melted:
This is done by moving import dirs that are prefixes to the current path to the front of the import dir list.

Also make a fuction that only looks for IBCs.

Both these changes should also be helpful for performance.

Fixes #3994

### DIFF
--- a/src/IRTS/Compiler.hs
+++ b/src/IRTS/Compiler.hs
@@ -76,7 +76,7 @@ compile codegen f mtm = do
         libs <- getLibs codegen
         flags <- getFlags codegen
         hdrs <- getHdrs codegen
-        impdirs <- allImportDirs
+        impdirs <- rankedImportDirs f
         ttDeclarations <- getDeclarations reachableNames
         defsIn <- mkDecls reachableNames
         -- if no 'main term' given, generate interface files
--- a/src/Idris/AbsSyntax.hs
+++ b/src/Idris/AbsSyntax.hs
@@ -1108,6 +1108,14 @@ allImportDirs = do i <- getIState
                    let optdirs = opt_importdirs (idris_options i)
                    return ("." : reverse optdirs)
 
+-- Like allImportDirs but the dirs that are a prefix of
+-- the files path first. This makes it look in the current
+-- package first.
+rankedImportDirs :: FilePath -> Idris [FilePath]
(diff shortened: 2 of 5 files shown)


jfdm:
I am wondering if it would be prudent to start documenting such observable aspects (i.e. how Idris load's and processes imports) of Idris' behaviour more in the reference documents.

melted:
I don't think anybody expected or relied on the previous behavior that imports from inside a package searched the current directory first. 

But it should probably have  a CHANGELOG note.

david-christiansen:
It would be nice to document these things, as much to discover weird bugs like this one as anything, but I don't think a desire for more docs should get in the way of fixing this clear bug.

david-christiansen:
This one has actually burned me once or twice when I was busy, and I forgot and didn't fix it. Thanks @melted  :)

jfdm:
> I don't think a desire for more docs should get in the way of fixing this clear bug.

I wasn't suggesting updating the documents to be a block on the PR;  the comment was made after the PR was merged. Documentation is something we need to do better in general, but time and effort... I'll see what I can do next week.

PR #4006, opened 2017-08-18
add System.system documentation

ivanov:


ahmadsalim:
Thanks!

alrunner4:
I think clarification is probably in order: https://github.com/idris-lang/Idris-dev/issues/3251

PR #4009, opened 2017-08-20
Add Cast instance for Nat to String

joom:


libs/prelude/Prelude/Nat.idr
```diff
@@ -238,6 +238,9 @@ Cast Integer Nat where
 Cast String Nat where
     cast str = cast (the Integer (cast str))
 
+Cast Nat String where
+  cast n = cast (the Integer (cast n))
```
david-christiansen: The indentation here should really match line 239.

david-christiansen:
This is a useful thing! But it seems to introduce some style inconsistencies.

libs/prelude/Prelude/Nat.idr
msmorgan: Line 239 is actually inconsistent with the rest of the surrounding lines.

joom:
All implementations indented 2 spaces except the one above. It is fixed now.

Melvar:
Personally, I don’t see the usefulness of a cast to `String` that does the same thing as `show`. As opposed to something like `Cast (List Char) String`, which does something different than `show`.

joom:
Fair point, but there are already `Cast Int String` and `Cast Integer String` instances, which do the same thing as `show`. One would expect a `Cast Nat String` in the same way.

melted:
Thanks!

msmorgan:
Would it be a better idea to implement `Show a` in terms of `Cast a String` throughout the library, establishing a pattern?

Melvar:
That’s not sufficient: `Show` includes `showPrec` to allow correct parenthesization. (It’s what probably most implementations should define, but lots of people miss it and just define `show`.)

msmorgan:
@Melvar I suspected that might be the case.

Hypothetically, is this possible?

```idris

module Prelude.Show
{- ... -}
Cast a String => Show a where
  %overridable
  show = cast

----

module Prelude.Str

Show String where
  show cs = strCons '"' (showLitString (cast cs) "\"")

```

PR #4012, opened 2017-08-21
Fix return check from idrnet_getaddrinfo

### LINKED ISSUE #4001: Rewrite doesn't work in proofs involving case statements
# Steps to Reproduce
Assume we have this program
```
%default total

data Foo = Bar | Baz

foo1 : Foo -> Foo
foo1 Bar = Baz
foo1 Baz = Bar

foo2 : Foo -> Bool
foo2 f = case foo1 f of 
  Bar => True
  Baz => False

fooProof : (f: Foo) -> (f=Bar) -> foo2 f = False
fooProof f prf = rewrite prf in ?fooProof_rhs
```
# Expected Behavior
`fooProof`'s type reduces to `False = False`.

# Observed Behavior
```
When checking an application of function rewrite__impl:
        Type mismatch between
                f = Bar (Type of prf)
        and
                False = Bar (Expected type)
        
        Specifically:
                Type mismatch between
                        f
                and
                        False
```

I understand that in this case this can be solved by simply pattern matching on `f` or perhaps calling `with ((foo1 f) proof ff` and doing some replaces, but in more complex cases this becomes very verbose.

### LINKED ISSUE #4011: Network.Socket `connect` swallows output
# Steps to Reproduce
```
makeRequest : IO ()
makeRequest = do
    case !(socket AF_INET Stream 0) of
      Left err   => putStrLn "Error"
      Right sock =>
        do
          putStrLn "Foom"
          connect sock (Hostname "host") 80
          putStrLn "Bar"
```

# Expected Behavior
```
Foom
Bar
```

# Observed Behavior
```
Foom
```

What codepath could Idris possibly take to not print "Bar"? Ofcourse "host" doesn't resolve to a valid host, but that's besides the point.

melted:
`getaddrinfo` returns non-zero on error, not -1.

Fixes #4011

### DIFF
--- a/rts/idris_net.c
+++ b/rts/idris_net.c
@@ -82,7 +82,7 @@ int idrnet_bind(int sockfd, int family, int socket_type, char* host, int port) {
 int idrnet_connect(int sockfd, int family, int socket_type, char* host, int port) {
     struct addrinfo* remote_host;
     int addr_res = idrnet_getaddrinfo(&remote_host, host, port, family, socket_type);
-    if (addr_res == -1) {
+    if (addr_res != 0) {
         return -1;
     }
 


melted:
Huh, Travis passed, with fast builds and all?

PR #4014, opened 2017-08-22
Travis: Don't wipe the whole cache because of a miss

melted:
Instead run with --upgrade-dependencies and repack the cabal dirs if there's new stuff. As the differences rarely are big, this will save much time.

In case cabal fails, we wipe the cache and try again. As it most likely will fail in the dependency resolution it will hopefully fail fast, and even if not, clearing the cache ensures the next run will start with a blank slate.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,6 @@ matrix:
 cache:
   directories:
     - $HOME/.cabsnap
-    - $HOME/.cabal/packages
     - $HOME/.stack
 
 before_cache:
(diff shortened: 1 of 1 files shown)


ahmadsalim:
I think you may need `--force-reinstalls` to upgrade transitive dependencies.

melted:
Yeah, there is some needlessly complicated logic there because I wanted to wipe the cache if it failed at first and then start over. I think that is unnecessary, it doesn't take a lot of time to save it again, we can do that every time.

melted:
I did something wrong with that update. I'll close this for now.

PR #4016, opened 2017-08-23
add startWith function for prepending a finite number of elements to a Stream

msmorgan:
This function inserts elements from a Foldable at the beginning of a Stream.

I was going to name this `prefix`, but that's a keyword. The following names were also candidates:

* `prepend`
* `insert`
* `prefixBy`
* `beginWith`

Should I rearrange the imports? I was going to alphabetize them, but I wasn't sure that would be appropriate.

### DIFF
--- a/libs/contrib/Data/Stream/Extra.idr
+++ b/libs/contrib/Data/Stream/Extra.idr
@@ -1,0 +1,10 @@
+module Data.Stream.Extra
+
+%access public export
+%default total
+
+||| Produce a Stream that is prefixed by elements from a Foldable
+||| @ pfx the Foldable containing elements to prepend
+||| @ stream the Stream to prepend the elements to
+startWith : Foldable t => (pfx : t a) -> (stream : Stream a) -> Stream a
+startWith pfx stream = foldr (\x, xs => x :: xs) stream pfx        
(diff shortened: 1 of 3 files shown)


melted:
Thanks!

The name seems fine to me.

msmorgan:
At the request of @ahmadsalim (via IRC) I have moved this function to **contrib**.

I chose the module `Data.Stream.Extra`. There is not currently a `Data.Stream` package, it would probably belong in **base** if there were.

libs/contrib/Data/Stream/Extra.idr
```diff
@@ -3,7 +3,7 @@ module Data.Stream.Extra
 %access public export
 %default total
 
-||| Produce a Stream that is prefixed by elements from a Foldable
+||| Insert elements from a Foldable at the start of an existing Stream
```
msmorgan: This wording seems a bit clearer.

msmorgan:
@melted Should I have squashed my changes into a single commit as I updated this PR? I had been planning to do it once it was was confirmed that the function name and module choice were approved.

Also, should my name be added to `CONTRIBUTORS` now, or is there some threshold I still need to meet? I can take care of it with my next PR if that's the normal procedure.

ahmadsalim:
@msmorgan There is no threshold, you are welcome to add your name.

PR #4018, opened 2017-08-24
add Interfaces.Zippable to contrib

msmorgan:
This module defines the interface `Zippable (t : Type -> Type)`, which
abstracts `zipWith` and related operations into an interface.

Included are implementations of Zippable for List, Stream, and Vect, in
addition to generalizations for Maybe and Either.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -151,6 +151,7 @@ Michael Gilliland
 Michael Hensler
 Michael Maloney
 Michael McGirr
+Michael Morgan
 Michael Thompson
 Michael Weiss
 Miikka Koskinen
--- a/libs/contrib/Interfaces/Zippable.idr
+++ b/libs/contrib/Interfaces/Zippable.idr
@@ -1,0 +1,54 @@
+module Interfaces.Zippable
+
+import Data.Vect
+
+%access public export
+%default total
+
+||| The Zippable interface describes the operation of combining
+||| two parameterized container elementwise using a provided function
+||| into a new container with the function's return values.
+interface Zippable (t : Type -> Type) where
+  ||| Combine two containers elementwise using some function.
+  ||| If the container type is Sized, the following should hold:
+  |||     size (zipWith f xs ys) = minimum (size xs) (size ys)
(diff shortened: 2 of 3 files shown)


clayrat:
Isn't this just a specific flavour of Applicative ("Naperian", if I recall correctly)?

msmorgan:
@clayrat Quite possibly. I approached this from a refactoring perspective, rather than a mathematical one. My goal was to do something about this repetition:

```
*Data\Vect> :printdef zip
Prelude.List.zip : List a -> List b -> List (a, b)
zip = zipWith (\x, y => (x, y))
Prelude.Stream.zip : Stream a -> Stream b -> Stream (a, b)
zip = zipWith (\x, y => (x, y))
Data.Vect.zip : Vect n a -> Vect n b -> Vect n (a, b)
zip = zipWith (\x, y => (x, y))
```

melted:
Thanks for the PR!

jfdm:
@msmorgan Thanks for the PR, having a 'Zippable' interface can be useful for generic programming with zippable things.

One thing to point out though with your comment about addressing repetition. The call you mentioned (`:printdef zip`) is returning valid results, you have asked Idris to print the definition for things called zip.  Obviously, Idris can discern between them at compile time by looking at their types, much like haskell, and will select the correct one. However, if you wish to print out a specific zip implementation, you will have to say which implementation you want printed. You can do this using namespace qualification. For example:

```
*Data/Vect> :printdef Vect.zip
zip : Vect n a -> Vect n b -> Vect n (a, b)
zip = zipWith (\x, y => (x, y))
```

msmorgan:
@jfdm Thanks, but for the above example I intended that the REPL show all the `zip`s, in order to demonstrate that the bodies are defined the same way aside from the version of `zipWith` they call. I meant to imply that if there was a common `zipWith` function defined for an interface, `zip` could be defined once in terms of that function instead of separately for each version of `zipWith`.

jfdm:
Thanks, I understood the rational behind the Zippable interface, but thought I would chime in that point anyway.

PR #4022, opened 2017-08-25
remove Interfaces.Zippable in favor of Interfaces.Pairwise

msmorgan:
`Pairwise` is a sub-interface of `Functor` with a single operator `(</>)`, which has
the same signature as `Prelude.Applicative.(<*>)`. The difference is
that a Pairwise functor must apply operations element-by-element. Many
Applicative implementations already work in this way, but notably List does
not (it uses the Cartesian product).

The name "Naperian" was suggested for this concept, but I was unable to find
any accessible literature indicating that this nomenclature would be correct.

Yes, I know I just contributed `Interfaces.Zippable` yesterday.


melted:
But what if somebody has built their startup on our Zippable technology? :-)

PR #4023, opened 2017-08-26
add Data.Pairs.Implicit for converting erased pairs

msmorgan:
The conversions are performed by the functions `evidence` and `element`.


melted:
Implicits can be confusing, but I guess the user should know what they do when they import a module called "Implicit". It shouldn't be a problem in `contrib`

There is deprecation warning on getProof and getWitness though.

msmorgan:
@melted That deprecation warning is only for the `DPair` versions of those functions are used (the new preference is `fst` and `snd` for DPairs).

melted:
ok, then I'll merge

PR #4024, opened 2017-08-26
remove Interfaces.Zippable in favor of Interfaces.Correlative

msmorgan:
Replaces #4022 

Correlative is a sub-interface of Functor with a single operator `(</>)`,
with the same signature as `Prelude.Applicative.(<*>)`. The difference is
that a Correlative functor must apply operations element-by-element to matching
elements. Many Applicative implementations already work in this way, but
notably List does not (it uses the Cartesian product).

The name "Naperian" was suggested for this concept, but I was unable to find
any accessible literature indicating that this nomenclature would be correct.

This is a better name, I think.


msmorgan:
[Here's a branch with a similar but slightly different idea.](/msmorgan/Idris-dev/commit/bb0a0ceccead6afaafac531d80e7022b959aa585) Is it possible to implement one of these in terms of the other?

melted:
@msmorgan Should I merge, or do you want to update it?

msmorgan:
@melted I haven't come up with a way to meld those two ideas together quite yet. I think `Correlative` (this PR) could be defined in terms of `Correlatable` (the link in my previous comment), but it probably requires a third interface, something like this:

```idris
interface Functor f => Reducible (f : Type -> Type) where
  reduce : (pred : a -> Bool) -> f a -> f a

Reducable List where
  reduce = Prelude.List.filter
```

It would probably be hard to define this for `Data.Vect` (though `Data.Vect.Quantifiers.All` may help).

Anyway, it's usable how it is. It's ready to merge, but there might be improvements to come.

ahmadsalim:
@msmorgan So this is like `Applicative` but without `pure`, right?

msmorgan:
@ahmadsalim Basically, yes. But the implementation of `</>` doesn't necessarily need to work the same way as `<*>`.

msmorgan:
This could be implemented for `SortedMap ` (I intend to do this soon) but  `Applicative`  cannot be, since there is no sensible definition for `pure`.

zenntenn:
This means it's basically the Idris version of Haskell's Data.Functor.Apply, right?

msmorgan:
@zenntenn It appears to be. Do you think I should add the operators `</` and `/>` ?

PR #4026, opened 2017-08-27
Package name abstraction & naming restrictions

vincenthz:
I was looking at the package code and strenghtening the package name from a `String` seems to be a good idea. This PR changes the following:

* Introduce a `PkgName` newtype on `String`
* only accept package name to be `[a-z][-_a-z0-9]*`
* change the ipkg parser to not accept a `filename` for package name, removing the support for quoted string that seems only used in one of the test suite (also fixed).

The choice of supported package name is very conservative since package name are used as filename in multiple places. it's conservative for the following reasons:

* avoid filesystem case insensitive issues by only supporting lowercase
* avoid filesystem unicode normalization issues by not supporting any unicode

Since it brings new restrictions, it could break existing code/package, so this is more for RFC for package naming than a finished PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@
 + The C backend now pass arguments to the C compiler in the same order
   as they were given in the source files.
 
+## Packaging Updates
++ Package names now only accept a restrictive charset of letters, numbers and the `-_` characters.
+  Package names are also case insensitive
+
 # New in 1.1.1
 
 + Erasure analysis is now faster thanks to a bit smarter constraint solving.
--- a/src/Idris/IBC.hs
+++ b/src/Idris/IBC.hs
@@ -144,9 +144,9 @@ loadIBC reexport phase fp
                             addImported reexport fp
 
 -- | Load an entire package from its index file
-loadPkgIndex :: String -> Idris ()
+loadPkgIndex :: PkgName -> Idris ()
 loadPkgIndex pkg = do ddir <- runIO getIdrisLibDir
-                      addImportDir (ddir </> pkg)
+                      addImportDir (ddir </> unPkgName pkg)
                       fp <- findPkgIndex pkg
                       loadIBC True IBC_Building fp
 
--- a/src/Idris/Imports.hs
+++ b/src/Idris/Imports.hs
@@ -18,7 +18,7 @@ import IRTS.System (getIdrisLibDir)
 
 import Control.Applicative ((<$>))
 import Control.Monad.State.Strict
-import Data.Char (isDigit, isAlpha, toLower)
+import Data.Char (isAlpha, isDigit, toLower)
 import Data.List (isSuffixOf)
 import System.Directory
 import System.FilePath
--- a/src/Idris/Imports.hs
+++ b/src/Idris/Imports.hs
@@ -18,7 +18,7 @@ import IRTS.System (getIdrisLibDir)
 
 import Control.Applicative ((<$>))
 import Control.Monad.State.Strict
-import Data.Char (isDigit, isLower)
+import Data.Char (isDigit, isAlpha, toLower)
 import Data.List (isSuffixOf)
 import System.Directory
 import System.FilePath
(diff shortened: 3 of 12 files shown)


vincenthz:
fixed the stylize issue. appveyor issue seems to be unrelated to this change (trifecta failing to build)

ahmadsalim:
@vincenthz Did this ever become completed? I think it was tagged in progress and we missed that you completed it.

vincenthz:
@ahmadsalim yes it is completed, it seems to be working and the test failures were AFAIS only spurious build that take too long to complete or some unrelated javascript lock error

I think the only thing missing is a decision whether the restriction on package name is a good thing, and if it is, whether the restrictions I've chosen is appropriate.

ahmadsalim:
I think restrictions on package names is good in general, but maybe @jfdm, @david-christiansen  or @edwinb can voice their opinions. I would prefer if the javascript tests did not fail.

Maybe @melted knows more about how to fix this?

vincenthz:
w.r.t the javascript failure, I don't think it's related to this PR, since I've seen the same failure in different PR; I could rebase to see if that's still happening..

melted:
I don't mind restrictions, but I think upper case A-Z should be allowed, it would break existing packages to remove it and it's not that hard to avoid trouble with them.

ahmadsalim:
@vincenthz Yeah, please rebase if possible. I agree with @melted about upper case letters.

vincenthz:
@melted related to capital letters, I agree that it should be allowed in package name, but I think capital letters should not appears on the filesystem. The 3 solutions I can think of:

* denies capital letters in package name (the simplest and what's used here)
* accept capital letters and normalize the package name (e.g. to lower cases)
* create an encoding for package name on filesystem

I think my goal here is to have `Pkg`, `pkg`  and `pKg` refer to the same package, and the second solution seems the best.

david-christiansen:
Do we really want to rule out package names in non-English languages? I see the reasoning here, but what's wrong with e.g. "økonomistyring"?

vincenthz:
@david-christiansen in spirit I think there's nothing wrong with a unicode package name, and it would be nice to have, but it's also a big can of worm:

* filesystem normalisation issue (NFC, NFD)
* character that looks the same for example `-` and `–` (http://www.unicode.org/Public/security/latest/confusables.txt)

unless there some existing package name relying on unicode, it's easier to not allow unicode name now and relax later when having worked out the details above

sjmackenzie:
Just a point to mention explicitly, `nix` doesn't support attribute names beginning with number. This creates a problem when referring to haskell packages starting with numerics in nix. Nor does `nix` supports utf8. Just adding a different point of view.

ahmadsalim:
@vincenthz I think allowing capital letters and normalizing the name might be an OK solution.

ahmadsalim:
While I in principle agree with @david-christiansen that having non-ASCII characters might be nice. I think it could be easy to accommodate as @vincenthz mentions in the future.
So I think it will be acceptable to merge the PR with a stricter restriction and expand it later.

ahmadsalim:
@vincenthz Can you please add the file name normalization procedure, and then we can merge it?

ahmadsalim:
Also an update to the `CHANGELOG` about the restriction please.

vincenthz:
@ahmadsalim I've updated the PR:

* package name now accept capital letters
* package name are kept as is but normalized to lowercase for Equality comparaison and when part of a filepath. I'ld think this is more user friendly when printing the package name (e.g. as part of a terminal message)

ahmadsalim:
Thanks for the PR.
Feel free to update CONTRIBUTORS if you want to in another PR.

sjmackenzie:
I'd just like to confirm that the `$filename.ipkg` can be a different name to the `package $packagename`

Why? Being able to set the file name to (i.e.) `agent.ipkg` means a build system can load the path and knows the entry point to idris agent code would be `agent.ipkg`. If the `$filename.ipkg` is the same as $packagename, then build system won't know what to look for. Yes one could load `*.ipkg` but it feels a bit dirty, besides having a convention of `agent.ipkg` is more suitable for users.

PR #4027, opened 2017-08-28
add opt combinator to Text.Lexer

msmorgan:
This combinator recognises a provided lexer, or recognises no input if
the lexer fails.

Also corrected an Americanism that I was responsible for introducing.


jfdm:
Thanks for the contributions. This looks fine to me.

PR #4028, opened 2017-08-28
Text.Lexer: introduce new combinators for surrounded sequences and escapes

msmorgan:
This pull request adds several useful combinators to `Text.Lexer`. It also fixes a bug in the behavior of `charLit`. The following lexers and combinators are included: 

* `surround`, `quote` - Recognise a sequence of sub-lexers between starting and ending boundaries.
* `escape` - Recognise a sub-lexer immediately following an escape character (such as `\`).

### DIFF
--- a/libs/contrib/Text/Lexer.idr
+++ b/libs/contrib/Text/Lexer.idr
@@ -164,21 +164,35 @@ export
 symbol : Lexer
 symbol = some (pred (\x => not (isAlphaNum x) && not (isSpace x)))
 
-strChar : Lexer
-strChar = (is '\\' <+> any) <|> isNot '"'
+||| Recognise zero or more occurrences of a sub-lexer between
+||| delimiting lexers
+export
+surround : (start : Lexer) -> (end : Lexer) -> (l : Lexer) -> Lexer
+surround start end l = start <+> manyTill l end
+
+||| Recognise zero or more occurrences of a sub-lexer surrounded
+||| by the same quote lexer on both sides (useful for strings)
+export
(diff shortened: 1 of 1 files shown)


msmorgan:
Let me know if I need to squash these, or if you need any of them removed.

Also, does anyone think it would be more useful if `escaped` and `escapeSequence` accepted an escape *Lexer*, rather than an escape character?

msmorgan:
This PR encompassed two distinct chunks of functionality, so this portion was split with #4029.

melted:
Hmm, I don't know about breaking changes, is that really necessary?

msmorgan:
@melted The breaking changes were moved to the other PR.

msmorgan:
I've updated the names and semantics a bit.

There's a single `escape` function now that just accepts a `Char` and a `Lexer` as arguments.

The other two functions were renamed to avoid confusion with the `between` combinator in `Text.Parser`. They are now called `surround` and `quote`.

jfdm:
Thanks for the contribution. The PR looks okay, however, it would be nice if a version of the `strChar` lexer was retained rather than removed. People may, or may not find `strChar` useful but I would rather keep it, or mark it as deprecated in case people are using it currently. Remember `strChar` is part of a public release, and although `contrib` is (by design) subject to change, it would be better if we ensure smooth changes.

msmorgan:
@jfdm `strChar` wasn't exported.

jfdm:
I thought it was. 

Okay, thanks for the contribution.

PR #4031, opened 2017-08-28
Text.Lexer: add quantified lexer combinators

msmorgan:
Add new combinators for counted sub-lexer repetition.

This includes the combinators `atLeast`, `atMost`, `between`, and `exactly`.

The `choice` combinator was also generalized to accept empty lists.

### DIFF
--- a/libs/contrib/Text/Lexer.idr
+++ b/libs/contrib/Text/Lexer.idr
@@ -113,11 +113,11 @@ export
 many : Lexer -> Recognise False
 many l = opt (some l)
 
-||| Recognise the first matching lexer from a non-empty list.
+||| Recognise the first matching lexer from a Foldable. Always consumes input
+||| if one of the options succeeds. Fails if the foldable is empty.
 export
-choice : (xs : List Lexer) -> {auto ok : NonEmpty xs} -> Lexer
-choice (x :: [])          = x
-choice (x :: xs@(_ :: _)) = x <|> choice xs
+choice : Foldable t => t Lexer -> Lexer
+choice xs = foldr Alt Fail xs
 
 ||| Recognise many instances of `l` until an instance of `end` is
 ||| encountered.
--- a/libs/contrib/Text/Lexer.idr
+++ b/libs/contrib/Text/Lexer.idr
@@ -127,6 +127,34 @@ export
 manyTill : (l : Lexer) -> (end : Lexer) -> Recognise False
 manyTill l end = end <|> opt (l <+> manyTill l end)
 
+||| Recognise a sub-lexer at least `min` times. Consumes input unless
+||| min is zero.
+export
+atLeast : (min : Nat) -> (l : Lexer) -> Recognise (min > 0)
+atLeast Z l       = many l
+atLeast (S min) l = l <+> atLeast min l
+
+||| Recognise a sub-lexer at most `max` times. Not guaranteed to
+||| consume input.
+export
+atMost : (max : Nat) -> (l : Lexer) -> Recognise False
+atMost Z _     = Empty
+atMost (S k) l = atMost k l <+> opt l
+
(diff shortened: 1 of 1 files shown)


msmorgan:
Yes, I've used the name `between` again. It's semantically very different from the `between` in Text.Parser. However, I think the form `between 3 5 digit` to mean "between 3 and 5 digits" is quite readable and makes up for it.

jfdm:

jfdm:
So having looked at this, I am unsure as to why we need quantified lexers. I always thought that such quantification should be dictated by the parser and not by the lexer. I am open to being convinced otherwise.

msmorgan:
```idris
unicodeEscape : Lexer
unicodeEscape = escape '\\' (exactly 4 digit)
```

jfdm:
For me, these examples are still not convincing enough to warrant inclusion of quantified lexers. The quantification presented should be handled by the parser.

msmorgan:
@jfdm They're useful when parsing with a layered (multiple pass) approach. Consider the following:

```idris

||| Remove the first `n` characters from a string, or return the empty string if
||| it's too short.
drop : (n : Nat) -> String -> String
drop n str = substr n (length str) str

||| Remove the first and last `n` characters from a string, or return the empty
||| string if it's too short.
shrink : (n : Nat) -> String -> String
shrink n str = reverse (drop n (reverse (drop n str)))

||| Convert input that was recognised as a "stringLit" into the String it
||| represents.
stringLitValue : String -> String
stringLitValue str = expandEscapes (shrink 1 str)
  where
    {- A real version of this function might use Text.Parser to understand the
       escape sequences. -}
    expandEscapes : String -> String
    expandEscapes = id

starRating : Lexer
starRating = surround (is '[') (is ']')
                      (between 1 5 (is '*'))

||| Convert input that was recognised as a `starRating` into its numeric rating.
starRatingValue : String -> Nat
starRatingValue str = length (shrink 1 str)

data MyToken
  = MTStringLit String
  | MTStarRating Nat
  | MTIgnore

myTokenMap : TokenMap MyToken
myTokenMap
  = [ (Text.Lexer.stringLit, MTStringLit . stringLitValue)
    , (starRating, MTStarRating . starRatingValue)
    , (any, MTIgnore)
    ]

```

msmorgan:
This style of approach allows distinct contexts of a grammar to be written as separate parsers (and potentially re-combined). It also minimizes the amount of different tokens the main parser has to handle.

Another note is that the above example is a reduction of the approach I'm talking about. Normally, instead of `stringLitValue` going straight from a string to a string, it would probably output e.g. a `StringLitResult`, which could have a fail state.

Then, the main parser would be freed from specifically processing string literals, and only needs to consider "is this a valid string literal or not".

This is most useful for things that are considered a "literal" within a particular grammar, but it could be helpful in other situations.

ahmadsalim:
@jfdm I think quantification is fairly useful for making custom literals, like e.g. regex or url literals, but I will let it up to you to decide.

jfdm:
@msmorgan I am sorry but your examples, are still not convincing enough for me. However, I will mull on the examples @ahmadsalim has suggested and decide whether I am capable of deciding on this PR or not.

msmorgan:
Text.Lexer is basically a regular expression library. Every lexer and combinator that currently exists can be expressed as an equivalent regex. (This is a fact I'm planning to use to improve the documentation after everything has been merged.)

I think we'd be remiss not to add an equivalent for `{n,m}` quantities. For instance, `starRating` above is `/\[\*{1,5}\]/`.

If you think they are too dangerous/abusable to include directly in the `Text.Lexer` module, I could relocate them to a `Text.Lexer.Quantity` module until they've proved their worth.

edwinb:
Given that the examples show tokenising, I would personally prefer to do this sort of thing in a lexer rather than a parser. So I think these are nice to have.

PR #4034, opened 2017-08-30
Text.Lexer: get a little fancy with default implicits

msmorgan:
`strLit`, `charLit`, were given new implict arguments defining their
quotations and valid escape sequences.

`intLit` has a new implicit argument denoting the valid prefix signs.

Adds a new function `escaped`, which is like `escape` but the character is
implicit.

### DIFF
--- a/libs/contrib/Text/Lexer.idr
+++ b/libs/contrib/Text/Lexer.idr
@@ -179,28 +179,48 @@ export
 quote : (q : Lexer) -> (l : Lexer) -> Lexer
 quote q l = surround q q l
 
-||| Recognise an escape character (often '\\') followed by a sub-lexer
+||| Recognise an escape character (such as `'\\'`) followed by a sub-lexer.
 export
-escape : (esc : Char) -> Lexer -> Lexer
-escape esc l = is esc <+> l
+escape : (c : Char) -> Lexer -> Lexer
+escape c l = is c <+> l
 
-||| Recognise a string literal, including escaped characters.
-||| (Note: doesn't yet handle escape sequences such as \123)
+||| Recognise a sub-lexer escaped by `'\\'`.
+||| Prefer using `escape` to changing the implicit argument of
+||| this function.
+||| @ c The escape character
 export
-stringLit : Lexer
-stringLit = quote (is '"') (escape '\\' any <|> any)
+escaped : {default '\\' c : Char} -> Lexer -> Lexer
+escaped {c} l = escape c l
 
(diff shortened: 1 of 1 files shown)


msmorgan:
My overall thought is that opinionated, narrow versions of `stringLit`, `charLit`, and `intLit` don't belong in `Text.Lexer` at all (sorry @edwinb). I decided to try this instead of removing them.

jfdm:
@msmorgan You can have both a generic lexer that accepts arguments, and one with a default set of arguments presented. Personally, that is my preferred option, it is also what you can achieve with default arguments but made much more explicit.

Also please bear in mind that the original constructions are not necessarily 'opinionated'. Sometimes libraries presented in contrib are best effort for what they are required to do. When one opens such libraries for general use the decisions might not reflect what can actually be achieved. After all, Idris is a research project.

msmorgan:
What I meant by opinionated, is that most languages will have a "string literal", and the majority of the time it won't look like this one. I've begun several lexers at this point with `%hide Text.Lexer.stringLit`.

Also consider how the functions in this commit come up using `:doc`, compared to how they did before.

msmorgan:
Depending on how this issue is resolved, I'd like to add `lineComment` and `blockComment`. If this is deemed an appropriate use of default implicits, I'll use the following types:

```idris
||| A line comment. Consumes from the start sequence until a newline or EOF.
lineComment : {default (exact "--") start : Lexer} -> Lexer

||| A block comment. Supports nesting, but must be balanced.
blockComment : {default (exact "{-") start : Lexer} -> {default (exact "-}") end : Lexer } ->
               Lexer
```

jfdm:
Line/Block comment levers should not have default arguments, it is making an unnecessary language design design. It would be simpler to leave the choice of deliminating lexers to the module user.

edwinb:
I'm not really keen on default implicits myself, they feel like a bit of a hack, and indeed that's what they were at first because auto implicits weren't introduced for a while. So while it'd be nice for these to exist, it seems to me that they'd be better as separately named functions.

msmorgan:
Okay. You mean separately named functions with explicit arguments instead of this approach, right?

iamrecursion:
My instinct says to have `lineComment` and `blockComment` without default args, but perhaps also include `idrisLineComment` and `idrisBlockComment`

ahmadsalim:
@msmorgan Yeah, please include new functions instead of using auto-implicits here.
I will close this PR, but feel free to open a new one!

PR #4036, opened 2017-08-31
Text.Lexer: positive and negative lookahead

msmorgan:
This required a new constructor for `Recognise`, called `Look`.

Exports new functions `look` and `unlook`.

### DIFF
--- a/libs/contrib/Text/Lexer.idr
+++ b/libs/contrib/Text/Lexer.idr
@@ -3,12 +3,13 @@ module Text.Lexer
 %default total
 
 ||| A language of token recognisers.
-||| The `consumes` flag is True is the recogniser is guaranteed to consume
-||| at least one character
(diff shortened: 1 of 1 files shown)


msmorgan:
Upon further reflection, I wasn't happy with the names I'd originally chosen, so I renamed the new identifiers to `Expect`, `expect` and `reject`. I think these are nice to read.

#### Usage examples
```idris
||| /[^A-Za-z]/
nonAlpha : Lexer
nonAlpha = reject alpha <+> any

||| /\\(?=[\r\n])/
newlineEscape : Lexer
newlineEscape = is '\\' <+> expect (oneOf "\r\n")
```

I've also added a second commit defining `non`, generalizing the form of `nonAlpha` in the above examples. That way, you can simply write `non alpha`.

jfdm:
The travis test failed due to a time out in a single build. As the Lexer module is not included in the tests, I think merging will be okay.

PR #4037, opened 2017-08-31
Text.Lexer: export a fail function

msmorgan:
Adds the function `fail`.

The expored API of this module currently allows the construction of
an always-failing `Lexer` in various ways. This function makes that
explicit.

The function `fail` has an implicit argument `{c}` like `Fail`, so it
can return both consuming and non-consuming recognisers.

I'm not certain an always-failing `Recognise False` is at all useful in a public
API, so let me know if I should change the return type to `Lexer`.

### DIFF
--- a/libs/contrib/Text/Lexer.idr
+++ b/libs/contrib/Text/Lexer.idr
@@ -39,6 +39,11 @@ export
 (<|>) : Recognise c1 -> Recognise c2 -> Recognise (c1 && c2)
 (<|>) = Alt
 
+||| A recogniser that always fails.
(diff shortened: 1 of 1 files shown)


jfdm:

jfdm:
The travis test failed due to a time out in a single build. As the `Lexer` module is not included in the tests, I think merging will be okay.

jfdm:
One merge conflicts are resolved, I don't see any reason to block this from being merged.

jfdm:
Thanks.

PR #4039, opened 2017-08-31
Text.Lexer: case-insensitive variants of `is` and `exact`.

msmorgan:
I have named these `like` and `approx`, respectively, which I think reads nicely.

### DIFF
--- a/libs/contrib/Text/Lexer.idr
+++ b/libs/contrib/Text/Lexer.idr
@@ -70,6 +70,14 @@ export
 isNot : Char -> Lexer
 isNot x = Pred (/=x)
 
+||| Recognise a specific string
+export
+exact : String -> Lexer
+exact str with (unpack str)
+  exact str | [] = Fail -- Not allowed, Lexer has to consume
+  exact str | (x :: xs)
+      = foldl SeqEmpty (is x) (map is xs)
+
 ||| Recognise a lexer or recognise no input. This is not guaranteed
 ||| to consume input
 export
@@ -189,14 +197,6 @@ digit = pred isDigit
 export
 digits : Lexer
 digits = some digit
-
-||| Recognise a specific string
-export
-exact : String -> Lexer
-exact str with (unpack str)
-  exact str | [] = Fail -- Not allowed, Lexer has to consume
-  exact str | (x :: xs)
-      = foldl SeqEmpty (is x) (map is xs)
 
 ||| Recognise a single hexidecimal digit
 export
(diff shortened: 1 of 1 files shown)


jfdm:

libs/contrib/Text/Lexer.idr
```diff
@@ -48,6 +48,40 @@ export
...
 isNot x = Pred (/=x)
 
+||| Recognise a character case-insensitively
+like : Char -> Lexer
+like x = Pred (\y => toUpper x == toUpper y)
+
+export
+is_ci : Char -> Lexer
+is_ci = like
+%deprecate Text.Lexer.is_ci "Please use `like`."
```
jfdm: The function `is_ci` and the thing it deprecates (`like`)are new additions. Would it not be worth just removing `is_ci`?

libs/contrib/Text/Lexer.idr
```diff
@@ -48,6 +48,40 @@ export
...
+export
+approx : String -> Lexer
+approx str with (unpack str)
+  approx str | [] = Fail -- Not allowed, Lexer has to consume
+  approx str | (x :: xs)
+      = foldl SeqEmpty (like x) (map like xs)
+
+export
+exact_ci : String -> Lexer
+exact_ci = approx
```
jfdm: ibid

libs/contrib/Text/Lexer.idr
msmorgan: I see why this might be confusing. I was trying to accommodate new users of the Lexer library, while still preferring names that maximize readability.

I prefer the names `like` and `approx`. They flow really well when reading lexer code. But they will be difficult to find for someone asking "is there a case insensitive version of `exact`?"

If you think the documentation comments on `like` and `approx` are enough to make them easy to find (via `:apropos`), then these can be removed. It will just be harder to find the case-insensitive functions via `browse`.

jfdm: TBH, i found `is_ci` to be confusing enough not to think `ci` implied case insensitive. I think function 'discovery' will be a issue regardless of name unless the name is precisely: `is_but_case_insensitive` or you place it under a different namespace e.g. `CaseInsensitive.is`. Both these options are not ideal. 

I think the names `like` and `approx` with sufficient documentation will be fine.

msmorgan: Fair enough. I'll take them out.

jfdm:
One the merge conflicts have been addressed, I do not see why this should not be merged.

PR #4040, opened 2017-08-31
Interfaces.Correlative: add the operators </ and />

msmorgan:
These correspond with the `<*` and `*>` operators defined in
Prelude.Applicative.


ahmadsalim:
Thanks

PR #4042, opened 2017-09-02
Text.Lexer: fix behavior of `surround`

msmorgan:
`manyTill` is a `Recognise False`. It will recognise either nothing,
or any number of `l` followed by an `end`. This makes it unsuited for
`surround`, because the end character is required.

### DIFF
--- a/libs/contrib/Text/Lexer.idr
+++ b/libs/contrib/Text/Lexer.idr
@@ -261,7 +261,7 @@ symbols = some symbol
 ||| delimiting lexers
 export
 surround : (start : Lexer) -> (end : Lexer) -> (l : Lexer) -> Lexer
-surround start end l = start <+> manyTill l end
+surround start end l = start <+> many (reject end <+> l) <+> end
 
 ||| Recognise zero or more occurrences of a sub-lexer surrounded
 ||| by the same quote lexer on both sides (useful for strings)


msmorgan:
A function similar to `manyTill` that always recognises at least its ending character, could be useful. Now, what to name it...

iamrecursion:
Some name bikeshedding: `manyAnd`, `manyThen`. I don’t _truly_ like either, however.

msmorgan:
@iamrecursion What do you think of this?

Deprecate `manyTill` in favor of `manyBefore`, and add `manyUntil`.

iamrecursion:
To me, `manyUntil x y` still implies ‘parse many `x` until a `y` is reached without consuming the `y`’. The same thing as `manyBefore`. This is probably because ‘till’ just expands to ‘until’ in my mind. 

Personally I like the combination of `manyUntil` and `manyThen`, but this is _very_ much in the region of bikeshedding here.

msmorgan:
@iamrecursion Consider currying though. What does `the (Lexer -> Lexer) (manyUntil a)` mean? To me it should mean that the argument `(manyUntil a)` accepts should be repeated until it finds `a`.

msmorgan:
This corresponds with the argument order of things like `Text.Parser.sepBy`, where the first argument is the "by".

msmorgan:
@iamrecursion Also, there are times when bikeshedding is appropriate. Designing a public API is one of those times. It's important not to get too carried away, but we're not even close.

iamrecursion:
Regarding bikeshedding you have a very good point. 

You’re definitely right on argument order considering currying. Consider it a late-night brain-fail on my behalf. 

What’re your thoughts on the pairing of `manyUntil` and `manyThen`?

msmorgan:
With `manyUntil` equivalent to `flip manuTill`, and `manyThen` is my new function? I like them both!

A few other incomplete ideas I was tossing around in case they inspire you: `repeatUntil`, `followedBy, `stopBefore`, `stopAfter`.

iamrecursion:
I'm not sure if `manyUntil` should be `flip manyTill` or just a wholesale replacement for `manyTill`. Despite my argument order above, I think I intended it to just be an effective rename to better match `manyThen` (which would be your new function, yes).

Probably not what you're intending, but `followedBy` put me in mind of an infix operator with low precedence that would allow one to write an expression similar to `many x followedBy y`, but that's not all that useful other than for the sake of easy reading of lexers and parsers.

msmorgan:
@iamrecursion I think I got it the two confused. I've settled on the following, unless you have any objections:

```idris
manyUntil : (stopBefore : Lexer) -> (l : Lexer) -> Recognise False
manyThen : (stopAfter : Recognise c) -> (l : Lexer) -> Recognise c
%deprecate Text.Lexer.manyTill "Prefer `manyThen`. The argument order is flipped."
```

iamrecursion:
Agreed on both of those! I think the order for `manyUntil` vs. the order for `manyTill` makes more sense due to how it reads as an expression in a curried context. 

Do you mean to have `manyThen` in the deprecation message? I was understanding the replacement to be `manyUntil`.

msmorgan:
@iamrecursion The equivalent is actually `manyTill l end = manyThen (opt end) l`.

iamrecursion:
Ah, yes. This is why I shouldn’t be commenting on PRs so late at night. I’ve not got any problems with the proposed changes above! It definitely makes sense to me from a user perspective.

msmorgan:
Wait a moment. I was wrong about the equivalence again. It's the same as `manyUntil end l <+> opt end`. I'm pretty sure I got it right this time.

iamrecursion:
Excellent stuf!

msmorgan:
Closing this. #4045 includes this fix.

PR #4045, opened 2017-09-03
Text.Lexer: deprecate `manyTill` and add more predictable replacements

msmorgan:
There is no exact replacement in this PR, but the behavior of `manyTill` is
inconsistent with what's expected from a lexer library that supports
lookahead. It consumes its terminator, but it succeeds if it never sees its
terminator and the input ends instead.

There are three new functions acting as more predictable versions of `manyTill`:
* `lineComment`, which recognises a given lexer and consumes until EOL or EOF.
* `manyTill`, `manyThen`: These new functions utilise lookahead to non-greedily consume a sub-lexer
until an endpoint lexer is recognised. `manyUntil` does not consume the end lexer, and `manyThen` does.

As a bonus, includes the lexer functions `newline`, `newlines`, and `blockComment`. `blockComment` works like `surround`, but supports balanced nesting.

### DIFF
--- a/libs/contrib/Text/Lexer.idr
+++ b/libs/contrib/Text/Lexer.idr
@@ -291,6 +291,17 @@ export
 spaces : Lexer
 spaces = some space
 
+||| Recognise a single newline sequence. Understands CRLF, CR, and LF
+export
+newline : Lexer
+newline = let crlf = "\r\n" in
+              exact crlf <|> oneOf crlf
+
+||| Recognise one or more newline sequences. Understands CRLF, CR, and LF
+export
+newlines : Lexer
+newlines = some newline
+
 ||| Recognise a single non-whitespace, non-alphanumeric character
 export
 symbol : Lexer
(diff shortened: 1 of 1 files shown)


msmorgan:
Depends on/replaces #4042 depending on how you look at it. Discussion of these changes occurred there.

Tagging @iamrecursion

ahmadsalim:
@msmorgan I am tagging @edwinb because of the breaking change.

msmorgan:
Is a deprecation a breaking change? My bad.

libs/contrib/Text/Lexer.idr
```diff
@@ -88,13 +88,30 @@ choice : (xs : List Lexer) -> {auto ok : NonEmpty xs} -> Lexer
...
+manyThen stopAfter l = manyUntil stopAfter l <+> stopAfter
+
 ||| Recognise many instances of `l` until an instance of `end` is
 ||| encountered.
 |||
 ||| Useful for defining comments.
 export
 manyTill : (l : Lexer) -> (end : Lexer) -> Recognise False
 manyTill l end = end <|> opt (l <+> manyTill l end)
+%deprecate manyTill "Prefer `manyUntil` and `manyThen`. Argument order is flipped. \
```
jfdm: Idris supports multiline strings using the following syntax: `""" """`. It would be preferable if such syntax is used.

msmorgan: `%deprecate` doesn't seem to work with triple quotes. Using them was my initial thought, too, but I couldn't the compiler it to accept it.

jfdm: Urm I thought it did... This must be a parser restriction...I am tempted to say reduce the statement so that it just points to the alternatives, or fix the parser....

msmorgan: I'll trim down this message once I've added `lineComment` (coming soon).

PR #4046, opened 2017-09-03
Enable test interactive017

steshaw:
Just happened to notice that `interactive017` wasn't being run. I've added it to `TestData.hs` and `chmod +x` a couple of files to make it work.


melted:
Thanks!

I wonder if we shouldn't tweak the test though. As it is now it always runs both node and C backend stuff. It would be more in line with the other tests if the run script checked if it was passed `--codegen node` and run node scripts  (they would have to have the same output as the C scripts).

steshaw:
The idea of this pull request is just to enable the test. I see what you mean about the test being different to other tests.

melted:
I'm mostly worried that it will fail in an environment without node even if run without "--codegen node".

steshaw:
That makes sense. I've disabled the interactive017 again on my branch (just commented out in `TestData.hs`). I'm currently working on a solution for #1209 (where I'd added a interactive018 and noticed that interactive017 wasn't enabled).

PR #4052, opened 2017-09-04
Text.Lexer: export `non`

msmorgan:
`non` was missing its `export` modifier.


jfdm:
The travis build stalled on the stylish check for some reason. All the other tests passed. Once, the check has been run this can be merged.

PR #4053, opened 2017-09-04
Text.Lexer: fix `reject` behavior at end-of-input

msmorgan:
Previously, `reject` was implemented using `expect`, and inverting the
predicate used with `Pred`. This means that `reject` would fail if it's the
final lexer in the input, instead of succeeding like it should.

To fix this, the constructor `Expect` was renamed to `Lookahead`, with a new
boolean argument indicating positive or negative lookahead. This boolean is
processed in `scan`. Now, `reject` succeeds if the sub-lexer fails for any
reason, including end of input.


edwinb:
Thanks!

Issue #3883, opened 2017-06-26
src/Idris/REPL.hs:(823,1)-(1460,70): Non-exhaustive patterns in function process


zilinc:
In version 1.0 REPL, if I do `:verbosity 1`, it splits out the above error.

ahmadsalim:
Thanks for reporting the error!

mdimjasevic:
Hi all!

I'd like to work on this issue as my first contribution to Idris.

I looked at patterns for the `process` function in `src/Idris/REPL.hs` and identified the following 7 patterns that are not matched at the moment:

```haskell
process fn Quit = ?process_quit
process fn Reload = ?process_reload
process fn Watch = ?process_watch
process fn (Load filePath maybeInt) = ?process_load
process fn Edit = ?process_edit
process fn Proofs = ?process_proofs
process fn (Verbosity i) = ?process_verbosity
```

User {AS} from the IRC channel suggested to provide a user facing error for these cases, as a starting point. I guess that would be something like:

```haskell
process fn Quit = iPrintError "Unimplemented pattern for command Quit"
```

Would that be a good implementation or should I implement something more useful?

mdimjasevic:
Pull request #4054 addressed this issue so it is ready to be closed.

PR #4054, opened 2017-09-05
Add missing patterns for a REPL function (#3883)

mdimjasevic:
Seven missing patterns for the `process` function (`src/Idris/REPL.hs`) were
added. This makes patterns in the function exhaustive.

### DIFF
--- a/src/Idris/REPL.hs
+++ b/src/Idris/REPL.hs
@@ -1474,6 +1474,17 @@ process fn (PPrint fmt width t)
             ty' = normaliseC ctxt [] ty
         iPrintResult =<< renderExternal fmt width (pprintDelab ist tm)
 
+
+process fn Quit = iPrintError "Unimplemented pattern for command Quit"
+process fn Reload = iPrintError "Unimplemented pattern for command Reload"
+process fn Watch = iPrintError "Unimplemented pattern for command Watch"
+process fn (Load _ _) = iPrintError "Unimplemented pattern for command Load"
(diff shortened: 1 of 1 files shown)


src/Idris/REPL.hs
```diff
@@ -1474,6 +1474,17 @@ process fn (PPrint fmt width t)
             ty' = normaliseC ctxt [] ty
         iPrintResult =<< renderExternal fmt width (pprintDelab ist tm)
 
+
+process fn Quit = iPrintError "Unimplemented pattern for command Quit"
```
ahmadsalim: I would use something like `iPrintError "Command ':quit'" is current unsupported` instead.

mdimjasevic:
Hi @ahmadsalim,

Thanks for the suggestion! It would definitely be more informative to the user, unlike with my current messages that rely on names of internal data constructors. I'll make the change and update the pull request.

ahmadsalim:
OK, the updated messages look good. I have tried to restart travis for additional safety, but I think it is good to merge in any case.

mdimjasevic:
@ahmadsalim, what is next, i.e. what do I need to do to finish this pull request?

ahmadsalim:
@mdimjasevic There is nothing left, I was just waiting to see if Travis could pass.

You are of course welcome to add your name to `CONTRIBUTORS` if you want to.

david-christiansen:
Thanks for the contribution, @mdimjasevic!

mdimjasevic:
You are welcome, @david-christiansen! I'm looking for the next thing to fix.

PR #4058, opened 2017-09-06
Text.Parser: implement `Functor (Grammar tok c)`

msmorgan:
Also included:

* Operators `(<*>)`, `(<*)`, `(*>)`, analogous to those defined in `Prelude.Applicative`.
* The functions `join` and `choice`, which resemble `Prelude.Monad.join` and `Prelude.Applicative.choice`.


msmorgan:
@edwinb Should I have used `%inline` with the applicative sequencing operators?

Also, do you think they have the correct associativity?

PR #4062, opened 2017-09-08
Prevent private functions showing up in autocomplete and :t

### LINKED ISSUE #1209: Private functions show up in autocomplete and work in :t

steshaw:
Fixes #1209.

Fixing this was a little bit like whack-a-mole. I'm not sure I've done it the easiest/best way and so am open to other approaches.

## `loadInputs`

The entry point for the fix is `ModeCommon.loadInputs`. It's here we meet the first problem. This is what happens when loading a file `Top.idr` that imports `Imported.idr`:

1. loadInputs `Bar.idr`
2. Build tree is calculated. Modules requiring recompilation are `Imported.idr` and `Bar.idr`.
3. Source files are built in `IBC_Building` phase/mode. i.e. `tryLoad IBC_BUILDING`.
4. Next the ibcs are loaded in "REPL" mode. i.e. `tryLoad (IBC_REPL True) Imported.ibc Bar.ibc`

Loading `Imported.ibc` in "REPL" mode is what causes the private functions to be made public by `loadIBC` (via loadIFile). So the first change here is to ensure that the only ibc files loaded in "REPL" mode are those that are specified as inputs to `loadInputs` — in this case, `Top.ibc`. Note that  I did this by mapping over the arguments to `loadInputs` using the following code:

```haskell
                   let fileToIFileType :: FilePath -> Idris IFileType
                       fileToIFileType file = do
                         ibcsd <- valIBCSubDir ist
                         ids <- rankedImportDirs file
                         findImport ids ibcsd file

                   ibcfiles <- mapM fileToIFileType inputs

                   tryLoad True (IBC_REPL True) ibcfiles
```

## `loadIBC`/`processImports`

The idea here, when loading in "REPL" mode is to ensure all definitions are made public. This happens in `processAccess`:

```haskell
        -- Everything should be available at the REPL from things imported publicly.
        when (phase == IBC_REPL True) $ do
            setAccessibility n Public
```

However, `processImports` changes the mode from `IBC_REPL True` to `IBC_REPL False` when recursively loading an imported file. This means that public definitions from those files will not be available in the REPL. The code was:

```haskell
       let phase' = case phase of
                        IBC_REPL _ -> IBC_REPL False
                        p -> p
```

In order to fix this problem, I used the `reexp`/`reexport` flag. The updated code still changes the phase from `IBC_REPL True` to `IBC_REPL False` but it passes along the `reexp` flag (whether the import is `import public` or not). This means that the import will be processed as if it were a `import public` — the public definitions will be public in the REPL!

A further complication here is the handling of imports that are `import public` (i.e. the `re` variable in `processImports`).

```haskell
        let (phase', reexp') = case phase of
                         IBC_REPL True -> (IBC_REPL False, reexp)
                         IBC_REPL False -> (IBC_Building, reexp && re)
                         p -> (p, reexp && re)
```

## Checking visibility

The final step is to filter out hidden definitions for completions and the REPL's type-check command (":t"). I added a visibleDefinitions function to `Idris.Core.Evaluate` for want of a better location. The completion tweak is in `Idris.Completion` and the REPL type-check command tweak is in `Idris.REPL`.

## Testing

I've added a test case, `interactive018`.

## Boyscouting (Campsite rule)

All other changes are just "boy scouting".

In particular, I enabled `interactive017` which was written but not enabled (a fact I noticed when adding `interactive018`). This required me to `chmod +x` a couple of files under `interactive017/`. I subsequently disabled `interactive017` again (in `TestData.hs`) because @melted has the legitimate concern that [`interactive017` currently depends on having nodejs installed](https://github.com/idris-lang/Idris-dev/pull/4046#issuecomment-326910042) (and needs more work).

### A misnomer

The phase `IBC_REPL` is a misnomer. It's also required for compiling to an executable/binary. For instance, I experimented early on, removing the code that makes definitions public when `IBC_REPL True`. Many tests failed to build a binary, unable to find a definition for `Main.main`. Even after adding a special case to export definitions of `Main.name`, there were still failures from modules using a `Main` namespace:

```idris
module Boo

namespace Main
  main : IO ()
  main = putStrLn $ "Boo"
```

### Potential refactoring?

It seems that it may be possible to collapse (or somehow tidy up) the arguments to loadIBC. i.e. reexport, and phase (`IBC_Building`, `IBC_REPL True`, `IBC_REPL False`).Maybe not but I just get a feeling. 😄 

I took another look and have a stronger feeling that `IBC_Building` and `IBC_REPL False` have the same behaviour. If that turns out to be the case, I might suggest:

```haskell
data IBCPhase
  = IBC_Building  -- ^ when building the module tree
  | IBC_Loading -- ^ when loading modules
```


ahmadsalim:
Wow, thanks for looking into this!

src/Idris/AbsSyntax.hs
```diff
@@ -7,6 +7,8 @@ Maintainer  : The Idris Community.
 -}
 
 {-# LANGUAGE DeriveFunctor, FlexibleContexts, PatternGuards #-}
+-- FIXME: {-# OPTIONS_GHC -fwarn-incomplete-patterns #-}
```
ahmadsalim: Is it possible to not have the `FIXME`?

src/Idris/AbsSyntax.hs
```diff
@@ -597,7 +598,7 @@ getHdrs :: Codegen -> Idris [String]
 getHdrs tgt = do i <- getIState; return (forCodegen tgt $ idris_hdrs i)
 
 getImported ::  Idris [(FilePath, Bool)]
-getImported = do i <- getIState; return (idris_imported i)
+getImported = idris_imported `fmap` getIState
```
ahmadsalim: Would it not be preferable to use `<$>` instead?

src/Idris/Completion.hs
```diff
@@ -28,19 +32,21 @@ import qualified Data.Text as T
...
 tactics = map fst tacticArgs
 
 -- | Get the user-visible names from the current interpreter state.
 names :: Idris [String]
-names = do i <- get
-           let ctxt = tt_ctxt i
+names = do ctxt <- tt_ctxt <$> getIState
            return $
-             mapMaybe nameString (allNames $ definitions ctxt) ++
+             mapMaybe nameString (allNames (visibleDefinitions ctxt)) ++
```
ahmadsalim: I think `allNames $ visibleDefinitions ctxt` is more readable (akin to what was before).

src/Idris/IBC.hs
```diff
@@ -517,13 +511,15 @@ processImports reexp phase fname ar = do
         ibcsd <- valIBCSubDir i
         ids <- rankedImportDirs fname
         putIState (i { imported = f : imported i })
-        let phase' = case phase of
-                         IBC_REPL _ -> IBC_REPL False
-                         p -> p
+        let (phase', reexp') = case phase of
```
ahmadsalim: Just a small note on style, I prefer not to have a dangling `case`.
Would it possible to use the following instead?

```haskell
let (phase', reexp') = 
    case phase of
        IBC_REPL ....
```

src/Idris/IBC.hs
```diff
@@ -735,8 +732,11 @@ processAccess :: Bool -- ^ Reexporting?
            -> IBCPhase
            -> Archive -> Idris ()
 processAccess reexp phase ar = do
+    logIBC 3 $ "processAccess (reexp, phase)" ++ show (reexp, phase)
     ds <- getEntry [] "ibc_access" ar
+    logIBC 3 $ "processAccess ds" ++ show ds
     mapM_ (\ (n, a_in) -> do
```
ahmadsalim: I know it is not you that introduced this, but would it be possible to use a `where`-clause to define the huge anonymous function in the lambda instead?

src/Idris/AbsSyntax.hs
steshaw: I tried to introduce `-fwarn-incomplete-patterns` here but there were too many warnings. 😢 

I'm not sure what you mean by your comment. Is your preference that I remove the entire line with the FIXME or just the word FIXME?

ahmadsalim: I think it would be best to avoid this line if possible.

src/Idris/AbsSyntax.hs
ahmadsalim: Do you know if `Util.Pretty` is used for anything?

ahmadsalim: I can't find its use.

src/Idris/IBC.hs
steshaw: No worries. I prefer that style too. I used `hindent` to help me format it like that (but I cherry picked the expression and then paste it back into place so as not to reformat the rest of the definition). A lot of the Idris code follows a different style though 😄 .

src/Idris/IBC.hs
steshaw: I thought about that myself but there are many other functions in `Idris.IBC` using precisely the same form.

ahmadsalim: I do not think it is blocking to fix this, but it may be useful if you have time.
Of course it would be nice to clean up the other functions as well, but maybe you could kindly just create an issue for those?

src/Idris/AbsSyntax.hs
steshaw: Try removing the import, that's what I did IIRC. Later I added `-fwarn-unused-imports` https://github.com/idris-lang/Idris-dev/pull/4062/files#diff-5348f61755edf5c96c5e2818d0ff2ba6R10. So, it's used somewhere.

src/Idris/IBC.hs
steshaw: I hope not otherwise the original code should have been rejected ;).

Happy to raise an issue. Was thinking it might be nicer even to use `forM_` and not having to introduce a local definition via where.

ahmadsalim: You are welcome to raise the issue.

steshaw: Might be better to work on a PR to refactor those functions. I'd be happy to put that together once this PR goes through. Less bureaucracy that way and I might get unlucky with merge conflicts if someone eagerly picks up the issue.

steshaw:
Thanks for taking the time to review this, @ahmadsalim!

ahmadsalim:
@steshaw Can we please get a `CHANGELOG` entry as well in another PR? Thanks!

steshaw:
No worries, @ahmadsalim — I've raised PR #4076.

PR #4070, opened 2017-09-14
Rework commit in Text.Parser

benclifford:
The previous commit/alternative behaviour seemed to be
trying to scope commits to a particular alternative,
rather than the parse as a whole, but it wasn't clear
to me how this was meant to work:

For example, the parser:

 Commit

would commit to the current alternative, while the
parser:

 (Commit <|> fail "X")

would not commit - seemingly because it was attempting
to commit only to the <|> used inside the parser
definition.

This violates a rule that I was expecting, that

  x <|> fail "X"

would behave the same, commitment-wise, as simply

  x

This reworked code never removes a commitment. When a
parser commits, it commits the entire parse tree.

In this model, I think it would make sense to have
a Try primitive that that removes commitment of a
failing sub-parser (see 'try' in Parsec).

I claim that whether a parser commits or not depends
only on that parser, not any parser evaluated before.

All simple parsers except Commit do not commit.

Parser combinators decide their commitness as functions
of the parameter sub-parsers, without reference to
anything else.

### DIFF
--- a/libs/contrib/Text/Parser.idr
+++ b/libs/contrib/Text/Parser.idr
@@ -111,67 +111,67 @@ shorter more [] = lteRefl
 shorter more (x :: xs) = LTESucc (lteSuccLeft (shorter more xs))
 
 doParse : {c : Bool} ->
-          (commit : Bool) -> (xs : List tok) -> (act : Grammar tok c ty) ->
+          (xs : List tok) -> (act : Grammar tok c ty) ->
           ParseResult xs c ty
-doParse com xs act with (sizeAccessible xs)
-  doParse com xs (Empty val) | sml = EmptyRes com val xs
-  doParse com [] (Fail str) | sml = Failure com str []
-  doParse com (x :: xs) (Fail str) | sml = Failure com str (x :: xs)
-  doParse com xs Commit | sml = EmptyRes True () xs
+doParse xs act with (sizeAccessible xs)
+  doParse xs (Empty val) | sml = EmptyRes False val xs
+  doParse [] (Fail str) | sml = Failure False str []
+  doParse (x :: xs) (Fail str) | sml = Failure False str (x :: xs)
+  doParse xs Commit | sml = EmptyRes True () xs
 
-  doParse com [] (Terminal f) | sml = Failure com "End of input" []
-  doParse com (x :: xs) (Terminal f) | sml
+  doParse [] (Terminal f) | sml = Failure False "End of input" []
+  doParse (x :: xs) (Terminal f) | sml
         = maybe
-             (Failure com "Unrecognised token" (x :: xs))
-             (\a => NonEmptyRes com {xs=[]} a xs)
+             (Failure False "Unrecognised token" (x :: xs))
+             (\a => NonEmptyRes False {xs=[]} a xs)
              (f x)
-  doParse com [] EOF | sml = EmptyRes com () []
-  doParse com (x :: xs) EOF | sml
-        = Failure com "Expected end of input" (x :: xs)
-  doParse com [] (NextIs f) | sml = Failure com "End of input" []
-  doParse com (x :: xs) (NextIs f) | sml
+  doParse [] EOF | sml = EmptyRes False () []
+  doParse (x :: xs) EOF | sml
+        = Failure False "Expected end of input" (x :: xs)
+  doParse [] (NextIs f) | sml = Failure False "End of input" []
+  doParse (x :: xs) (NextIs f) | sml
         = if f x
-             then EmptyRes com x (x :: xs)
-             else Failure com "Unrecognised token" (x :: xs)
-  doParse com xs (Alt x y) | sml with (doParse False xs x | sml)
-    doParse com xs (Alt x y) | sml | Failure com' msg ts
+             then EmptyRes False x (x :: xs)
+             else Failure False "Unrecognised token" (x :: xs)
+  doParse xs (Alt x y) | sml with (doParse xs x | sml)
+    doParse xs (Alt x y) | sml | Failure com' msg ts
           = if com' -- If the alternative had committed, don't try the
-                    -- other branch (and reset commit flag)
-               then Failure com msg ts
-               else weakenRes (doParse False xs y | sml)
-    -- Successfully parsed the first option, so use the outer commit flag
-    doParse com xs (Alt x y) | sml | (EmptyRes _ val xs)
-          = EmptyRes com val xs
-    doParse com (z :: (ys ++ more)) (Alt x y) | sml | (NonEmptyRes _ val more)
-          = NonEmptyRes com val more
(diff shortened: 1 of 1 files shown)


msmorgan:
Text.Parser is backtracking by default (similar to [Lightyear](/ziman/lightyear)). `Commit` is meant to be used after a consuming grammar within a branch of an `Alt`. This indicates that the parser should NOT try to backtrack and use the next alternative.

This eliminates the need for `try`.

benclifford:
@msmorgan - I didn't need try in the parser that I was writing, which is why it isn't in this PR.

But the use case I see for it (which is possibly contrived for real world examples?) is where you want to use commit inside a sub-parser but don't want it to commit the *whole* parser; instead you want the sub-parser to behave in a backtracking manner with respect to the whole parser, but internally do commits (for better error messages, for example).

Pieces of Text.Parser as exists before this PR make me think there was some intention to support that sort of behaviour (but I'm unsure) - for example, `(and reset commit flag)` in the `Alt` case of `doParse`.

msmorgan:
Here's an example.

```idris
import Text.Parser

-- using `Char` as our token type`
Token : Type
Token = Char

Rule : Type -> Type
Rule ty = Grammar Token True ty

||| Parse a specific character.
char : Token -> Rule Char
char c = terminal (\x => if x == c then Just x else Nothing)

p_abc : Rule Char
p_abc = do char 'a'
           char 'b'
           commit
           char 'c'
           pure '!'

||| Parse any character repeated 1 or more times.
p_repeated : Rule Char
p_repeated = do c <- peek
                char c
                commit
                many (char c)
                pure c

||| Parse a non-empty sequence of the above two rules.
p_all : Rule (List Char)
p_all = some (p_abc <|> p_repeated)

demo : Either (ParseError Token) (List Char, List Token)
demo = parse (unpack "aaabbbcccabcaaa") p_all
```

msmorgan:
The "and reset commit flag" part was referring to the following:

After trying the first alternative, `Alt` knows if it committed. If it both committed and failed, the parser won't try the second alternative, and the failure bubbles up to the grammar that contained this `Alt`. The commit flag is then reset to the state it was in before the first alternative was tried.

benclifford:
OK, so in my mental model, the below `demo` should return a Left failure, not a Right success, because `p_repeated` should be stopping the alternative p_abc from being tried once it sees the first 'a'.

```
import Text.Parser

-- using `Char` as our token type`
Token : Type
Token = Char

Rule : Type -> Type
Rule ty = Grammar Token True ty

||| Parse a specific character.
char : Token -> Rule Char
char c = terminal (\x => if x == c then Just x else Nothing)

p_abc : Rule Char
p_abc = do char 'a'
           commit
           char 'b'
           pure '!'

||| Parse any character repeated 1 or more times.
p_repeated : Rule Char
p_repeated = do c <- peek
                char c
                commit
                many (char c)
                char 'X'
                pure c

||| Parse a non-empty sequence of the above two rules.
p_all : Rule Char
p_all = p_repeated <|> p_abc

demo : Either (ParseError Token) (Char, List Token)
demo = parse (unpack "ab") p_all
```

msmorgan:
@benclifford In your example, the reason it succeeds is because `many` doesn't need to consume any input, and `repeated` is the first rule that's tried. So it should succeed after parsing 'a', repeated once. The character 'b' will remain unparsed.

benclifford:
> So it should succeed after parsing 'a', repeated once. The character 'b' will remain unparsed.

What version are you working against? With my 1.1.1 install, it's returning `Right ('!', [])` which suggests it *is* consuming the 'b' (and taking the `p_abc` branch).

benclifford:
> EDIT2: Now that you added char 'X' to p_repeated, is it failing like you expected?

With this PR, yes - with 1.1.1, no.

msmorgan:
Hm, that does seem like a probable bug. I'll look into it some more.

benclifford:
@msmorgan I don't understand what information is being passing around in the commit flag  parameter to doParse at all: a parser either commits or doesn't; if it's a terminal it doesn't unless it is Commit; and if its a combinator, it knows that based on any sub-parsers it ran and no other information; and the only place that information is needed is in <|> when you need to observe the commit flag output from one of your sub-parsers, not from any other part of the parse tree.

msmorgan:
@benclifford As far as I understand, nothing commits except `Commit`. The only time this matters is when parsing the first alternative of an `Alt`. If the flag is `True`, the `Alt` reports a failure instead of trying the second grammar.

benclifford:
@msmorgan My (opinion/understanding) is that a <|> and a >>= should also commit (based on what their component parsers do): if my child committed, then I pass the commit upwards to my parent - and somewhere I might have an ancestor that is another <|> that cares about it; and every <|> above a Commit in the parse tree should get the `True` commit message and abandon alternatives. It feels like the code in 1.1.1 is somehow trying to avoid that.

msmorgan:
@benclifford After looking into it a bit further, I think there is most likely a problem with how the `doParse` clauses for `SeqEmpty` and `SeqEat` handle the `com` flag.

I'm out of time to look at this right now, but I'll get back to it this evening.

msmorgan:
I've reduced the bug to an issue in `doParse`, when a sequence contains an `Alt`.

```idris
-- `next` and `next'` should work equivalently (aside from the error message),
-- but they do not.
||| Parse the next character (or fail with a standard message).
next : Grammar Char True Char
next = terminal Just

||| Parse the next character or fail with a custom message.
next' : Grammar Char True Char
next' = terminal Just <|> fail "failed"

Input : List Char
Input = ['a']

-- uses `next` after commit
gram : Grammar Char False String
gram = do { next; commit; next; pure "success"}
   <|> pure "never"

-- uses `next'` after commit
gram' : Grammar Char False String
gram' = do { next; commit; next'; pure "success"}
    <|> pure "never"

-- Note: ParseResult is not exported from Text.Parser
demo : ParseResult Input False String
demo = doParse False Input gram

demo' : ParseResult Input False String
demo' = doParse False Input gram'
```

benclifford:
> I get the impression that everyone here has a different idea of what it’s supposed to do.

yes definitely that.

edwinb:
`commit` is supposed to mean that if the current alternative fails to parse, the whole thing fails. So if you have (a <|> b), and the `a` branch fails after a `commit` then `b` won't be tried.

The issue in the examples above is that if there's another `<|>` after a commit, then the parser forgets the value of the commit flag. There's a simple fix for this, which I'm currently testing.

ahmadsalim:
@edwinb any news for this?

msmorgan:
@ahmadsalim It seems he fixed it [in this commit](/idris-lang/Idris-dev/commit/be8523f834faf04709cdf8e63bcf09ba942d20f3).

ahmadsalim:
I think I was told that this was solved, so I will close the PR.
Thanks for submitting it though!

PR #4075, opened 2017-09-14
Text.Parser: right-associative-esque alternative operator

msmorgan:
This PR changes the behavior of `Text.Parser.(<|>)`, so that if an alternative in a chain of alternatives fails, the parser will not try remaining ones.

This causes `a <|> b <|> c` to evaluate to `Alt a (Alt b c)`, even though `<|>` remains left-associative.

### DIFF
--- a/libs/contrib/Text/Parser.idr
+++ b/libs/contrib/Text/Parser.idr
@@ -1,267 +1,21 @@
 module Text.Parser
 
+import Text.Parser.Core
+
 %default total
 
--- TODO: Add some primitives for helping with error messages.
--- e.g. perhaps set a string to state what we're currently trying to
--- parse, or to say what the next expected token is in words
+mutual
+  ||| Parse one or more things
+  export
+  some : Grammar tok True a ->
+         Grammar tok True (List a)
+  some p = pure (!p :: !(many p))
 
-||| Description of a language's grammar. The `tok` parameter is the type
-||| of tokens, and the `consumes` flag is True if the language is guaranteed
-||| to be non-empty - that is, successfully parsing the language is guaranteed
-||| to consume some input.
-export
-data Grammar : (tok : Type) -> (consumes : Bool) -> Type -> Type where
-     Empty : (val : ty) -> Grammar tok False ty
-     Terminal : (tok -> Maybe a) -> Grammar tok True a
-     NextIs : (tok -> Bool) -> Grammar tok False tok
-     EOF : Grammar tok False ()
-
-     Fail : String -> Grammar tok c ty
-     Commit : Grammar tok False ()
-
-     SeqEat : Grammar tok True a -> Inf (a -> Grammar tok c2 b) ->
-              Grammar tok True b
-     SeqEmpty : {c1, c2 : Bool} ->
-                Grammar tok c1 a -> (a -> Grammar tok c2 b) ->
-                Grammar tok (c1 || c2) b
-     Alt : {c1, c2 : Bool} ->
-           Grammar tok c1 ty -> Grammar tok c2 ty ->
-           Grammar tok (c1 && c2) ty
-
-public export
-inf : Bool -> Type -> Type
-inf True t = Inf t
-inf False t = t
-
-||| Sequence two grammars. If either consumes some input, the sequence is
-||| guaranteed to consume some input. If the first one consumes input, the
-||| second is allowed to be recursive (because it means some input has been
-||| consumed and therefore the input is smaller)
-export %inline
-(>>=) : {c1 : Bool} ->
-        Grammar tok c1 a ->
-        inf c1 (a -> Grammar tok c2 b) ->
-        Grammar tok (c1 || c2) b
-(>>=) {c1 = False} = SeqEmpty
-(>>=) {c1 = True} = SeqEat
-
-||| Sequence a grammar followed by the grammar it returns.
-export
-join : {c1 : Bool} ->
-       Grammar tok c1 (Grammar tok c2 a) ->
-       Grammar tok (c1 || c2) a
-join {c1 = False} p = SeqEmpty p id
-join {c1 = True} p = SeqEat p id
-
-||| Give two alternative grammars. If both consume, the combination is
-||| guaranteed to consume.
-export
-(<|>) : Grammar tok c1 ty ->
-        Grammar tok c2 ty ->
-        Grammar tok (c1 && c2) ty
-(<|>) (Alt {c1 = True} x y) z = Alt x (y <|> z)
-(<|>) (Alt {c1 = False} x y) z = Alt x (y <|> z)
-(<|>) x y = Alt x y
-
-||| Allows the result of a grammar to be mapped to a different value.
-export
-Functor (Grammar tok c) where
-  map f (Empty val)  = Empty (f val)
-  map f (Fail msg)   = Fail msg
-  map f (Terminal g) = Terminal (\t => map f (g t))
-  map f (Alt x y)    = Alt (map f x) (map f y)
-  map f (SeqEat act next)
-      = SeqEat act (\val => map f (next val))
-  map f (SeqEmpty act next)
-      = SeqEmpty act (\val => map f (next val))
-  -- The remaining constructors (NextIs, EOF, Commit) have a fixed type,
-  -- so a sequence must be used.
-  map {c = False} f p = SeqEmpty p (Empty . f)
-
-||| Sequence a grammar with value type `a -> b` and a grammar
-||| with value type `a`. If both succeed, apply the function
-||| from the first grammar to the value from the second grammar.
-||| Guaranteed to consume if either grammar consumes.
-export
-(<*>) : {c1 : Bool} ->
-        Grammar tok c1 (a -> b) ->
-        inf c1 (Grammar tok c2 a) ->
-        Grammar tok (c1 || c2) b
-(<*>) {c1 = False} x y = SeqEmpty x (\f => map f y)
-(<*>) {c1 = True} x y = SeqEat x (\f => map f y)
-
-||| Sequence two grammars. If both succeed, use the value of the first one.
-||| Guaranteed to consume if either grammar consumes.
-export
-(<*) : Grammar tok c1 a ->
-       inf c1 (Grammar tok c2 b) ->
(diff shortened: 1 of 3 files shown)


benclifford:
I'm uncomfortable with this - the below contrived example should (in my opinion) fail `A` but actually returns pure `c`.

```
import Text.Parser

-- using `Char` as our token type`
Token : Type
Token = Char

Rule : Type -> Type
Rule ty = Grammar Token True ty

||| Parse a specific character.
char : Token -> Rule Char
char c = terminal (\x => if x == c then Just x else Nothing)

p_a : Rule Char
p_a = do
  char 'a'
  commit
  the (Rule Char) (fail "A")

p_b : Rule Char
p_b = do
  char 'a'
  pure 'B'

p_c : Rule Char
p_c = do
  char 'a'
  pure 'C'

||| Parse a non-empty sequence of the above two rules.
p_all : Rule Char
p_all = p_a <|> p_b <|> p_c

demo : Either (ParseError Token) (Char, List Token)
demo = parse (unpack "ab") p_all

```

msmorgan:
@benclifford You're right, this does not fix the whole issue. Thanks for making the example.

It's somewhat problematic however, because there would be no problem if `<|>` were a right-associative operator... I think using a right-associative operator is the "correct" way to fix to me, but it's impossible to change the associativity of (<|>).

msmorgan:
I've committed a proposed solution. I'll let the code speak for itself.

msmorgan:
On the subject, would `>>=` benefit from being right-associative as well? This would result in `SeqEat` being used quite a bit more often, I think. I'm not sure if that's a benefit, though.

At the moment, as I understand it, operator fixity cannot be overloaded because associativity is determined at the parser level. Could operator chaining be resolved in the elaborator somehow instead?

msmorgan:
Is there a way to make `>>=` right-associative as well? This might result in better utilization of `SeqEat`'s laziness.

As far as I understand, fixity can't be overloaded because associativity is determined in the parser. Can this be deferred to the elaborator level for operator chains somehow?

benclifford:
There's another thing that I'm still unclear about commits. #4070 takes the approach that when you commit, you've committed the whole parse so far, but I can't find anything that says if that is meant to be the behaviour or not - so I can't say whether this is a bug or not. It's the behaviour I want though.

In the below program, `p_object` should (in the #4070 approach) commit after two `{` symbols, but it does not - although there is a commit inside the `many p_val` call, that doesn't "propagate upwards" and commit the enclosing parser, and the entire inner `{X}` is left unconsumed.

I've also supplied `p_object'` which *does* commit after two `{{` symbols.

*c> parse (unpack "{{X}}") p_object
Left (Error "Unrecognised char, was expecting }" ['{', 'X', '}', '}']) : Either (ParseError Char) (Char, List Char)
*c> parse (unpack "{{X}}") p_object'
Left (Error "Unrecognised char, was expecting }" ['X', '}', '}']) : Either (ParseError Char) (Char, List Char)

```
import Text.Parser

-- using `Char` as our token type`
Token : Type
Token = Char

Rule : Type -> Type
Rule ty = Grammar Token True ty

||| Parse a specific character.
char : Token -> Rule Char
char c = terminal (\x => if x == c then Just x else Nothing)
    <||> (fail ("Unrecognised char, was expecting " ++ pack [c]))

mutual

 p_object : Rule Char
 p_object = do
  char '{'
  commit
  many p_val
  char '}'
  pure 'O'

 p_object' : Rule Char
 p_object' = do
  char '{'
  commit
  many p_object'
  char '}'
  pure 'O'

 p_array : Rule Char
 p_array = do
  char '['
  commit
  char ']'
  pure 'A'

 p_val : Rule Char
 p_val = p_object <||> p_array <||> fail "F"

demo : Either (ParseError Token) (Char, List Token)
demo = parse (unpack "{{X}}") p_val

```

msmorgan:
Here's a more JSON-esque example that illustrates how I expect this module to work.

```idris
import Text.Parser

%default total

-- using `Char` as our token type`
Token : Type
Token = Char

Rule : Type -> Type
Rule ty = Grammar Token True ty

||| Parse a specific character.
char : Token -> Rule Char
char c = terminal (\x => if x == c then Just x else Nothing)
    <||> (fail ("Unrecognised char, was expecting " ++ pack [c]))

alpha : Rule Char
alpha = terminal (\x => if isAlpha x then Just x else Nothing)
   <||> fail "Expected an alpha character."

comma : Rule ()
comma = do char ','
           pure ()

data AST
  = Object (List (Char, AST))
  | Array (List AST)
  | Letter Char

mutual
  p_pair : Rule (Char, AST)
  p_pair = do
    k <- alpha
    commit
    char ':'
    v <- p_val
    pure (k, v)

  p_object : Rule AST
  p_object = do
    char '{'
    commit
    kvs <- sepBy comma p_pair
    char '}'
    pure (Object kvs)

  p_arrayvals : Grammar Token False (List AST)
  p_arrayvals = sepBy comma p_val

  p_array : Rule AST
  p_array = do
    char '['
    commit
    xs <- p_arrayvals
    char ']'
    pure (Array xs)

  p_letter : Rule AST
  p_letter = pure (Letter !alpha)

  p_val : Rule AST
  p_val = p_object
     <||> p_array
     <||> p_letter
     <||> fail "Could not parse a value."

-- succeeds
demo : Either (ParseError Token) (AST, List Token)
demo = parse (unpack "{a:{b:[c,d,e,{f:g},h]},i:{j:{},k:[]}}") p_val

-- fails within the first array
-- note: a proper parser would include information about the token's position
-- in the token type, allowing errors to be less vague
demo2 : Either (ParseError Token) (AST, List Token)
demo2 = parse (unpack "{a:{b:[c,d,e,{f:g},h,]},i:{j:{},k:[]}}") p_val
```

msmorgan:
@edwinb Fixed the alt/commit issue, but I think there may be a lingering bug with the associativity of `<|>`.

```idris
-- expected behavior:
-- this call to `parse` fails with the error "End of input"
parse [] $ the (Grammar Char True Char) (do {commit; terminal Just} <|> fail {c=True} "never1")

-- confusing behavior:
-- this fails with the error "never2"
parse [] $ the (Grammar Char True Char) (do {commit; terminal Just} <|> fail {c=True} "never1" <|> fail "never2")
```

Another potential solution, instead of this heavy-handed replacement of the `<|>` operator, would potentially be to define `(<|>) = flip Alt`, and try the right side first in `doParse`.

ahmadsalim:
@msmorgan Thanks for the response, but using `flip Alt` also changes the semantics I think, not just the associativity.

msmorgan:
@ahmadsalim The `Alt` constructor is private, so as long as the right alternative is processed first within `doParse`, I don't think there'd be any semantic difference for `<|>`, other changing the behavior in the above snippet.

I have not yet tested this approach, but I'll update with the results once I have.

ahmadsalim:
@msmorgan Another alternative is just to flatten on parse (so define Alt as taking a list).

msmorgan:
@ahmadsalim I was being dumb with the `flip Alt` thing. The associativity issue can't be solved by that. See below.

```idris
(<|>) = Alt
(a <|> b) <|> c = Alt (Alt a b) c
-- vs --
x <|> y = Alt y x
(a <|> b) <|> c = Alt c (Alt b a)
```

"Flattening" `Alt` could be done via pattern matching in the definition of `(<|>)` as well, like so:

```idris
(<|>) (Alt x y) z = Alt x (Alt y z)
(<|>) x y = Alt x y
```

I see a potential issue here: doing this would make it impossible to create something of the form `Alt (Alt a b) c` using `(<|>)`. Whether that's ever a useful thing to do is debatable.

ahmadsalim:
@msmorgan To be honest, I think alternatives should always be read as a list from left to right.

ahmadsalim:
So it is OK to ensure that they are always right-associated.

msmorgan:
For some reason, adding the pattern match clauses to the definition of `<|>` has caused `some` and `sepBy1` to fail to check totality. I was able to fix `some`, but not `sepBy1`. If the two pattern match lines for `<|>` are commented out, `sepBy1` is total again.

I've split `Text.Parser` into two modules to speed up the compiler when working on parser combinators.

@ahmadsalim Did you see this coming when you suggested flattening during the parse stage instead of in the definition of the operator? At this point, I'm not sure there's a better way than that to fix this issue.

ahmadsalim:
@msmorgan I did not see that coming no.

ahmadsalim:
@msmorgan Usually totality should be modular, so changing the definition of `<|>` which is not mutually recursive with `sepBy` or `some` , should not have an effect on their totality.

@edwinb Do you have an idea why this might be the case?

ahmadsalim:
@msmorgan It seems that you have changed the definition of `sepBy` and `some` in your PR as well?

msmorgan:
I managed to fix the totality of `some`, but I didn't change `sepBy`. I tried for hours to fix `sepBy1`, but I wasn't able to get anywhere.

msmorgan:
@ahmadsalim I know why it's happening. TDDWI, page 297, halfway down:

> Idris considers a function to be total if there are patterns that cover all well-typed inputs, and it can determine that one of the following conditions holds:
>
> * When there is a recursive call (or a sequence of mutually recursive calls), there's a decreasing argument that converges toward a base case.
> * When there is a recursive call as an argument to `Delay`, the delayed call will always be an argument to a data constructor (or sequence of nested data constructors) after evaluation, for all inputs.

The second bullet is the reason `sepBy1` is not total. It appears that Idris lost the ability to determine this now that the operator isn't simply an alias for a constructor.

ahmadsalim:
@msmorgan Maybe I missed that delay was used somewhere?

msmorgan:
@ahmadsalim `SeqEat`

ahmadsalim:
@msmorgan I see, I think the the second one regards productivity.
Could one alternatively not make the behaviour of commit the same under associativity?

Maybe the issue is not with alternatives per se, but rather with something else.

ahmadsalim:
Postsumed by #4111

ahmadsalim:
@msmorgan you are welcome to reopen it if #4111 does not go through :).

msmorgan:
Fingers crossed I won't have to.

PR #4084, opened 2017-09-22
Make type of `withAlloc` more general

asandroq:
This change makes the whole `withAlloc` block return the same value of
the function given as a parameter.


melted:
Looks good! Thanks!

PR #4086, opened 2017-09-28
Make UnmatchedCase errors lazy (#4049)

### LINKED ISSUE #4049: `:exec main` crashes with `No location:No pattern clauses`
There appears to be a compiler bug that causes a crash for the following code:

https://gist.github.com/jdevuyst/ea523209bf8f26aebb894d35c50b55e6

```
Type checking /Users/jonasdevuyst/Downloads/NoPatternClausesBug.idr
*NoPatternClausesBug> :exec main
No location:No pattern clauses
*NoPatternClausesBug> :x main
2
MkIO (\w => prim_io_pure ()) : IO' (MkFFI C_Types String String) ()
*NoPatternClausesBug> fst example
2 : Nat
```

This seems to be a variant on #3088. Tested with Idris 1.1.1.

jdevuyst:
Failure in erasure analysis may lead to UnmatchedCase errors being included in the generated code. Such errors should be evaluated lazily.

This fixes #4049

### DIFF
--- a/src/IRTS/Compiler.hs
+++ b/src/IRTS/Compiler.hs
@@ -527,7 +527,7 @@ irTree top args tree = do
 
 irSC :: Name -> Vars -> SC -> Idris LExp
 irSC top vs (STerm t) = irTerm top vs [] t
-irSC top vs (UnmatchedCase str) = return $ LError str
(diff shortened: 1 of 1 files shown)


melted:
Thanks for the PR!

PR #4088, opened 2017-09-29
Add discriminate tactic to Pruviloj

joom:
A port of the `discriminate` tactic from Coq is added to Pruviloj.


ahmadsalim:
Thanks for the PR! Could you kindly update the CHANGELOG as well, and possibly add your name to CONTRIBUTORS?

PR #4091, opened 2017-10-01
Implement primitiveNotEq using really_believe_me (not postulate)

### LINKED ISSUE #4090: 'reachable postulates' errors for Decidable.Equality.Int
I get the following compile error:

```
reachable postulates:
  Decidable.Equality.Decidable.Equality.Int implementation of Decidable.Equality.DecEq, method decEq, primitiveNotEq
```

when I try to compile the file https://gist.github.com/jdevuyst/1598490bfb1db7e812cb2157963ad9fc

Some context:
- Part I of the code is an implementation of `Ordered` for `Int`
- Part II of the code in the gist was previously discussed in issue #4049. There is some failure in erasure analysis in this part of the code.
- Part III is a simple operation that triggers the `reachable postulates` error. The operation is an `Int`-based counterpart of the example discussed in #4049.

The code can be made to run fine by defining `DecEq Int` as follows:

```
implementation DecEq Int where
    decEq x y = case x == y of -- Blocks if x or y not concrete
                     True => Yes primitiveEq
                     False => No primitiveNotEq
       where primitiveEq : x = y
             primitiveEq = really_believe_me (Refl {x})
             primitiveNotEq : x = y -> Void
             primitiveNotEq _ = really_believe_me ()
```

(whereas currently `primitiveNotEq` is defined as `postulate primitiveNotEq : x = y -> Void`)

I'm not sure if redefining `primitiveNotEq` like this is an option. Naively I do find it surprising that a program would fail to compile due to a failure in erasure analysis.

jdevuyst:
This diff fixes #4090, although it does not address the underlying failure in erasure analysis.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -119,6 +119,7 @@ John Connor
 John Dobronszki
 John Wiegley
 Jon Atack
+Jonas De Vuyst
 Jonas Westerlund
 Jonathan Sterling
 Joomy Korkut
--- a/libs/prelude/Decidable/Equality.idr
+++ b/libs/prelude/Decidable/Equality.idr
@@ -169,7 +169,8 @@ implementation DecEq Int where
                      False => No primitiveNotEq
        where primitiveEq : x = y
              primitiveEq = really_believe_me (Refl {x})
-             postulate primitiveNotEq : x = y -> Void
+             primitiveNotEq : x = y -> Void
+             primitiveNotEq = really_believe_me (id {a = x = y})
 
 --------------------------------------------------------------------------------
 -- Char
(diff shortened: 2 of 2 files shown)


melted:
Thanks for the PR!

I'm a bit puzzled as to why it would make difference though. I'll take a look at the code before I merge.

ahmadsalim:
@jdevuyst I guess I would prefer `really_believe_me (\x =>x)` instead.

ahmadsalim:
Since that is not likely to break subject reduction.

jdevuyst:
@ahmadsalim I tried and it will work with `\prf : x = y => prf` or `id {a = x = y}`. Not sure I understand what the theoretic advantages are of this approach but I can update my pull request tonight (GMT+8).

ahmadsalim:
@jdevuyst The idea is that `()` is a canonical constructor of `Unit`, and so the value would mismatch. with `id` we are taking a seemingly impossible proof `x = y` for the use case presumably `x != y` (so an impossible element) and converting it to an impossible element of `Void`.
The latter would seem to less break subject reduction.

ahmadsalim:
@jdevuyst Any updates?

jdevuyst:
@ahmadsalim Thanks for explaining & sorry for the delay. I'll update my pull request within the hour.

PR #4094, opened 2017-10-03
Add splitPair tactic to Pruviloj

joom:


david-christiansen:
Cool!

What about a generalization to any inductive type with a single constructor? It shouldn't make the code much longer, but it might make a different name more appropriate.

joom:
That would be like this: (It needs a better name)
```idris
splitWithConstr : Elab (List TTName)
splitWithConstr =
  do g <- goalType
     case headName g of
       Nothing =>
         fail [RawPart g, TextPart "is not a datatype declared with the data keyword"]
       Just h =>
         case !(constructors <$> lookupDatatypeExact h) of
           [(n, _, _)] => refine (Var n)
           _ => fail [NamePart h, TextPart "does not have exactly one constructor"]
```

I was wondering if it would be better to define something like Coq's `constructor` tactic instead of having a tactic only for inductive types with a single constructor.

ahmadsalim:
Thanks for the PR!
It fails due to other reasons, so I will merge it.

ahmadsalim:
@joom You are welcome to add a constructor tactic if you feel so instead in another PR.

Issue #4078, opened 2017-09-19
Idris calls C compiler with arguments in wrong order


asandroq:
After parsing `%flag` directives the Idris backend calls the C compiler with the flags in the opposite order, which breaks some order-dependent flags. For instance, the following directive:

```
%flag C "-L/usr/local/IUP/lib -Wl,-Bstatic -Wl,-liup -Wl,-Bdynamic"
```
will result in the following arguments to the C compiler:

```
"-Wl,-Bdynamic","-Wl,-liup","-Wl,-Bstatic","-L/usr/local/IUP/lib"
```
which obviously has the opposite effect.

ahmadsalim:
Thanks for reporting the issue. We do welcome PRs!

PR #4095, opened 2017-10-03
Pass directives to C backend in given order (#4078)

asandroq:
Directives for compiling and linking C code are order-sensitive and
therefore the order given in the input files must be preserved.


test/directives003/expected
```diff
@@ -0,0 +1,6 @@
+nuyahcha
+mohgiide
+apeekair
+eeraimam
+rahhahwi
+goocheiv
```
ahmadsalim: Are these just random words or do they mean something in some language?

ahmadsalim:
Thanks for the PR. I will try to fix travis and hopefully your PR should succeed

test/directives003/expected
asandroq: They are random words generated with `pwgen -AB0`.

ahmadsalim:
It passes now.
Can you please update the CHANGELOG and CONTRIBUTORS if needed?

I will merge afterwards

PR #4099, opened 2017-10-03
Replace splitPair with construct tactic

joom:
As @david-christiansen suggested in the comment thread of #4094, this is a generalized version of `splitPair`. It behaves like `constructor` in Coq.

Now we can do this:
```idris
data Even : Nat -> Type where
  evenZ : Even Z
  evenSS : Even n -> Even (S (S n))

even6 : Even 6
even6 = %runElab (repeatUntilFail (ignore construct))
```

### DIFF
--- a/libs/pruviloj/Pruviloj/Core.idr
+++ b/libs/pruviloj/Pruviloj/Core.idr
@@ -256,21 +256,6 @@ unproduct tm =
        try (unproduct (Var n1))
        try (unproduct (Var n2))
 
-||| If the goal is a tuple or a dependent pair, then split it into
-||| two goals and return the new goal names.
-||| It fails if the goal is neither a `Pair` nor a `DPair`.
-splitPair : Elab (TTName, TTName)
-splitPair =
-  let bug = [TextPart "Splitting created wrong number of holes"] in
-  case !goalType of
-    `(Pair ~A ~B)   => do [x, y] <- apply `(MkPair {A=~A} {B=~B}) [False, False]
-                                | _ => fail bug
-                          solve ; pure (x, y)
-    `(DPair ~A ~Pr) => do [x, y] <- apply `(MkDPair {a=~A} {P=~Pr}) [False, False]
-                              | _ => fail bug
-                          solve ; pure (x, y)
-    _ => fail [TextPart "Goal is not a pair"]
-
 ||| Try to apply the constructors of the goal data type one by one,
 ||| and apply the first one that works.
 ||| Similar to `constructor` in Coq.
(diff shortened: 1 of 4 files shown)


libs/pruviloj/Pruviloj/Core.idr
```diff
@@ -256,20 +256,23 @@ unproduct tm =
...
-                              | _ => fail bug
-                          solve ; pure (x, y)
-    _ => fail [TextPart "Goal is not a pair"]
+||| Try to apply the constructors of the goal data type one by one,
+||| and apply the first one that works.
+||| Similar to `constructor` in Coq.
+construct : Elab (List TTName)
+construct =
+    do g <- goalType
+       case headName g of
```
david-christiansen: As a minor matter of style, the `do` could be eliminated by using `case headName !g of`.

libs/pruviloj/Pruviloj/Core.idr
```diff
@@ -256,20 +256,23 @@ unproduct tm =
...
+         Nothing =>
+           fail [RawPart g, TextPart "is not a type declared with the data keyword"]
+         Just h =>
+           choiceMap (\(n, xs, _) => apply (Var n) (map shouldUnify xs) <* solve)
+                     !(constructors <$> lookupDatatypeExact h)
+  where
+    shouldUnify : CtorArg -> Bool
```
david-christiansen: I like this heuristic. But it should be mentioned in the docstring that explicit constructor arguments will not be solved by unification.

libs/pruviloj/Pruviloj/Core.idr
```diff
@@ -256,20 +256,23 @@ unproduct tm =
...
-  let bug = [TextPart "Splitting created wrong number of holes"] in
-  case !goalType of
-    `(Pair ~A ~B)   => do [x, y] <- apply `(MkPair {A=~A} {B=~B}) [False, False]
-                                | _ => fail bug
-                          solve ; pure (x, y)
-    `(DPair ~A ~Pr) => do [x, y] <- apply `(MkDPair {a=~A} {P=~Pr}) [False, False]
-                              | _ => fail bug
-                          solve ; pure (x, y)
-    _ => fail [TextPart "Goal is not a pair"]
+||| Try to apply the constructors of the goal data type one by one,
```
david-christiansen: The docs should mention that the return value is the list of names of new holes induced by the tactic.

ahmadsalim:
@joom It is just a few tests that have the wrong line number etc.

joom:
Fixed it. Is there any way to turn off this kind of warning messages?
```
pruviloj001.idr:73:9:
Main.evenZ has a name which may be implicitly bound.
This is likely to lead to problems!
pruviloj001.idr:74:10:
Main.evenSS has a name which may be implicitly bound.
This is likely to lead to problems!
```

ahmadsalim:
@joom Yeah, name them `EvenZ` and `EvenSS`

david-christiansen:
Thank you very much!

PR #4102, opened 2017-10-04
Add styling for metavariables in LaTeX and HTML pretty printer

joom:
The metavariable styling case is missing in the pretty printer for LaTeX and HTML, which means `\IdrisMetavar` command is undefined in LaTeX, which means the pretty printer can return buggy LaTeX. This PR fixes that.


ahmadsalim:
Thanks

PR #4106, opened 2017-10-07
Prelude.Nat: add IsSucc predicate type

msmorgan:
By popular demand.

If this should be moved to `Data.Nat` in `base`, say the word. There is no
such module at the moment, however.


ahmadsalim:
@msmorgan Thanks!

Can you move it under syntactic tests please? Now it is under comparisons.

ahmadsalim:
Also, please add an entry to the changelog

msmorgan:
@ahmadsalim Can do. My rationale for putting it where I did is that it's comparing to zero, and its function is very similar to NotBothZero.

ahmadsalim:
@msmorgan I would still prefer the former please :).
I do not think it is strictly a comparison.

Issue #4107, opened 2017-10-08
`stack-shell.nix` doesn't work with NixOS 17.09


vaibhavsagar:
# Steps to Reproduce
```bash
$ stack --nix build
```

# Expected Behavior
A working executable.

# Observed Behavior
`error: attribute ‘lts-9_0’ missing`

ahmadsalim:
Thanks for reporting the issue.

I am unfortunately not familiar with the platforms, so I can not provide a workaround but hopefully someone will come here who is part of maintaining the Nix package or stack package.

PR #4108, opened 2017-10-08
Update resolver to match current version of GHC

vaibhavsagar:
LTS package sets are no longer part of Nixpkgs.

### DIFF
--- a/stack-shell.nix
+++ b/stack-shell.nix
@@ -1,8 +1,8 @@
 with (import <nixpkgs> {});
 
 let
-  # MUST match resolver in stack.yaml
-  resolver = haskell.packages.lts-9_0.ghc;
+  # MUST match GHC of resolver in stack.yaml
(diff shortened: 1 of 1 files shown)


vaibhavsagar:
This addresses #4107.

Melvar:
What about the comment immediately above that line that it must match the resolver in `stack.yaml`?

vaibhavsagar:
I think that's either misleading or out of date. Nixpkgs hasn't had these package sets for at least 6 months and towards the end they decided to follow the LTS packages [loosely instead of strictly](https://nixos.org/nix-dev/2016-June/020585.html). AFAICT Stack selects the package versions to use so it's enough to match the GHC version used in the corresponding LTS. I have updated my PR accordingly.

ahmadsalim:
Thanks for the PR.

vaibhavsagar:
My pleasure :smile:

PR #4111, opened 2017-10-09
change `<|>` operator to be right-associative

msmorgan:
This is a difference from Haskell's `<|>` operator, but the semantics seem clearer this way to me.

Although this PR successfully passes the test suite, it should probably be classified as a **BREAKING CHANGE**.

#### Rationale

> I define the following function for notational convenience.
> ```idris
> ||| Prettier prefix notation for `<|>`
> Alt : Alternative f => f a -> f a -> f a
> Alt = (<|>)
> ```

Consider the expression `a <|> b <|> c`. Though only three terms are used here, the same logic applies when more alternatives are added.

With a left-associative operator (the current behavior), this will expand to `(a <|> b) <|> c`, equivalent to `Alt (Alt a b) c`. Translating the top-level call to plain English, it seems to say: "if all alternatives before `c` fail, finally try `c`".

If the operator is right-associative, it instead expands to `a <|> (b <|> c)`, or `Alt a (Alt b c)`. This seems to have a clearer English translation: "if `a` fails, try the remaining alternatives".

I think the second one reflects what the programmer intends more often than not.

Another benefit of the new fixity is that both `choice` and chained application of `<|>` will produce the same call structure. `choice` is implemented with `foldr`, so `choice [a, b]` expands to `a <|> (b <|> empty)`.


msmorgan:
This PR obsoletes #4075.

david-christiansen:
How do these new rules compare to other languages with similar notations, such as Haskell?

msmorgan:
@david-christiansen In Haskell, the operator is `infixl 3`. However, the documentation states that `Alternative` is a monoid on applicative functors, which should mean that the fixity shouldn't matter in practice as long as alternatives are attempted left-to-right. Irrespective of the fixity chosen in Idris, its version of `Alternative` will remain a monoid.

Since it's idiomatic in Idris to define dependent versions of well-known operators (as has been done in Text.Parser), fixity needs to be considered more carefully than in Haskell. Using `infixr` results in a parse tree that better represents the intended meaning of a chain of `<|>`.

ahmadsalim:
@msmorgan Can you please solve the merge conflicts? :)

PR #4112, opened 2017-10-10
Refactor stack-shell.nix to use buildStackProject

vaibhavsagar:
Simpler and more robust to future changes, see https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/haskell-modules/generic-stack-builder.nix

### DIFF
--- a/stack-shell.nix
+++ b/stack-shell.nix
@@ -1,9 +1,7 @@
+{ ghc }:
 with (import <nixpkgs> {});
 
 let
-  # MUST match GHC of resolver in stack.yaml
-  ghc = haskell.packages.ghc802.ghc;
-
   native_libs = [
     libffi
     zlib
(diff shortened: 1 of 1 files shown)


ahmadsalim:
@vaibhavsagar Can you please provide a more elaborate description of what this PR does?
I am not too familiar with NixOS.

vaibhavsagar:
This is purely a refactoring that uses the Stack support in Nixpkgs instead of reimplementing it in `stack-shell.nix`. If you look at [the linked file in Nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/haskell-modules/generic-stack-builder.nix), it defines the same functionality and works the same.

ahmadsalim:
I will merge it then.
Thanks for the PR!

vaibhavsagar:
My pleasure :smile:

PR #4113, opened 2017-10-10
Prelude: rework operator fixity for comparison and boolean operators

msmorgan:
This PR includes the following **BREAKING CHANGE**.

`&&` has been given higher precedence than `||`, and both have been made right-associative.

The associativity of `Bool`-producing comparison operators has been removed, and they all have a higher precedence than `&&`.


david-christiansen:
Changing operator precedence and associativity for core things that are used in most programs is a pretty big deal, with the potential to break lots and lots of programs. I think that such pull requests should have some reasoning attached about why this is a good idea, and why the benefits outweigh the drawbacks. A similar question applies to #4111. Even better would be a preceding discussion on the mailing list.

msmorgan:
@david-christiansen I submitted this after a discussion of operators with @ahmadsalim on IRC. I'll add a writeup here to summarize some of the points that were made, along with some further rationale and examples. Afterwards I can submit links to both PRs to the mailing list for further comments.

Do you think the reasoning attached to the other PR is sufficient?

david-christiansen:
The other one makes sense, though it'd be good to have some idea both of how this compares to Haskell and some idea of which programs this will break. We definitely would need to advertise major breaking changes on the core language such as this on the mailing list before doing it.

msmorgan:
### `&&` and `||`

There are two changes to these operators.

Firstly, they have been made `infixr` instead of `infixl`. Since `&&` and `||` are both associative operations this change will not affect functionality. Here are the two ways that `a && b && c && d` could expand.
```idris
((a && Delay b) && Delay c) && Delay d -- current behavior
a && Delay (b && Delay (c && Delay d)) -- this PR
```
Since both `&&` and `||` are pattern-matched on the first argument, with a lazy second argument, the second version appears to be a more appropriate parse tree.

The other change is that the precedence of `&&` has been bumped up from 4 to 5. In most languages I have used (including Haskell) "and" has a higher precedence than "or". This changes the meaning of the expression `a && b || c && d` in the following way.
```idris
((a && Delay b) || Delay c) && Delay d -- current behavior
(a && Delay b) || Delay (c && Delay d) -- this PR
```

msmorgan:
### `==`, `/=`, `<`, `<=`, `>=`, and `>`

`==` and `/=` have had their precedence bumped up to the same level as the others. This has been done for consistency, since all these operators represent comparison, and have the form `a -> a -> Bool`.

This PR changes these operators to be declared as `infix` (that is, neither left- nor right-associative). This means an expression cannot mix these operators without grouping them within parentheses. In all cases except when comparing `Bool`s, chained comparisons make no sense (and even then it's dubious), so failing as early as possible seems like the correct thing to do.

For example, the expression `a < b == c < d` will no longer parse. It is ambiguous to the parser, since it could hypothetically mean either of the following, depending on the types involved. This is by no means an exhaustive list of such groupings.
```idris
sameOrder : (Ord t1, Ord t2) => (a, b : t1) -> (c, d : t2) -> Bool
sameOrder a b c d = (a < b) == (c < d)

weird : Ord t => (a : Bool) -> (b, c : t) -> (d : Bool) -> Bool
weird a b c d = a < ((b == c) < d)
```

There were two expressions, both within `libs/prelude/Prelude.idr`, that were ambiguous in this way (they are using `==` to mean "XNOR"). They have been corrected by adding explicit parentheses. So, although this change may break existing code, the fix is to be more explicit about what you intend to compare.

NOTE: A better error message for this sort of error should be implemented in the parser. The current error is "operator without known fixity" which is confusing. Haskell's error is much clearer. Furthermore, if the error occurs with an `implementation` block, as was the case with the above two errors, the compiler's error message is much less informative.

msmorgan:
After thinking about it some more, perhaps the comparison operators should be `infixl`, after all. It would make it possible to write a module that supports chained comparisons, such as `a < b <= c`, using the default operators.

I've written a prototype of this behavior in [this gist](https://gist.github.com/msmorgan/895c9435ea8023ea4082b9d47204dedb). In order to make the example easier to run and try out, I've used different interface names (starting with `Chain`) and operator names (surrounded by `|`s).

I also made [this version](https://gist.github.com/msmorgan/895c9435ea8023ea4082b9d47204dedb/fb572175470e6ac64904b4ddf3a3a6f131e2c531), which requires the chains of comparisons to be in strictly ascending or descending order. That is, `a > b < c` would not work.

ahmadsalim:
@msmorgan Can you please solve the merge conflicts? I can merge afterwards, since no one complained

ahmadsalim:
@david-christiansen @msmorgan I am still a bit unsure about this one.

Maybe @edwinb can voice his opinion, if he has any?

eraserhd:
FYI, wrt `&&` having a higher precedence than `||` - I very much expect this, and while I haven't hit this with Idris (probably cause I don't do a lot of boolean computation), it feels like it would be a papercut.

I also expect all the comparisons to have the same precedence, though not sure if I would ever encounter that.

ahmadsalim:
OK, I will merge this because there is no protest.
Thanks @msmorgan

msmorgan:
Thanks @ahmadsalim for resolving the conflicts in this and the other branch. I didn't have access to a PC this weekend so it was much appreciated.

PR #4114, opened 2017-10-10
Fix LaTeX pretty printing of backslash, braces and percent

### LINKED ISSUE #1876: Pretty print produces invalid LaTeX for anonymous functions.
The following `Idris` snippet:

``` idris
mkNodeList : List (Document a) -> {auto prf : ValidNode a} -> List $ Document NODE
mkNodeList Nil = Nil
mkNodeList xs = map (\x => Node x) xs
```

when pretty printed using `:pp latex 80`, produces invalid LaTeX for the Verbatim mode used. This is a result of command chars being specified using `commandchars=\\\{\}`. As a result braces and backslashes, used in implicit arguments and anonymous functions, cause Verbatim to think there is a LaTeX command. 

The fix will be to ensure that braces are escaped properly (`\{auto implic\}`), and backslashes are created using the `\textbackslash{}` command.

joom:
Resolves #1876.

I'm not sure about the `toSExp (AnnSyntax c) = SexpList []` part so I need feedback on that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@
 + The --interface option now creates CommonJS modules on the node backend.
 + The C backend now pass arguments to the C compiler in the same order
   as they were given in the source files.
++ Backslash, braces and percent symbols are now correctly pretty printed
+  in LaTeX.
 
 # New in 1.1.1
 
--- a/src/Idris/AbsSyntaxTree.hs
+++ b/src/Idris/AbsSyntaxTree.hs
@@ -1697,6 +1697,9 @@ pprintPTerm ppo bnd docArgs infixes = prettySe (ppopt_depth ppo) startPrec bnd
   where
     startPrec = 0
     funcAppPrec = 10
+    lbrace = annotate (AnnSyntax "{") (text "{")
+    rbrace = annotate (AnnSyntax "}") (text "}")
+    percent = annotate (AnnSyntax "%") (text "%")
 
     prettySe :: Maybe Int -> Int -> [(Name, Bool)] -> PTerm -> Doc OutputAnnotation
     prettySe d p bnd (PQuote r) =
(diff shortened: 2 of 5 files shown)


joom:
Following @ahmadsalim's suggestion, I tried writing a test for this, but my test failed. When I diff the output and the expected, I get this:
```
15c15
< \begin{Verbatim}[commandchars=/\{\}]
---
> \begin{Verbatim}[commandchars=\\\{\}]
```

That is unusual because even when I do `./run > expected` and then run the tests, I get the same result. In other words, the problem is not with my code or the test I wrote, but with the testing system itself. Then I discovered [the `normalise` function in `TestRun.hs`.](https://github.com/idris-lang/Idris-dev/blob/4e00dc2fa433ae6d58310fda0e1f4e9a234f3ecd/test/TestRun.hs#L99-L104) I can't think of any way of bypassing it with small changes to `TestRun.hs`. Any suggestions?

I also don't think testing pretty printed code generation makes much sense, given that even changes to the CSS in the HTML, or spacing in LaTeX command definitions can break the test.

ahmadsalim:
@joom Hmm, the `normalize` function seems a bit weird indeed.

OK, I will merge your PR without a test then. Perhaps, just an update to the CHANGELOG please? 😄

melted:
Normalize just makes filepaths compare the same. If you create the expected with update it's automatically taken into account.

joom:
@melted Does creating the expected with update mean manually changing `\\\` to `/` in every pretty printing test? I think that might get confusing in the future for other people who might want to write more pretty printing tests.

I updated the changelog. I haven't added the test yet, but I can if you think what I said above seems fine to you.

david-christiansen:
Interesting solution, using the annotation system. I hadn't thought to do it that way.

Can we use this annotation for anything else that's interesting? E.g. balancing delimiters in console output a la Emacs's rainbow parens?

melted:
@joom I don't think a test is required here., so I'll merge. 

It used to be that the normalization was done on both expected and output when they were compared. So I'm a bit surprised that there is a problem with it.

PR #4115, opened 2017-10-11
Add Alternative implementation for StateT

joom:


ahmadsalim:
Thanks for the PR.
Could this not work with all `MonadTrans` ?

joom:
For the definition of `empty = lift empty`, yes. But I don't see how that would work for `(<|>)`, it seems like we need a way to access the inner structure of `StateT`, either with the constructor `ST` directly, or with `runStateT`.

I checked Haskell's `mtl` library, and they also just have a `(Functor m, MonadPlus m) => Alternative (StateT s m)` instance, which is pretty much the same thing.

ahmadsalim:
Ah, right, thanks!

PR #4116, opened 2017-10-11
 Added various combinators, most of which are also present in Haskell

Synthetica9:


libs/contrib/Data/Combinators/Applicative.idr
```diff
@@ -0,0 +1,27 @@
...
+||| for function composition:
+module Data.Combinators.Applicative
+
+import Data.Combinators
+
+%default total
+%access public export
+
+||| Pure is just const, also known as the K combinator
+pure : a -> b -> a
```
LeifW: `pure : Applicative f => a -> f a` is already defined as a member of the `Applicative` typeclass in Prelude.

eraserhd:
I think the Arrow ones are all already present in `Control.Arrow` in `base`.

Synthetica9:
While you are right, these are only defined or `Morphism`, not over bare
`(->)`, because the compiler doesn't allow this.

ahmadsalim:
Thanks for the PR.
Since these are in `contrib` I will just merge them.

Consider adding dependent versions as well.

PR #4117, opened 2017-10-11
Added a Haskell-like Down interface

Synthetica9:


libs/prelude/Prelude/Interfaces.idr
```diff
@@ -73,6 +73,21 @@ Eq Ordering where
...
+inverse : Ordering -> Ordering
+inverse LT = GT
+inverse EQ = EQ
+inverse GT = LT
+
+-- https://hackage.haskell.org/package/base-4.10.0.0/docs/Data-Ord.html#t:Down
+data Down a = MkDown a
+
+Eq a => Eq (Down a) where
+  Down x == Down y = x == y
```
msmorgan: Needs to be `(Down x) == (Down y)`.

msmorgan: Should be `(Down x) == (Down y)`

msmorgan: Should be `(Down x) == (Down y)`.

ahmadsalim:
Can you please provide a description of what this does?

libs/prelude/Prelude/Interfaces.idr
Synthetica9: Fixed.

Synthetica9:
@ahmadsalim It allows you tocompare on the inverse of an ordering. This is useful for (for example) inversing sorts.

For example: 

```
> MkDown 3 < MkDown 2
True : Bool
```
It also adds the `inverse` function, which is basically `not` on orderings. (`GT` -> `LT`, `EQ` -> `EQ`, `LT` -> `GT`)

ahmadsalim:
@Synthetica9 Thanks for the PR! I think this belongs more to `contrib` than `prelude`.

If you want it in `prelude` then there is a more strict review requirements, and you should send a mailing list update, etc.

ahmadsalim:
I closed the PR, you are welcome to add it to contrib if you prefer instead with a new PR.

PR #4127, opened 2017-10-16
Fix tests for ASCII characters in RTS

### LINKED ISSUE #3767: Incorrect behaviour when using 'singleton' and 'chr'
I compiled and executed this code with the `c` backend using Idris 1.0
This bug doesn't happen in the REPL and I only see the issue when combing `singleton` and `chr`.

```
module main
main : IO ()
main = do
  printLn "\x141"
  printLn (chr 0x141)
  printLn (singleton '\x141')
  printLn (singleton (chr 0x141))
```
The output is this:
```
"\321"
'\321'
"\321"
"A"
```
However if you replace `x141` with `x1B1` the output is this:
```
"\433"
'\433'
"\433"
"\433"
```
I looked into it and whenever bit 8 is zero on the input integer, the integer is masked so that only the 8 least significant bits are used. So in the above example, `0x141` was changed to `0x041` which is the letter `A`, but `0x1B1` has bit 8 set to one so it doesn't happen.

asandroq:
Since characters are encoded as ints, instead of just testing if a bit
is not set, the condition must test if the character is in range.

Fixes #3767


melted:
Thanks for the PR!

PR #4130, opened 2017-10-17
Sort module lists in package files

msmorgan:
Also updated indentation to be consistent throughout.

### DIFF
--- a/libs/base/base.ipkg
+++ b/libs/base/base.ipkg
@@ -1,48 +1,43 @@
 package base
 
 opts = "--nobasepkgs --total --no-elim-deprecation-warnings --partial-eval -i ../prelude"
-modules = System,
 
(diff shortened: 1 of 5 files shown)


libs/base/base.ipkg
```diff
@@ -1,48 +1,42 @@
...
+        , Data.Morphisms
+        , Data.Nat.Views
+        , Data.Primitives.Views
+        , Data.So
+        , Data.String
+        , Data.String.Views
+        , Data.Vect
+        , Data.Vect.Quantifiers
+        , Data.Vect.Views
+        , Debug.Error Debug.Trace
```
melted: Crashes here

msmorgan: Thanks, fixed!

ahmadsalim:
Thanks for the PR

PR #4132, opened 2017-10-18
Rename `contrib:Data.String`, was shadowing `base:Data.String`

Melvar:
Also add the resulting `Data.String.Extra` to `contrib.ipkg`, as the module was never actually listed and was thus inaccessible outside of `contrib` itself.


ahmadsalim:
Thanks!

Issue #4134, opened 2017-10-19
Functions using `divides` do not reduce


clayrat:
# Steps to Reproduce
```
import Data.Primitives.Views

%default total
%access public export

half : Integer -> Integer
half n with (divides n 2)
  half ((2 * div) + mod) | DivBy prf = div

test : half 4 = 2
test = Refl
```
# Expected Behavior
Everything typechecks.
# Observed Behavior
On version `1.1.1-git:7936fbb` (commit from 19/10/17) this gives: 
```
bug.idr:11:6:When checking right hand side of test with expected type 
        half 4 = 2                                                    
                                                                      
Type mismatch between                                                 
        2 = 2 (Type of Refl)                                          
and                                                                   
        with block in Main.half 4 (divides 4 2) = 2 (Expected type)   
                                                                      
Specifically:                                                         
        Type mismatch between                                         
                2                                                     
        and                                                           
                with block in Main.half 4 (divides 4 2)               
```
Last version this worked on was that I checked was `1.1.1-git-f436c5f` (commit from 13/09/17).

ahmadsalim:
@clayrat This is because `divides` is marked as `export`, so this is expected behaviour.

You are welcome to change it to `public export` if you prefer.

PR #4135, opened 2017-10-20
Make `divides` views public

clayrat:
So that functions using them can reduce in typechecking (fixes https://github.com/idris-lang/Idris-dev/issues/4134).

I'm not sure why `int{eger}Rec` and corresponding structures do not have any modifiers, should we add `public export` to these as well?


ahmadsalim:
@clayrat You are welcome to add modifiers, but please also use `really_believe_me` so they can reduce in types.

ahmadsalim:
Thanks!

Issue #3740, opened 2017-04-05
Inconsistent behaviour of div between REPL and compiled version


dalcde:
If I run
 
    > div (-3) 5

in the REPL, then the result is `-1`. If I compile the file

    main : IO()
    main = putStrLn . show $ div (-3) 5

and run it, then it prints `0`.

This may or may not be related to #3592, but in light of that, the behaviour of the compiled version is desired, I think.

ctford:
A good way to quickly check this is to use `:x` vs `:exec`:

```
Idris> :x div (-3) 5
-1 : Integer
Idris> :exec div (-3) 5
0
```
To me, the first duty of the REPL is to accurately reflect the real C runtime. Anything else is misleading to users.

There's a separate question of updating the C runtime as mentioned in #3592. Perhaps that could be fixed at the same time if it's easy enough and doesn't have any extra implications.

david-christiansen:
> To me, the first duty of the REPL is to accurately reflect the real C runtime

Rather than the C backend, the REPL should accurately represent the type checker's notion of computation. For all total closed programs, the C backend should additionally accurately implement the REPL's functional semantics, but not necessarily extra-functional things like evaluation order and cost models. The needs of normalizing open terms inside types to compare them for equality are different from the needs of running closed programs in compiled code, after all.

Likewise, because the interpreter in `:x` is really designed for the type provider feature, it should accurately emulate the C backend's I/O behavior, but it does not follow the same cost model at all. This is because a big part of compiling Idris is erasing things used only for type checking, but we need to preserve those things in type providers because they are intended to be used in type checking later.

However, `:x` and `:exec` should certainly provide the same answers for closed expressions, so this is definitely a bug.

PR #4137, opened 2017-10-20
Make `div` and `mod` behave consistently

### LINKED ISSUE #3592: Incorrect implementation of div and mod
The functions `div` and `mod` of `Integer` (and `Int` etc.) do not satisfy

    (div a b) * b + (mod a b) = a

since in the backend, the Haskell function `div` is used for `div` while `rem` is used for `mod`, where in fact `div` should be paired with `mod` and `quot` should be paired with `rem` for the above equation to hold.

This causes issues when implementing, say, the Euclidean algorithm.

asandroq:
These functions now satisfy the following equality for all builtin
integral types:

```
(div a b) * b + (mod a b) = a
```

Fixes #3592, #3740


ahmadsalim:
Thanks for the PR! Can you please also fix the node tests :).

asandroq:
@ahmadsalim It seems it's going to work this time. :)

ahmadsalim:
Thanks!

PR #4143, opened 2017-10-23
Fix generation of integer literals in the C backend

### LINKED ISSUE #4142: C backend generate invalid big integer literal
# Steps to Reproduce

Compile the following source file with the C backend:

```idris
module Main

main : IO ()
main = do
  printLn $ div (-432642342742368327462378462387) 36473264372
```

# Expected Behavior

Idris will generate correct C code to be compiled by a C compiler.

# Observed Behavior

The C code cannot be compiled:

```
[alex@bree test (master *>)]$ idris -o testcase testcase.idr
/var/folders/6s/8k5_5k5n5bd5l3180xg5jl4w0000gn/T/idris25619-0.c:270:21: error: integer literal is too large to be represented in any integer type
    LOC(3) = MKINT(-432642342742368327462378462387);
                    ^
1 error generated.
```

asandroq:
Negative bignum literals were being generated as Ints.

Fixes #4142.


melted:
Thanks!

src/IRTS/CodegenC.hs
```diff
@@ -179,7 +179,7 @@ bcc i (ASSIGNCONST l c)
     = indent i ++ creg l ++ " = " ++ mkConst c ++ ";\n"
   where
     mkConst (I i) = "MKINT(" ++ show i ++ ")"
-    mkConst (BI i) | i < (2^30) = "MKINT(" ++ show i ++ ")"
+    mkConst (BI i) | i > ((-2)^30) && i < (2^30) = "MKINT(" ++ show i ++ ")"
```
byorgey: @asandroq @melted Isn't `(-2)^30 == 2^30`?  So this condition is never true, and the code generator will now *always* generate a `MKBIGC`.  It should probably say `-(2^30)` instead?  (As an aside, why is it `2^30` rather than `2^31`?)

asandroq: Well, this is embarrassing. I am going to make another PR to fix this one.

asandroq: I pushed a fix in another PR, I hope I got this right this time. The reason for using 2^30 is that the last bit of a C int is used to tag it so it can be unboxed in the runtime.

PR #4147, opened 2017-10-24
Fix generation of bignum literals

asandroq:
My previous attempt added an impossible condition that output all
bignum literals using the `MKBIGC` macro even when the value would fit
into a C int. This change fixes that bug.


src/IRTS/CodegenC.hs
```diff
@@ -179,8 +179,10 @@ bcc i (ASSIGNCONST l c)
     = indent i ++ creg l ++ " = " ++ mkConst c ++ ";\n"
   where
     mkConst (I i) = "MKINT(" ++ show i ++ ")"
-    mkConst (BI i) | i > ((-2)^30) && i < (2^30) = "MKINT(" ++ show i ++ ")"
-                   | otherwise = "MKBIGC(vm,\"" ++ show i ++ "\")"
+    mkConst (BI i) = let maxInt = 2^30
```
ahmadsalim: On C at least the maximum integer is `INT_MAX` which is equal to 2^31-1.
The minimum integer is `INT_MIN` which is equal to -2^31.

see https://www.tutorialspoint.com/c_standard_library/limits_h.htm

ahmadsalim: Well for 32-bit integers anyway.

asandroq: That's true, but the last bit of an int is used to tag it as an unboxed value in the runtime.

ahmadsalim: Ah, I see.

Melvar: Well, since the size is based on a pointer, it’s either 31 or 63 bits depending on platform.

melted:
It's slightly irritating that we can't reliably know if a long is 32 or 64 bits at codegen time. But big literals are rare, so it's way better to go with 32 bits than to hack something that could fail.

src/IRTS/CodegenC.hs
```diff
@@ -179,8 +179,10 @@ bcc i (ASSIGNCONST l c)
...
+                     in if i >= -maxInt && i < maxInt
```
melted: You could use `if i < abs 2^30` for simplicity

ahmadsalim: You mean `abs i < abs 2^30`

asandroq: The test is different on either side of the interval.

melted: No I put the abs on the wrong side, 2^30 don't need an abs, it's a constant :-)

ahmadsalim: or `abs i < 2^30`

clayrat: spoiler alert: `x^30 = abs (x^30)` for any real `x` :D

ahmadsalim:
Thanks

PR #4151, opened 2017-10-24
add some new functions to Data.String.Extra

msmorgan:
Includes `drop`, `dropLast`, `take`, `takeLast`, `shrink`, `join`, and `index`.


melted:
Thanks!

PR #4154, opened 2017-10-26
refactor Text.Parser and Text.Lexer

msmorgan:
There's a lot in this one.

* Split mostly-private core modules into Text.Parser.Core and
  Text.Lexer.Core.
* Refactor many combinators in Text.Parser to accept more inputs.
* New parser combinators, including `choose`, `someTill`, `manyTill`,
  `afterSome`, `afterMany`, `sepEndBy1`, and `endBy1` (and friends).
* Flip the argument order of `Text.Parser.parse`. The old version still
  exitsts, but is deprecated in a different namespace.
* Rename `optional` and `maybe` in Text.Parser to `option` and `optional`,
  to be more in line with Haskell parser combinator library semantics.
* `concat` for `Lexer`s.
* Various documentation updates.


libs/contrib/Data/Bool/Extra.idr
```diff
@@ -0,0 +1,28 @@
...
+%default total
+
+andDelaySameNeutral : (x : Bool) -> x && Delay x = x
+andDelaySameNeutral False = Refl
+andDelaySameNeutral True = Refl
+
+andDelayFalseFalse : (x : Bool) -> x && Delay False = False
+andDelayFalseFalse False = Refl
+andDelayFalseFalse True = Refl
+
```
clayrat: Why do you have explicit `Delay`s here, don't logical ops automatically insert that?

msmorgan: I assumed that the errors were due to the second argument being lazy, so I wrote it that way for clarity. This assumption may have been incorrect, so I'll do a little digging and revise if I was wrong.

msmorgan: Turns out it was a regular old reduction issue. I tried strict versions of `&&` and `||` and had the same result. I'll rename the functions accordingly and fix their types.

Do you think it would be a bad idea to declare the equality in the type in reverse, to avoid using `sym` everywhere? I'm afraid it wouldn't be as semantically correct.

libs/contrib/Control/Delayed.idr
```diff
@@ -0,0 +1,29 @@
+||| Utility functions for contitionally delaying values.
+|||
+||| NOTE: Import this module with `import public` if exporting functions
```
ahmadsalim: I think this NOTE should be removed, and converted to an issue.

ahmadsalim:
@msmorgan Thanks for the refactoring, most things seem useful.

A note however: I can not particularly see the benefit of the swap of arguments for parse. In any case, the deprecation is still going to break code because it is moved to a new namespace, so I am unsure whether this is the way to go.
Would it be possible to have this change in another separate PR?

libs/contrib/Text/Lexer/Core.idr
```diff
@@ -0,0 +1,190 @@
...
+
+||| Combine each lexer in a container into a lexer that sequences them.
+||| The resulting lexer always fails if the container was empty.
+export
+concat : Foldable t => t Lexer -> Lexer
+concat xs = foldl op Fail xs
+  where
+    op : Lexer -> Lexer -> Lexer
+    op Fail x = x
+    op x y = x <+> y
```
Melvar: Normally, `concat` and `choice` do the same thing, except that `concat` involves a `Type` and `choice` involves a `Type -> Type`. It happens that `Lexer` has two monoidal structures on it; the “additive” one with the operation being alternation and the identity element failure, and the “multiplicative” one with the operation being concatenation and the identity element being, not failure, but a lexer that always succeeds with no result.

Melvar: See [Wikipedia on Semiring of formal languages](https://en.wikipedia.org/wiki/Semiring#formal_languages). In the context of `Applicative` and `Alternative`, the multiplicative counterpart of `choice` would be `sequence_`/`sequence`.

libs/contrib/Data/Bool/Extra.idr
clayrat: Why would you use `sym`, are you rewriting `False` back to `x && False`?

msmorgan: I don't think so... But each use in libs/contrib/Text/Parser.idr needs a `sym`.

libs/contrib/Text/Lexer/Core.idr
msmorgan: I did it this way so that the return value could be `Lexer`. Always succeeding would mean that the return type would have to be `Recognize False` instead, which isn't a problem except I will have to check that the strings are not empty in `exact` and `approx`.

libs/contrib/Data/Bool/Extra.idr
clayrat: Ah, so you indeed are rewriting backwards: e.g., in `option` you convert `Grammar tok False a` to `Grammar tok (c && False) a` so that `<|>` could be applied. You could also use `replace (andFalseFalse c) (p <|> pure def)` but that would most likely require writing the `P` implcit term, which I feel is somewhat ugly.

msmorgan:
@ahmadsalim I could move the deprecated `parse` to Text.Parser instead of Text.Parser.Core.Deprecated.

I did it because it was inconsistent with lex. Consider that the grammar specifies how to parse, and the token list specifies what to parse. Similarly, the token map specifies how to lex, and the string specifies what to lex. Previously they had the forms `parse what how` vs `lex how what`. I flipped parse because I think the lex version is clearer and works better with currying.

libs/contrib/Text/Lexer/Core.idr
msmorgan: Actually, there is not a way to succeed if the foldable is empty, unless the return type is `(c : Bool ** Recognise c)`.

Melvar: There are at least two ways to fix it, but both require you take a `List` rather than any `Foldable`. One is using `NonEmpty`, and the other is specifying the return value using `isCons`.

Also, do you not want to be able to sequence a collection of `Recognise False`s?

msmorgan: I went with the `isCons` variant. Both that and `NonEmpty` are somewhat cumbersome to use, but `isCons` is more flexible.

Melvar:
My comment on the name of `concat` still applies: I think it should be named `sequence` to indicate its relationship to `choice`.

msmorgan:
@Melvar I disagree about the name. Lexers do not have result values, and semantically sequence involves collecting result values.

I chose the names I did because of the following (perhaps too simplistic) logic:
* `choice`: fold using `<|>`
* `concat`: fold using `<+>`

melted:
That function is nothing like (++), concat is a confusing name for it.

msmorgan:
What does concat have to do with `++`, other than the String implementation of Monoid?
https://github.com/idris-lang/Idris-dev/blob/master/libs/prelude/Prelude/Foldable.idr#L38

Furthermore, I think it is similar. `is 'a'` will recognise the input `"a"`.
`is 'b'` will recognise the input `"b"`.
`is 'a' <+> is 'b'` will recognise the input `"a" <+> "b"` (`"ab"`).

Melvar:
Well, `<+>` is really also the wrong name. In any type implementing both `Monoid` and `Alternative`, they should function identically; essentially `Alternative` is a workaround for the fact that a constraint of `{a : Type} -> Monoid (f a)` can’t be found by the implementation search.
Thus, the obvious operator for sequencing of recognisers should be `*>`, and the function you name `concat` should be `sequence_`, but having `sequence_` without a corresponding `sequence` seemed odd to me, so I proposed using `sequence` instead. Your new `concatMap` corresponds to `traverse_`.

msmorgan:
@Melvar You'll have to take it up with @edwinb about `<+>`. I maintain that it's a better choice than `*>`, since the latter operator has the semantics of discarding the value of the first action. But since recognisers produce no value upon success&mdash;they simply succeed, having consumed some amount of input&mdash;this doesn't seem to fit.

If either one of the operators defined for `Recognise` is incorrect, I would say it's `<|>`, since `Recognise` isn't a `Type -> Type`. The boolean `consumes` flag is not determined when the lexer runs, but rather when it is assembled, and states whether the recogniser will always consume input upon success.

However, there's a key difference between  the Alternative operator (`<|>`) and the Applicative sequencing operators (`<*`, and `*>`) that makes `<|>` more semantically consistent for recognisers: `<|>` requires its arguments to produce the same type of value upon success. Since recognisers never produce a value, all recognisers vacuously have the same value type.

I think it's important to note that `Recognise` does not strictly conform to any Haskell typeclass because of its dependent nature. It is monoid-like&mdash;in two modes, like you mentioned earlier&mdash;but not really a monoid in either case because both operations are dependent. This means the operator choice is somewhat arbitrary, and falls to the library author to determine what is most semantically correct. It seems most appropriate to treat `Recognise` as an almost-monoid, rather than as an almost-applicative.

Melvar:
Okay, so if you don’t want to treat `Recognise` as morally an `Applicative`+`Alternative`, you must treat it as morally a `Semiring`, so alternation is denoted by `<+>` and sequencing by `<.>`, and `choice` and `sequence` are replaced by `sum` and `product`.

msmorgan:
I understand your concern. However, `Recognise` itself is not really a single type. It is two related types, `Recognise False` and `Recognise True`, which can be combined in dependent ways. The structures you are describing do not have this property, so it comes down to what are we trying to be consistent with. I think it's most important to be consistent in semantic meaning of operations, and what we have now achieves this. You're letting dubious mathematical rigidity get in the way of semantic elegance.

Melvar:
If I was going for mathematical rigidity, I would insist on the `Semiring`-like interpretation. No, but the use of certain operators implies that they follow an approximation of the rules that their interface counterparts do. In this case, I believe that `<+>` and `<|>` both existing but doing different things is *surprising* and will trip up someone who knows their interface counterparts.

ahmadsalim:
@Melvar Is there anything blocking this besides the choice of `<+>`, which you said should be a separate issue?

Melvar:
@ahmadsalim Not from my view, feel free to merge if you are satisfied.

ahmadsalim:
@Melvar Thanks

@msmorgan I think what I said about parse still holds, it would be not useful to have the old version under Deprecated since it would break existing code anyway.

ahmadsalim:
@msmorgan Thanks, can we remove all the things under the `Deprecated` namespace please, and just deprecate them under their previous namespace?

msmorgan:
@ahmadsalim Unfortunately that is not possible with `optional`, since there's another function called `optional` in that namespace already.

msmorgan:
The new module `Text.Quantity` could also be used by a future Regex module.

ahmadsalim:
@msmorgan The test fail now.

PR #4156, opened 2017-10-27
adjust fixities of several common operators

msmorgan:
Corrects several fixity problems throughout the standard library. Most notably, `<|>` had a higher precedence than `<*>`, and `>>=` had a higher precedence than both.

### DIFF
--- a/libs/contrib/Interfaces/Correlative.idr
+++ b/libs/contrib/Interfaces/Correlative.idr
@@ -5,7 +5,7 @@ import Data.Vect
 %access public export
 %default total
 
-infixl 2 </>
+infixl 3 </>
 
 ||| A Correlative functor is a functor where, given `(xs, ys : Correlative f)`,
 ||| certain elements from xs and ys can be paired with one another due to some
@@ -13,11 +13,11 @@ infixl 2 </>
 interface Functor f => Correlative (f : Type -> Type) where
   (</>) : f (a -> b) -> f a -> f b
 
-infixl 2 </
+infixl 3 </
 (</) : Correlative f => f a -> f b -> f a
 x </ y = map const x </> y
 
-infixl 2 />
+infixl 3 />
 (/>) : Correlative f => f a -> f b -> f b
 x /> y = map (const id) x </> y
 
--- a/libs/prelude/Prelude/Applicative.idr
+++ b/libs/prelude/Prelude/Applicative.idr
@@ -12,17 +12,17 @@ import Prelude.Functor
 
 ---- Applicative functors/Idioms
 
-infixl 2 <*>
+infixl 3 <*>
 
 interface Functor f => Applicative (f : Type -> Type) where
     pure  : a -> f a
     (<*>) : f (a -> b) -> f a -> f b
 
-infixl 2 <*
+infixl 3 <*
 (<*) : Applicative f => f a -> f b -> f a
 a <* b = map const a <*> b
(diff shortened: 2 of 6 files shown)


libs/prelude/Prelude/Basics.idr
```diff
@@ -29,7 +29,7 @@ fst (x, y) = x
 snd : (a, b) -> b
 snd (x, y) = y
 
-infixl 9 .
+infixr 9 .
```
ahmadsalim: Why the re-association?

libs/prelude/Prelude/Functor.idr
```diff
@@ -6,15 +6,15 @@ import Prelude.Basics
...
 interface Functor (f : Type -> Type) where
-    ||| Apply a function across everything of type 'a' in a 
+    ||| Apply a function across everything of type 'a' in a
     ||| parameterised type
     ||| @ f the parameterised type
     ||| @ func the function to apply
     map : (func : a -> b) -> f a -> f b
 
-infixl 4 <$>
+infixr 4 <$>
```
msmorgan: Since `(->) a` is not a functor in Idris, chaining doesn't work with the current fixity.
```idris example
-- this now works:
(+4) <$> (*2) <$> Just (the Int 10)
```
An alternate way to solve this would be to define another version of `(<$>)`, as an alias for `(.)`.

libs/prelude/Prelude/Monad.idr
```diff
@@ -11,7 +11,7 @@ import IO
 
 %access public export
 
-infixl 5 >>=
+infixl 1 >>=
```
msmorgan: In Haskell this operator has a lower precedence than `<*>` and `<|>`. There's good arguments for both lower and higher precedence, but having one of the lowest precedences means it will not conflict with most other operators.

It's not used much outside of `do` notation, so any impact this change has should be minor.

libs/prelude/Prelude/Basics.idr
msmorgan: Three minor reasons, which may or may not add up to a good enough reason to change it.
* It now matches Haskell's associativity of this operator.
* It better represents the order in which functions will be applied (`f . (g . (h . i))` vs old `((f . g) . h) . i`).
* There is no difference in functionality (no pun intended).

libs/contrib/Interfaces/Correlative.idr
```diff
@@ -5,19 +5,19 @@ import Data.Vect
 %access public export
 %default total
 
-infixl 2 </>
+infixl 3 </>
```
msmorgan: To match `<*>`.

libs/prelude/Prelude/Applicative.idr
```diff
@@ -12,17 +12,17 @@ import Prelude.Functor
 
 ---- Applicative functors/Idioms
 
-infixl 2 <*>
+infixl 3 <*>
```
msmorgan: It makes sense for the alternative operator to have lower precedence than an operator that might be used to create an alternative. `pa *> a <|> pb *> b` meaning `(pa *> a) <|> (pb *> b)` seems to make more sense intuitively than `p1 <|> p2 *> x` meaning `(p1 <|> p2) *> x`. Also, it's this way in Haskell.

ahmadsalim:
Thanks for the comments.

PR #4160, opened 2017-10-28
added ST.Random utilitary functions

rbarreiro:
Added
+ rndInt
+ rndFin
+ rndSelect'
+ rndSelect

### DIFF
--- a/libs/contrib/Control/ST/Random.idr
+++ b/libs/contrib/Control/ST/Random.idr
@@ -21,7 +21,7 @@ rndInt : (rnd : Var) -> Integer -> Integer -> ST m Integer [rnd ::: Random]
 rndInt rnd lower upper = do v <- getRandom rnd
                             pure $ assert_total ((v `prim__sremBigInt` (upper - lower)) + lower)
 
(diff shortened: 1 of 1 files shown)


Melvar:
`rndSelect` should probably require `NonEmpty` and leave off the `Maybe`.
Also, documentation is in Markdown, I’m not sure what you thought it was.

rbarreiro:
This is a copy of Effects.Random

I can try to switch to NonEmpty instead of Maybe.
Regarding de documentation I just copy pasted it, what is wrong with it?

ahmadsalim:
@rbarreiro Please change to `NonEmpty` instead.

I can't see what is immediately wrong with the documentation.
@Melvar can you please provide a hint there?

Melvar:
```
Generate a random number in Fin (S `k`)
```
This uses backticks on only a variable name, rather than the entire code span `Fin (S k)`, suggesting the person writing it was expecting them to mean something different than they mean in Markdown. Other code spans are not marked up at all.

ahmadsalim:
Thanks

PR #4165, opened 2017-10-31
Data.List.isElem: Extracted mkNo to a top-level function Fixes #4161

### LINKED ISSUE #4161: Data.List.isElem extremely slow for large Lists (k>10)
The function ```Data.List.isElem``` is getting extremely slow with an increasing number of elements in the list. The 
reason for this is the inlining of the ```mkNo``` function. Extracting ```mkNo``` to a top-level function solves the problem.
For ```Data.Vect.isElem```, this is actually already implemented like this (see function ```neitherHereNorThere``` ) and makes ```Data.Vect.isElem``` much faster than ```Data.List.isElem``` (see Expected Behavior).

# Steps to Reproduce
It can be reproduced easily in the repl following these steps:
```idris
-- fast
isElem 'A' ['a'..'c'] 

-- starting to get slow
isElem 'A' ['a'..'k'] 

-- not usable any more
isElem 'A' ['a'..'z'] 
```
# Expected Behavior
Data.List.isElem should be as fast as Data.Vect.isElem.
```idris
:module Data.Vect
-- fast
Data.Vect.isElem 'A' (Data.Vect.fromList ['a'..'c'])

-- fast
Data.Vect.isElem 'A' (Data.Vect.fromList ['a'..'k'])

-- still fast
Data.Vect.isElem 'A' (Data.Vect.fromList ['a'..'z'])
```
# Observed Behavior
Data.List.isElem starting to get slow for lists > 10 elements.

janschultecom:
Fixes #4161. Extracted mkNo to a top-level function for performance improvements

### DIFF
--- a/libs/base/Data/List.idr
+++ b/libs/base/Data/List.idr
@@ -19,6 +19,11 @@ implementation Uninhabited (Elem {a} x []) where
      uninhabited Here impossible
      uninhabited (There p) impossible
 
+||| An item not in the head and not in the tail is not in the List at all
+neitherHereNorThere : {x, y : a} -> {xs : List a} -> Not (x = y) -> Not (Elem x xs) -> Not (Elem x (y :: xs))
+neitherHereNorThere xneqy xninxs Here = xneqy Refl
+neitherHereNorThere xneqy xninxs (There xinxs) = xninxs xinxs
+
 ||| Is the given element a member of the given list.
 |||
 ||| @x The element to be tested.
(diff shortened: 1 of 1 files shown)


ahmadsalim:
There seems to be an error with Travis, but since AppVeyor succeeded and it is a simple library change I will merge. Thanks for the PR!

janschultecom:
Awesome! 👍 Thx!

janschultecom:
@mpickering I agree there must be some deeper, underlying problem here that probably might show up in other places as well. I'm not really into the compiler internals, but I guess it might have something to do with the repeated declaration of mkNo in the where block due to the recursion!? You can clearly note some kind of exponential explosion with the mkNo definition.

PR #4171, opened 2017-11-02
Use a modern ghc for appveyor

melted:


melted:
We should create a ghc 8.2 target for travis too.

melted:
Ok, so pkg004 flakes out. I can see that locally too.

eraserhd:
@melted 
Haha, this error is expected, I think.  Because it's checking whether there are any system directory separators, and apparently in the new GHC, "/" is not a system directory separator, and the backslash in `test/pkg004/test.ipkg` is escaping the "a" and is therefore an "invalid character".

It should pass if you push a commit doubling the backslash.

melted:
Fixed in #4170

PR #4174, opened 2017-11-03
Use native `atan2` function from backend

### LINKED ISSUE #3497: Wrong implementation of atan2 in Doubles.idr
Idris provides two arc-tangent functions:
- `atan: Double -> Double` and
- `atan2: Double -> Double -> Double`.

Both return an angle, but their images in the codomain differ.  `atan` is limited to half the unit circle, while `atan2` covers the whole unit circle.  `atan2` is able to determine the correct quadrant of an angle by looking at the sign of each of its arguments.  When both arguments are positive, then `atan2 y x = atan (y/x)`, but as soon as one is negative, the results differ.

The current implementation of `atan` and `atan2` in `Doubles.idr` are

```
atan : Double -> Double
atan x = prim__floatATan x

atan2 : Double -> Double -> Double
atan2 y x = atan (y/x)
```

By calling `atan(y/x)` in all cases, even when `x` or `y` is negative, this version of `atan2` loses sign information and therefore does not fulfill its job.

I do not know whether there exists a `prim__floatATan2` that you can use.  If not, Wikipedia describes the algorithm clearly https://en.wikipedia.org/wiki/Atan2 .

(Edit: grammar)

asandroq:
The implementation of `atan2` in Idris was broken. Instead of fixing
the implementation the native function of the backend was used instead
because it's both correct and faster.

Fixes #3497.


ahmadsalim:
Thanks, can you please also make an entry to the CHANGELOG?

PR #4177, opened 2017-11-03
rename primitive types and add aliases

### LINKED ISSUE #4175: MkDPair shorthand cannot use primitive types to the left of `**`
# Steps to Reproduce

Try the following expressions in the REPL:
```idris example
-- working
the (DPair Type id) (MkDPair String "hello")
the (DPair Type id) (MkDPair Int 42)
the (a ** a) (MkDPair Double 3.14)
the (a ** a) (MkDPair Integer 100)
the (a ** a) (Nat ** 3)
the (a ** a) (Float ** 3.14) --deprecation warning

-- not working
the (DPair Type id) (String ** "hello")
the (a ** a) (Int ** 42)
the (a ** a) (Double ** 3.14)
the (a ** a) (Integer ** 100)
```

# Expected Behavior

Expressions in both categories should work

# Observed Behavior

Each of expressions in the "not working" category fails with an error in the following form:
```
When checking argument x to constructor Builtins.MkDPair:
        No such variable String
```

msmorgan:
Fixes #4175.

They all highlight green now, so maybe this isn't the best idea...

### DIFF
--- a/libs/prelude/Builtins.idr
+++ b/libs/prelude/Builtins.idr
@@ -78,38 +78,47 @@ void : Void -> a
 ||| Arbitrary-precision integers
 Integer : Type
 Integer = prim__Integer
+%used Integer
 
 ||| Fixed-precision integers of undefined size
 Int : Type
 Int = prim__Int
+%used Int
 
 ||| Characters in some unspecified encoding
 Char : Type
 Char = prim__Char
+%used Char
 
 ||| Double-precision floating-point numbers
 Double : Type
 Double = prim__Double
+%used Double
 
 ||| Strings in some unspecified encoding
 String : Type
 String = prim__String
+%used String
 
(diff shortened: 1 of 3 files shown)


ahmadsalim:
@msmorgan would it not now show `prim__Integer` etc. in the REPL?

msmorgan:
@ahmadsalim No, I think that is determined by the `Show` instance for `Const` in TT.

The build error should be fixable with `%used`, right?

I'm currently investigating a different approach that allows these named constants to be looked up by name in the Context, eliminating the special treatment by the parser.

ahmadsalim:
@msmorgan Sorry, I have not worked much with primitives, so I do not know.

msmorgan:
Clearly I have no idea what I'm doing.

PR #4178, opened 2017-11-06
Add Functor implementation for FunClause and FunDefn

joom:
If `Binder` has a `Functor` implementation, why don't `FunClause` and`FunDefn`? It's helpful if the user is trying to move between `FunDefn Raw` and `FunDefn TT`.


david-christiansen:
Thanks!

PR #4181, opened 2017-11-07
Language.JSON: total JSON parsing

msmorgan:
At the moment, does not display parse errors to the user (that is, `parse` returns `Maybe JSON`).

### DIFF
--- a/libs/contrib/Data/String/Extra.idr
+++ b/libs/contrib/Data/String/Extra.idr
@@ -73,3 +73,16 @@ index n str with (unpack str)
   index n str | [] = Nothing
   index Z str | (x :: xs) = Just x
   index (S n) str | (x :: xs) = index n str | xs
+
+||| Produce a string by repeating the character `c` `n` times.
+replicate : (n : Nat) -> (c : Char) -> String
+replicate n c = pack $ replicate n c
+
+||| Indent a given string by `n` spaces.
+indent : (n : Nat) -> String -> String
+indent n x = replicate n ' ' ++ x
+
+||| Indent each line of a given string by `n` spaces.
+indentLines : (n : Nat) -> String -> String
+indentLines n str = unlines $ map (indent n) $ lines str
+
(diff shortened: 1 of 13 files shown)


jfdm:
I think it would be prudent to split this into two PRs: one with the changes to the Lexer and Parser; and the other with the JSON parser contributions. 

There are one or two things in the changes to lexer/parser libraries that at first glance would require some further thought, and I want to consider it for a little before merging or not.

msmorgan:
@jfdm Can do.

Additionally, I'm running into an error with what I thought would be "private" modules. Several of the submodules in Language.JSON are deliberately not included in contrib.ipkg, since a user of the contrib library would likely not want to import them alone.

However, without these packages in the ipkg, the following error occurs when a build is attempted.

```
$ idris -p contrib -o main Main.idr
idris.exe: Erasure/getArity: definition not found for Language.JSON.Lexer.lexJSON
CallStack (from HasCallStack):
  error, called at src\Idris\Erasure.hs:602:20 in idris-1.1.1-6LYZyZduqp2CPjeZoUWKa1:Idris.Erasure
```

To fix this, I'll add the missing modules to the ipkg, and potentially merge them together at a later time.

Is this intended behavior?

jfdm:
One thing to note, if a module is not listed in the ipkg file then it is not installed.

msmorgan:
@jfdm I opened a PR on my own fork, with a version of this library that doesn't require Text.Token.

https://github.com/msmorgan/Idris-dev/pull/5

I wrote it as a basis for comparison, but if you prefer that version I can update this PR with that code.

ahmadsalim:
@msmorgan Can you please rebase this on `master`? 😄

msmorgan:
I've rebased onto master, and cleaned up the history a bit.

ahmadsalim:
@msmorgan Since @jfdm approved merging this when ready, I will merge it now 😄.
Thanks again for the PR!

PR #4182, opened 2017-11-07
Text.Token: token utilities for lexing and parsing

msmorgan:
Split from #4181. To be reviewed separately.

### DIFF
--- a/libs/contrib/Text/Parser.idr
+++ b/libs/contrib/Text/Parser.idr
@@ -4,9 +4,19 @@ import Data.Bool.Extra
 
 import public Text.Parser.Core
 import public Text.Quantity
+import public Text.Token
 
 %access export
 %default total
+
+||| Parse a terminal based on a kind of token.
+match : (Eq k, TokenKind k) =>
+        (kind : k) ->
+        Grammar (Token k) True (tokType kind)
+match kind = terminal $
+  \(Tok kind' text) => if kind' == kind
+                          then Just $ tokValue kind text
+                          else Nothing
 
 ||| Optionally parse a thing, with a default value if the grammar doesn't
 ||| match. May match the empty input.
--- a/libs/contrib/Text/Parser.idr
+++ b/libs/contrib/Text/Parser.idr
@@ -12,7 +12,7 @@ import public Text.Token
 ||| Parse a terminal based on a kind of token.
 match : (Eq k, TokenKind k) =>
         (kind : k) ->
-        Grammar (Token k) True (tokType kind)
+        Grammar (Token k) True (TokType kind)
 match kind = terminal $
   \(Tok kind' text) => if kind' == kind
                           then Just $ tokValue kind text
--- a/libs/contrib/Text/Parser/Core.idr
+++ b/libs/contrib/Text/Parser/Core.idr
@@ -109,6 +109,20 @@ export
        Grammar tok (c1 || c2) b
 (*>) x y = map (const id) x <*> y
 
+||| Produce a grammar that can parse a different type of token by providing a
+||| function converting the new token type into the original one.
+export
+mapToken : (a -> b) -> Grammar b c ty -> Grammar a c ty
+mapToken f (Empty val) = Empty val
+mapToken f (Terminal g) = Terminal (g . f)
+mapToken f (NextIs g) = SeqEmpty (NextIs (g . f)) (Empty . f)
+mapToken f EOF = EOF
+mapToken f (Fail msg) = Fail msg
+mapToken f Commit = Commit
+mapToken f (SeqEat act next) = SeqEat (mapToken f act) (\x => mapToken f (next x))
+mapToken f (SeqEmpty act next) = SeqEmpty (mapToken f act) (\x => mapToken f (next x))
+mapToken f (Alt x y) = Alt (mapToken f x) (mapToken f y)
+
 ||| Always succeed with the given value.
 export
 pure : (val : ty) -> Grammar tok False ty
--- a/libs/contrib/Text/Token.idr
+++ b/libs/contrib/Text/Token.idr
@@ -10,9 +10,9 @@ module Text.Token
 ||| data SimpleKind = SKString | SKInt | SKComma
 |||
 ||| TokenKind SimpleKind where
-|||   tokType SKString = String
-|||   tokType SKInt = Int
-|||   tokType SKComma = ()
+|||   TokType SKString = String
+|||   TokType SKInt = Int
+|||   TokType SKComma = ()
 |||
 |||   tokValue SKString x = x
 |||   tokValue SKInt x = cast x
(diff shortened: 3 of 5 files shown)


libs/contrib/Text/Parser/Core.idr
```diff
@@ -109,6 +109,20 @@ export
        Grammar tok (c1 || c2) b
 (*>) x y = map (const id) x <*> y
 
+||| Produce a grammar that can parse a different type of token by providing a
+||| function converting the new token type into the original one.
+export
+mapToken : (a -> b) -> Grammar b c ty -> Grammar a c ty
```
msmorgan: Dubious name choice . Other ideas for consideration: `tmap`, `lmap`, `contramap`.

libs/contrib/Text/Token.idr
```diff
@@ -0,0 +1,45 @@
...
+import Text.Lexer
+import Text.Parser
+
+%access public export
+%default total
+
+||| Convert a grammar with the token type `t` into one with a token type
+||| `TokenData t`.
+export
+wrapTokenData : Grammar t c a -> Grammar (TokenData t) c a
```
msmorgan: This is included mostly as an example of how to use `mapToken`, but could be useful. I'm not a fan of this name either, though.

msmorgan:
I arrived at the current Text.Token module after several iterations of writing JSON parsers. It accomplishes three main goals:

1.  **Create a system that emits a compiler error when a new type of token has been added.**  
    The `TokenKind` interface leverages dependent types to handle conversion from a recognised string into a value. When a new constructor has been added to the "kind" type, it must be accounted for in the implementation.
1.  **Reduce the amount of boilerplate code required to implement a parser.**  
    The `match` syntax rule allows library users to avoid writing a `terminal` grammar for each type of token. I would prefer for `match` to be a function, but that would require the type implementing `TokenKind` to also implement either `Eq` or `DecEq`. `DecEq` is stricter, but it is tedious to implement for types with many constructors without using elaborator reflection.
1. **Keep the information about the text that was recognised around for as long as possible.**  
    The `Token` record type contains the kind of token it represents as well as the text that was recognised. If a parser or lexer error occurs, this preserves as much information as possible about the input.

I'm happy to change whatever needs to be changed, but hopefully the above rationale makes it a little clearer what I was doing with this module.

libs/contrib/Text/Parser/Core.idr
jfdm: name is fine.

edwinb:
Just one observation I'd  make, without having looked closely at this, is that at least in contrib it's probably better to err on the side of merging and just see what works out. That was a lot of the point of having a contrib in the diary place, so experiments are good. I only put the parser in because I was using it in two projects, and so it seemed other people might want it too...

jfdm:
I've had a look at this, I am going to merge it, however, there are some comments that should be addressed at somepoint in the future.

1. I would like the commit messages to be improved to include the rationale, or at least more information to have context for the changes. Personally I am not a fan of PRs with little to no supporting information in the commit. 

1. I am not too keen on use of Syntax rules, even though they are there, I would rather that we try to avoid their use. They add an unnecessary layer of indirection. Moreover, for `match`  I would say it is more advantageous for it to be a proper function.

2.  While you are trying to improve the error messages, I feel that the `Token` data structure, and it's uses, are nonetheless external to the Lexer and parser itself. While this doesn't impact on existing lexer's and parsers,  I feel at somepoint the features should be integrated in. 

On that final note it is worth bearing in mind that there are active users of the lexer and parser so while changes are welcome, please remember the changes will have an impact.

Once you've updated the commit messages then we can see about merging.

ahmadsalim:
I agree with @jfdm about syntax rules here.
I would prefer a function with an `Eq` constraint on `k`, instead of hackily pattern matching against its value in a `syntax` rule.

ahmadsalim:
In particular, there is the use of `syntax` as a way of providing DSL-like syntax, which is currently non-optimal in idris but an OK use, but also where you use it as a hack as you did now, which I don't think we should accept in the provided libraries.

libs/contrib/Text/Token.idr
msmorgan: I'm removing this.

msmorgan:
I made the requested changes, but Travis stalled yesterday. It's rerunning now.

libs/contrib/Text/Parser.idr
```diff
@@ -4,10 +4,20 @@ import Data.Bool.Extra
 
 import public Text.Parser.Core
 import public Text.Quantity
+import public Text.Token
 
 %access export
 %default total
 
+||| Parse a terminal based on a kind of token.
+match : (Eq k, TokenKind k) =>
```
jfdm: It might be better to add the `Eq` interface constraint to the definition for `TokenKind` itself. This way you are forced to define an `Eq` implementation during the definition of the `TokenKind` implementation.

msmorgan: I considered this, but decided against it because it seems like a valid option to use `Token` and `TokenKind` without using `match`.

ahmadsalim:
@jfdm Is it good to merge now? 😄

msmorgan:
Should `tokType` begin with a capital letter since it's type-level?

ahmadsalim:
@msmorgan You are welcome to do that if you like.

msmorgan:
@jfdm @ahmadsalim 

What would you think about abstracting the "discriminated conversion" pattern present in `TokenKind` to something more general:

```idris
interface Convert from discriminator where
  Target : discriminator -> Type
  convert : from -> (d : discriminator) -> Target d
```

ahmadsalim:
@msmorgan Please let us just keep it on the same abstraction level for now. Let us make this PR merge-ready first before changing the content too much 😄

msmorgan:
I had a concept ready here:

https://github.com/msmorgan/Idris-dev/tree/text-token-convert

jfdm:
I agree with @ahmadsalim on this. Moreover, I think having an extra interface, that is really a specialised variant of `Cast` is not required.

msmorgan:
It's not a specialized version of `Cast`, but a dependent one. I'll defer to your judgment if you think it's useless.

jfdm:
I didn't say it was 'useless', I said it wasn't required for this PR. Two separate concepts.

msmorgan:
@jfdm I see what you intended now, but I initially read your comment as a statement about the worth of the `Convert` interface in general. I didn't mean to put words in your mouth.

It's fine with me if you merge this PR as-is. Thinking a few steps ahead, however: If an interface such as `Convert` did exist in the library, `TokenKind` is a specialized version of it, allowing conversion from `String` only. In my opinion, it may not be wise to add a specialized interface like `TokenKind` if it can be superseded by something more flexible. The parent commit to the branch I linked does not have the `TokenKind` interface at all, and instead defines `match` and `value` in terms of `Convert`. That is the version I prefer.

libs/contrib/Text/Parser/Core.idr
clayrat: `contralmapToken` ? :)

msmorgan: `hocusTokens`

msmorgan:
@jfdm What's the status on this? Is there anything further you need from me?

jfdm:
Sorry, I've been snowed under with things at the moment. I think it is fine, but will double check later this week.

jfdm:
Sorry about the wait. I've had things pop up that had to be dealt with. Tis merged. If you rebase your other PR, then that can be merged if there are no problems with it.

PR #4185, opened 2017-11-08
Ensure only injectivity if not named

### LINKED ISSUE #3727: Implementation arguments containing injective functions

ahmadsalim:
Closes #3727

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,14 +15,14 @@
   undeclared fixity are treated as non-associative with precedence lower
   than all declared operators.
 + Allow non-injective implementations if explicitly named, e.g.,
-```idris
-LTB : Nat -> Type
-LTB b = DPair Nat (\ n  => LT n b)
+  ```idris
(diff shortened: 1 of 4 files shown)


ahmadsalim:
@edwinb Any objections?

I will try to add a test and an entry to the CHANGELOG.

PR #4188, opened 2017-11-09
Fix docs

jfdm:
Fix minor mistakes in reST documentation.


jfdm:
Our travis test's don't check these documents, so no point in letting them run. If there is a problem RTD will let us know. Thus, I am merging

PR #4191, opened 2017-11-10
Remove long deprecated aspects of Idris' shipped libraries.

jfdm:
Several releases ago we decided to deprecate the use of `Float` in
favour of `Double` to describe 64bit floating point numbers. For the
rationale please see our FAQs [1].

Given that the decision was made several releases ago I think it is
time that we remove the type alias that point's `Float` to `Double`.

There has been enough time for people to move on.

[1] http://idris.readthedocs.io/en/latest/faq/faq.html#why-does-idris-use-double-instead-of-float64

I've updated the PR to remove other deprecated parts of idris' libraries that have been deprecated for several releases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@
 + `Prelude.Doubles.atan2` is now implemented as a primitive instead of
   being coded in Idris.
 + Added `Test.Unit` to `contrib` for simple unit testing.
++ Removed several deprecated items from the libraries shipped with Idris.
 
 ## Tool Updates
 
--- a/libs/effects/Effects.idr
+++ b/libs/effects/Effects.idr
@@ -306,11 +306,6 @@ toEff xs' = id
 pure : a -> EffM m a xs (\v => xs)
 pure = Value
 
-return : a -> EffM m a xs (\v => xs)
-return = pure
-
-%deprecate Effects.return "Please use `pure`."
-
 pureM : (val : a) -> EffM m a (xs val) xs
 pureM = Value
 
--- a/libs/prelude/Builtins.idr
+++ b/libs/prelude/Builtins.idr
@@ -130,14 +130,14 @@ data Delayed : DelayReason -> Type -> Type where
 Force : {t, a : _} -> Delayed t a -> a
 Force (Delay x) = x
 
-||| Lazily evaluated values. 
+||| Lazily evaluated values.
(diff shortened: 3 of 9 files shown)


david-christiansen:
It's probably time now :-)

david-christiansen:
I think we should leave `return` sitting around. It's not really there to be deprecated per se, but rather to help Haskellers get started more quickly. The others are all genuine deprecations, and I think they should go.

Good initiative!

jfdm:
I kind a agree with leaving `return` in there, however, i don't think the deprecation directive is suitable. I'll reintroduce it using the `fragile` directive, which potentionally be a better meaning directive for this purpose.

ahmadsalim:
I agree with @jfdm, that using something other than deprecated might be good.
@jfdm @david-christiansen How about introducing a more general `%hint` directive, which takes two arguments, one the type of hint e.g. `fragile`, `deprecated`, etc. which is extensible.

Then `%deprecated` would just have the same meaning as `%hint deprecated`.

jfdm:
@ahmadsalim Maybe that would be a good idea. However, we cannot use `%hint` as that is already taken. We use it to pass functions to the proof search mechanism to help with proof construction.

Yeah our directive documentation needs improving...

ahmadsalim:
@jfdm Everyday I learn something 😄

eraserhd:
There's a number of deprecations in the parser.  Is it useful to remove those?  (It would be a separate PR):

- 'public' is deprecated. Use 'public export' instead.
- The 'abstract' keyword is deprecated. Use 'export' instead.
- %assert_total is deprecated. Use the 'assert_total' function instead.
- The 'class' keyword is deprecated. Use 'interface' instead. (I suspect this one is too new.)
- This style of tactic proof is deprecated. See %runElab for the replacement.
- The use of [static] is deprecated, use %static instead.

jfdm:
I think separate PR, and announcement to the mailing list would be better.  If/when you do update, it would be good to make sure the examples and documentation are consistent.

The only thing I would keep is the deprecation related to `%runElab`. It is not deprecated per se, elaborator reflection offers something better. Further, there *is* some existing code that uses tactic based proofs, and I don't want to annoy the author.

ahmadsalim:
@jfdm Would it be possible to use `%fragile` internally for the `%runElab` thing?

jfdm:
@ahmadsalim Maybe...

PR #4199, opened 2017-11-16
Adds scanl1 to Data.Vect

robstewart57:
This adds scanl1, matching the scanl and scanl1 scans in Data.List.

The documentation for scanl1, and added documentation for scanl, in
Data.Vect is borrowed from Data.List.


ahmadsalim:
Thanks for the PR

PR #4200, opened 2017-11-16
Support trailing quotes in verbatim string literals

msmorgan:
Since verbatim strings in Idris don't support escapes, it seems appropriate
to have them consume all `'"'` characters before the closing `"\"\"\""`. This is
the approach taken by Scala.

This means the following verbatim strings now parse, and produce the parenthesized
values, which are expressed as standard strings.
* `"""""""` (`"\""`)
* `""""hello""""` (`"\"hello\""`)
* `"""""a"""""""` (`"\"\"a\"\"\"\""`)
* `"""trailing """"` (`"trailing \""`)


asandroq:
What about adding a few tests?

msmorgan:
@asandroq I added a few test cases, but I'm not sure if they're in the right place.

ahmadsalim:
This looks good, thanks!

PR #4201, opened 2017-11-17
Add parity functions for naturals.

vmchale:
This adds `even` and `odd`, which return a boolean value given a natural.


ahmadsalim:
Thanks for the PR!

However, I am unsure whether this should be accepted directly in the prelude. Could you kindly consider moving it to `contrib` instead, which is where we usually accept library changes?

vmchale:
Sure thing! The latest commit should address that.

ahmadsalim:
Thanks!

PR #4204, opened 2017-11-19
Set length when accepting on socket

A1kmm:
Without this change, the sockaddr will be left unmodified and
uninitialised memory is interpreted as a sockaddr.

### DIFF
--- a/rts/idris_net.c
+++ b/rts/idris_net.c
@@ -150,7 +150,7 @@ void* idrnet_create_sockaddr() {
 
 int idrnet_accept(int sockfd, void* sockaddr) {
     struct sockaddr* addr = (struct sockaddr*) sockaddr;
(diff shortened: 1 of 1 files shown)


ahmadsalim:
Thanks

ahmadsalim:
@A1kmm Just to be sure, should it not be `sizeof(struct sockaddr)` instead?

A1kmm:
sockaddr is polymorphic, in the sense that sockaddr is the base type and only has the common fields, and there are different types for different address families.

sockaddr_storage is defined to be big enough to handle any of the other address families (see
 https://www.unix.com/man-page/posix/7posix/socket.h/ for details). The code allocates the buffer as sizeof(sockaddr_storage), so that is the most appropriate size to use.

ahmadsalim:
Thanks for the response.

PR #4213, opened 2017-11-22
Small doc. error in Interfaces.

xekoukou:


ahmadsalim:
Thanks for the PR, but actually this is correct/not an error.
Since applicatives take a funtion wrapped in `f` and applies it to a value wrapped in `f`.

PR #4216, opened 2017-11-22
Prelude.Nat: make `fib` and `fact` tail-recursive

msmorgan:
Uses default implicits as accumulators.

The performance of `fib` in particular is drastically improved.

### DIFF
--- a/libs/prelude/Prelude/Nat.idr
+++ b/libs/prelude/Prelude/Nat.idr
@@ -326,15 +326,16 @@ pred (S n) = n
 --------------------------------------------------------------------------------
 
 ||| Fibonacci numbers
-total fib : Nat -> Nat
-fib Z         = Z
-fib (S Z)     = S Z
-fib (S (S n)) = fib (S n) + fib n
+total
+fib : {default Z a : Nat} -> {default (S Z) b : Nat} -> Nat -> Nat
+fib {a} {b} Z = a
+fib {a} {b} (S n) = fib {a=b} {b=a + b} n
 
 ||| Factorial function
-total fact : Nat -> Nat
-fact Z     = S Z
-fact (S n) = (S n) * fact n
+total
+fact : {default (S Z) acc : Nat} -> Nat -> Nat
+fact {acc} Z = acc
+fact {acc} (S n) = fact {acc=S n * acc} n
 
 --------------------------------------------------------------------------------
 -- Division and modulus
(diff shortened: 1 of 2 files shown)


ahmadsalim:
@msmorgan Thanks for the PR, but I believe that the ordinary versions are much nicer for proofs.

If you need efficiency, please do not rely on the `Nat` instances anyway 😄

ahmadsalim:
In any case, these are publicly exported functions in the `prelude` and I would not want to merge a breaking change in the `prelude`. As for the `fibs` stream, you are welcome to make a new PR in `contrib`

jfdm:
One thing to bear in mind with the constructions of some functions, is that the tail recursive versions do not lend them well to proofs as their structure is not the best. However, this doesn't mean that we should not have 'efficient' versions. I think if we are to add in efficient versions of `fib` to the prelude (or contrib for that matter) then we should add them with a different name (`fib'` as we do with safe and unsafe version's of `head`) or under a different namespace. 

Further, from a style perspective, I am not a fan of default implicits in prelude functions. I think it would be better to have them explicit but the calculating function be part of a where clause that does the construction.

msmorgan:
So, this is a breaking change because any proofs written against the existing functions cease to work? That's fair, I guess.

I must have misunderstood the priorities of the library. On my machine, computing the 45th fibonacci number took over 50 seconds with the current version, and under 200 milliseconds with the version in this PR. In most other software projects, poor performance resulting from diverging recursion would be a problem worth fixing.

If we were to add `fib'`, why shouldn't it use default implicits? Such an implementation seems easier to write proofs about than a version that hides its functionality in a where function. It also highlights an interesting and useful feature of Idris that is currently unrepresented in the library code. The tutorial includes a section that is essentially identical to the code I submitted.

ahmadsalim:
@msmorgan Because Idris is a system that allows you to do proofs as well as programs, sometimes we have functions which are more meant to be nice specifications (like `fib`) than used to efficiently compute with large numbers.

As @jfdm mentioned you are welcome to contribute `fib'` in the `contrib` library, which provides a version that is more meant to be performant than nice with proofs 😄 .

msmorgan:
Is it okay to use default implicit arguments in `contrib`?

ahmadsalim:
@msmorgan To be honest, it is not a big deal for me (whether or not you have them).
But if you open a PR, it should be easy to change 😄 when the other contributors take a look.

PR #4218, opened 2017-11-23
Prefix for pasting in identifiers in C FFI

melted:


jfdm:
Does this require a changelog entry?

melted:
Yes, it should have one, I’ll submit one after work.

PR #4219, opened 2017-11-24
Documents how to include debugging symbols

robstewart57:
This documents the possibility of using

%flag C "-g"

to include debug symbols to improve the backtrace information when
debugging runtime errors (i.e. segfaults) of Idris programs with gdb.


ahmadsalim:
Thanks for the PR!

PR #4224, opened 2017-11-26
First attempt at implementing ST alternative for File.idr in Effects.

XeFias:
The Effects tutorial for Idris recommend using Control.ST. At the moment most of the effects which are currently implemented for Effect (like the FILE effect) do not have an alternative implementation for Control.ST. The “_libs/contrib/Control/ST/File.idr_” file implements the same behavior as “_libs/effects/Effect/File.idr_”.

Any recommendations on how to improve the code, tests, documentation, etc. is welcome.


XeFias:
It seems I incorrectly merged changes from upstream/master into this change. Sorry.

PR #4231, opened 2017-11-30
Implenting Data.List.Permutations

Average-user:
I noticed that theres no built-in implementation of  ```permutations``` so I'm attempting to implement it.

I would also like to know what you think about this, and hear some possible improvements that I'm sure there are.

Only in case it is not clear by now, ```permutations``` should have the type ```List a -> List (List a)``` and should return all possible rearrangements of a list. ```permutations [1,2] == [[1,2], [2,1]]```. Every element of the list would be treated as unique, even if there are two or more repeated: ```permutations [1,1] = [[1,1],[1,1]]```

### DIFF
--- a/libs/contrib/Data/List/Permutations.idr
+++ b/libs/contrib/Data/List/Permutations.idr
@@ -1,0 +1,10 @@
+module Permutations
+
+-- | All permutations of a list.
+permutations : List a -> List (List a)
+permutations [] = [[]]
+permutations xs = [y :: zs | (y, ys) <- select xs, zs <- permutations ys]
+  where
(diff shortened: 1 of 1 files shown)


ahmadsalim:
Thanks for the PR! 👍

Average-user:
There is a possibility to add the permutations function directly to Data.List?

ahmadsalim:
@Average-user Usually, we are careful in accepting PRs directly to the `prelude` since it can potentially break people's code since all functions are imported implicitly.

You are welcome to write on the mailing list to see whether people have any objections.

PR #4232, opened 2017-11-30
Reorder the type parameters on BoundedList.

sellout:
Now we can have Functor and Foldable instances, rather than just defining overloaded versions.

Also, I don’t know how I’ve had these changes just sitting on my fork for over three years.


sellout:
Um, *cough*, it seems like I got these changes merged like three years ago, too …

PR #4235, opened 2017-12-01
contrib: add groupBy and group

vaibhavsagar:


msmorgan:
Should this be in `contrib`?

ahmadsalim:
@vaibhavsagar Thanks for the PR. Yes, please put them in `contrib`.

vaibhavsagar:
Hi, is this what you had in mind?

libs/contrib/Data/List/Group.idr
```diff
@@ -0,0 +1,24 @@
+module Group
```
ahmadsalim: I think this should be `Data.List.Group`

ahmadsalim:
@vaibhavsagar Yeah, thanks. There is a small correction with regards to module name but otherwise it is fine.

vaibhavsagar:
Thanks, I've updated the module name.

vaibhavsagar:
Thanks!

PR #4249, opened 2017-12-07
Make sure a point release of idris builds with LTS Haskell.

jfdm:


jfdm:
If we are going to to do a release, I think it would be prudent that the `stack.yaml` file points to an LTS version of haskell rather than a nightly version. Post release, we can then see how to manage a stable and nightly stack configuration.

Issue #3479, opened 2016-10-06
No defined implementation Cast ty ty


Termina1:
For example, there is a code that does something like:
```idris
let x : Integer = 10 in the Integer (cast x)
```
And I get error like:

```
Can't cast from Integer to Integer 
```

Which is strange, but we could solve this with this implementation:
```idris
Cast ty ty where
    cast = id 
```
I encountered this problem while doing exercise for the book Type Driven Development with Idris.
I need to implement Cast interface for some custom type.
```idris
data Expr num = Val num
              | Add (Expr num) (Expr num)
              | Sub (Expr num) (Expr num)
              | Mul (Expr num) (Expr num)
              | Div (Expr num) (Expr num)
              | Abs (Expr num)
eval : (Neg num, Integral num) => Expr num -> num
eval (Val x) = x
eval (Add x y) = eval x + eval y
eval (Sub x y) = eval x - eval y
eval (Mul x y) = eval x * eval y
eval (Div x y) = eval x `div` eval y
eval (Abs x) = abs (eval x)
```

I defined it as follows:

```idris
(Integral ty, Neg ty, Cast ty num) => Cast (Expr ty) num where
    cast orig = cast (eval orig) 
```
But the test example from the book failed because of the reason I described earlier:

```idris
let x : Expr Integer = 6 * 3 + 12 in the Integer (cast x) 
```

Am I getting something wrong? Why casting from type ty to ty haven't been implemented as a part of standard library for idris?

xekoukou:
There is no reason to cast to the same type. This is why there is no implementation.
For the remaining questions, you could ask them at the idris irc channel or the google group if there is noone at the irc channel.

Termina1:
@xekoukou so, I think I provided a valid use case? Why you think there is "no reason" to do this?

xekoukou:
If your initial piece of code is given by the Book, then you would indeed need to define Cast.
If it is not, then notice that cast can be removed, since it of the type we already have.

For the same reason, cast is not necessary in your cast from ```Expr num``` to ```num``` since eval returns ```num```.

Your last piece of code needs to be like this:

``` let x : Expr Integer = Add (Mul (Val 6) (Val 3)) (Val 12) in the Integer (cast x)   ```

**Comment by mitchellwrosen** [comment]
> There is no reason to cast to the same type

What if you have a function that is polymorphic, like

```
foo : Cast a Int => a -> Int -> Int
foo x y = cast x + y
```

[...]

PR #4252, opened 2017-12-11
add implementation for `Cast a a`

msmorgan:
When `from` and `to` are the same type, `cast = id`.

### DIFF
--- a/libs/prelude/Prelude/Cast.idr
+++ b/libs/prelude/Prelude/Cast.idr
@@ -1,5 +1,6 @@
 module Prelude.Cast
 
+import Prelude.Basics
 import Prelude.Bool
 import public Builtins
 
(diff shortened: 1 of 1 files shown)


Melvar:
If you want this implementation, you’re misusing `Cast`. You should never have a constraint using it. It’s only to save on remembering the names of every obvious conversion function.

Melvar:
See in particular https://github.com/idris-lang/Idris-dev/issues/3479#issuecomment-254211975 for the previous argument on this.

msmorgan:
I wanted it for the following helper function for Advent of Code:

```idris
||| Convert a line of text into a list of data by splitting on spaces
||| and casting the resulting strings.
dataRow : Cast String ty => String -> List ty
dataRow = map cast . split isSpace
```

Sometimes, the input should be a list of strings.

I can always implement `Cast String String` locally, so it's no big deal if this doesn't get merged.

ahmadsalim:
I agree with @Melvar, this is not the purpose of `cast`, and since we have discussed this before (see the link by @Melvar), I will close the PR.

PR #4253, opened 2017-12-11
add precedence for infix `div` and `mod`

msmorgan:
These functions have the same precedence as `/`.


ahmadsalim:
Thanks for the PR.

PR #4260, opened 2017-12-19
Add alternatives syntax to `let` in `do` blocks

### LINKED ISSUE #3158: “Bind” alternatives for let
The syntax

```
do pat1 <- act | pat2 => other
   rest
```

is, if I recall correctly, equivalent to

```
do case !act of
     pat1 => rest
     pat2 => other
```

I now find myself using `pat1 <- pure $ val | pat2 => other` to save a nesting depth if I need to match on a pure value in a `do` block. It would be nice if

```
do let pat1 = val | pat2 => other
   rest
```

would analogously act like

```
do case val of
     pat1 => rest
     pat2 => other
```

to streamline this case.

This could be generalized to `let … in …`, but I think that might be less useful and harder to read.

Melvar:
This works analogously to the syntax for alternatives on
pattern-matching bind. Fixes #3158.

Most of the changes cause `DoLetP` to be handled analogously to `DoBindP`.

There are differences to the implementation for alternatives in `let … in …`:
- `with`-applications are always disallowed, rather than being allowed for single-name patterns (since it’s not possible to positively distinguish a regular `let` from a pattern-matching `let` matching a nullary constructor in the parser).
- Patterns which are not single names may not co-occur with an explicit type (doing better would require passing on the information whether a type was present and erroring in a later pass, similar to the above).
- Alternatives may not co-occur with an explicit type.

The implementation for `let … in …` silently ignores any explicit type if it ends up being a pattern-matching `let`. Should I make it as restrictive as the one for `do let`, relax the latter, or leave it as is?

### DIFF
--- a/src/Idris/Parser/Expr.hs
+++ b/src/Idris/Parser/Expr.hs
@@ -211,7 +211,9 @@ updateSynMatch = update
             upd ns (DoBindP fc i t ts)
                     = DoBindP fc (update ns i) (update ns t)
                                  (map (\(l,r) -> (update ns l, update ns r)) ts)
-            upd ns (DoLetP fc i t) = DoLetP fc (update ns i) (update ns t)
+            upd ns (DoLetP fc i t ts)
+                    = DoLetP fc (update ns i) (update ns t)
+                                (map (\(l,r) -> (update ns l, update ns r)) ts)
             upd ns (DoRewrite fc h) = DoRewrite fc (update ns h)
     update ns (PIdiom fc t) = PIdiom fc $ update ns t
     update ns (PMetavar fc n) = uncurry (flip PMetavar) $ updateB ns (n, fc)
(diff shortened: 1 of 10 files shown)


eraserhd:
- Test case
- CHANGELOG
- Documentation

src/Idris/Parser/Expr.hs
```diff
@@ -1288,16 +1290,26 @@ do_ :: SyntaxInfo -> IdrisParser PDo
...
                  (i, ifc) <- withExtent name
-                 ty <- P.option Placeholder (do lchar ':'
-                                                expr' syn)
+                 ty' <- P.optional (do lchar ':'
+                                       expr' syn)
                  reservedOp "="
-                 (e, fc) <- withExtent $ expr syn
-                 return (DoLet fc RigW i ifc ty e))
+                 (e, fc) <- withExtent $ expr (syn { withAppAllowed = False })
+                 -- If there is an explicit type, this can’t be a pattern-matching let, so do not parse alternatives
```
eraserhd: Should it be possible to say, for example, `let Nothing : Maybe (Int, String) = listToMaybe (reads "42") | Just _ => ...`?

src/Idris/Parser/Expr.hs
```diff
@@ -1288,16 +1290,26 @@ do_ :: SyntaxInfo -> IdrisParser PDo
...
+                 case ty' of
+                   Just ty -> return (DoLet fc RigW i ifc ty e)
+                   Nothing ->
+                     P.option (DoLet fc RigW i ifc Placeholder e)
+                              (do lchar '|'
+                                  ts <- P.sepBy1 (do_alt (syn { withAppAllowed = False })) (lchar '|')
```
eraserhd: If there's a reason not to support the above case, then putting `when (isJust ty') $ fail "<reason>"` above this line instead of using the `case` will explain to the user what happened.

src/Idris/Parser/Expr.hs
Melvar: In the case of `let … in …` it *is* possible, but the type annotation is ignored, and allowing it not to be ignored is far beyond the scope of this PR, so I decided to disallow the combination here since it doesn’t add anything but looks confusing if the type happens to be wrong.

jfdm:
@Melvar As an update, the reason the travis tests are failing are due to a bug introduce in a previous commit. If you rebase on master *all* tests should now pass as expected.

Further, it would be beneficial to have at least a CHANGELOG entry added to the PR so as to not block 1.2 by much. While it *would* be nice to have a test and documentation updates I think they *can* come later but would still be preferable before 1.2 is released.

Thanks!

Melvar:
I have added an error message for the combination of an explicit type annotation and alternatives, as suggested by @eraserhd (suggestions for improvement of its contents welcome), written a changelog entry, and extended the test for alternatives in the contexts where they already worked to test this functionality as well.
As for documentation, I feel a bit lost in that subtree.

src/Idris/Parser/Expr.hs
ahmadsalim: What worries me is that `let` pattern matches on unary constructors, but since this is not something introduced by this PR, I will not block this.

ahmadsalim:
@Melvar Feel free to merge when you feel like it. The changes seem good

PR #4263, opened 2017-12-20
Add helper functions for Monad

clayrat:
Namely, `=<<`, `<=<`, `>=>`.

### DIFF
--- a/libs/base/Control/Monad.idr
+++ b/libs/base/Control/Monad.idr
@@ -1,0 +1,18 @@
+module Control.Monad
+
+%default total
+%access public export
+
+infixr 1 =<<, <=<, >=>
+
+||| Left-to-right Kleisli composition of monads.
+(>=>) : Monad m => (a -> m b) -> (b -> m c) -> (a -> m c)
+(>=>) f g = \x => f x >>= g
+
+||| Right-to-left Kleisli composition of monads, flipped version of `>=>`.
+(<=<) : Monad m => (b -> m c) -> (a -> m b) -> (a -> m c)
+(<=<) = flip (>=>)
+
+||| Right-to-left monadic bind, flipped version of `>>=`.
+(=<<) : Monad m => (a -> m b) -> m a -> m b
+(=<<) = flip (>>=)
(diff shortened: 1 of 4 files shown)


libs/base/Control/Monad.idr
```diff
@@ -0,0 +1,18 @@
...
+(>=>) : Monad m => (a -> m b) -> (b -> m c) -> (a -> m c)
+(>=>) f g = \x => f x >>= g
+
+||| Right-to-left Kleisli composition of monads, flipped version of `>=>`.
+(<=<) : Monad m => (b -> m c) -> (a -> m b) -> (a -> m c)
+(<=<) = flip (>=>)
+
+||| Right-to-left monadic bind, flipped version of `>>=`.
+(=<<) : Monad m => (a -> m b) -> m a -> m b
+(=<<) = flip (>>=)
```
asandroq: Missing newline

msmorgan:
Do you think it might be worthwhile to make `>=>` an `infixl` operator, since it's the reverse of `<=<`?

clayrat:
@msmorgan Not sure, for now I've just copied these from Haskell since I needed them for porting `Freer`/`Cofreer`. I guess we can try?

ahmadsalim:
@clayrat Can we please have these in `contrib`? 😄

msmorgan:
@ahmadsalim It seems wrong to have "sub-modules" like `Control.Monad.Identity`, etc. residing in `base`, and `Control.Monad` in `contrib`. If this is moved to `contrib`, it should be under a different module name.

ahmadsalim:
@msmorgan Yeah, I agree the name should be different

clayrat:
And what should it be then?

ahmadsalim:
@clayrat How about `Control.Monad.Syntax`?

clayrat:
Done

clayrat:
Not sure about making them left-associative, any preferences from others?

ahmadsalim:
Thanks!

`>>=` is meant to be associative, so it should not matter what parsing order should be used. Since it is in `contrib` we probably can change it anyway later if needed.

PR #4276, opened 2018-01-03
Implement DecEq for Elem

i-am-tom:
This commit adds decidable equality to the `Elem` type in `Data.List`. Let me know if anything could be cleaned up - I'm very new to Idris :)


david-christiansen:
Looks great! Thanks!

PR #4277, opened 2018-01-03
Replace '-' with '_' in LHS of Node requires

namanbharadwaj:
Currently, the following

```
module Main
%lib Node "node-pty"

main : JS_IO ()
main = pure ()
```

produces the invalid javascript `var node-pty = require("node-pty");`. This is a quick fix that matches the way we handle other invalid characters.


ahmadsalim:
Thanks!

PR #4280, opened 2018-01-04
Set environment with Idris flags while running `make`

asandroq:
When building a package with foreign code and a makefile, the makefile
itself must find the compilation flags that Idris already knows
about. Moreover, the `idris` executable found by the makefile may not
be the same used to build the package. This change passes the build
flags to the makefile though environment variables.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -109,7 +109,10 @@
 ## Packaging Updates
 
 + Package names now only accept a restrictive charset of letters, numbers and the `-_` characters.
-  Package names are also case insensitive
+  Package names are also case insensitive.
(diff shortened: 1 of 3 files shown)


ahmadsalim:
@asandroq Does this solve any existing issue?

asandroq:
I've noticed the issue while working on a private codebase of my own. Eventually I'll have more than one build of Idris installed (for cross-compilation) and each build must have the correct build flags set. The best way of doing this is letting Idris export the flags itself.

edwinb:
I don't entirely understand the need for this (though I'm willing to believe it exists :)). Could you add a note to the documentation please? The environment variables section is in `docs/reference/compilation.rst`

Issue #4297, opened 2018-01-19
Support let-binding for linear variables


clayrat:
# Steps to Reproduce
```
%language LinearTypes
%default total

foo : (1 x : Int) -> ((1 i : Int) -> String) -> String
foo x f = f x

bar : (1 x : Int) -> ((1 i : Int) -> String) -> String
bar x f = let res = f x in res
```
# Expected Behavior
Both `foo` and `bar` type check (possibly, with some additional syntax, e.g. `let (1 res) = ...`.

# Observed Behavior
On Idris 1.2.0:
```
  |
8 | bar x f = let res = f x in res
  |           ~~~~~~~~~~~~~~~~~~~~
Trying to use linear name x in non-linear context
```

@Melvar on IRC had more details on how this could be implemented, unfortunately I have lost access to those logs.

Melvar:
The `PTerm` constructor `PLet` representing true `let` already has a `RigCount` field that presumably, when set to `Rig1`, would do the right thing, but the current parser function `let_` does not parse a rigcount and always passes `RigW`, i.e. unlimited use, as that field.

ahmadsalim:
Fixed by #4298 .

PR #4298, opened 2018-01-19
WIP: parse RigCount for let-bindings

clayrat:
Solves https://github.com/idris-lang/Idris-dev/issues/4297. This seems to work for the original issue, should anything else be added? The new syntax simply allows to add `0`/`1` before the name in let, e.g.: 
```
bar : (1 x : Int) -> ((1 i : Int) -> String) -> String
bar x f = let 1 res = f x in
          res
```

### DIFF
--- a/src/Idris/Parser/Expr.hs
+++ b/src/Idris/Parser/Expr.hs
@@ -939,7 +939,7 @@ rewriteTerm syn = do keyword "rewrite"
                      return (PRewrite fc using prf sc giving)
                   <?> "term rewrite expression"
 
--- | Parse a constant and its source span
+-- | Parse rig count for linear types
 rigCount :: Parsing m => m RigCount
 rigCount = P.option RigW $ do lchar '1'; return Rig1
                        <|> do lchar '0'; return Rig0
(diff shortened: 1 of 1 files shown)


Melvar:
This looks like what I expected. 👍

ahmadsalim:
@clayrat Looks good, but a few small questions.

Does this work with `let`-bindings in `do` ? Also, should `<-` bindings also have this type of rig-count?

Melvar:
@ahmadsalim I can answer those: It does not work for `do let` which is parsed elsewhere, and `<-` should probably inherit this option if lambdas had it – however they do not currently, and the `PLam` constructor does not even have a field for it, so it requires much deeper changes to add it.

ahmadsalim:
@Melvar Thanks!

Melvar:
Personally I think the parsing of `let` could use a big rewrite to exclude combinations that don’t make sense, such as pattern-matching with explicit types (or producing a new `PLetP` constructor so such a type can be checked instead of ignored), and reduce/eliminate the duplication between `let in` and `do let`, but that’s a bigger project that would want a decent amount of discussion all on its own.

clayrat:
Looks like the tests are failing @edwinb hasn't actually added `test/regression002/reg001.idr` in https://github.com/idris-lang/Idris-dev/commit/06db7dc171686fdc4ea29953d7cdf2ef2ecde1d2 Most likely it's the same as the snippet in the linked issue?

edwinb:
Sorry about that, I didn't notice the failure. I've been getting so many false negatives from Travis that I must be starting to subconsciously ignore them... I've added the file now anyway.

ahmadsalim:
Thanks for the PR!

Issue #4299, opened 2018-01-20
Totality checking says obviously not total function is total


2426021684:
# Steps to Reproduce
Write this in a file:
```idris
module Main
import Data.Vect
%default total

p : Elem (Maybe x) (map Maybe xs) -> Elem x xs
p Here impossible
p (There l) impossible
```
Load it in the Idris REPL

# Expected Behavior

It says that p is not total.

# Observed Behavior

Idris accepts it as total. Even `:total` thinks that `p` is total.

luochen1990:
@2426021684 Do you mean to use `Just` instead of `Maybe`?

When you use `Maybe`, you means that `x` is a type, and the unification of them is obviously impossible.
If you use `Just`, then the type checking result is as you expected.

2426021684:
@luochen1990 I have no idea what you are saying. There is a valid total definition of p. I did not mean to use `Just`.

```
p : Elem (Maybe x) (map Maybe xs) -> Elem x xs
p {xs = _ :: _} Here = Here
p {xs = _ :: _} (There l) = There (p l)
```

luochen1990:
@2426021684 sorry, I think you are right, this is indeed something weird.

luochen1990:
And I also find that the following will throw an error as expected:

```
p : Elem (Maybe x) (map Maybe xs) -> Elem x xs
p {xs = _ :: _} Here impossible
p {xs = _ :: _} (There l) impossible
```

yanok:
Related to #4282 (in fact, it's the same problem).

Issue #4282, opened 2018-01-06
Proof of Void with equalities in Type


anton-trunov:
# Steps to Reproduce
```
total
bad1 : (f : Type -> Type) -> (x, y : Type) -> f x = f y -> Void
bad1 _ _ _ Refl impossible

total
ohoh : Void
ohoh = bad1 id Unit Unit Refl

-- of course, these shouldn't be provable as well:
total
bad2 : (f, g : Type -> Type) -> (x : Type) -> f x = g x -> Void
bad2 _ _ _ Refl impossible

total
bad3 : (f, g : Type -> Type) -> (x, y : Type) -> f x = g y -> Void
bad3 _ _ _ _ Refl impossible
```
# Expected Behavior

`bad1`, `bad2`, `bad3` are not provable

# Observed Behavior

`bad1`, `bad2`, `bad3` are provable

# Notes

Tested with Idris 1.1.1-git:05634d71 (macOS 10.13.2 (17C88), High Sierra)

ahmadsalim:
Thanks for reporting the issue!

It should not recognize these cases as impossible since `f x = f y` should get stuck and not fail.
Agda gives the following error:

```
I'm not sure if there should be a case for the constructor refl,
because I get stuck when trying to solve the following unification
problems (inferred index ≟ expected index):
  f x ≟ f y
when checking that the pattern refl has type f x ≡ f y
```

anton-trunov:
Thanks for checking it with Agda. Coq also rejects this thing.

melted:
elabClause for an impossible clause is

```idris
elabClause info opts (_, PClause fc fname lhs_in [] PImpossible [])
   = do let tcgen = Dictionary `elem` opts
        i <- get
        let lhs = addImpl [] i lhs_in
        b <- checkPossible info fc tcgen fname lhs_in
        case b of
            Just _ -> tclift $ tfail (At fc
                                 (Msg $ show lhs_in ++ " is a valid case"))
            Nothing -> do ptm <- mkPatTm lhs_in
                          logElab 5 $ "Elaborated impossible case " ++ showTmImpls lhs ++
                                      "\n" ++ show ptm
                          return (Left ptm, lhs)
```

problem is that checkPossible returns `Nothing`not because it is impossible but because typechecking fails:
```
Impossible case .\.\i4282.idr:3:6:When checking an application of Main.bad1:
        Type mismatch between
                f y = f y (Type of Refl)
        and
                f x = f y (Expected type)

[...]

PR #4303, opened 2018-01-24
Small fix for #4282 and #4299

yanok:


ahmadsalim:
@yanok Thanks for the PR. Can you please explicate a bit more the reasoning behind your fix?

yanok:
@ahmadsalim Well, I have to admit it was the first time I looked at the Idris internals, so I might be missing something. But my reasoning was along these lines:

The only thing changed is the
```haskell
checkRec (TType _) (TType _) = return False
```
case, which I changed from `False` to `True`.

1. Why returning `False` is a bad idea? I tried to explain this in the commit message, but I'll try to re-state it here. In case of constructor application `checkRec` works by recursively checking all the arguments. Now imagine a constructor applied (among other things) to `Type`. This clause for `TType` would result in the whole application reported as unrecoverable, even if all the other arguments are recoverable. For example, in #4282 `checkRec` gets called with `(=) Type Type (f x) (f y)` and `(=) Type Type (f x) (f x)`. Here pairs `f x` , `f x` and `f y` , `f x` are (correctly) reported as recoverable but because of the `Type`s the whole application is reported as not recoverable.
2. Why it's ok to always return `True`? Well, first of all, I think it's much better to fail to recognise valid impossible pattern, then to accept the invalid one. Then, `checkRec` is only called if there were an unification error. Correct me if I'm wrong, but I think problems with `Type`s are reported as universe errors, not unification errors so in `checkRec` it's safe to assume that two matching `TType`s agree with each other.

ahmadsalim:
@yanok Thanks for the explanation, it sounds reasonable

ahmadsalim:
@edwinb I have merged this, but if something goes wrong please tell me so that I can revert it.

edwinb:
Seems a reasonable fix to me. Thanks!

(I did the now usual thing of checking if the same problem exists in Blodwen - fortunately not! Its approach to type checking makes this rule much simpler to implement because the notion of "recoverable error" just results in holes rather than an error message we have to decode)

PR #4306, opened 2018-01-25
[ fix #4304 ] Make the Module Name parser explicit

### LINKED ISSUE #4304: idris --install: "Uncaught error: [filepath] copyFile: does not exists"
The issue is that the parser uses [`iName`](https://github.com/idris-lang/Idris-dev/blob/master/src/Idris/Package/Parser.hs#L120) (that is an `identifier`) to parse a module's
file name but it may not be a valid one (e.g. it can use a reserved keyword such as
`Type` in `Data.Tree.Type`).

gallais:
and share it between Idris.Parser and Idris.Package.Parser

### DIFF
--- a/src/Idris/Package/Parser.hs
+++ b/src/Idris/Package/Parser.hs
@@ -11,6 +11,7 @@ module Idris.Package.Parser where
 import Idris.CmdOptions
 import Idris.Imports
 import Idris.Package.Common
+import Idris.Parser (moduleName)
 import Idris.Parser.Helpers (Parser, Parsing, eol, iName, identifier, isEol,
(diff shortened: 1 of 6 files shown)


ahmadsalim:
Thanks!

gallais:
There seems to be an issue with the test case on windows: the `expected` file
contains paths which use `/` as their delimiters but on windows the delimiter is
`\`. Is there a canonical way of dealing with this in the test-suite?

ahmadsalim:
@melted Any idea? ^

melted:
The test runner should disregard the difference between / and \. So it’s a bit weird.

PR #4308, opened 2018-01-26
Set C language standard to C99 in runtime and codegen

asandroq:
Currently the exact C language used by the runtime and the C code generator is not set and therefore different compilers may use different standards depending on their default behaviour. This may lead to inconsistencies in builds, such a sucessful build in one machine failing in another.

### DIFF
--- a/src/IRTS/CodegenC.hs
+++ b/src/IRTS/CodegenC.hs
@@ -83,7 +83,7 @@ codegenC' defs out exec incs objs libs flags exports iface dbg
              let args = gccDbg dbg ++
                         gccFlags iface ++
                         -- # Any flags defined here which alter the RTS API must also be added to config.mk
-                        ["-DHAS_PTHREAD", "-DIDRIS_ENABLE_STATS",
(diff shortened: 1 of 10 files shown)


src/IRTS/CodegenC.hs
```diff
@@ -83,7 +83,7 @@ codegenC' defs out exec incs objs libs flags exports iface dbg
              let args = gccDbg dbg ++
                         gccFlags iface ++
                         -- # Any flags defined here which alter the RTS API must also be added to config.mk
-                        ["-DHAS_PTHREAD", "-DIDRIS_ENABLE_STATS",
+                        ["-std=c99", "-D_POSIX_C_SOURCE=200809L", "-DHAS_PTHREAD", "-DIDRIS_ENABLE_STATS",
```
ahmadsalim: What is the `"-D_POSIX_C_SOURCE=200809L"` flag?

asandroq: The multi thread code uses some very recent POSIX functions such as `pthread_cond_timedwait` and `pthread_mutexattr_settype` that are only available after this feature is selected.

ahmadsalim:
Thanks @asandroq !

PR #4309, opened 2018-01-27
Update documentation to add the options for CATS in logging

joom:
In the [documentation about internal logging](http://docs.idris-lang.org/en/latest/reference/compilation.html#logging-internal-operation), the options for `CATS` are listed but there is no way to know what the syntax is for those options. It would be nice to have that in the documentation instead of looking up `strLogCat` in `Idris.Options` every time.


ahmadsalim:
Thanks @joom !

jfdm:
There is also the command line option:

```
idris --listlogcats
```

Issue #2099, opened 2015-04-02
Need more list theorems


treeowl:
I'm going to attempt some of these, but I figured I'd open a ticket to track.

1. Cancellation laws for `++`
2. `init xs ++ [last xs] = xs`
3. `inits [] = [[]]` and `inits ys@(_:_) = inits (init ys) ++ [ys]`
4. `inits xs = reverse . tails . reverse $ xs`, if it's worth the trouble
5. `sort` is actually a (stable) sort—this will have to wait for a proven-total version.

One 1 and 2 are complete, it should be safe to hide the choice of `init` implementation, which I think would probably be wise. Once 3 is complete, it should be safe to hide the choice of `inits` implementation, which I think would certainly be wise.

david-christiansen:
Hiding implementations is a bit of a tradeoff. Because our type and term languages coincide, ordinary computation becomes visible in types, and things that are equal are just equal. If the implementation isn't available (eg if the function is abstract) then manual rewriting with equality proofs becomes necessary, which is quite tedious. In essence, treating functions that occur in types as black boxes makes type checking painful, while not treating them as black boxes is perhaps ugly software engineering.

treeowl:
I don't understand all the details, but it seems pretty clear to me that
exposing the implementation allows people to write all sorts of code that
breaks as soon as the implementation changes! I'd much rather see black
boxes, except for some functions exposed explicitly for this purpose. But
I'd much rather get, say, a sorting function whose type explains what its
result means than a sorting function whose code I need to dig into.

treeowl:
Note, for example, that the current inits implementation is asymptotically
bad, so it needs to be replaced. There are several good ways to do it.
Letting everyone depend on one particular one seems to limit future
improvements.

david-christiansen:
This is one of the software engineering challenges of dependent types, that we don't yet have a good answer for.

I'm not saying we shouldn't do it in this case - just that it's not a generally applicable principle.

What we _really_ need is a way to have run time and compile time versions of a function, one used in types and the other in programs, given proofs that they compute the same result for all inputs.

treeowl:
Here are some basics I worked out (some badly) that would be very good to have in some form:

```idris
reverseOnto : List a -> List a -> List a
reverseOnto acc [] = acc
reverseOnto acc (x::xs) = reverseOnto (x::acc) xs

myreverse : List a -> List a
myreverse xs = reverseOnto [] xs

[...]

PR #4310, opened 2018-01-29
Reverse improvements

stephen-smith:
Inspired by my recent frustration of having to write my own reverse in order to prove anything about it and the earlier discussion in #2099.

Last commit is totally optional, but a good idea I think; it hides the implementation of reverse (and dependents).  With the way reverse was implemented previously, I don't think external proofs could have been made about it (or isSuffixOf) because there was no way to reference the local function definition, so I doubt it will break anyone.  The "specification" properties should be enough to make any proofs needed about reverse or reverseOnto without exposing their implementation, which gives us some flexibility if (for some reason) we need to change their implementation again.  (Unfortunately, there's no "specification" for isSuffixOf or isSuffixOfBy, so any proofs about them will have to be in this module, but that was the old status quo as well.)

### DIFF
--- a/libs/contrib/Data/List/Extra.idr
+++ b/libs/contrib/Data/List/Extra.idr
@@ -1,0 +1,22 @@
+module Data.List.Extra
+
+%default total
+%access export
+
+||| Serves as a specification for reverseOnto.
+reverseOntoSpec : (xs, ys : List a) -> reverseOnto xs ys = reverse ys ++ xs
+reverseOntoSpec _  [] = Refl
+reverseOntoSpec xs (y::ys) =
+  trans (reverseOntoSpec (y::xs) ys)
+        (trans (appendAssociative (reverse ys) [y] xs)
+               (cong {f = \l => l ++ xs} (sym (reverseOntoSpec [y] ys))))
+
+||| The reverse of an empty list is an empty list.  Together with reverseCons,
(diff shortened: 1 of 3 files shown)


stephen-smith:
I can move the non-specification properties over to the contrib or base library instead of prelude, if that would be cleaner.  (Moving reverse would be an API change, reverse depends on reverseOnto, the "specification" properties need to "see" the reverse/reverseOnto implementations [which either means reverse remains `public export` (drop last commit) or they need to stay in the same module].)

ahmadsalim:
@stephen-smith Thank you for the PR. 
Yes, please do move the properties outside of prelude to contrib.

stephen-smith:
@ahmadsalim Done.

msmorgan:
The properties shouldn't be in the `Data.List` module of `contrib`, since that module exists in `base` and there would be a conflict. Consider adding a new submodule instead, such as `Data.List.Extra`. Remember to add it to the list of modules in `contrib.ipkg`.

stephen-smith:
@msmorgan Done.

stephen-smith:
I'm pretty sure the failed test_js is unrelated to the new code.

stephen-smith:
Rebased on master and got a good test run, @ahmadsalim and @msmorgan

PR #4315, opened 2018-01-31
Add bitsToHexStr to Data.Bits

clayrat:
Add a helper for printing the hex value of wrapped `Bits`, as it's usually more convenient than the binary. Changing the `Show` instance here would be breaking, I guess.


ahmadsalim:
Thanks!

Issue #4312, opened 2018-01-31
:exec power 10 1000000 returns with no output and does not report error


lpadukana:
# Steps to Reproduce

```
~/code/play/idris $ idris --nobanner
Idris> :exec power 10 1000000
Idris>
```

# Expected Behavior

### One of the following

- Prints `1` followed by `1000000` zeros
- Produces a stack trace in case it fails

# Observed Behavior

Prints nothing, which is wrong

lpadukana:
Symptom is very similar to #4311

ahmadsalim:
@lpadukana Thanks for reporting the issue, but can you please collect them all in #4311 since they are related?

lpadukana:
I reported separately because although the symptoms are similar, I thought they might have different root causes. Ex: this one might be due to a loop terminating prematurely vs the other might be something to do with the type system. Perfectly fine to collect them together if that helps.

yanok:
Well, this one is simple: compiled executable crashes with Segmentation fault because `power` is implemented as naive recursion. So I guess there should be two issues: one to catch and reported crashes from `:exec` and another one to re-implement power using tail recursion (even better using logarithmic power algorithm but it's not structurally recursive).

PR #4320, opened 2018-02-04
Print error message if :exec'ed program exits with failure

yanok:
Check the result code of the compiled program and print an error message if it
didn't return success.

Partially solves #4312.


ahmadsalim:
@yanok Thank you for the PR!

Can you please add a test case and a CHANGELOG entry? 😄

yanok:
@ahmadsalim do you think it is really worth a CHANGELOG entry?

ahmadsalim:
@yanok Sure 👍

yanok:
@ahmadsalim I've updated the PR.

ahmadsalim:
@yanok Thanks! Please, exclude your test from the javascript suite, since it is failing there 😄

yanok:
@ahmadsalim done! but what's the general rule? Tests that rely on :exec should be `C_CG`? Or something else? I wonder how other tests that try to run the compiled results work with the JS backend...

ahmadsalim:
@yanok I think you are relying on the C IO somehow. Maybe @edwinb know more?

PR #4324, opened 2018-02-06
Fix looped temporary allocations in mini-GMP

### LINKED ISSUE #4311: :exec power 10 100000 prints 0 as the output
# Steps to Reproduce

```
~/code/play/idris $ idris --nobanner
Idris> :exec power 10 100000
0
```

# Expected Behavior

### One of the following

- Prints `1` followed by `100000` zeros
- Produces a stack trace in case it fails

# Observed Behavior

Prints `0`, which is wrong

yanok:
Mini-GMP is repeatedly allocating/freeing temporary memory while converting
`mpz_t` to string: to convert some big integer to a string it does a temporary
allocation for each decimal character (and in fact it does so twice: once from
`mpz_sizeinbase` function and then to produce the actual string output).

This results in GC being run while `mpz_get_str` is still working if a number is
big enough. As a result the code may crash or produce an incorrect result
(see #4311).

This patch renames `mpn_div_qr_1_preinv` function to `mpn_div_qr_1_preinv_tmp` and
adds an extra argument to it to pass the pre-allocated temporary buffer.
The old function is available as macro with the old name.
`mpz_sizeinbase` and `mpz_get_str_other` functions (places where the original
function was called inside the loops) are changed to use the new function with
the temporary buffer allocated outside of the loop.

Fixes #4311.


ahmadsalim:
@yanok Thank you for the PR.

I see that you changed the license from LGPL to GPL. 
This is unfortunately an issue, since we would like to keep Idris as BSD3 licensed.

ahmadsalim:
@yanok Or actually, I am unsure whether you did. Is it still LGPL licensed?

yanok:
@ahmadsalim It definitely was not me who changed the license ;) It now says

> he GNU MP Library is free software; you can redistribute it and/or modify
it under the terms of either:
>
>  * the GNU Lesser General Public License as published by the Free
    Software Foundation; either version 3 of the License, or (at your
    option) any later version.
>
> or
>
>  * the GNU General Public License as published by the Free Software
    Foundation; either version 2 of the License, or (at your option) any
    later version.
>
> or both in parallel, as here.

So we are free to choose between LGPL/GPL as we wish. I think there is no problem here but I have to say I'm not a legal expert.

ahmadsalim:
@yanok I see 😄

ahmadsalim:
@yanok Maybe try `IO' ffi ()` instead of `IO` and see if that fixes JS tests?

yanok:
That one exceeds JS maximum stack depth, so I've disabled JS for this one too.

yanok:
After some discussion in gmp mailing list, the gmp developers came up with a nicer way to avoid unnecessary allocations in `mpn_div_qr_1_preinv`. I've updated the PR to be a (trivial) back port the patch merged into gmp development branch.

ahmadsalim:
@yanok Sounds good :)

ahmadsalim:
Thanks!

PR #4333, opened 2018-02-15
Makefiles and changes needed for the C RTS on the seL4 microkernel

mokshasoft:
I have run the following:
git diff --check
And the test suite for both master and with my commit. They both fail on 12 test cases:
'''
12 out of 241 tests failed (449.73s)
Test suite regression-and-feature-tests: FAIL
Test suite logged to: dist/test/idris-1.2.0-regression-and-feature-tests.log
0 of 1 test suites (0 of 1 test cases) passed.
make: *** [Makefile:33: test_c] Error 1
'''
From diffing the test logs it looks that it is only the execution times that are different, so the commit looks ok at least from the test suit perspective.

Check https://github.com/mokshasoft/sel4-idris-manifest for an example of an Idris app running on seL4.

### DIFF
--- a/rts/idris_rts.c
+++ b/rts/idris_rts.c
@@ -114,7 +114,7 @@ void init_threaddata(VM *vm) {
 }
 
 void init_signals(void) {
-#if (__linux__ || __APPLE__ || __FreeBSD__ || __DragonFly__)
+#if defined(__linux__) || defined(__APPLE__) || defined(__FreeBSD__) || defined(__DragonFly__)
     signal(SIGPIPE, SIG_IGN);
 #endif
 }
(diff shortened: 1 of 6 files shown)


ahmadsalim:
Thanks for the PR.
I would let @edwinb review this, since it changes the C RTS .

melted:
I don’t think it needs a review from busy edwin, it just adds a new target, and ifdefs out a thing for that. As long as it works on the CI systems we should be fine :-)

edwinb:
That would indeed be the nature of my review :)

mokshasoft:
The travis-ci failure comes from:
'''
Still running (1 of 20): make TEST-JOBS=2 test_js

The job exceeded the maximum time limit for jobs, and has been terminated.
'''
The only test that failed is test_js which should not have been affected by the changes in the C RTS.

melted:
Thanks for the PR!

PR #4336, opened 2018-02-17
Add `Identity a` basic instances

aiya000:
I want the basic instances for `Identity a` :relaxed:

### DIFF
--- a/libs/base/Control/Monad/Identity.idr
+++ b/libs/base/Control/Monad/Identity.idr
@@ -11,8 +11,21 @@ implementation Functor Identity where
 
 implementation Applicative Identity where
     pure x = Id x
-
     (Id f) <*> (Id g) = Id (f g)
 
 implementation Monad Identity where
     (Id x) >>= k = k x
+
+implementation Show a => Show (Identity a) where
+  showPrec d (Id x) = showCon d "Id" $ showArg x
+
+implementation Eq a => Eq (Identity a) where
+  (Id x) == (Id y) = x == y
+
+implementation Ord a => Ord (Identity a) where
+  compare (Id x) (Id y) = compare x y
+
+implementation Enum a => Enum (Identity a) where
+  toNat (Id x) = toNat x
+  fromNat n = Id $ fromNat n
+  pred (Id n) = Id $ pred n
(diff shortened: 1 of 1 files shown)


Melvar:
Would you please make the `Show` instance parenthesize when necessary by implementing and using `showPrec` instead of `show`? Read the docs in `Prelude.Show`, there are helpers.

aiya000:
@Melvar  I fixed now :+1:

I have a question, (I'm Idris beginner :joy:).

An `implementation` keyword is often shown in instance definitions,
but I think the instances can be defined without `implementation`.

What is it?
and shouldn't I use `implementation`?

Melvar:
It needs to be `showArg x` instead of `show x`.
The implementation keyword is optional, I think in prelude and base libraries it’s usually left off.

aiya000:
@Melvar Can I remove `implementation` in Identity.idr?
And are other things OK?
(sorry, I'm English beginner :joy:)

libs/base/Control/Monad/Identity.idr
```diff
@@ -11,8 +11,21 @@ implementation Functor Identity where
...
 implementation Applicative Identity where
     pure x = Id x
-
     (Id f) <*> (Id g) = Id (f g)
 
 implementation Monad Identity where
     (Id x) >>= k = k x
+
+implementation Show a => Show (Identity a) where
+  showPrec d (Id x) = showCon d "Id" $ show x
```
Melvar: The `show x` here needs to be `showArg x` instead.

Melvar:
You can remove the `implementation` keyword.

aiya000:
Thanks, I finished to add `showArg` and remove `implementation` keyword 👍

ahmadsalim:
@aiya000 I think the idea was that you should use the `showArg` function on the argument (like you used `showCon` on the constructor), and not rename `show` to `showArg` 😄

aiya000:
@ahmadsalim Thank you! I might understand 😄 
@Melvar The task maybe is done, would you please check it :D

Melvar:
Looks correct now.

ahmadsalim:
Cool, thanks!

PR #4337, opened 2018-02-17
Fin.restrict fix

trsutton:
The following changes resolve a bug in Fin.restrict where the Fin value wasn't being bound correctly, resulting in the inability to create a Fin representing the last value of the bounded integer (e.g. using the result of `restrict 1 1` to index a Vect of length 2 was not possible). This fix explicitly passes the implicit bound to the `Fin.fromInteger`. I also wrote a test to capture the behavior.

My apologies in advance. The commits were not as clean as I would have liked: 

(1) I was using Travis to build the changes and run the tests, resulting in a subsequent commit to resolve the test issues. 
(2) Additionally, after committing I realized that my user name had a typo so I committed a fix where I reset the author of the commit.


ahmadsalim:
Thanks!

PR #4338, opened 2018-02-18
Add Effects.runPureEnv

aiya000:
This allows programs that are like below

```idris
import Effect.State
import Effects

-- runPureEnv : (env : Env Basics.id xs) -> (prog : EffM Basics.id a xs xs') -> (x : a ** Env Basics.id (xs' x))
-- runPureEnv env prog = eff env prog (\r, env => (r ** env))

main : IO ()
main = do
  let (_ ** [s]) = runPureEnv [default] (do update (+1))
  printLn s
```

This maybe used in any effect, like Haskell's `Reader` :smile:

Also a document comment of `runEnv` is added as the preparing :dog2:

PS.
`%no_implicit` is inserted on `runPureEnv`, but I don't know what is `%no_implicit` :bow:
Are problems nothing?

### DIFF
--- a/libs/effects/Effects.idr
+++ b/libs/effects/Effects.idr
@@ -440,6 +440,11 @@ runEnv : Applicative m => Env m xs -> EffM m a xs xs' ->
          m (x : a ** Env m (xs' x))
 runEnv env prog = eff env prog (\r, env => pure (r ** env))
 
+||| Similar to 'runEnv', but the context (m) is 'pure'
+%no_implicit
+runPureEnv : (env : Env Basics.id xs) -> (prog : EffM Basics.id a xs xs') -> (x : a ** Env Basics.id (xs' x))
+runPureEnv env prog = eff env prog (\r, env => (r ** env))
+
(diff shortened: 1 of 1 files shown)


tbeddy:
This doesn't invalidate the PR, but isn't Effects being deprecated in favor of [Control.ST](https://github.com/idris-lang/Idris-dev/blob/master/libs/contrib/Control/ST.idr)?

aiya000:
Is it meaning what the whole of `Effects` are deprecated ? 🐱

tbeddy:
Yeah. From [the Effects tutorial:](http://docs.idris-lang.org/en/latest/effects/index.html)

> There is a new module in the contrib package, Control.ST, which provides the resource tracking facilities of Effects but with better support for creating and deleting resources, and implementing resources in terms of other resources. Unless you have a particular reason to use Effects you are strongly recommended to use Control.ST instead.

aiya000:
mmm, Thanks, I understood :joy:

Should this be closed?

ahmadsalim:
Thanks! 😄

PR #4339, opened 2018-02-19
fix for odd-even bug in random

martinbaker:
Fix for the following bug and also add comments.

When I call (rndInt 0 100) repeatedly the result does not look very psudorandom, for instance, it gives alternating odd and even numbers.

When I look at the Idris code here:
https://github.com/idris-lang/Idris-dev/blob/master/libs/effects/Effect/Random.idr
and I enter the code manually I also get alternating odd and even numbers:

```
Idris> :let seed = 1
Idris> :let seed=((1664525 * seed + 1013904223) `prim__sremBigInt` (pow 2 32))
Idris> seed
1015568748 : Integer
Idris> :let seed=((1664525 * seed + 1013904223) `prim__sremBigInt` (pow 2 32))
Idris> seed
1586005467 : Integer
Idris> :let seed=((1664525 * seed + 1013904223) `prim__sremBigInt` (pow 2 32))
Idris> seed
2165703038 : Integer
Idris> :let seed=((1664525 * seed + 1013904223) `prim__sremBigInt` (pow 2 32))
Idris> seed
3027450565 : Integer
Idris> :let seed=((1664525 * seed + 1013904223) `prim__sremBigInt` (pow 2 32))
Idris> seed
217083232 : Integer
Idris> :let seed=((1664525 * seed + 1013904223) `prim__sremBigInt` (pow 2 32))
Idris> seed
1587069247 : Integer
Idris> :let seed=((1664525 * seed + 1013904223) `prim__sremBigInt` (pow 2 32))
Idris> seed
3327581586 : Integer
```

### DIFF
--- a/libs/effects/Effect/Random.idr
+++ b/libs/effects/Effect/Random.idr
@@ -5,6 +5,19 @@ import Data.Vect
 
 %access public export
 
+{-
+ - This code implements a 'Linear Congruential Generator' as described here
+ - https://en.wikipedia.org/wiki/Linear_congruential_generator
+ -
+ - Looking at the wiki page for 'linear congruential generator' it looks like the
+ - parameters used by Idris come from the 'Numerical Recipes' books. The web page
+ - does not give any known issues with these parameters, however another example is
+ - known to have this bug: "If Xn is even then Xn+1 will be odd, and vice versa-the
+ - lowest bit oscillates at each step." So although the 'Numerical Recipes' parameters
+ - are not listed with this issue they do have it. Some of the implementations do not
+ - use the lower significant bits so divBigInt v 2 has been added to do this.
+ -}
+
 data Random : Effect where 
      GetRandom : sig Random Integer Integer
      SetSeed   : Integer -> sig Random () Integer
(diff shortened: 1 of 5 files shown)


libs/effects/Effect/Random.idr
```diff
@@ -33,7 +48,8 @@ rndFin k = do let v = assert_total $ !(call GetRandom) `prim__sremBigInt` (cast
        toFin x = case integerToFin x (S k) of
                       Just v => v
                       Nothing => toFin (assert_smaller x (x - cast (S k)))
-
+       -- TODO insert divBigInt x 2 to prevent low bit alternating between
```
ahmadsalim: Please remove TODO

ahmadsalim:
@martinbaker It seems some tests are broken, can you please take a look at them?

martinbaker:
test/effects001
test/effects003

are hardcoded to expect particular outputs from the pseudorandom generator.
Since this has changed the tests have to be modified to expect the new values.

I have changed the tests to what I think they should be but I don't really know much about AppVeyor and Travis so I might have got it wrong.

Martin

ahmadsalim:
@martinbaker It seems that `effect003` still fails, but now on a more minor thing.

You can try to run the test locally (by using the script inside the folder).

ahmadsalim:
Thanks!

PR #4343, opened 2018-02-21
Various additions to contrib

Synthetica9:
* `Data.SortedBag`: Bag (or Multiset) implemention based on `Data.SortedMap`.
  * `Data.PosNat`: A `Nat` paired with a proof that it is positive.
  * `Data.Chain`: A function with an arbitrary number of arguments, plus combinators for working with them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 + REPL now prints an error message if program compiled by `:exec` terminates
   abnormally.
++ Additions to `contrib`:
+  * `Data.SortedBag`: Bag (or Multiset) implemention based on `Data.SortedMap`.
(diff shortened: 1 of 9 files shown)


ahmadsalim:
Thanks!

Melvar:
`Data.Chain.Chain` appears to be a special case of `Data.Fun.Fun`. Why was this added as a separate module?

Synthetica9:
@Melvar The case where all arguments are of the same type allows for more combinators to be defined, see for example `reflexes` and `before`.

Melvar:
Sure, but there is a duplicate, and they’re closely related … a separate module seems overkill. Also I’m suspicious about the usefulness – remember that Idris can’t invert `Chain`, so a lot of specifying implicits is likely to be required. (The same with `Fun`.)

PR #4348, opened 2018-02-23
Fixed `Data.SortedBag.contains` and added docs to Data.PosNat

Synthetica9:
Turns out my last PR wasn't as polished as I thought.


ahmadsalim:
Thanks

PR #4353, opened 2018-02-25
Fix #4346 - add Network.Socket.recvAll

### LINKED ISSUE #4346: Receive-all function for Socket?
`Socket` has a `recv` function that reads into a buffer of the desired length, but very often you simply want to read until the end of the input. Can we have a function like:

`recvAll : (sock : Socket) -> IO (Either SocketError (String, ResultCode))`

That calls `recv` until the socket is exhausted and then returns the data read in a single value?

I am happy to send a pull request, if there is a chance of it being merged?

Dretch:


melted:
Thanks for the PR!

melted:
Huh, why didn't I merge this?

Issue #4352, opened 2018-02-25
readFile implementation in Prelude.File partially broken with binary files


22459:
The current implementation uses `addToStringBuffer` to concatenate a read file chunk to a buffer.
Explicit length parameters seem to be used almost everywhere, however not for `addStringToBuffer`,
which uses C `strlen` in the underlying implementation to deduce the length of the concatenee. 
`strlen` uses C style strings, resulting in truncation of the read chunk at null characters.

Proposal: require an explcit length parameter for addStringToBuffer.

https://github.com/idris-lang/Idris-dev/blob/5eb8f8393f068fb04aaf02d0a8617d16f5092d93/libs/prelude/Prelude/File.idr#L273

https://github.com/idris-lang/Idris-dev/blob/e0198402903117227d0d319e8a8388d1093a09ec/libs/prelude/Prelude/Strings.idr#L57-L59

https://github.com/idris-lang/Idris-dev/blob/f92ecd25d48a9120abfbb7d6af81f9b194ab98f1/src/Idris/Core/Execute.hs#L441-L449

https://github.com/idris-lang/Idris-dev/blob/a4c183c83762c8ef75c0c927fde3479ce356b0a4/rts/idris_stdfgn.c#L206-L212

A more grand proposal?: Should Idris globally switch to netstring style strings instead of C strings.

22459:
I think `fGetChars` should be modified to return a tuple of `(String, Int)` where the `Int` is the return value of `fread()`.

22459:
All of a sudden I'm having trouble replicating my original issue on an unmodified branch.

22459:
Test case for compiled:

```
$ cat asd.idr
main : IO ()
main =
  do
    (Right l) <- readFile "main"
    putStr l

$ xxd main | head -n 1
00000000: 7f45 4c46 0201 0100 0000 0000 0000 0000  .ELF............

$ idris asd.idr -o asd && ./asd | xxd | head -n 1
00000000: 7f45 4c46 0201 01                        .ELF...


```

22459:
It has been brought to my attention that `MKSTR()` also uses `strlen` and `strcpy`; with `fGetChars` using `MKSTR()` in it's underlying implementation.

Melvar:
Also note that primitives can’t take or return values of non-primitive types, and that the C FFI also only supports a limited range of types which doesn’t include pairs, so neither can return a pair.

ahmadsalim:
[...]

PR #4354, opened 2018-02-26
WIP: making readFile work correctly with binary data when compiled

22459:
WIP:
make fGetChars return a tuple of (string, length) instead of just the string result,
make addToStringBuffer take a length argument for the length of its string parameter


22459:
PR for https://github.com/idris-lang/Idris-dev/issues/4352

22459:
I got stuck on trying to figure out how to return a tuple in the RTS for `idris_readChars`.

PR #4361, opened 2018-03-01
stack test --nix should work now

jacereda:
`stack test --nix` was failing on Mac. Looks like this makes it pass, although I'm not very experienced with Cabal and maybe there's a better approach.

### DIFF
--- a/Setup.hs
+++ b/Setup.hs
@@ -18,6 +18,7 @@ import Distribution.Simple.Utils (createDirectoryIfMissingVerbose, rewriteFile, 
 import Distribution.Compiler
 import Distribution.PackageDescription
 import Distribution.Text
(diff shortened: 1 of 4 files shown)


jacereda:
This is what the output looked like before the patch:

```
             
Progress: 1/2Regression and feature tests
  Base
    base001:           
Error: test/basic007
ld: library not found for -lgmp
clang: error: linker command failed with exit code 1 (use -v to see invocation)
run: line 3: ./test033: No such file or directory


Error: test/basic002
ld: library not found for -lgmp
clang: error: linker command failed with exit code 1 (use -v to see invocation)
run: line 3: ./test006: No such file or directory

...
```

melted:
Thanks!

PR #4363, opened 2018-03-02
Fix shell return code, closes #4362

### LINKED ISSUE #4362: build.pl doesn't return the right error code to the shell
Looks like fasta.idr isn't building properly right now, but CI is still happy because build.pl doesn't return the right error code to the shell.

jacereda:


melted:
Hmmm, and then Travis fails as the fasta benchmark fails.

jacereda:
Yes, should be OK again when https://github.com/idris-lang/Idris-dev/issues/4364 gets fixed.

jfdm:
So I am going to merge this. I've committed a patch that gets the fasta benchmark passing again by rewriting the errorful code. Given that we have #4364 to track the original issue for why fasta was failing i think it was prudent to get the fix by @jacereda for the benchmark suite integrated.

jfdm:
Oh @jacereda thanks for the patch, please add yourself to the contributors file, and if you want find other things in idris to fix.

PR #4366, opened 2018-03-03
Allow capitalized pattern variables via switch

### LINKED ISSUE #4359: Not a valid name for a pattern variable
The program
```
> %default total
> %access public export
> %auto_implicits off


> Unique : Type -> Type
> Unique t = (p : t) -> (q : t) -> p = q

> Unique1 : {A : Type} -> (A -> Type) -> Type
> Unique1 {A} P = (a : A) -> Unique (P a)
```
does not type check with 1.2.0-git:507a7312. The error message is
```
- + Errors (1)
 `-- ./issues/not_a_valid_name.lidr line 10 col 2:
     When checking left hand side of Unique1:
     When checking argument A to Main.Unique1:
             A is not a valid name for a pattern variable
```
There are hundreds of lines of code in https://gitlab.pik-potsdam.de/botta/IdrisLibs/tree/towards1.1.1 that do not type check anymore with similar error messages. Until a few weeks ago and under 1.2.0, they typed checked fine. Perhaps a regression?

ahmadsalim:
Fixes #4359

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ## Library updates
 
-+ changed rndInt in Effect.Random so that it does not alternate between odd
++ Changed rndInt in Effect.Random so that it does not alternate between odd
   and even.
 + Additions to `contrib`:
   * `Data.SortedBag`: Bag (or Multiset) implemention based on `Data.SortedMap`.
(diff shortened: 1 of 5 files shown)


edwinb:
I just wonder if that's the right solution even if it works for now...

PR #4374, opened 2018-03-09
Patched fasta benchmark so it compiles.

jfdm:
Underlying problem with string literals needs to be addressed.

that can come later.


jfdm:
This is going to pass all checks, but the last two are remaining. I am impatient so merging this.

PR #4376, opened 2018-03-11
Add tryCatch as an Elab action in Language.Reflection.Elab

joom:
There already is `tryCatch :: Elab' aux a -> (Err -> Elab' aux a) -> Elab' aux a` in Idris.Core.Elaborate in the compiler, but there's no matching Elab action in Language.Reflection.Elab in the Idris prelude.

Now we have a new Elab action `tryCatch : Elab a -> (Err -> Elab a) -> Elab a` that lets us learn *why* the first Elab action failed and do different things based on the error.

## Example

```idris
import Language.Reflection.Elab

%language ElabReflection

f : Err -> Elab ()
f (Msg _) = fill `("message error")
f (CantUnify _ _ _ _ _ _) = fill `("unification error")
f _ = fill `("other")

s1 : String
s1 = %runElab (do tryCatch (fail []) f ; solve)

s2 : String
s2 = %runElab (do tryCatch (fill `(True)) f ; solve)

main : IO ()
main = do putStrLn s1 ; putStrLn s2
```

This program would print
```
message error
unification error
```

### DIFF
--- a/libs/prelude/Language/Reflection/Elab.idr
+++ b/libs/prelude/Language/Reflection/Elab.idr
@@ -16,6 +16,7 @@ import Prelude.Maybe
 import Prelude.Monad
 import Prelude.Nat
 import Language.Reflection
+import Language.Reflection.Errors
 
 %access public export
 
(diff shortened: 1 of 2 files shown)


melted:
Thanks for the PR!

@david-christiansen Any comments?

david-christiansen:
I will read it in detail before bed tonight.

david-christiansen:
Only one comment: it would be great if the example from the PR was included as a test.

Thanks!

david-christiansen:
I seem to have failed Github that this was merged when I did it from the command line with the added test, but it is there in 09fe4917058955f47d23ea59244d1d0729608f1a.

david-christiansen:
Thanks again!

joom:
My intention when writing the example was to add it as a test, but then I couldn't find the other elaborator reflection tests. Apparently they are in `meta***`. Thanks!

PR #4377, opened 2018-03-12
Add conversions between BoundedList and Vect

brainrape:
A small addition to `contrib`, suggested in a StackOverflow [comment](https://stackoverflow.com/questions/49193132/how-to-write-an-idris-function-that-accepts-a-vect-of-length-less-than-3#comment85397084_49195458).


melted:
Thanks for the PR!

PR #4385, opened 2018-03-18
Smaller closure

jacereda:
Looks like something like this could be useful to reduce GC frequency. Thoughts?

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -14,7 +14,7 @@ RANLIB          ?=ranlib
 CABAL           :=cabal
 # IDRIS_ENABLE_STATS should not be set in final release
 # Any flags defined here which alter the RTS API must also be added to src/IRTS/CodegenC.hs
-CFLAGS          :=-O2 -Wall -std=c99 -D_POSIX_C_SOURCE=200809L -DHAS_PTHREAD -DIDRIS_ENABLE_STATS $(CFLAGS)
+CFLAGS          :=-O2 -Wall -g -std=c99 -D_POSIX_C_SOURCE=200809L -DHAS_PTHREAD -DIDRIS_ENABLE_STATS $(CFLAGS)
 
 # CABALFLAGS	:=
 CABALFLAGS      += --enable-tests
--- a/config.mk
+++ b/config.mk
@@ -14,7 +14,7 @@ RANLIB          ?=ranlib
 CABAL           :=cabal
 # IDRIS_ENABLE_STATS should not be set in final release
 # Any flags defined here which alter the RTS API must also be added to src/IRTS/CodegenC.hs
-CFLAGS          :=-O2 -Wall -g -std=c99 -D_POSIX_C_SOURCE=200809L -DHAS_PTHREAD -DIDRIS_ENABLE_STATS $(CFLAGS)
+CFLAGS          :=-O2 -Wall -std=c99 -D_POSIX_C_SOURCE=200809L -DHAS_PTHREAD -DIDRIS_ENABLE_STATS $(CFLAGS)
 
 # CABALFLAGS	:=
 CABALFLAGS      += --enable-tests
--- a/rts/idris_gc.c
+++ b/rts/idris_gc.c
@@ -17,22 +17,22 @@ VAL copy(VM* vm, VAL x) {
             return x;
         } else {
             allocCon(cl, vm, CTAG(x), ar, 1);
-            memcpy(cl->info.cargs, x->info.cargs, sizeof(VAL)*ar);
+            memcpy(cl->extra.cargs, x->extra.cargs, sizeof(VAL)*ar);
         }
         break;
     case CT_ARRAY:
-        len = x->sz;
+        len = CELEM(x);
 	allocArray(cl, vm, len, 1);
-        memcpy(cl->info.array, x->info.array, sizeof(VAL)*len);
+        memcpy(cl->extra.array, x->extra.array, len * sizeof(VAL));
         break;
     case CT_FLOAT:
         cl = MKFLOATc(vm, GETFLOAT(x));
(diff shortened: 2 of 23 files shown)


rts/idris_rts.c
```diff
@@ -7,6 +7,11 @@
 #include "idris_bitstring.h"
 #include "getline.h"
 
+#define STATIC_ASSERT(COND,MSG) typedef char static_assertion_##MSG[(COND)?1:-1]
+
+
+STATIC_ASSERT(sizeof(Closure) == 16, sizeofClosure);
```
melted: This could be false on 32-bit systems depending on how double is aligned.

melted:
Thanks for the PR!

melted:
I’ll take a serious look at it after work. Please try to test a 32-bit build also.

rts/idris_rts.c
jacereda: I guess a 32-bit "test_c" job in the CI would be nice... I have a 32-bit Nix system at work, I can try there anyway.

jacereda:
I don't expect this to be merged yet, I opened it to get some feedback on whether this is worth pursuing. I think I can make some further improvements to reduce memory usage even more (and maybe simplify it at the same time). Please, wait before doing a real review.

jacereda:
I'm trying an alternative approach using a header instead of a union at https://github.com/jacereda/Idris-dev/tree/smaller-closure-3 and the result seems promising. Should I open a separate PR or merge into this branch?

melted:
Better to keep it in one PR as only one of the ways will get in.

I don't think there should be bitfields. It makes it a pain to access from other languages than C, and it would be nice if one could mess with Idris internals from Idris code. Also, we don't have to reserve 56 bits for size, let's say that people who want to allocate more than 4GB have to use a pointer.

jacereda:
@melted, I think this is ready. I have a couple of branches on top of this one that make further improvements like reducing Bits8/Bits16 to 8 bytes and improving support for developing on NixOS/Nixpkgs, but I guess it would be better to go in small steps.

Also, I guess I should have opened a separate PR for https://github.com/idris-lang/Idris-dev/pull/4385/commits/11130b194d49a4ce158cb4b91f2716688b92c65c, since it's a bug.

The 32-bit test suite is still broken, but that's also the case in master.

config.mk
```diff
@@ -14,7 +14,7 @@ RANLIB          ?=ranlib
 CABAL           :=cabal
 # IDRIS_ENABLE_STATS should not be set in final release
 # Any flags defined here which alter the RTS API must also be added to src/IRTS/CodegenC.hs
-CFLAGS          :=-O2 -Wall -std=c99 -D_POSIX_C_SOURCE=200809L -DHAS_PTHREAD -DIDRIS_ENABLE_STATS $(CFLAGS)
+CFLAGS          :=-O2 -Wall -g -std=c99 -D_POSIX_C_SOURCE=200809L -DHAS_PTHREAD -DIDRIS_ENABLE_STATS $(CFLAGS)
```
melted: remove

rts/idris_gc.c
```diff
@@ -5,115 +5,85 @@
...
-           }
-           break;
-       case CT_REF:
-           heap_item->info.ptr = copy(vm, (VAL)(heap_item->info.ptr));
-           break;
-       case CT_STROFFSET:
-           heap_item->info.str_offset->str
-               = copy(vm, heap_item->info.str_offset->str);
-           break;
+       case CT_CON: {
```
melted: uneven indentation in the cases.

rts/idris_gmp.c
```diff
@@ -22,264 +22,141 @@ VAL MKBIGI(int val) {
...
+    mpz_fdiv_q_2exp(*getmpz(cl), *getmpz(getbig(vm,x)), GETINT(y));
+    return (VAL)cl;
 }
 
 VAL idris_bigAnd(VM* vm, VAL x, VAL y) {
     if (ISINT(x) && ISINT(y)) {
         return INTOP(&, x, y);
     } else {
-        return bigAnd(vm, GETBIG(vm, x), GETBIG(vm, y));
+        return bigAnd(vm, (VAL)GETBIG(vm, x), (VAL)GETBIG(vm, y));
```
melted: GETBIG is defined with a cast to VAL already.

Further occurences below.

rts/idris_rts.h
```diff
@@ -25,68 +23,115 @@
...
+
+typedef struct Float {
+  Hdr hdr;
+  double f;
+} Float;
+
+typedef struct String {
+  Hdr hdr;
+  size_t slen;
+#define _null hdr.u8
```
melted: Remove this and substitute `hdr.u8` in the use sites and put in a comment that String uses the hdr.u8 field for null.

rts/idris_rts.h
```diff
@@ -25,68 +23,115 @@
...
+typedef struct Val {
+    Hdr hdr;
+} Val;
+
+typedef struct Val * VAL;
+
+typedef struct Con {
+  Hdr hdr;
+  uint32_t tag;
+  uint32_t arity;
```
melted: Arity could use one of the fields in the header. That would save four bytes on 32-bit.

jacereda: Not now, since the aligned() function is now requiring 8 bytes alignment, but aligned() could as well align to sizeof(intptr_t). I'll change that.

rts/idris_gc.c
jacereda: Can I add a .clang-format file that tries to be consistent with current formatting?

melted:
Let's merge this. I'll note it in the change log.

Thanks for the contribution!

PR #4389, opened 2018-03-21
Support skipping any number of shebang lines.

srdqty:
Hello :) I just started learning Idris and I wanted to be able to run my scripts with nix-shell and found that the second shebang line was causing a parse error with Idris. I ran tests locally with `stack --nix test` and everything passed. Thanks for any feedback.

Below is what I'm talking about:

```
#! /usr/bin/env nix-shell
#! nix-shell -i idris -p idris

main : IO ()
main = putStrLn "Hello world from nix-shell shebang!"
```

Additional changes:

Updates `stack-shell.nix` to include perl and nodejs.

I was getting errors about perl and node missing when I tried to run `stack --nix test.`

### DIFF
--- a/src/Idris/Parser.hs
+++ b/src/Idris/Parser.hs
@@ -1497,7 +1497,7 @@ parseImports fname input
                                  [ImportInfo],
                                  Maybe Mark),
                                 [(FC, OutputAnnotation)], IState)
-        imports = do optional shebang
+        imports = do many shebang
                      whiteSpace
                      (mdoc, mname, annots) <- moduleHeader
                      ps_exp        <- many import_
--- a/stack-shell.nix
+++ b/stack-shell.nix
@@ -7,6 +7,8 @@ let
     zlib
     ncurses
     gmp
+    nodejs
+    perl
   ];
   native_libs = lib.optionals stdenv.isDarwin (with darwin.apple_sdk.frameworks; [
     Cocoa
--- a/test/interactive017/expected
+++ b/test/interactive017/expected
@@ -5,3 +5,5 @@ hello from node
 Hello ["aaa"]
 Hello ["bbb", "aaa"]
 Hello from a module
+hello from C with multiple shebangs
+hello from node with multiple shebangs
--- a/test/interactive017/run
+++ b/test/interactive017/run
@@ -7,3 +7,5 @@ PATH="../../scripts:$PATH"
 ./shebang-args.idr aaa
 ./shebang-args.idr bbb aaa
 ./shebang-import.idr
+./shebang-multiple.idr
+./shebang-node-multiple.idr
--- a/test/interactive017/shebang-multiple.idr
+++ b/test/interactive017/shebang-multiple.idr
@@ -1,0 +1,5 @@
+#!/usr/bin/env runidris
+#! shebang 2
+
+main : IO ()
+main = putStrLn "hello from C with multiple shebangs"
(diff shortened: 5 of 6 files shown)


Melvar:
So … what exactly is the point of having a second shebang line?

NobbZ:
I'm not sure what this nix-shell is about, but isn't it a common pattern to use `--` to ignore following flags and switches and pass them down the road, such that `#!/usr/bin/env nix-shell -- -i idris -p idris` might work?

Synthetica9:
@NobbZ Also doesn't work

Melvar:
My interpretation of this is that the second line isn’t actually a shebang line in any useful sense; it’s a directive to `nix-shell` which means it’s `nix-shell`’s responsibility to recognize it independently of the comment syntax of the language of the file, and not the language’s responsibility to support an unrelated tool’s choice of directive syntax.

srdqty:
Hello, checking in about this PR. I understand the inclination to perhaps not merge it on principle. I guess in my mind, skipping 1 vs many lines with a shebang prefix is the same level of hackiness.

For more context, here is an example of when I was using this back when I was learning haskell. The reason it's such a nice to have is because you can so easily run your one-off learning scripts that require libraries.

```
#! /usr/bin/env nix-shell
#! nix-shell -i runhaskell -p 'pkgs.haskell.packages.ghc841.ghcWithPackages (ps: with ps; [hspec])'

import Data.Monoid (Sum (Sum), Product (Product))
import Data.Semigroup (Semigroup ((<>)))

import Test.Hspec
    ( context
    , describe
    , hspec
    , it
    , shouldBe
    )

data Optional a
    = Nada
    | Only a
    deriving (Eq, Show)

instance Semigroup a => Semigroup (Optional a) where
    Nada   <> Nada = Nada
    Nada   <> x = x
    x      <> Nada = x
    Only x <> Only y = Only (x <> y)

instance Monoid a => Monoid (Optional a) where
    mempty = Nada

main :: IO ()
main = hspec $
    describe "Optional Monoid" $ do
        context "both Only" $ do
            it "mappends Sum" $
                shouldBe
                    (Only (Sum 1) `mappend` Only (Sum 1))
                    (Only (Sum 2))
            it "mappends Product" $
                shouldBe
                    (Only (Product 4) `mappend` Only (Product 2))
                    (Only (Product 8))
        context "with Nada as second argument" $ do
            it "returns Only Sum" $
                shouldBe
                    (Only (Sum 1) `mappend` Nada)
                    (Only (Sum 1))
            it "return Only List" $
                shouldBe
                    (Only [1] `mappend` Nada)
                    (Only [1])
        context "with Nada as first argument" $ do
            it "returns Only Sum" $
                shouldBe
                    (Nada `mappend` Only (Sum 1))
                    (Only (Sum 1))
            it "return Only List" $
                shouldBe
                    (Only [1] `mappend` Nada)
                    (Only [1])
```

melted:
@NobbZ shebangs can only take one argument, it's a terrible hack.

The cost would probably be acceptable, since it wouldn't do any harm to normal code.

I wonder why nix doesn't just use an ordinary comment like emacs and vim file configs (or stack) though. Because patching all the languages to support non-standard shebangs sounds  harder than that.

srdqty:
Hmm, maybe I should look into adding support to nix-shell for parsing more styles of single line comment prefixes. The change in this PR was so attractive because it was only a one word change lol.

PR #4396, opened 2018-03-28
Fix compilation with GHC 8.4

melted:


melted:
Not elegant at all. But there isn't a nice way to solve this.

PR #4398, opened 2018-03-30
Allow building on GHC 8.4.1

RyanGlScott:
Thankfully, only the `Setup` script needs to be changed:

* `FlagAssignment` is now an abstract type, so we need to use the `lookupFlagAssignment` function instead of `lookup`.
* `rewriteFile` is now deprecated in favor of `rewriteFileEx`.

I also needed to bump the upper version bounds on `Cabal` and `async` to support GHC 8.4.1.

### DIFF
--- a/Setup.hs
+++ b/Setup.hs
@@ -14,7 +14,7 @@ import Distribution.Simple.InstallDirs as I
 import Distribution.Simple.LocalBuildInfo as L
 import qualified Distribution.Simple.Setup as S
 import qualified Distribution.Simple.Program as P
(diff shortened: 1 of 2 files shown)


melted:
Hmmm, I wonder why didn't need this when I did https://github.com/idris-lang/Idris-dev/pull/4396

melted:
I did use `--allow-newer=base` to build though.

RyanGlScott:
It's possible you installed an older version of `Cabal` somehow. Note that I'm using the version of `Cabal` shipped with GHC 8.4.1 itself, `Cabal-2.2.0.0`.

melted:
Ah, it installs Cabal-2.0 because of the previous bound.

Thanks for the patch!

PR #4405, opened 2018-04-02
Build on Nixpkgs/NixOS

jacereda:
Some problems with this PR:

- Still contains a crude hack (a chmod on Darwin after compiling). I think this is due to a bug in clang-wrapper, but it could well be some silly error of my own.
- Enabled FFI+GMP by default due to my low skills with cabal and Nix.
- Removes the TIMEOUT invocation for some scripts, I still need to figure out why those failed when invoked through it.
- I didn't test anything on systems without Nixpkgs.

Maybe someone has ideas on how to deal with those.


david-christiansen:
I'm not a Nix expert, so sorry if this is misguided. How do we ensure that updates to Idris don't break the Nix build script? Can we reasonably run it from Travis, for instance?

Makefile
```diff
@@ -15,68 +19,86 @@ endif
...
-	$(MAKE) -C docs latexpdf
+	$(MK) -C docs latexpdf
 
 fast:
-	$(CABAL) install $(CABALFLAGS) --ghc-option=-O0
+	$(CB) install $(CABALFLAGS) --ghc-option=-O0
 
 dist/setup-config:
-	$(CABAL) configure $(CABALFLAGS)
+	$(CB) configure $(CABALFLAGS)
```
david-christiansen: What is the purpose of renaming the CABAL and MAKE variables? These shorter versions are (slightly) more difficult for me to understand.

idris.cabal
```diff
@@ -108,12 +108,12 @@ custom-setup
 
 Flag FFI
   Description:  Build support for libffi
-  Default:      False
```
david-christiansen: These flags should stay as they are, because libffi can be difficult for Mac users and some people don't like the license of GMP.

idris.nix
```diff
@@ -0,0 +1,61 @@
+{ mkDerivation, aeson, annotated-wl-pprint, ansi-terminal
```
david-christiansen: Is this output from cabal2nix, or is it hand-written? If it's generated code, should it be in the repo, or should the repo contain instructions to generate it? I'm not sure what best practices are around that.

libs/Makefile
```diff
@@ -1,37 +1,45 @@
-build:
```
david-christiansen: These changes seem unrelated to the task of building Idris with nix. Could they be in a separate PR/commit? And do they maintain compatibility with BSD make?

test/basic010/run
```diff
@@ -1,10 +1,4 @@
 #!/usr/bin/env bash
-
```
david-christiansen: Why change this script? And likewise for reg039.

Makefile
jacereda: Those try to ensure the correct `idris` is invoked via `env IDRIS=$(IDRIS) ...` and I simply used those. I'm not particularly attached to any of those names though, what would you suggest?

idris.cabal
jacereda: And I agree, I'm sort of asking how should this be handled in the second bullet point.

idris.nix
jacereda: It is the output from `cabal2nix` with some hand-edits to make sure `perl`, `nodejs` and `cabal` are available and to fix some problems with `fsnotify` (IIRC) that forgets to add some frameworks to the propagated dependencies. I'm a newbie with Nix, so I don't have an answer for the best practices.

libs/Makefile
jacereda: It doesn't try to maintain compatibility with BSD make because Setup.hs already assumes GNU make (see `mymake` definition).

I can extract that to a separate PR if desired, but I'd prefer to keep it in this PR as well, since I had problems with missed dependencies until I changed the Makefiles.

test/basic010/run
jacereda: Third bullet point. I don't know yet why, it fails on Nixpkgs/Darwin when invoked via the TIMEOUT script.

jacereda:
A travis job building via nixpkgs would be the next logical step to ensure this doesn't bitrot. They already have infrastructure in place for Nix jobs.

https://docs.travis-ci.com/user/languages/nix/

Makefile
david-christiansen: Why not leave them as they were?

idris.cabal
david-christiansen: I think that their behavior should remain unchanged in the .cabal file at least. Nix users are a minority of Idris users, and I don't think that we should make it more difficult to build it the common way to make Nix packaging easier.

idris.nix
david-christiansen: @shlevy , I seem to recall you being an experienced Nixer. Is this the right way to do things? (sorry if I remembered wrong)

libs/Makefile
david-christiansen: OK. @markuspf , is this likely to cause a problem for BSD support?

test/basic010/run
david-christiansen: The point of that script is to ensure that there's not an infinite loop. Getting rid of it breaks these tests, so that shouldn't be done. Better to figure out why Nix isn't working with these.

david-christiansen:
Good to hear that Travis supports Nix.

It looks to me like there's already an Idris package in nixpkgs: https://github.com/idris-lang/Idris-dev/wiki/Idris-on-NixOS Why can't that be used here?

david-christiansen:
(and thanks for this, by the way - I'm rushing through comments and have just realized that I'm likely coming across as a bit gruff)

idris.nix
shlevy: I don't think we've quite reached a point yet where there are best practices here exactly, but my preferred approach would be:

a) Autogenerate the cabal-based Nix expression on-the-fly with `callCabal2nix`
b) Add any necessary modifications using `overrideCabal`, `overrideDerivation`, etc. after importing the base generated expression.

This won't be helpful to you now, but we're developing a set of libraries internally at Target that we're hoping to open source over the next few months that will make this workflow clean and easy and hopefully get it adopted more widely.

shlevy: I'm happy to make these changes if/when the PR is in an otherwise-acceptable state.

shlevy:
@david-christiansen Re the build expression in nixpkgs and reducing duplication:

* We can use the expression in nixpkgs here as long as the build process doesn't differ significantly for a release build vs a dev build. This isn't always true, but if it is for Idris then great!
* We can't use the expression here in nixpkgs as of now, because there are unresolved issues with `recursive nix`, where a Nix evaluation depends on the results of a build (in this case, the build being the fetch of the tarball containing Idris's custom Nix expressions). In the long run we *definitely* want to support expressions provided by upstream packages, though.

Makefile
jacereda: What do you mean? Repeat explicitly like in `env IDRIS=$(IDRIS) $(MAKE)`?

david-christiansen: I was commenting on renaming the CABAL variable to CB here.

idris.nix
david-christiansen: Thanks @shlevy! @jacereda, does the approach of importing the unmodified cabal2nix output make sense to you?

david-christiansen:
@jacereda Have you tried the expression from nixpkgs, instead of this one? Presumably it won't require all the changes to the Cabal flags and Makefiles.

david-christiansen:
@shlevy Thanks!

jacereda:
@david-christiansen I have successfully installed idris via nixpkgs, but I haven't figured out how to use the expressions in nixpkgs for interactive development and things like running a single test after making a change. Maybe @shlevy can help with that?

idris.nix
jacereda: Sure, I just need to figure out how.

Makefile
jacereda: Uses the original names now.

jacereda:
Oh, I found there's a `callCabal2nix`, I guess I should use that instead of a make rule.

idris.nix
```diff
@@ -0,0 +1,22 @@
...
+      ]);
+    testHaskellDepends = orig.testHaskellDepends ++ [ nodejs ];
+    preBuild = ''
+      export DYLD_LIBRARY_PATH=`pwd`/dist/build
+    '';
+    preCheck = ''
+      export DYLD_LIBRARY_PATH=`pwd`/dist/build
+      export TARGET=`pwd`
+    '';
+    src = if lib.inNixShell then null else orig.src;
```
shlevy: Haskell packages have an `env` attribute that is appropriate for nix-shell environments, have you tried that?

jacereda: I'm not sure I understand. Do you mean the `src = ...` part is redundant?

shlevy: I mean rather than doing `src = ...` here you can do something like `let pkg = nixpkgs.pkgs.haskell.packages.${compiler}.callPackage ./idris.nix { }; in if inNixShell then pkg.env else pkg`. See the output of `cabal2nix --shell` for an example.

jacereda: OK, I'll get rid of shell.nix and use that instead to avoid polluting the base directory.

PR #4410, opened 2018-04-04
Update travis with newer GHC versions

melted:


melted:
`tasty-rerun` stops testing on GHC 8.4. We should probably ditch that. Just more evidence that tight upper bounds on base are useless.

jfdm:
some people use tasty-rerun...

melted:
@jfdm Yeah, I commented out the GHC 8.4 jobs for now. But it's not good to not test the latest on travis, and I don't want to put `--allow-newer=base` in there.

PR #4412, opened 2018-04-08
Bump to LTS 11.3 and fingertree-0.1.4.1

joom:
The types in the fingertree module have Generic instances now. In order to create NFData instances, we need that. (We currently don't have an NFData instance for any fingertree type but my next pull request will require that)


melted:
Thanks!

PR #4413, opened 2018-04-08
getMsg leak

jacereda:
Shouldn't `getMsg` call `idris_freeMsg`?


melted:
I don't think this will work, x is a pointer into the Msg structure at that point. It needs to be copied before the message can be freed.

melted:
Thanks!

PR #4428, opened 2018-04-25
Add not LTE implies GT, and LTE implies maximum.

LeifW:
On the existing `notLTImpliesGTE` - wasn't sure if there was a way I could use it (besides cribbing from its implementation), so I just implemented `notLTEImpliesGT`, so I could apply it to the disproof returned by `isLTE`, as in:
```idris
eitherLTEorGT : (i, j : Nat) -> Either (i `LTE` j) (i `GT` j)
eitherLTEorGT i j = case isLTE i j of
                         Yes lte => Left lte
                         No notLte => Right (notLTEImpliesGT notLte)
```
As for `lteImpliesMax`, I figure knowing which number is greater should enable you to say something about the maximum, or at least they should be related, somehow. I needed it for something I was working on.

### DIFF
--- a/libs/prelude/Prelude/Nat.idr
+++ b/libs/prelude/Prelude/Nat.idr
@@ -149,9 +149,9 @@ fromLteSucc (LTESucc x) = x
 isLTE : (m, n : Nat) -> Dec (m `LTE` n)
 isLTE Z n = Yes LTEZero
 isLTE (S k) Z = No succNotLTEzero
-isLTE (S k) (S j) with (isLTE k j)
-  isLTE (S k) (S j) | (Yes prf) = Yes (LTESucc prf)
-  isLTE (S k) (S j) | (No contra) = No (contra . fromLteSucc)
+isLTE (S k) (S j) = case isLTE k j of
+  Yes prf => Yes (LTESucc prf)
+  No contra => No (contra . fromLteSucc)
 
 ||| `LTE` is reflexive.
 lteRefl : LTE n n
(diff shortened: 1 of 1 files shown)


libs/prelude/Prelude/Nat.idr
```diff
@@ -149,9 +149,9 @@ fromLteSucc (LTESucc x) = x
 isLTE : (m, n : Nat) -> Dec (m `LTE` n)
 isLTE Z n = Yes LTEZero
 isLTE (S k) Z = No succNotLTEzero
-isLTE (S k) (S j) with (isLTE k j)
-  isLTE (S k) (S j) | (Yes prf) = Yes (LTESucc prf)
-  isLTE (S k) (S j) | (No contra) = No (contra . fromLteSucc)
+isLTE (S k) (S j) = case isLTE k j of
```
ahmadsalim: Is there any reason for the change here? 😄

libs/prelude/Prelude/Nat.idr
```diff
@@ -840,3 +846,9 @@ sucMinL (S l) = cong (sucMinL l)
 total sucMinR : (l : Nat) -> minimum l (S l) = l
 sucMinR Z = Refl
 sucMinR (S l) = cong (sucMinR l)
+
+||| If one number is biggre than another, you know what the maximum is.
```
ahmadsalim: There is a small typo `biggre` -> `bigger`

libs/prelude/Prelude/Nat.idr
```diff
@@ -840,3 +846,9 @@ sucMinL (S l) = cong (sucMinL l)
...
+total lteImpliesMax : i `LTE` j -> maximum i j = j
```
ahmadsalim: Maybe this should be added to `contrib` instead of prelude?

PR #4429, opened 2018-04-26
Unboxed Bits8/Bits16 using Int as the underlying representation

jacereda:


melted:
On first glance it looks good. I'll do a proper review soon.

rts/idris_gc.c
```diff
@@ -4,17 +4,27 @@
 #include "idris_bitstring.h"
 #include <assert.h>
 
+static inline VAL plain(VM* vm, VAL x, size_t sz) {
```
melted: call this function `copy_plain`

rts/idris_rts.c
```diff
@@ -978,7 +965,7 @@ Msg* idris_checkInitMessages(VM* vm) {
     Msg* msg;
 
     for (msg = vm->inbox; msg < vm->inbox_end && msg->msg != NULL; ++msg) {
-	if ((msg->channel_id & 1) == 1) { // init bit set
+        if (msg->channel_id && 1 == 1) { // init bit set
```
melted: this change is wrong, it should be bitwise and.

rts/idris_rts.c
```diff
@@ -417,11 +409,10 @@ void dumpStack(VM* vm) {
 void dumpVal(VAL v) {
     if (v==NULL) return;
     int i;
-    if (ISINT(v)) {
-        printf("%d ", (int)(GETINT(v)));
-        return;
-    }
     switch(GETTY(v)) {
+    case CT_INT:
+        printf("%d ", (int)(GETINT(v)));
```
melted: This was wrong before, but this always prints an int as a 32-bit number, might as well fix that.

rts/idris_rts.c
jacereda: Oops, this PR is based on another branch of mine that didn't contain some of my other fixes, sorry, let me fix the mess.

melted:
Looks good! Thanks!

PR #4434, opened 2018-05-01
Strip unreachable code/data

jacereda:
This should yield smaller executables. I added also `-pipe`, it could help with compilation times.


melted:
Thanks! Rerunning a travis job.

PR #4443, opened 2018-05-16
Fixes a few typos in the tutorial on views

mdimjasevic:
This pull request replaces "ride" with "right", as in the right side of the `=` sign in the tutorial on views.


jfdm:
@mdimjasevic Thanks for the contributions. Our own proof reading cannot always be ride, and we ultimately will make mistakes.

Thanks for spotting, and fixing, the errors!

mdimjasevic:
:)

All ride!

PR #4448, opened 2018-05-21
Finds the difference between two Nats

zenntenn:
Total function that finds the difference between two Nats, order independent.	

new file:   libs/contrib/Data/Nat/Difference.idr


david-christiansen:
Thanks!

This seems like it will be a very small module, when all is said and done. Why not put it in a more general module?

zenntenn:
I couldn't find any appropriate modules in Contrib. Do you have a
suggestion of which one?

david-christiansen:
What about calling it Data.Nat?

zenntenn:
@david-christiansen K, I think I added the module correctly to the pull request, but I'm still pretty new to Github so I might have done something wrong.

david-christiansen:
One more step: it needs to be added to the package file:
https://github.com/idris-lang/Idris-dev/blob/master/libs/contrib/contrib.ipkg#L46-L48

thanks!

david-christiansen:
Thanks!

PR #4456, opened 2018-05-29
Correct link to "Design and Implementation" paper

steshaw:
Previously using markdown style hyperlinking. Switched to restructured text.


melted:
Thanks!

PR #4460, opened 2018-05-30
A paper by David R. Christiansen has been moved.

XeFias:
The paper by David Raymond Christiansen referenced in this document is no longer available at the url http://www.itu.dk/people/drc/pubs/dependent-type-providers.pdf but has been moved to url http://www.davidchristiansen.dk/pubs/dependent-type-providers.pdf.


jfdm:
Thanks

PR #4474, opened 2018-06-09
Don't cache after a failed build

melted:


melted:
Problem with this is that we could time out and manage to set up the cache before. Now we can't. So it's probably not worth it. It would be nice if we could detect time outs.

PR #4475, opened 2018-06-09
Use a newer cabal with 7.10

melted:
The 7.10 build fails when the cabal executable crashes in the configure step.

It works with a new cabal, so let's use that. Any 7.10 user can install a newer
cabal-install if it would turn out to be more than a Travis environment issue.

Update comment for why 8.4 is disabled with new roadblock (sigh).


melted:
One of the other travis jobs is timing out, that's what I get for removing the cache. But 7.10 works, which was the intention.

Issue #4480, opened 2018-06-18
toBinary from the book exercises behaves weird


skibaa:
A trivial example from the "TDD with Idris" produces strange results. Happened in Idris  Version 1.3.0.
```
toBinary: Nat -> String
toBinary k with (halfRec k)
  toBinary Z | HalfRecZ = ""
  toBinary (n + n) | (HalfRecEven rec) = toBinary n | rec ++ "0"
  toBinary (S (n + n)) | (HalfRecOdd rec) = toBinary n | rec ++ "1"

```

# Steps to Reproduce
evaluate toBinary 2 in REPL

# Expected Behavior
outputs
"10" : String
# Observed Behavior

```
*../TypeDD-Samples/Chapter10/Exercises/ex_10_2> toBinary 2
prim__concat (with block in Main.toBinary 1
                                          (with block in with block in Data.Nat.Views.halfRec 0
                                                                                              (half 0)
                                                                                              (Access (\z, zLTy =>
                                                                                                         Prelude.WellFounded.sizeAccessible, acc Nat
                                                                                                                                                 2
                                                                                                                                                 (constructor of Prelude.WellFounded.Sized (\meth =>
                                                                                                                                                                                              meth))
                                                                                                                                                 1
                                                                                                                                                 z
                                                                                                                                                 (lteTransitive zLTy
                                                                                                                                                                (LTESucc LTEZero))))))
             "0" : String
```

clayrat:
See http://docs.idris-lang.org/en/latest/reference/repl.html#evaluation for explanation of this.

To get what you expect, write `:exec toBinary 2`

skibaa:
*try1> :total toBinary
Main.toBinary is Total

I don't really understand. toBinary is a total function, returning String. Isn't it supposed to return String on evaluation? 

Besides that, this is the code which is supposed to be the solution of an exercise from the textbook TypeDD with Idris - and the exercise says

```
If you have a correct implementation, you should see this:
*ex_10_2> toBinary 42
"101010" : String
*ex_10_2> toBinary 94
"1011110" : String
```

No mention of `:exec toBinary`

BTW `toBinary 0` and `toBinary 1` work as expected.

skibaa:
Done, thank you.

[https://github.com/idris-lang/Idris-dev/pull/4481](https://github.com/idris-lang/Idris-dev/pull/4481)

corazza:
this fix is not in 1.3 right?

clayrat:
[...]

PR #4481, opened 2018-06-18
Make covering function public export

skibaa:
Without public export it's not evaluated well in REPL, see issue #4480


LeifW:
The CI failure is just from the build timing out on GHC 8.0

melted:
Jenkins failure is unrelated, let’s merge. 

Thanks for the PR!

PR #4488, opened 2018-06-22
Updated to latest LTS Haskell snapshot.

jfdm:


jfdm:
The commit introduces a change not covered by the CI tests we have. This worked locally on my machine, and given the nature of Stack it should work locally for others to. I shall merge.

PR #4496, opened 2018-06-27
Flushing progress reports after every newline

Infinisil:
This makes progress reporting work even on non-tty outputs, e.g. when piping to a file.

My main motivation is to have this for [nixpkgs](https://github.com/NixOS/nixpkgs), which currently doesn't have any indication of progress throughout a package build, see https://github.com/NixOS/nixpkgs/pull/42647 for a hacky fix.


melted:
Ok, thanks for working on it!

Infinisil:
@melted Proper fix: #4498

PR #4498, opened 2018-06-27
Set stdout line buffering before building inputs

Infinisil:
This makes progress reporting work even on non-tty outputs, e.g. when piping to a file. For real this time. The problem was that line buffering was turned on only after the packages have been built.

The problematic lines were these: https://github.com/idris-lang/Idris-dev/blob/a675cdb9a712c05486a17d6139e5b135b5924873/src/Idris/Main.hs#L179-L184

Line 179 there apparently builds packages, but only later down at line 184 line buffering gets turned on.

My main motivation is to have this for nixpkgs, which currently doesn't have any indication of progress throughout a package build, see NixOS/nixpkgs#42647 for a hacky fix.


Infinisil:
I tested this by running `idris --build Bi.ipkg > foo` (from [idris-bi](https://github.com/sbp/idris-bi)) and `tail -f foo` in another terminal. Before this commit, you'd only get any output after the package was built completely. With this commit you get continuous output.

Infinisil:
Relevant commit: f59ae217eb9a3afb7b22ae034520e4f8cdfe55c5 which set up line buffering initially

PR #4499, opened 2018-06-29
fix the type of end grammar in someTill'

clayrat:
The type of `someTill'` currently assumes that the `end` marker has the same type as the `List` being collected. This doesn't align with the types of `someTill` and `manyTill` and seems like a typo.


msmorgan:
Typo confirmed.

Issue #4391, opened 2018-03-22
modifyIORef causes segmentation fault in FFI callback


thalerjonathan:
# Steps to Reproduce
```Idris
import Data.IORef

%include C "idris_clbks_test.h"
%link C "idris_clbks_test.o"

TestCallback : Type
TestCallback = Int -> ()

testCallbackPure : IORef Integer -> Int -> ()
testCallbackPure ref x = unsafePerformIO $ do 
  putStrLn "before modifyIORef"
  modifyIORef ref (+1) -- segmentation fault occurs here
  putStrLn "after modifyIORef"

testCallbackPurePtr : IORef Integer -> IO Ptr
testCallbackPurePtr ref
  = foreign FFI_C "%wrapper" (CFnPtr TestCallback -> IO Ptr) (MkCFnPtr $ testCallbackPure ref)

setTestCallbackWithPtr : IORef Integer -> IO ()
setTestCallbackWithPtr ref = do
  ptr <- testCallbackPurePtr ref
  foreign FFI_C "registerCallback" (Ptr -> IO ()) ptr

runCallback : IO ()
runCallback = foreign FFI_C "runCallback" (IO ())

export
run : IO ()
run = do
  ref <- newIORef 42
  setTestCallbackWithPtr ref
  runCallback
  ret <- readIORef ref
  putStrLn $ "IORef = " ++ show ret
```
Note that is is not possible to reproduce this problem with just this code, one needs the c files as well and compile the whole thing into an executable. For this purpose I created a repository which is available here https://github.com/thalerjonathan/idris-callbacks-tests. The relevant file for this problem is https://github.com/thalerjonathan/idris-callbacks-tests/blob/master/TestModifyIORef.idr
To compile this example run 'make' in the cloned folder. Make sure that in the Main.idr the TestModifyIORef.run function is called.

# Expected Behavior
Program should not crash.

# Observed Behavior
Program crashes with segmentation fault when modifyIORef is called.

thalerjonathan:
Is also caused by readIORef and writeIORef alone.

thalerjonathan:
The crash occurs in idris_readRef (). I uploaded the crashing binary https://github.com/thalerjonathan/idris-callbacks-tests/blob/master/idrisclbktest and the coredump https://github.com/thalerjonathan/idris-callbacks-tests/blob/master/core.6975 for further inspection.

ivanperez-keera:
[...]

PR #4500, opened 2018-06-30
Error when using closure as callback in C backend

melted:
Document it.


melted:
Related to #4391. Fixes the run-time error by making it compile-time.

PR #4505, opened 2018-07-05
Add a slew of proofs (mostly Isos) for (mostly) Vect

howtonotwin:
Also makes `Iso` a record (so you can actually run one without ceremoniously unwrapping it) and gives it a `Monoid` (for endoisomorphisms). Together with `stimes` and `mtimes`, this means you get to say such fun things as ``to (permuted (2 `stimes` rotatedDown))``.

I wasn't entirely sure where to put the many of the proofs, as there appears to be a precedent of sending proofs of `base` functions into `base` without going into `contrib`, so I will happily shuffle them around.


melted:
Thanks for the PR! Looks like useful stuff.

I'll review it later today.

howtonotwin:
Can I change `Fin.strengthen` to return a `Maybe`? It really bothers me that its type doesn't accurately describe what it does. So far, I think I've managed to avoid making breaking changes to existing type signatures, though I have broken proofs by changing implementations. Would such a change would be acceptable?

melted:
Looks good. Thanks!

PR #4509, opened 2018-07-07
fix param name in comment.

MarisaKirisame:
I think the render function has an error.
It seems like the lvl argument never get changed, and will always stay at w.
Judging by the documentation, it seems like
    best lvl col curr (Nest x y) k = best lvl col (curr+x) y k
should be
    best lvl col curr (Nest x y) k = best (lvl-x) col (curr+x) y k.
maybe it should also start at 0 and count up instead, to match the name 'lvl'.


melted:
Thanks for the PR!

PR #4513, opened 2018-07-12
Bump the network upper bounds.

LeifW:
This is already being done to build the Idris package on Arch Linux:
https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/idris#n30


melted:
Travis fails because the new network has depreciated stuff used in Util.Net (we have -Werror). Could you update that with the preferred API?

melted:
I'm working on a PR to drop the deprecated APIs.

melted:
Replaced by #4540

PR #4515, opened 2018-07-14
More interface implementations for Data.Morphisms

howtonotwin:
* `data` types + accessor functions turned into `record`s
* `Monad` constraint on `Kleislimorphism` nixed
   * Side note: is it possible to remove the duplication between `Kleislimorphism` and `ReaderT`?
* "Reader-style" `Semigroup` and `Monoid` for (`Kleisli`)`Morphism`
* `Functor`/`Applicative`/`Monad` for `Kleislimorphism`


melted:
Thanks for the PR!

PR #4526, opened 2018-07-25
add `get current directory name`

doofin:


rts/idris_stdfgn.c
```diff
@@ -259,3 +259,11 @@ VAL idris_getString(VM* vm, void* buffer) {
     free(sb);
     return str;
 }
+
+char* idris_currentDir() {
+   char cwd[1024];
+   if (getcwd(cwd, sizeof(cwd)) != NULL)
+     return cwd;
```
melted: Returning a pointer to a local buffer is wrong. It will kind of work until something else is called and that part of the callstack overwritten. 

Call MKSTR to create an idris string and return it as a VAL. Then set the return type in the FFI declaration on the IO side to IO (Raw String) and unpack it by matching on MkRaw.

melted:
Thanks for the PR! To be able to merge this the C side needs to be fixed to handle allocation right. The best way is to create an idris value so that the garbage collector will handle the deallocation.

rts/idris_stdfgn.c
```diff
@@ -259,3 +259,11 @@ VAL idris_getString(VM* vm, void* buffer) {
     free(sb);
     return str;
 }
+
+VAL idris_currentDir() {
+   char cwd[1024];
+   if (getcwd(cwd, sizeof(cwd)) != NULL)
+     return MKSTRlen(get_vm(),cwd,1024);
```
melted: Use MKSTR, it will use strlen to allocate the right size string

rts/idris_stdfgn.c
```diff
@@ -259,3 +259,11 @@ VAL idris_getString(VM* vm, void* buffer) {
...
+   else
+     return MKSTR(get_vm(),"getcwd() error");
```
melted: Use `MKSTR(get_vm(), "").

doofin: OK

PR #4527, opened 2018-07-30
Add perl, nodejs to stack-shell.nix

zgrannan:
Currently, `stack --nix test` fails because the tests depend on `nodejs` and `perl`.

This pull-request adds them to the build dependencies for the stack shell.


melted:
Thanks!

Issue #4525, opened 2018-07-22
assert_smaller doesn't work under "with" statements.


Raleigh-Foster:
In the following code, the commented out code fails totality, where the uncommented code passes.
For some reason if you use "with", assert_smaller doesn't know which case it is in.
This is unfortunate, since there are other cases, where "with" results in the typechecker knowing more.

` {-
 total
 foo : Nat -> (Nat, (Nat,Nat)) -> Nat
 foo x (y,z) with (z)
   | (_, Z) = 0
   | (a, S k) = foo x (assert_smaller (y, z) (y, (a, k)))
   -}
 
 total
 foo : Nat -> (Nat, (Nat,Nat)) -> Nat
 foo x (y,z) = case z of
   (_, Z) => 0
   (a, S k) => foo x (assert_smaller (y, z) (y, (a, k)))
`

See also this issue, which may be related:

https://github.com/idris-lang/Idris-dev/issues/2872

jfdm:
Hi @Raleigh-Foster I am not sure if this would be helpful for your needs, but Idris support's syntax to pass proof that something is getting smaller. I think the thing has to provide an instance os SizeAccessible. I don't know how well documented the feature is, but you can find an example in Text.Parser [here](https://github.com/idris-lang/Idris-dev/blob/master/libs/contrib/Text/Parser/Core.idr#L224) and [here](https://github.com/idris-lang/Idris-dev/blob/master/libs/contrib/Text/Parser/Core.idr#L217) and [here](https://github.com/idris-lang/Idris-dev/blob/master/libs/contrib/Text/Parser/Core.idr#L238) in `contrib`

Hope this helps.

Raleigh-Foster:
Thanks @jfdm It looks like this isn't a bug at all.
I have some usability concerns though: when I change my code to use "with (sizeAccessible z)", I get this error:
</code>
  |
4 | foo x (y,z) with (sizeAccessible z)
  |             ~~~~~~~~~~~~~~~~~~~~~~~
When checking with value in foo:
Can't find implementation for Sized (Nat, Nat)
</code>

Of course, I can implement sized for the type I am working with, but it adds more work than when using a case statement.

Perhaps I should accept this, but if this is the desired design, should sizeAccessible be mentioned here?
http://docs.idris-lang.org/en/latest/reference/misc.html

I read about assert_smaller there, but sizeAccessible appears to only appear in the 1.1 change log:
https://www.idris-lang.org/idris-1-1-0-released/

PR #4528, opened 2018-07-30
Add implementation of Sized for Pair type

zgrannan:
As mentioned in https://github.com/idris-lang/Idris-dev/issues/4525, there is no implementation of `Sized` for the `Pair` type.

This PR adds a generic implementation that I think is sensible.


melted:
Looks reasonable.

PR #4539, opened 2018-08-19
Raise an error if too many "with" arguments are provided

### LINKED ISSUE #4238: Idris accepts too many "with" arguments
The check in https://github.com/idris-lang/Idris-dev/blob/4064dcf1c871bfb7efa926efdc20b9af7b36d74c/src/Idris/Elab/Clause.hs#L664-L666 is naive, it only checks that we're either outside a with or have with arguments, not that we have the correct number of with arguments.

# Steps to Reproduce

```idris
foo : Int -> Bool
foo n with (succ n)
  foo _ | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 = True
```

# Expected Behavior

Idris errors at the first additional with argument.

# Observed Behavior

No errors.

zgrannan:
Fixes #4238 

It appears that during elaboration, any `withs` in a  `PClause` are extraneous, so we can throw an error if they are present. I'm not 100% confident about that however, so please let me know if I'm missing any case.

### DIFF
--- a/src/Idris/Elab/Clause.hs
+++ b/src/Idris/Elab/Clause.hs
@@ -660,11 +660,13 @@ elabClause info opts (cnum, PClause fc fname lhs_in_as withs rhs_in_as wherebloc
         i <- getIState
         inf <- isTyInferred fname
 
-        -- Check if we have "with" patterns outside of "with" block
-        when (isOutsideWith lhs_in && (not $ null withs)) $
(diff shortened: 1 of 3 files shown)


melted:
Hmmm, how have this been here for so long?

I'll check it out tonight

melted:
Looks good! Thanks!

PR #4547, opened 2018-09-01
Strengthen `elemSmallerThanBound`

JulianKniephoff:
I found myself proving something along the lines of `Data.Fin.Extra.elemSmallerThanBound`, before stumbling over this implementation. However, my version was a tiny bit stronger, and I think this one should be, too. :wink:

For a `n : Fin m`, we have `n < m`, whereas `elemSmallerThanBound` only proves `n <= m`, even though the name and the docstring suggest otherwise.

If we want the old behavior back, something like

```idris
elemLteBound : (n : Fin m) -> LTE (finToNat n) m
elemLteBound = lteSuccLeft . elemSmallerThanBound
```

should suffice. However, I don't think that it's worth including that in the library, but that's not a strong opinion.

### DIFF
--- a/libs/contrib/Data/Fin/Extra.idr
+++ b/libs/contrib/Data/Fin/Extra.idr
@@ -4,6 +4,6 @@ import Data.Fin
 
 ||| Proof that an element **n** of Fin **m** , when converted to Nat is smaller than the bound **m**.
 public export
-elemSmallerThanBound : (n : Fin m) -> LTE (finToNat n) m
-elemSmallerThanBound FZ = LTEZero
+elemSmallerThanBound : (n : Fin m) -> LT (finToNat n) m
+elemSmallerThanBound FZ = LTESucc LTEZero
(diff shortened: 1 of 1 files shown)


jfdm:
@JulianKniephoff Thanks for the PR, one thing that would be great would be to include the contents of the github comment in the git commit itself. Having the extra reasoning there as well would be good for completeness. Otherwise, what 'strengthen' means might be lost in interpretation.

JulianKniephoff:
@jfdm thanks for the quick reply. I reworded the commit and basically added the second paragraph of my comment. I think the rest of the message was mostly fluff and would not add any value to the commit. I hope, this is okay for you. :smile:

JulianKniephoff:
Is there anything else I can do here? I don't think the Travis failure is my fault. :thinking:

melted:
It's fine, thanks for the PR!

PR #4557, opened 2018-09-12
[ docs ] definitions in the same order as in the source file

gallais:
I got annoyed with the fact that docs were listing definitions in
reverse order with respect to the source file. This mean that they
were presenting traversals before the type they are working on has
been introduced. This fixes that problem.

Also, we don't need to go *via* lists to get rid of the `Nothing`s.

### DIFF
--- a/src/Idris/IdrisDoc.hs
+++ b/src/Idris/IdrisDoc.hs
@@ -427,13 +427,12 @@ createNsDoc :: IState   -- ^ Needed to determine the types of names
                         --   documentation will be written.
             -> IO ()
 createNsDoc ist ns content out =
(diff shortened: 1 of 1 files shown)


melted:
Thanks for the PR!  Looks like travis trouble and not something with the commit. Let’s restart.

gallais:
The GHC 8.0.1 has been consistently failing on master for a few commits now.
I doubt it'll magically succeed on this PR.

melted:
Yeah, it was 7.10 failing that bothered me.

gallais:
Ah ok, I didn't see that one. My bad.

melted:
OK, 7.10 succeeded

PR #4563, opened 2018-09-16
Disable travis job that always times out

melted:


melted:
Looks correct on Travis, no need to wait for the result of the run.

PR #4564, opened 2018-09-17
[ travis ] tests for 8.4.3

gallais:
There has been a report of a build failure for 8.4.3 on the mailing list so I figured
it was worth adding a new test case for it. I have guessed the stack number so it may
need to be changed.
Also threw in a tag for the 8.4.1 one which was showing up under `C` so far on the
travis logs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,11 @@ env:
 
 matrix:
   include:
+    - env: CABALVER="2.4" GHCVER="8.4.3" STACKVER="12.9" STYLISH=YES
+      compiler: " #GHC 8.4.3"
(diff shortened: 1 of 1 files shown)


melted:
There are already commented out 8.4 tests further down in the travis config. Looks like these because of the same libffi thing.

melted:
It would be very good to get a functioning job up for GHC 8.4, so please try.

There should still be a STYLISH=YES job. That runs stylish-haskell.

tuncer:
Related: #4604

gallais:
I'm not planning to investigate this one any time soon so if #4604 covers the same
topic, feel free to close this one.

tuncer:
#4604 only updates stack.yaml to select a 8.4.4-based resolver, no CI changes at all. Since #4604 modifies a different file than this pr, the pull requests can be merged separately.

PR #4565, opened 2018-09-18
add `VerifiedMonad (Either a)`, `VerifiedMonad List`

stop-cran:
Added implementations of VerifiedFunctor, VerifiedApplicative and VerifiedMonad for (Either a) and List. 
Added VerifiedAlternative with implementations for Maybe and List.
Also added some auxiliary functions for reasoning about lists.


melted:
Thanks for the PR!

PR #4571, opened 2018-10-12
Proven some basic statements about Data.Fin.

stop-cran:
I've started with an idea of having a view of Nat, that represents a number as a quotient and a remainder in division by some fixed non-zero number. This, as well as some basic facts about Fin proven here, can be used for a variety of purposes, like defining and reasoning about prime numbers, finite rings and fields.


melted:
Looks good, thanks!

Issue #4570, opened 2018-10-11
Math library feature request.


thyrgle:
Hello,

This isn't exactly a feature request for the language itself, but a lot of the computer verification languages have a library with the ultimate goal of providing a rich set of theorems and definitions that would allow a mathematician to perform daily mathematics. From my understanding, a library like this would be simply too large for `base` or `prelude`. For instance, [`mathlib`](https://github.com/leanprover/mathlib) is an example of a such a library for the lean verification language. Is anyone aware if anyone is working on something like this? If not, is anybody interested working on such a library?

Thanks.

PR #4572, opened 2018-10-12
Updated contributing based on comment in #4570.

thyrgle:
Based off of @melted comment to my feature request.


ahmadsalim:
@thyrgle I think the current formulation is great. Basically, we like to get feature requests, but only if you are willing to implement them yourself as soon as possible. If your feature request is something you plan to implement, we would still welcome it 😄 . Please use IRC or the mailing list for extended discussion.

PR #4575, opened 2018-10-18
Added Singleton interface.

stop-cran:
Suggested a standard proof, that some type has at most one value. It can be useful for some equality proofs, e.g. https://github.com/stop-cran/Idris-Fraction/blob/master/Fraction.idr


Melvar:
Two things.
1. Does this have a particular reason for being in the prelude rather than being added to `contrib` or `base`?
2. I find the name odd. To me, “singleton” usually means “a container containing exactly one element”. In this context, it could equally well mean that a type has *exactly* one inhabitant (contractible, true proposition), rather than *at most* one (collapsible, mere proposition).

stop-cran:
1. Will move to `contrib`.
2. How about this naming?

PR #4576, opened 2018-10-19
Added Proposition interface.

stop-cran:
Suggested a standard proof, that some type has at most one value. It can be useful for some equality proofs, e.g. in [this example](https://github.com/stop-cran/Idris-Fraction/blob/master/Fraction.idr#L105).
As per [previous discussion](https://github.com/idris-lang/Idris-dev/pull/4575) moved the definition to `contrib` and renamed the interface:

```
interface Proposition t where
  atMostOneValue : (x : t) -> (y : t) -> x = y
```

### DIFF
--- a/libs/contrib/Interfaces/Proposition.idr
+++ b/libs/contrib/Interfaces/Proposition.idr
@@ -1,0 +1,32 @@
+module Interfaces.Proposition
+
+%default total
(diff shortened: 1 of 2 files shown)


melted:
Thanks for the PR

PR #4580, opened 2018-10-29
Export interface implementations of Network.Cgi

pykello:
Without exporting these, the users get errors like:
```
  Can't find implementation for Monad CGI.
```
when compiling a program like:
```
import Network.Cgi

prog: CGI ()
prog = do
  vars <- queryVars
  output "Hello World!\n"
```


melted:
Thanks for the PR! I’ll restart the travis job.

PR #4584, opened 2018-11-06
Add regression test for #2349

### LINKED ISSUE #2349: Associate types problem
Now that we've decided we should support associated types, I think the following should work:

    class ListLike a where
        Elem : Type
        fromList : List Elem -> a
  
    data Silly = None | Cons Int Silly
  
    instance ListLike Silly where
          Elem = Int
          fromList [] = None
          fromList (x :: xs) = Cons x (fromList xs)

But, alas, I get an error:

        Type mismatch between
                List Main.Silly instance of Main.ListLike, method Elem (Type of xs)
        and
                List Elem (Expected type)
        
        Specifically:
                Type mismatch between
                        Int
                and
                        Elem

I suspect this is a type class resolution problem in the body of the instance - it can be worked around by defining a helper function of type `List Int -> Silly`.

RyanGlScott:
This program appears to work now with contemporary Idris, so let's add a regression test for it to ensure that it remains fixed.

Fixes #2349.


RyanGlScott:
It appears that the Travis build fails due to stylization problems (which #4581 would fix). Should I rebase this patch on top of that commit, or just wait for the issue to be fixed upstream first?

melted:
I merged that so let’s restart travis

melted:
Thanks!

PR #4607, opened 2018-12-09
Allow newer version of aeson and containers.

LeifW:
It builds fine against these, and these are already the versions it's distributed with as a package on Arch Linux: https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/idris#n29


melted:
Thanks!

PR #4610, opened 2018-12-11
Megaparsec 7

jmorag:
Same as #4609, but from a different branch of my fork and with ./stylize.sh run.

### DIFF
--- a/idris.cabal
+++ b/idris.cabal
@@ -289,10 +289,11 @@ Library
                 , fingertree >= 0.1.4.1 && < 0.2
                 , haskeline >= 0.7 && < 0.8
                 , ieee754 >= 0.7 && < 0.9
-                , megaparsec >= 6.2 && < 7
+                , megaparsec >= 7.0.4 && < 8
                 , mtl >= 2.1 && < 2.3
                 , network < 2.8
                 , optparse-applicative >= 0.13 && < 0.15
+                , parser-combinators >= 1.0.0
                 , pretty < 1.2
                 , process < 1.7
                 , regex-tdfa >= 1.2
--- a/src/Idris/Package/Parser.hs
+++ b/src/Idris/Package/Parser.hs
@@ -26,14 +26,13 @@ import System.Exit
 import System.FilePath (isValid, takeExtension, takeFileName)
 import Text.Megaparsec ((<?>))
 import qualified Text.Megaparsec as P
-import qualified Text.Megaparsec.Char as P
 import qualified Text.PrettyPrint.ANSI.Leijen as PP
 
 type PParser = Parser PkgDesc
 
 parseDesc :: FilePath -> IO PkgDesc
 parseDesc fp = do
-    when (not $ takeExtension fp == ".ipkg") $ do
+    unless (takeExtension fp == ".ipkg") $ do
         putStrLn $ unwords ["The presented iPKG file does not have a '.ipkg' extension:", show fp]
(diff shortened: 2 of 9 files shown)


jfdm:
Thanks for the PR. 

It looks like Travis is failing due to some warnings turned into errors (tests have the flag `-Werror`.

AppyVeyor is failing because the output from Idris appears to have changed with an upgrade to MegaParsec 7. It would be worth noting and finding out how the error message's have changed and if this change is for the better.

My only reservation with this PR is not the migration to MegaParsec 7 itself, but rather that we are now modifying the `stack.yaml` file. Part of me believes that we should depend on LTS Haskell librarys without the need to specify specific packages. However, as our stack support is experimental the impact of this shouldn't be too high. I'll certainly clean up the `stack.yaml` once MegaParsec7 is in a supported LTS version.

Thanks

Jan

On a side note, force pushing the results of `./stylise.sh` to #4609 would have been acceptable as well as creating a new PR.

LeifW:
That megaparsec-7.0.4 is in Stackage nightly.
That's also the megaparsec version on Arch Linux. A "megaparsec6" package was made to support Idris  - the other packages are all on megaparsec 7.
We install in a lot of place outside of what's specified in that stackage version, which specifies GHC 8.2.2. E.g. the Arch package is built on GHC 8.6.3.

jmorag:
A lot of the warnings are in parts of the code that have nothing to do with the parser. I think nearly all of them are `unused-toplevel-binding`. I can silence those in the relevant files if you'd like. Deleting the bindings entirely is probably a bad idea. 

As for modifying the stack.yaml file, it seemed to be the least invasive way to bump the megaparsec version without also dragging every other dependency along to whichever nightly has 7.0.4, but I could try to do that as well, if you'd like.

I'll go through the AppyVeyor fails more carefully now, see what the differences are exactly, and report back.

jfdm:
@jmorag 

The specific error I saw was:

```
src/Idris/Parser.hs:63:1: Warning:
    The qualified import of ‘Text.Megaparsec.Char’ is redundant
      except perhaps to import instances from ‘Text.Megaparsec.Char’
    To import instances alone, use: import Text.Megaparsec.Char()
```

TBH, and IMHO, the stack modification would be a bad reason to stop this PR being merged. So if we see what is happening with the AppVeyor results and the travis builds then we can see about merging.

jmorag:
Strange, I removed the offending import and travis is still complaining.

jmorag:
How exactly do you run the tests that appveyor runs with stack?

melted:
Yeah, that it looks for node before it looks if it should test with node is backward (and we've had complaints about that before). I'll fix it.

melted:
I have merged #4613 to not require node always.

jfdm:
The point of having the tests fail if you haven't node was done on purpose. To force people who run test's locally to test both the C and Node/JS codegen targets. We had cases of people submitting PRs and not testing on both and travis discovering a discrepancy.

This is an okay alternative but obviously not as strict as what was there before.

melted:
I think it’s reasonable that it should be possible to run the tests for the C backend without having node. 

make test still tries to run the node tests. So it will fail in the same manner as before for a full test.

jmorag:
A lot of the test errors are related to printing the source name of the file twice. Is there anywhere else in the codebase that this would happen other than `Parser/Stack.hs`?

PR #4611, opened 2018-12-11
Enable GHC 8.4 in travis

melted:
Let's see if 8.4.4 fixes the libffi problem

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: trusty
+dist: xenial
 sudo: false
 
 language: c


jfdm:
It would be good to review the versions of GHC that Travis, at a minimum, should test with. 

From an quick search of two known LTS like operating systems:

+ Debian stable is on 8.0.X series.
+ Ubuntu LTS 18.01 is on 8.0.X as well.

I am not sure of any other Linux distro's (or other OS) that ship with older versions of GHC---i.e. 7.X series. I think this gives us grounds to comfortably drop all ghc versions prior to the 8.X series in our testing matrix.

melted:
I saw that Travis has Xenial now. So let's test that.

ilyakooo0:
This got resolved in #4827

PR #4617, opened 2018-12-16
Fixing error message on partial pattern matching failure.

unalos:


melted:
Thanks for the PR!

The JavaScript wouldn’t be suitable in a browser. You can make a function in the jsrts that is different for browser or node if you want to handle it that way in node.

unalos:
It should now be fixed.

melted:
Looks good!

Issue #4618, opened 2018-12-16
ipkg sourcedir can't be more than 1 level deep


shmish111:
If your source code is in `src/sub` and you add `sourcedir = "src/sub"` you will get `unexpected '"'` or with `sourcedir = src/sub` you will get `unexpected '/'`

# Steps to Reproduce

Create a module called Main.idr in the directory `issue/src`

```
module Main

main : IO ()
main = pure ()
```

Create issue.ipkg in the parent directory of `issue/src`

```
package issue

modules = Main

sourcedir = idris/issue
```

run `idris --build issue.ipkg`

# Expected Behavior

compilation completes

# Observed Behavior

idris fails to parse ipkg file

PR #4619, opened 2018-12-16
Allow ipkg sourcedir to have deep location

shmish111:
https://github.com/idris-lang/Idris-dev/issues/4618

### DIFF
--- a/src/Idris/Package/Parser.hs
+++ b/src/Idris/Package/Parser.hs
@@ -122,7 +122,7 @@ pOptions = do
 pClause :: PParser ()
 pClause = clause "executable" filename (\st v -> st { execout = Just v })
       <|> clause "main" (iName []) (\st v -> st { idris_main = Just v })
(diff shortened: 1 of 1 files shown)


melted:
Hmmm.  Can we just read to the newline, trim whitespace and parse that as a path?

Thanks for the PR anyway!

shmish111:
Good point @melted so I've implemented the easy option in this new commit. I find it quite normal to allow simple things unquoted and complicated things quoted.

shmish111:
What tends to be the release schedule for Idris? Not being pushy here, just want to get some idea of how long bugs like this tend to get fixed. Also @melted is my PR ok now that I have enabled quoted and unquoted parsing?

melted:
@shmish111 Work on Idris has slowed down (with all the effort Edwin is putting into Blodwen, Idris in Idris), but there have been several releases a year even so. But the exact timing depends on need, a new release of GHC usually provides motivation with breaking code.

shmish111:
That's awesome thanks @melted

PR #4626, opened 2018-12-22
Documentation for partial, covering and total.

unalos:


melted:
Looks good! Thanks!

jfdm:
We should really have a copy of this text in the tutorial somewhere. For instance, when describing functions or later on in one of the later sections where partial and total are used.

PR #4632, opened 2018-12-24
Update ffi.rst to show example of correct including C code and exampl…

bigos:
…e Makefile

That would save a lot of time since simple examples are hard to find.

### DIFF
--- a/docs/reference/ffi.rst
+++ b/docs/reference/ffi.rst
@@ -115,6 +115,24 @@ For C, this is:
     FFI_C : FFI
         FFI_C = MkFFI C_Types String String
 
+Linking foreign code
+====================
+
+This is the example of linking C code. 
+
+.. code-block:: idris
+    %include C "mylib.h"
+    %link C "mylib.o"
+
+Example Makefile
+
+.. code-block:: shell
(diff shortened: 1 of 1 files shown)


unalos:
Ideally, I'd rather have documentation on how to use idris --build rather than a Makefile. As you do not really need a Makefile for this.

unalos:
I also took time to figure it out. This part of the documentation needs improvement IMHO.

The problem I see is that using a Makefile in this way ultimately is misleading, as it encourages not to use the .ipkg files which provide a standard way to bundle the code in packages and install them locally with idris --install. For instance https://github.com/unalos/idris-mongo/blob/master/Makefile

(I won't be the one deciding to accept or reject your PR. I just wanted to provide my take on it.)

bigos:
I did not read about .ipkg files yet. I think creating a simple minimalistic project to illustrate the point and linking to it from the tutorial would be a better idea.

melted:
It would be valuable to have an ipkg example too, and how to it just from the command line too. Still, some documentation is better than no documentation. We can expand it in further PRs.

Thanks!

Issue #4635, opened 2019-01-07
Make Identity an instance of Semigroup/Monoid (for Semigroup/Monoid types)?


alaendle:
I wonder if the following lines should be attached to Identity.idr:

```
(Semigroup a) => Semigroup (Identity a) where
  (<+>) x y = Id (runIdentity x <+> runIdentity y)

(Monoid a) => Monoid (Identity a) where
  neutral = Id (neutral)
```

Unfortunately I have no profound knowledge of the underlying type theory; but if it makes sense I willingly provide a pull request. As far as I understand the implementation is equivalent to the haskell one (generated by haskell's deriving mechanism).

The downside of the implementation are for sure the type constraints - so I'm unsure if this is generic enough to justify a presence in the base library. Nonetheless I find it useful enough (maybe just because it's present in haskell) to create this issue :smiley:

david-christiansen:
This seems reasonable enough to me - I see the type Identity as mostly just a hack to let type class resolution work better than it would for `\x => x`, so we might as well make it invisible in more contexts.

alaendle:
Resolved long time ago (and still present in Idris2 😄) -> closed!

PR #4636, opened 2019-01-08
Make Identity an instance of Semigroup/Monoid

alaendle:
Make Identity an instance of Semigroup/Monoid (for Semigroup/Monoid datatypes).

#4635


melted:
Thanks for the PR!

PR #4648, opened 2019-01-25
Fix badly formatted DOI link in doc

beardhatcode:
Fix a an ill formated DOI link and and use [CrossRef's DOI display guideline](https://www.crossref.org/display-guidelines/) (= use HTTPS, don't use `dx`, don't use DOI prefix)

This applies to http://docs.idris-lang.org/en/latest/tutorial/views.html


melted:
Thanks for the PR!

PR #4649, opened 2019-01-26
Makingtotal

zenntenn:
I'm focusing on adding %default total where possible in Blodwen, including updating the code as needed to make it total with the same functionality.  There's going to be this branch and the corresponding branch on Blodwen as I work.  

I don't know if each chunk of work should be merged separately to master, or if I should just keep working on the branches and merge in everything when I'm done.  If the second option I'll just keep adding to the PR.


zenntenn:
I'll be really impressed if I managed to accidentally trigger that GHC bug in appveyor.

melted:
Thanks for the PR!

PR #4652, opened 2019-01-30
Fix sample code

### LINKED ISSUE #4651: Well Typed Interpreter example does not typecheck
# Steps to Reproduce

Following the steps at
http://docs.idris-lang.org/en/latest/tutorial/interp.html
the example does not typecheck.

The first error seems to be obvious. In:

    interpTy : Ty -> Type
    interpTy TyInt       = Integer
    interpTy TyBool      = Bool
    interpTy (TyFun A T) = interpTy A -> interpTy T

`A` and `T` are capital pattern match values which are not allowed anymore.
New problems arise at:

    lookup : HasType i G t -> Env G -> interpTy t
    lookup Stop    (x :: xs) = x
    lookup (Pop k) (x :: xs) = lookup k xs

which raises the error 

    When checking right hand side of Lambda.Term.lookup with expected type
            interpTy t
    
    Type mismatch between
            interpTy a (Type of x)
    and
            interpTy t (Expected type)

for the line `lookup Stop    (x :: xs) = x`
Please also note that idris is not able to case split on `x` at 

     lookup : HasType i G t -> Env G -> interpTy t
     lookup x y = ?hole

This has been confirmed with Atom and emacs idris mode.

I created a gist with the source as far as I got it at https://gist.github.com/justjoheinz/de294fb184c75cb305eb00c91bfa96dc

# Expected Behavior

The example should typecheck and be free of errors

# Observed Behavior

see above

justjoheinz:
Fixes #4651

### DIFF
--- a/docs/tutorial/interp.rst
+++ b/docs/tutorial/interp.rst
@@ -30,7 +30,7 @@ calculated just like any other value:
     interpTy : Ty -> Type
     interpTy TyInt       = Integer
     interpTy TyBool      = Bool
-    interpTy (TyFun A T) = interpTy A -> interpTy T
+    interpTy (TyFun a t) = interpTy a -> interpTy t
 
 We're going to define a representation of our language in such a way
 that only well-typed programs can be represented. We'll index the


melted:
Thanks for the PR!

PR #4653, opened 2019-01-31
Add correct version of the `network` pkg as extra dep

asandroq:
The version of the `network` package in the chosen Stackage snapshot does not match the one needed by Idris. This change adds the needed package as an external package.


melted:
Is this still valid?

asandroq:
I think the stack.yaml has changed since then.

PR #4654, opened 2019-02-05
Add takeUpto and dropUpto functions to Data.Vect

locallycompact:
Hi, I thought these would be useful to place here. What do you think?

### DIFF
--- a/libs/base/Data/Vect.idr
+++ b/libs/base/Data/Vect.idr
@@ -110,6 +110,20 @@ drop : (n : Nat) -> Vect (n + m) elem -> Vect m elem
 drop Z     xs        = xs
 drop (S k) (x :: xs) = drop k xs
 
+||| Take up to the first n elements of a Vect if they exist.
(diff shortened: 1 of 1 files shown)


libs/base/Data/Vect.idr
```diff
@@ -110,6 +110,21 @@ drop : (n : Nat) -> Vect (n + m) elem -> Vect m elem
 drop Z     xs        = xs
 drop (S k) (x :: xs) = drop k xs
 
+||| Take up to the first n elements of a Vect if they exist.
+||| @ n the maximum number of elements to take
+takeUpto : (n : Nat) -> Vect n elem -> Vect (minimum x n) elem
```
clayrat: Duplicate signature

locallycompact: Sorry, corrected.

locallycompact:
Fixed another silly error. Any further comments?

melted:
Thanks for the PR!

PR #4659, opened 2019-02-25
Initialise error while opening a directory.

jinbijin:
On at least some machines, the field error of DirInfo isn't actually automatically initialised as zero; this is enforced in this commit.


melted:
That makes sense. Thanks!

PR #4660, opened 2019-02-25
fix #4656

### LINKED ISSUE #4656: JS Codegen Bug with --interface
There seems to be a codegen bug with the JS backend having to do with the combination of using `--interface`, `do` blocks, and `let` statements in `do` blocks.

# Steps to Reproduce

Consider the following code:

```idris
module Main

export
f : JS_IO ()
f = do
    let res = ()
    let res2 = res
    let res3 = res
    putStrLn' "Hello?"

lib : FFI_Export FFI_JS "" []
lib =
    Fun f "f" $
    End
```

Then, compile the code to a node module like so:

```bash
idris --codegen node --interface bug.idr -o bug.js
```

If we require the module and execute `f` from `node`, we get the following:

```bash
$ node
> l = require('./bug.js');
{ f: [Function: Main__f] }
> l.f()
[Function]
> 
```

# Expected Behavior

Calling `f()` should print `"Hello?"`.

# Observed Behavior

Calling `f()` returns an unapplied function. Note that it is pretty easy to see where this unapplied function comes from in the generated JS code.

# Triaging Efforts

I've already minimized this as much as I can. I also have an idea at the specific situation that triggers this. If we consider the slightly different code:

```idris
module Main

export
f : JS_IO ()
f = do
    let res = ()
    let res2 = res
    let res3 = res2 -- Here we use res2 instead of res. This code DOES print "Hello?"
    putStrLn' "Hello?"

lib : FFI_Export FFI_JS "" []
lib =
    Fun f "f" $
    End
```
then the bug is not exhibited (that is, the code above does correctly print "Hello?"). 

For another variation, consider:

```idris
module Main

export
f : JS_IO ()
f = do
    let res = "Hello?"
    putStrLn' res
    putStrLn' res

lib : FFI_Export FFI_JS "" []
lib =
    Fun f "f" $
    End
```

This code also triggers the bug. Similar variations show the pattern that the bug is triggered when a `let x = ...` statement is used in a `do` block, and `x` is then referenced 2 or more times below.

I'm using Idris `1.3.1` on macOS.

# Attached Code
I've attached a zip file with 2 failing cases, 2 passing cases, and a test running script `test.sh` which takes 1 argument, the name of the `.idr` file to test:

[bug.zip](https://github.com/idris-lang/Idris-dev/files/2881249/bug.zip)

rbarreiro:
fix for #4656


melted:
Thanks for the fix!

PR #4670, opened 2019-03-12
Make binary files the default when opening.

melted:
The reason is that most things on windows handles unix line endings just
fine and this leads to terrible bugs when writing binary files.

fopen still allows opening of files without the "b".

This only has effect on Windows, but POSIX systems allows the "b" as a
noop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # New in next version
 
+
+## Library updates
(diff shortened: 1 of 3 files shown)


clayrat:
Oh, I've ran into this this issue recently and thought that's it's rather a biug in the TTC producing code itself :+1:

PR #4675, opened 2019-03-16
Algebra proofs

Sventimir:
The module `Control.Algebra.Laws` introduces proofs of some basic basic laws of algebra, such as that anything multiplied by zero is zero or inverse composed with inverse is identity. These proofs might be helpful in proving more complex facts about algebraic structures.

I'm not sure if the coding style complies well with the rest of the package. If it doesn't or there are some similarly basic proofs that I overlooked, I'll be happy to fix it.


clayrat:
You can put multiple bindings under one `let` block, i.e.
```
let
  foo = 1
  bar = "a"
  ...
 in
...
```

Sventimir:
Sure, why not.

melted:
Thanks!

Sventimir:
Your welcome :)

PR #4684, opened 2019-04-01
Fixes error in type definition for So.

### LINKED ISSUE #4683: Error in Documentation for Effects
The documentation of [Simple Effects](https://github.com/idris-lang/Idris-dev/blob/master/docs/effects/simpleeff.rst) is incorrect. As documented by [Philipe de Rochambeau](https://groups.google.com/forum/#!topic/idris-lang/L0GWo7va7dc), it uses the syntax:

```
data So : Bool -> Type = Oh : So True 
```

Where it should say:

```
data So : Bool -> Type where
  Oh : So True
```

ivanperez-keera:
The type definition for So used the syntax:
```
data So : Bool -> Type = Oh : So True 
```
which is not valid idris.

Fixes #4683

### DIFF
--- a/docs/effects/simpleeff.rst
+++ b/docs/effects/simpleeff.rst
@@ -212,7 +212,8 @@ from the result of a dynamic check:
 
 .. code-block:: idris
 
-    data So : Bool -> Type = Oh : So True
+    data So : Bool -> Type where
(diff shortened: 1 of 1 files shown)


melted:
Thanks for the PR!

PR #4685, opened 2019-04-03
Add `--optimise-nat-like-types`

ziman:
This patch compiles all type families that end up looking like `Nat` after erasure into bigints.

This includes `Nat`, but also `Fin`, various `Elem` predicates, etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@
   interfaces (i.e. types for names declared `export` and definitions of names
   declared `public export`) are unchanged.
 
+## Miscellaneous updates
++ Compiler flag `--optimise-nat-like-types` enables compilation
+  of `Nat`-like type families to big integers. A type family
+  is `Nat`-like if, after erasure, it has two constructors,
+  one nullary, the other one with exactly one recursive field.
+
 # New in 1.3.1
 
 ## Tool updates
--- a/man/idris.1
+++ b/man/idris.1
@@ -99,6 +99,10 @@ should not necessarily be seen as production ready nor for industrial use.
                            e.g. corei7 or cortex-m3.
   --color,--colour         Force coloured output
   --nocolor,--nocolour     Disable coloured output
+  --optimise-nat-like-types,--optimize-nat-like-types
+                           Compile Nat-like types to big ints
+  --no-optimise-nat-like-types,--no-optimize-nat-like-types
+                           Do not compile Nat-like types to big ints
   --consolewidth WIDTH     Select console width: auto, infinite, nat
   --highlight              Emit source code highlighting
   --no-tactic-deprecation-warnings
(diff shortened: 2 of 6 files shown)


jfdm:
@ziman Thanks for the PR! If this is above board and passes the test, I don't see why this should not be merged. Prior to merging we need to think of house keeping and update:

+ CHANGELOG
+ Documentation in `docs` were appropriate.
+ Man page

david-christiansen:
This is really cool!

Is the eventual intention that it be turned on by default?

melted:
Great feature!

The expected output of pkg010 needs to be updated.

Shouldn't it be spelled generalized?

david-christiansen:
> Shouldn't it be spelled generalized?

Typical UK rules use an "s" here, and Idris prefers UK-exclusive spellings.
Though for some other options, like "color/colour", we do support the US
standard as an alternative, so we might want to do that here too.


Den ons. 3. apr. 2019 kl. 09.19 skrev Niklas Larsson <
notifications@github.com>:

clayrat:
Maybe remove the `-hack` part too? The fact that it has to be enabled with a separate flag already indicates this is something non-standard.

ziman:
> Is the eventual intention that it be turned on by default?

AFAIK, there are no concrete plans. I implemented this feature in an attempt to speed up Blodwen (but it seems that it does not have as much impact as we would have liked to see) and now just adding it as an option since every once in a while someone brings this up.
> Shouldn't it be spelled generalized?
> Maybe remove the `-hack` part too?

I propose renaming the flag to `--opt-nat-likes`, which avoids both spelling questions.

david-christiansen:
Sorry for the bikeshedding. But when I see that name, I think of Facebook
likes :-)

What about "--optimise-nat-like-types" and "--optimize-nat-like-types" for
consistency with our handling of color/colour?

Den tor. 4. apr. 2019 kl. 08.21 skrev Matúš Tejiščák <
notifications@github.com>:

ziman:
> What about "--optimise-nat-like-types" and "--optimize-nat-like-types" for consistency with our handling of color/colour?

I like that, thank you!

ziman:
Some GHC versions still crash on a multiway `if`. I'll rewrite it in a different way.

melted:
All green!

natewind:
Does this patch replace bounded (dependent) Nats by ints (instead of BigInt) if the upper limit is smaller that the corresponding power of 2? For example, a Nat that is guaranteed to be smaller than 1000 would be a uint16_t because 2^8 < 1000 < 2^16.

ziman:
No, this patch always uses bigints.

natewind:
Ok :(

natewind:
Anyway, I hope someone will implement this and similar optimizations at some point, so just pitching the idea :) To have the speed of C++, Haskell-like expressiveness and correctness proofs in one language would be incredible.

PR #4687, opened 2019-04-16
rewrite documentation for interactive theorem proving

martinbaker:
I think its bad that the documentation about interactive theorem proving uses deprecated code. I have therefore rewritten this page to use elaborator-reflection instead of the old interactive tactics.


jfdm:
Thanks for the contribution, I will look at this in more detail later this week

Just one immediate observation, it would be good to keep the old tutorial around, as well as having the new one. Although, the old system is *deprecated*, it still exists. 

Maybe it would be better to place a copy of the old tutorial in the tutorial section of the docs, whilst promoting your contribution as the *official line*.

Thanks.

david-christiansen:
Thanks for working on this!

I would very much prefer that elaborator reflection not be positioned as a technique for proof automation per se, but rather as a metaprogramming feature that has some primitives that look a bit like proof tactics. This is because the elaboration mechanism that it exposes to user code is designed for elaborating Idris, rather than proving things, which leads to various differences in API.

I also agree with @jfdm that it would be good to keep the docs for the old proof setup around as long as that feature exists.

How would you feel about retargeting this text to be an elaborator reflection tutorial, rather than as a replacement for the old section?

melted:
@martinbaker We shouldn't remove the documentation for a still existing feature despite the depreciation. It should note that it's depreciated. People will come across or work with programs that uses old-style proofs and removing the documentation will not help them.

I like the idea of a guide for proving with elaborator reflection even though the mechanism is more general than that.

martinbaker:
OK, I have changed the pull request to include documentation pages for both the old and the new code.

I have also added the following text:

"The primary purpose of the elaboration mechanism is to elaborate Idris and so it
is not optimised to work as a proof assistant, however it can interactively
construct proofs as described on this page."

docs/proofs/interactive.rst
```diff
@@ -25,145 +123,147 @@ hypothesis, and insert *holes* for the rest of the definition:
...
+    Holes: Main.plusredZ_S, Main.plusredZ_Z
 
-    *theorems> :t plusredZ_S
-    plusredZ_S : (k : Nat) -> (k = plus k Z) -> S k = plus (S k) Z
+This tells us that we have two holes Main.plusredZ_S and Main.plusredZ_Z. We can solve
+these separately, ``plusredZ_Z`` is the simplest so we will do that first.
 
-The ``:p`` command enters interactive proof mode, which can be used to
-complete the missing definitions.
+The ``:elab plusredZ_Z`` command enters interactive proof mode, which can be used to
```
david-christiansen: Instead of calling it "interactive proof mode", I'd call it "interactive elaboration mode".

docs/proofs/interactive.rst
```diff
@@ -25,145 +123,147 @@ hypothesis, and insert *holes* for the rest of the definition:
...
-    {hole0} : Z = plus Z Z
+    ----------                 Goal:                  ----------
+    {hole_0} : 0 = 0
 
-This gives us a list of premises (above the line; there are none here)
-and the current goal (below the line; named ``{hole0}`` here). At the
-prompt we can enter tactics to direct the construction of the proof. In
-this case, we can normalise the goal with the ``compute`` tactic:
+This has been normalised to ``0 = 0`` so now we have to prove that ``0`` equals ``0``, which
+is easy to prove by reflexivity from pruviloj library:
```
david-christiansen: I think there's a missing "the" after "from".

docs/proofs/interactive.rst
```diff
@@ -25,145 +123,147 @@ hypothesis, and insert *holes* for the rest of the definition:
...
+    -Main.plusredZ_S> intro `{{ih}}
+
+    ----------              Assumptions:              ----------
+     k : Nat
+     ih : k = plus k 0
+    ----------                 Goal:                  ----------
+    {hole_0} : S k = S (plus k 0)
 
 Since plus is defined by recursion on its first argument, the term
 ``plus (S k) Z`` in the goal can be simplified, so we use ``compute``.
```
david-christiansen: It looks to me like "compute" didn't do anything here.

docs/proofs/interactive.rst
```diff
@@ -25,145 +123,147 @@ hypothesis, and insert *holes* for the rest of the definition:
...
 
-    -plusredZ_S>
-
-The ``rewrite`` tactic takes an equality proof as an argument, and tries
-to rewrite the goal using that proof. Here, it results in an equality
-which is trivially provable:
+The ``rewriteWith`` tactic takes an equality proof as an argument, and tries
+to rewrite the goal using that proof. The ih value is entered as a constant
+of type ``TTName`` which is entered as a backtick with double braces `{{ih}} but
+``rewriteWith`` requires a ``raw`` value so we use ``Var`` to do this conversion.
```
david-christiansen: `raw` should be capitalized.

I'd phrase this something like "`rewriteWith` requires an expression, rather than just a name, so the `Var` constructor is used to make a variable".

david-christiansen:
Thanks for doing this!

martinbaker:
I agree with the detailed points made by David Thrane Christiansen, I believe I have now addressed these points in the update I have just committed.

jfdm:
Ah thanks to all for commenting on this. The PR is in a better position now, Ill merge this!.

Thanks @martinbaker 

Documentation might not be seen by some as a 'trendy' improvement to a language but I certainly do. Unfortunately, some of us who use Idris heavily, and also wish to contribute, do not have the time to contribute as much as we should. We have to prioritise, and contributions like this are more than appreciated.

Thanks

david-christiansen:
Yes, thank you very much!

PR #4689, opened 2019-04-18
Initial transfer of {AS} grammar to Sphinx Documentation.

jfdm:
When @ahmadsalim originally moved Idris' parser from parsec to Trifecta, a grammar describing the surface syntax was contributed as source code comments.
This is good for developers, but not necessarily helpful for those not wanting to have a reference.
Given that we have moved documentation from LaTeX to sphix-doc we can leverage the `productionlist` environment to provide a clickable grammar.

I have split the grammar into sections, but over time this should be pruned, sorted, and formatted properly.
Further I think a good contribution would be to merge this file with the proceeding section to provide examples alongside.
It might be work spinning out these sections as a top level document in its own right, with the 'Language Reference' being placed as a 'Idris Tooling Guide'.

### DIFF
--- a/docs/reference/index.rst
+++ b/docs/reference/index.rst
@@ -26,6 +26,7 @@ This will tell you how Idris works, for using it you should read the Idris Tutor
    uniqueness-types
    ffi
    syntax-guide
+   syntax-reference
    erasure
    ide-protocol
    semantic-highlighting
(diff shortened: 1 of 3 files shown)


ahmadsalim:
Oh, this looks cool! 😄

PR #4701, opened 2019-05-08
Implement idris_clock on windows

melted:


melted:
Hmm, this has a Y2038 problem as MKBIGI takes a 32-bit int. But I'll fix that later.

PR #4705, opened 2019-05-13
Simplify reverse proofs

nickdrozd:


nickdrozd:
Any takers on this one? There are two small changes:

1. Add theorem `reverseOntoAcc`, which is a generalization of `reverseOntoSpec` with a much simpler proof. Use the former to simplify the proof of the latter.

2. Rewrite the proofs of `reverseAppend` and `reverseReverseId` to use `rewrite` instead of repeated applications of `trans` and `cong`. I guess this is basically a style change, but I think it improves readability.

melted:
I’ll look at it after work.

melted:
Looks good!

Issue #4717, opened 2019-05-24
Missing screenshot in Interactive Theorem Proving Documentation


Phiroc:
A screenshot is missing in the Interactive Theorem Proving documentation (http://docs.idris-lang.org/en/latest/proofs/interactive.html), right after the following paragraph:

```
This has been normalised to 0 = 0 so now we have to prove that 0 equals 0, which is easy to prove by reflexivity from the pruviloj library:
```
Here's what is should look like:

```
-Pruv.plusredZ_Z> reflexivity
plusredZ_Z: No more goals.
```

martinbaker:
oops, sorry schoolboy error.

In the rst file I wrote the code block like this:

.. code-block:: idris
    -Main.plusredZ_Z> reflexivity
    plusredZ_Z: No more goals.

But this does not display unless there is an empty line after '.. code-block:: idris'.
So I have sent a pull request which adds in the blank line.
https://github.com/idris-lang/Idris-dev/pull/4718

PR #4718, opened 2019-05-24
fix formatting error which caused missing code block

martinbaker:
The fixes an issue reported by Phiroc here:
https://github.com/idris-lang/Idris-dev/issues/4717

There was no empty line after the code-block in the rst file.
This prevented the code block from being displayed in the documentation.

This patch adds the blank line and the code should now be displayed correctly.


jfdm:
Thanks.

PR #4719, opened 2019-05-27
Added examples to the Vect doc

klangner:
This PR contains only new examples in doc comments in Vect type.


jfdm:
Ah nice. I don't see any reason not to accept this. Thanks!

PR #4724, opened 2019-05-31
Add some propositional logic theorems

nickdrozd:
These aren't useful for anything, but they might be of interest as a reference.

### DIFF
--- a/libs/contrib/Data/Logic/Propositional.idr
+++ b/libs/contrib/Data/Logic/Propositional.idr
@@ -1,0 +1,161 @@
+module Data.Logic.Propositional
+
+%default total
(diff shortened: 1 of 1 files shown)


nickdrozd:
Appveyor failure seems unreleated.

jfdm:
I have not looked at this in detail, however, two things stand immediately out:

1. There is a missing module definition.
2. It would be better if the module's file name matched the style used elsewhere in `libs/`.

Once these are fixed, then we can look at the rest of the code.

nickdrozd:
@jfdm Done

melted:
Thanks for the PR!

PR #4728, opened 2019-06-12
Rewrite fib with provably equivalent tail recursive definition

nickdrozd:


nickdrozd:
@melted Could you take a look? This is an easy one, I hope :)

msmorgan:
I tried doing this previously in #4216, but my change was turned down because it's a breaking change in prelude.

melted:
This doesn't use default implicits though.

Frankly, I have no idea why fib is in the prelude at all. That split will be way better in Idris 2 :-)

nickdrozd:
I don't particularly care how the function is implemented in the prelude (why is it defined there at all???), so I've amended this PR to only add a proof that recursive and iterative Fibonacci are equivalent without any reference to the "official" `fib` function.

nickdrozd:
The build failure looks unrelated.

melted:
Looks good!

Issue #4730, opened 2019-06-19
Can't use %runElab in 'let ... in' statement


Phiroc:
Hello,
if you cut and paste the second Elab proof displayed at the bottom of the Automatic Theorem Proving documentation (http://docs.idris-lang.org/en/latest/proofs/interactive.html), into the second 'hole' in the example code, an error message stating that "in" is "unexpected" is displayed.

# Steps to Reproduce

```
import Pruviloj
import Pruviloj.Induction

%language ElabReflection

plusReducesZ' : (n:Nat) -> n = plus n Z
plusReducesZ' Z = %runElab (do reflexivity)
plusReducesZ' (S k) = let ih = plusReducesZ' k in
                        %runElab (do intro `{{k}}
                                    intro `{{ih}}
                                    compute
                                    rewriteWith (Var `{{ih}})
                                    reflexivity)
```

# Expected Behavior

The compiler should not generate any errors 
(cf. "Again, we can cut & paste this into the hole in the original file." statement at the bottom of the aforementioned documentation page).

# Observed Behavior

> 16 |                                     intro `{{ih}}
>    |                                     ^
> unexpected "in"
> expecting dependent type signature

Phiroc:
NB : the 'compute' command should be removed in the source code above

PR #4731, opened 2019-06-22
Correct documentation showing 'compute' tactic where it should not be.

martinbaker:
This pull request changes documentation only.

The documentation shows a 'compute' command being generated by Elab which it does not do in this example, so this PR removes the 'compute' tactic output line.

This may have caused the issue in this forum discussion:

https://groups.google.com/forum/#!topic/idris-lang/ivH9fpTmMKE

It is also documented here:

https://github.com/idris-lang/Idris-dev/issues/4730

Is there an underlying issue with the code? In that Elab can generate output without generating errors, but if cut&pasted into the hole will produce an error? This PR does not address this issue so it may not resolve the above linked issue.

Martin

### DIFF
--- a/docs/proofs/interactive.rst
+++ b/docs/proofs/interactive.rst
@@ -270,7 +270,7 @@ because this gives the following error:
   unexpected "in"
   expecting dependent type signature
 
-We can try putting the proof into a seperate function like this:
+However if we put the proof into a seperate function like this:
 
 .. code-block:: idris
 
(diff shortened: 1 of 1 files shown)


martinbaker:
I have also modified the documentation to reflect the problems that people are reporting in putting the proofs generated by :elab REPL into a file and compiling.
However this documentation needs to have more useful advise to potential users about these issues they are finding.
@david-christiansen
help please!

melted:
Thanks for the PR! Looks good.

PR #4732, opened 2019-07-02
Stop ignoring dots for implicits

ziman:
`JanBessai` on IRC reported the following MFE: https://gist.github.com/JanBessai/999c7162f35eb782a53de0d2bf61fe9c

The compiler does not emit an erasure warning, despite `--warnreach`. I looked into it and found that implicits are hardcoded to be always *un*dotted, so I changed it.


ziman:
This is the original diff rebased on top of `master`. Let's see what happens this time.

melted:
The error is unrelated anyway. Let's merge.

PR #4733, opened 2019-07-04
expand Bool laws, add (verified) lattice instances for Bool & Nat

clayrat:


clayrat:
Looks like there's some kind of stack issue, only for GHC 8.4.1

melted:
Yes, there is something unrelated with the stack build.

But reg024 fails on Appveyor, I'll restart it to see if it was intermittent.

clayrat:
Now `st002` fails, seems random

melted:
Let's merge. CI failures are unrelated.

PR #4737, opened 2019-07-10
add fRemove FFI call

clayrat:


melted:
Looks good!

I guess the fileRemove wrapper is useful if someone has an oddball system with another API.

clayrat:
Hm I wonder how `make` has succeeded for me locally if there's actually a type mismatch. Should I catch the IO error in `Core.Execute`?

clayrat:
Ugh, not sure why cabal just swallows these errors, gonna try building with stack.

melted:
Do you have a newer GHC than Travis?

clayrat:
@melted The build on Travis seems to finally work, but I still have a couple of questions:
* Should I catch the error in `Core.Execute` a-la the `fileOpen` case? Currently it seems to abort execution e.g. when file is not there which is probably not what we want?
* I wanted to test both the regular function and the exec handler in a single test but it looks like the output from the test gets overwritten, so I get only a single `success`, is that expected or am I doing something wrong?

clayrat:
@melted Nevermind, I think I finally figured it out.

clayrat:
Everything passes now except that weird 8.4.1 stack thing.

PR #4734, opened 2019-07-07
Fix Travis build

ziman:
The Travis build [seems to be complaining](https://travis-ci.org/idris-lang/Idris-dev/jobs/554976050) about something in the `package`s field. Instead of trying to find a formulation that works across various GHC/stack versions, I removed it because  `.` is the default value, anyway.

PR #4738, opened 2019-07-12
Fix stack file

wilbowma:
`stack build` gave syntax errors and missing dependencies until I modified it as such.
Also, some trouble with libffi which I gather people know about.

Credit for this to Ohad.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -2,9 +2,10 @@
 resolver: lts-13.21
 
 packages:
-  - location: .
+  - '.'
 
 extra-deps:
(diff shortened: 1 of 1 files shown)


jfdm:
@wilbowma  Thanks for the PR, it is pretty much identical to #4734, which I did just merge, aside from turning libFFI support off. 

I am not sure what our position should be wrt to libFFI issues and turning support off in the build---maybe something for @edwinb to chime in on. We do provide an alternative stack yaml config in the mean time.

wilbowma:
Ah didn't notice that one. Thanks!

ilyakooo0:
Should we close this PR then?

PR #4746, opened 2019-07-27
Add Ackermann function definitions and proof

nickdrozd:
Two different Ackermann defintions and a proof that they're equivalent. These aren't useful for anything, although they can serve as good test cases for the totality checker.


nickdrozd:
@melted Ping :)

The build failure looks unrelated.

melted:
Yes, stack on travis is known broken.

Looks good otherwise, so I'll merge.

PR #4754, opened 2019-08-14
Fix file handle leaks and simplify code

clayrat:


melted:
Looks good!

Travis and stack still don't get along. But that's not this PR's problem.

Issue #4757, opened 2019-08-27
Build with GHC 8.8


melted:
New-style cabal installs are broken because sdist hooks were removed and cabal's wildcards are unnecessarily restricted: https://github.com/haskell/cabal/issues/5883. I'm inclined to just stick v1- on the relevant cabal commands and not implement new-style commands at all. Because we don't want to name every single test in the cabal file again.

Otherwise, I guess it's just trying to build it until all deps are in order. I don't know any more roadblocks at least.

melted:
The build fails on regex-base. Relevant ticket: https://github.com/ChrisKuklewicz/regex-tdfa/issues/23

melted:
The regex packages will get new maintenance:

https://mail.haskell.org/pipermail/libraries/2019-September/029927.html

PR #4758, opened 2019-08-27
Compatibility with Cabal 3.0

melted:
Use the v1 cabal commands


melted:
Building with GHC 8.8 is still blocked by dependencies (see #4757), but this build fine with 8.6

melted:
Same stack issue as always.

PR #4769, opened 2019-10-20
Update tutorial PDF

sysvinit:
The current compiled `idris-tutorial.pdf` shipped in the Idris repository hasn't been updated in a while, as it was last modified just after the 1.3.0 release last year. I've regenerated the PDF from the sources in `docs/` and copied it to the repository root, so it should now be up to date with documentation changes made between 1.3.0 and the current master commit.


melted:
I wonder why it's so much smaller in size. Still, it looks good.

Thanks!

PR #4776, opened 2019-11-07
Proofs docs were fixed a little

buzden:
Mostly, backticks were added to set the code where needed.
Also, couple of code blocks were added, "i.e." was added for clarification and one misprint fixed.
In couple places text was made to be styled like the rest of the docs.


melted:
Looks good! Thanks!

PR #4780, opened 2019-11-19
An attempt to fix recently failing travis CI build job

buzden:
Version of stack used for the style check was bumped


melted:
It's all green so it must be good. Thanks!

PR #4781, opened 2019-11-20
Add equality functions for Vectors of potentially different lengths.

jfdm:
Sized lists are useful, however, interfaces require that the length value is fixed.
The added functions allow for vectors of potentually different lengths to be compared.

### DIFF
--- a/libs/base/Data/Vect.idr
+++ b/libs/base/Data/Vect.idr
@@ -380,6 +380,14 @@ implementation (Eq elem) => Eq (Vect len elem) where
   (==) (x::xs) (y::ys) = x == y && xs == ys
 
 
+||| Compare two vectors of potentially different length for equality.
+|||
+vecEq : Eq type => Vect n type -> Vect m type -> Bool
+vecEq [] [] = True
+vecEq [] (x :: xs) = False
+vecEq (x :: xs) [] = False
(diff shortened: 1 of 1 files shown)


melted:
CI reports 

```
Type checking .\Pruviloj\Derive\DecEq.idr
.\Pruviloj\Derive\DecEq.idr:97:43-66:
   |
97 |                           (tm2', ty2') <- forget tm2 >>= check env
   |                                           ~~~~~~~~~~~~~~~~~~~~~~~~
When checking right hand side of Pruviloj.Derive.DecEq.case block in case block in case block in case block in case block in case block in mkRhs at .\Pruviloj\Derive\DecEq.idr:84:10-21 at .\Pruviloj\Derive\DecEq.idr:86:13-23 at .\Pruviloj\Derive\DecEq.idr:87:13-23 at .\Pruviloj\Derive\DecEq.idr:94:56-65 at .\Pruviloj\Derive\DecEq.idr:96:43-66 at .\Pruviloj\Derive\DecEq.idr:97:43-66 with expected type
        Elab Raw
When checking an application of constructor Language.Reflection.Var:
        Can't disambiguate name: Data.Vect.DiffLength.decEq, 
                                 Decidable.Equality.decEq
make[1]: *** [Makefile:5: build] Error 1
```

PR #4783, opened 2019-11-23
Corrections in the `docs/reference`

buzden:
Mostly backticks, but also old wiki artifacts (incl. references) were corrected. Also, an attempt of structuring in pretty non-structured file.


melted:
Thanks!

PR #4784, opened 2019-11-23
Effects tutorial fix and additon of small clarification

buzden:
The summary of the `effects` library used incorrect (maybe old?) type of the `Eff` type with explicit `m : Type -> Type` argument. Also, some of functions in the summary required explicitly `Handler`s (which they should not require). I've fixed it.

Also, the styling (indenting) of the `FILE_IO` part was made to be consistent with the rest of the summary.

Also, I made one existing note as a _ReST's note_ and extended it for those people who may want to use the `FILE` effect after the effects tutorial reading (to make the wonder a bit less).


buzden:
I have no idea why `primitives005` test fails in appveyor.

melted:
Thanks! It's obviously unrelated.

PR #4785, opened 2019-11-23
Couple of convenience absurd-related functions were added to `Data.So`

buzden:
These functions may shorten a little proofs when you have situation of `So b` and `So (not b)` simultaneously.


melted:
Looks like reasonable stuff to have in base. Thanks!

Travis just timed out.

PR #4788, opened 2019-12-01
Functions to operation with So of conjunctions

buzden:
Several functions that allow to operate with `Data.So` of conjunctions. They allow to write external properties proofs for functions that use conjunctions inside (similar to e.g. `sorted`) more or less easily.

This PR partially depends on #4787. Is contains the same commit with one `Bool` property (for the case when this PR is accepted but those one is not). It's okay for this commit to be rebased in case if #4787 is accepted.


buzden:
Now this PR was changed to depend on #4789 to not to duplicate the code of `Bool` properties.

This is done to put similar functions for `So` near (i.e. to the `Data.So` module of the `base` package).  But one of suggested functions depends on `Bool` property that resides in the `contrib` library and is moved to `base` by #4789 (that's why this PR depends on that one). An alternative solution would be to put all functions that depend on `Bool` properties to the `contrib` library to module somewhat like `Data.So.Extra`. But I personally do not see why some of the functions should lie in `Data.So` itself but some in `Data.So.Extra`. What does the community think?

PR #4789, opened 2019-12-02
Module with properties of `Bool` was moved from `contrib` to `base`

buzden:
This allows to use nicely written properties of `Bool` type in functons in the `base` package.
It looks like this is a better place for it than `contrib`.


buzden:
Travis-CI seems to fail because of Ubuntu keyserver problems (according to [this](https://travis-ci.community/t/packages-cannot-be-authenticated-gpg-error-google-chrome-stable/3000) and [that](https://github.com/travis-ci/travis-ci/issues/9822)) :-(

PR #4794, opened 2019-12-15
Rearrange Documentation for Elaborator Reflection

martinbaker:
Rearrange documentation for elaborator reflection

This PR changes documentation only.

Currently elaborator reflection documentation is split over two sections, there is an introduction in the 'reference' section plus the stuff that I added in the 'proofs' section. This PR just moves these together in a new 'elaborator reflection' section.

I think this will make it easier for the reader to find the information they need.


melted:
Looks good!

PR #4796, opened 2019-12-18
Added missing RESERVENOALLOC case to ToJSON BC definition

mkleczek:
Hello,
I have found a missing case in toJSON implementation when trying to build idris-jvm from mmhelloworld.
This PR is a simple fix.


melted:
Thanks for the PR!

PR #4806, opened 2020-01-21
Rts stdfgn

mokshasoft:
Move the usleep implementation to the idris_stdfgn file. It seems to belong more there than in idris_rts. Also remove some unused headers in the C RTS.


melted:
Thanks for the PR!

PR #4808, opened 2020-01-22
Compile with GHC 8.8

melted:
Now with cabal 3.0 updates


melted:
I put GHC 8.8 on appveyor as you still cant build Idris with libffi support on Ubuntu.

More information here: https://gitlab.haskell.org/ghc/ghc/issues/15397

It seems insurmountable without building a new GHC using the system libffi and we can't do that in travis. I guess we could and put them up as PPA, but then we would have to figure out how to package that. The simplest solution would be to disable the tests requiring libffi and build idris without the support.

If we had another CI I assume we could start from arch or something with up to date GHC and libffi.

PR #4818, opened 2020-02-21
Simplify some Nat proofs

nickdrozd:


nickdrozd:
@melted I think this one is ready now.

melted:
Looks good!

PR #4820, opened 2020-02-29
More proofs, simpler proofs

nickdrozd:


nickdrozd:
@melted Here's another one. Let me know if you want me to start bothering someone else.

libs/base/Control/Isomorphism.idr
```diff
@@ -81,14 +82,12 @@ eitherAssoc = MkIso eitherAssoc1 eitherAssoc2 ok1 ok2
...
 ||| Disjunction with false is a no-op
 eitherBotLeft : Iso (Either Void a) a
-eitherBotLeft = MkIso to from ok1 ok2
+eitherBotLeft = MkIso to from (\_ => Refl) ok2
   where to : Either Void a -> a
         to (Left x) = void x
         to (Right x) = x
         from : a -> Either Void a
         from = Right
-        ok1 : (x : a) -> to (from x) = x
```
LeifW: Leaving this as a separate function is more verbose, but I think the type signature helps you follow what's going on.

libs/base/Control/Isomorphism.idr
```diff
@@ -248,9 +241,7 @@ maybeEither = MkIso to from iso1 iso2
 
 ||| Maybe of void is just unit
 maybeVoidUnit : Iso (Maybe Void) ()
-maybeVoidUnit = (Maybe Void)     ={ maybeEither   }=
-                (Either Void ()) ={ eitherBotLeft }=
-                ()              QED
```
LeifW: I think I prefer the preorder reasoning syntax form, as it shows the types of the intermediate values (despite being more verbose).

libs/base/Control/Isomorphism.idr
nickdrozd: Fair enough, changed back.

libs/base/Control/Isomorphism.idr
nickdrozd: I actually didn't know this syntax existed until I looked at this file today. Is it something that should be used more? Does it make more sense for some kinds of proofs than for others? In any case, I changed it back.

nickdrozd: Links:
- https://gist.github.com/david-christiansen/7962927
- https://trillioneyescoding.wordpress.com/2014/10/28/preorder-reasoning-in-idris/
- https://groups.google.com/forum/#!msg/idris-lang/JcxsNHHbptc/gkhmuVruBAAJ

LeifW: I think it's intended to look like how you would write out a proof on pencil & paper with regular math, e.g.
https://proofwiki.org/wiki/Natural_Number_Addition_is_Associative/Proof_2#Induction_Step

It's somewhat documented at http://docs.idris-lang.org/en/latest/reference/misc.html#preorder-reasoning, but I like how it's described in its comment:
> `foo ={ prf }= bar ={ prf' }= fnord QED`
> is a proof that `foo` is related to `fnord`, with the intermediate step being `bar`, justified by `prf` and `prf'`
https://github.com/idris-lang/Idris-dev/blob/master/libs/base/Syntax/PreorderReasoning.idr
Like how you might write down something like:
*    a + 0  
* = 0 + a  -by commutativity
* = a        -by left identity
You could instead just write down something like "commutativity ∘ left identity", but that could be harder for people to follow.
I think it makes more sense for the cases where you have a lot of intermediate steps, i.e. a bunch of nested "rewrite in... rewrite in...", etc.
As far as a proof that is just e.g. `a + b = { plusCommutative }= b + a QED`, I think that syntax isn't adding anything, because you already have the types of the beginning and end from the top-level signatures probably. It's more for when you have a lot of steps, to document the intermediate types. Also it can be a helpful way of constructing proofs - you can write out the types, and leave metavariables for the implementations (as on the trillioneyes blog post), to have the compiler-driven editor interactivity help you figure out how to fill those.
I have some more examples of using that syntax if you'd be curious.
As far as preferred style, I don't know that we have a consensus on preferred style of proofs. Also, the language has been changing, and many proofs were perhaps done when "tactics" were the preferred way of doing them, and since then other language features have been added, or even other libraries / syntaxes like that "preorder reasoning".

I suppose it depends on who is maintaining / reading these proofs. Personally I like not just using them, but also reading them, and having them to point people to example proofs for them to walk through. I think of type signatures as not just stuff to guide the compiler and improve error messages when things don't line up, but also as documentation for humans reading the code.
Idris can do a lot automatically for you, but also I like to write a bunch of stuff out explicitly. E.g. when using `cong` I like to explicitly supply the function to be applied to both sides of the equation, e.g. `cong {f = S}`, as documentation to myself, if nothing else.

It can be frustrating developing proofs with just "rewrite" by itself, sometimes, as the compile failure is juts "rewrite did not change the type", and it's hard to tell where it went wrong. We even have other syntax in the language, `==>` and `<==`, to add extra type annotations to rewrite or to figure out the arguments of a equality to be used in a rewrite for you: http://docs.idris-lang.org/en/latest/reference/misc.html#match-application
But they don't seem to be used much, and I'm having a hard time figuring them out because they're so little used and I can't really find examples using them. Makes me wonder if they should even be in the language, if they're not really used. At any rate, I think I prefer the preorder reasoning syntax to ascribing the goal type of a rewrite with `==>`, though maybe that's just because I can't seem to get the `==>` working the way I expect or don't really understand it.

nickdrozd: I'm going to give preorder reasoning syntax a shot going forward (though I don't feel like going back and redoing this PR to use it).

nickdrozd:
This CI is real flaky. The Linux one always times out, and the Windows one fails some side-effect test. But the code in this PR is fine.

nickdrozd:
@melted This one is ready to rock I think.

melted:
Thanks!

Issue #1524, opened 2014-09-14
Can't find import Effect/StdIO


TorosFanny:
I recently started to learn effect. However, when I want to "import Effect.StdIO", it just report "Can't find import Effect/StdIO" both in the command line and in the "Try Idris" website.

PR #4829, opened 2020-03-17
Update introduction.rst with required `-p` flag

alexhumphreys:
Got the error `Can't find import Effects` when attempting this part of the tutorial. Managed to solve it after finding issue #1524, but would have been faster if this flag was mentioned in the tutorial.


melted:
Looks good! Thanks!

PR #4830, opened 2020-03-18
Fix build on GHC 8.6

melted:


ilyakooo0:
@melted just FYI: it still doesn't pass tests with GHC 8.6:

```
Type checking ./Control/Isomorphism/Vect.idr
/Users/iko/Developer/Idris-dev/Control/Isomorphism/Vect.idr:11:9:
   |
11 | nilUnit : Iso (Vect Z a) ()
   |         ^
```

PR #4837, opened 2020-04-02
Debian packaging

jfdm:
This PR adds an experimental script to generate deb packages, and updates the macOS script to use cabal 2.4 v1 style commands.

I have also fixed the issue in which the resulting paths are styled for cabal (i.e. with abi information) rather than for installation in the standard POSIX heirarchies.


melted:
So now we are back where we started with a huge list of files :-)

If that's the only way to keep it working on all cabal versions I guess that's OK.

ilyakooo0:
I didn’t investigate it, but this feels like a problem that hpack would solve. 

It would be ideal if the paths in the cabal file got statically generated by walking the directory structures. 

(Yes, I am aware that there used to be a fragile script that did something like that)

Having expanded file lists is the only way to reliably have the project building with cabal versions prior to 2.0.

LeifW:
For comparison, the equivalent script to make a package for Arch Linux (which installs into /usr as opposed to /usr/local or wherever) is in https://git.archlinux.org/svntogit/community.git/tree/trunk?h=packages/idris
The cabal configure, build, and copy invocations are in the PKGBUILD file.
I'm curious why the separate `install --only-dependencies` needed to be added? Is that just to keep more separate stages?

LeifW:
Once this is done, perhaps we could publish this to a PPA or something? We're already building Idris on an Ubuntu box on Travis, though perhaps there's more appropriate systems for building Ubuntu packages?

jfdm:
Thanks for the comments

> So now we are back where we started with a huge list of files :-)

> Having expanded file lists is the only way to reliably have the project building with cabal versions prior to 2.0.

Indeed, this is indeed a result of wanting to use cabal v1 commands.

> It would be ideal if the paths in the cabal file got statically generated by walking the directory structures.

Indeed that is one way to do it.

> I didn’t investigate it, but this feels like a problem that hpack would solve.

I am a bit reluctant to include more tooling, as that increases the maintenance dependence of the project.

> For comparison, the equivalent script to make a package for Arch Linux 

Thanks! I will update the script with some of their configuration. I forgot about our release and freestanding commands...

> The cabal configure, build, and copy invocations are in the PKGBUILD file.

Notice that they don't use cabal copy.

We 'need' to use cabal copy to avoid having to manually move all important files and exes across to the staging area. Unfortunately, cabal new style builds do not support a version of v1 copy. So maybe we have to do manual copying eventually if new style builds don't support copying for building binary releases. This wil also allow us to 'just' have an executable in the package.

> I'm curious why the separate install --only-dependencies needed to be added? Is that just to keep more separate stages?

cabal v1 build failed to get dependencies on my system with their calculate build plan. Sandboxing and install was the only way to get it to work.

> Once this is done, perhaps we could publish this to a PPA or something? We're already building Idris on an Ubuntu box on Travis, though perhaps there's more appropriate systems for building Ubuntu packages?

I am not going to look into PPAs as I don't have time to do so. Like cabal, I find the documentation lacking enought to quickly do something. That being said, we can use Travis to push releases to github itself.

PR #4844, opened 2020-04-12
XDG Base Directory Specification compliance

juanmeleiro:
Idris currently stores it's configuration files and cache (history file) on the `~/.idris` directory. To comply with the [XDG Base Directory Specification](https://wiki.archlinux.org/index.php/XDG_Base_Directory), it should use `$XDG_CONFIG_HOME/idris` (`~/.config/idris` by default) for configurations and `$XDG_CACHE_HOME/idris` (`~/.cache/idris` by default) for cache data.

However, doing that in a backwards compatible manner requries a work-around.  With this commits, the REPL would use the old directory, *if it exists*, for both configuration and cache data, using the XDG ones otherwise (*c.f.* [this comment](https://github.com/haskell/directory/issues/6#issuecomment-96521020)).

### DIFF
--- a/docs/reference/repl.rst
+++ b/docs/reference/repl.rst
@@ -40,12 +40,13 @@ Idris supports initialisation scripts.
 Initialisation scripts
 ~~~~~~~~~~~~~~~~~~~~~~
 
-When the Idris REPL starts up, it will attempt to open the file
-repl/init in Idris's application data directory. The application data
-directory is the result of the Haskell function call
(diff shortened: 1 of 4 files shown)


juanmeleiro:
I'm not really sure how to fix this issues. I can't really find anything about them on the contribution guide, so I'd appreciate some guidance.

juanmeleiro:
Just noticed there's another PR open for this same fix.

PR #4854, opened 2020-05-09
Simplify style in a few modules

nickdrozd:
This PR makes strictly nonfunctional style changes to a few modules, with an eye towards reducing verbosity and using less text overall. I think it makes things nicer to look at.


melted:
Thanks!

PR #4855, opened 2020-05-09
Typo of `<` instead of `<<` was fixed in the C RTS

buzden:
I have no idea whether this part of RTS is really used, but if it is, it seems that integer multiplications would be done more optimally.


melted:
Thanks!

PR #4856, opened 2020-05-09
Making `idris_main.c` for arduino work again

buzden:
Naming scheme was changed at 3e4752fc9f but not reflected in arduino's main. BTW, when corrected (with little patches to RTS) it still works nicely at Arduino even on AVR (at least, on atmega2560).


melted:
Thanks!

PR #4857, opened 2020-05-12
Lack of C types correspondence was fixed in C RTS

buzden:
`usigned` was used to fill fields declared as `uint32_t` and `uint16_t`, but these types not always correspond nicely to each other. In rare circumstances this can lead to warnings during compilation of a `*.c` file generated from `*.idr`, leading to ending of execution instead of calling a function at runtime.


melted:
Thanks!

PR #4859, opened 2020-05-14
Factorisation

Sventimir:
This pull request adds `Data.Nat.Factor` module to `contrib` package. It defines notions of a factor (a natural number, which divides another number without a remainder), and common factors of two natural numbers. It proves several basic laws of factors and implmements a decision procedure as to whether a certain number is a factor of another number of not. Finally, the notion of a greatest common divisor is formally defined and Euclid's algorithm for finding GCDs is implemented and proven correct.

This is a major step towards implementing a decent `Rational` type, which will be my next goal.

### DIFF
--- a/libs/contrib/Data/Nat.idr
+++ b/libs/contrib/Data/Nat.idr
@@ -10,6 +10,19 @@ diff k Z = k
 diff Z j = j
 diff (S k) (S j) = diff k j
 
+succPred : (n : Nat) -> {auto ok : LTE 1 n} -> S (pred n) = n
+succPred Z {ok} = absurd $ succNotLTEzero ok
+succPred (S k) {ok} = Refl
+
+plusSuccIsNotIdentity : {a, b : Nat} -> a + S b = a -> Void
+plusSuccIsNotIdentity {a = Z} {b} prf = absurd prf
+plusSuccIsNotIdentity {a = (S k)} {b} prf =
+        plusSuccIsNotIdentity $ succInjective (k + S b) k prf
+
+lteToSucc : LTE 1 n -> (m : Nat ** n = S m)
+lteToSucc {n = Z} prf impossible
+lteToSucc {n = S k} prf = (k ** Refl)
+
 ltePlus : LTE m1 n1 -> LTE m2 n2 -> LTE (m1 + m2) (n1 + n2)
 ltePlus {n1=Z} LTEZero lte = lte
 ltePlus {n1=S k} LTEZero lte = lteSuccRight $ ltePlus {n1=k} LTEZero lte
@@ -76,3 +89,15 @@ rightFactorLteProduct {a} {b} {c} prf =
         rewrite multRightSuccPlus b a in
         rewrite multCommutative b a in
         prf
+
+nonZeroProdNonZeroFactors : (a, b : Nat) -> LTE 1 (S a * S b)
+nonZeroProdNonZeroFactors a b = LTESucc LTEZero
+
+nonZeroLeftFactor : {a, b : Nat} -> a * b = S n -> LTE 1 a
+nonZeroLeftFactor {a = Z} eq = void . SIsNotZ $ sym eq
+nonZeroLeftFactor {a = S k} _ = LTESucc LTEZero
+
+nonZeroRightFactor : {a, b : Nat} -> a * b = S n -> LTE 1 b
+nonZeroRightFactor {a} {b = Z} eq {n} =
+        void . SIsNotZ $ replace (multZeroRightZero a) (sym eq)
+nonZeroRightFactor {b = S k} prf = LTESucc LTEZero
--- a/libs/contrib/Data/Nat.idr
+++ b/libs/contrib/Data/Nat.idr
@@ -44,20 +44,16 @@ lteMult (LTESucc lte1) (LTESucc lte2) = LTESucc $ ltePlus lte2 $ lteMult lte1 $ 
 lteCongMult : (k : Nat) -> LTE m n -> LTE (m * k) (n * k)
 lteCongMult k lte = lteMult lte lteRefl
 
-addLteRight : LTE a b -> LTE (a + c) (b + c)
-addLteRight {a} {b} {c = Z} prf =
-        rewrite plusZeroRightNeutral a in
-        rewrite plusZeroRightNeutral b in
-        prf
-addLteRight {a} {b} {c = S k} prf =
-        rewrite sym $ plusSuccRightSucc a k in
-        rewrite sym $ plusSuccRightSucc b k in
-        LTESucc $ addLteRight prf
-
 addLteLeft : LTE b c -> LTE (a + b) (a + c)
 addLteLeft {a = Z} {b} {c} prf = prf
 addLteLeft {a = S k} {b} {c} prf =
         LTESucc $ addLteLeft prf
+
+addLteRight : LTE a b -> LTE (a + c) (b + c)
+addLteRight {a} {b} {c} prf =
+        rewrite plusCommutative a c in
+        rewrite plusCommutative b c in
+        addLteLeft {a = c} {b = a} {c = b} prf
 
 subtractLteLeft : LTE (a + b) (a + c) -> LTE b c
 subtractLteLeft {a = Z} {b} {c} prf = prf
@@ -68,15 +64,11 @@ subtractEqLeft {a = Z} prf = prf
 subtractEqLeft {a = S k} {b} {c} prf = subtractEqLeft $ succInjective (k + b) (k + c) prf
 
 subtractEqRight : {a, b, c : Nat} -> a + c = b + c -> a = b
-subtractEqRight {a} {b} {c = Z} prf =
-        rewrite sym $ plusZeroRightNeutral a in
-        rewrite sym $ plusZeroRightNeutral b in
-        prf
-subtractEqRight {a} {b} {c = S k} prf =
-        subtractEqRight $
-        succInjective (a + k) (b + k) .
-        replace {P = \x => S (a + k) = x} (sym $ plusSuccRightSucc b k) $
-        replace {P = \x => x = b + S k} (sym $ plusSuccRightSucc a k) prf
+subtractEqRight {a} {b} {c} prf =
+        subtractEqLeft {a = c} {b = a} {c = b} .
+        trans (plusCommutative c a) .
+        trans prf $
+        plusCommutative b c
 
 leftFactorLteProduct : a * S b = c -> LTE a c
 leftFactorLteProduct {a} {b} {c} prf =
--- a/libs/contrib/Data/Nat.idr
+++ b/libs/contrib/Data/Nat.idr
@@ -1,5 +1,7 @@
 module Data.Nat
 
+import Syntax.PreorderReasoning
+
 %default total
 %access public export
 
@@ -23,3 +25,54 @@ lteMult (LTESucc lte1) (LTESucc lte2) = LTESucc $ ltePlus lte2 $ lteMult lte1 $ 
 
 lteCongMult : (k : Nat) -> LTE m n -> LTE (m * k) (n * k)
 lteCongMult k lte = lteMult lte lteRefl
+
+addLteRight : LTE a b -> LTE (a + c) (b + c)
+addLteRight {a} {b} {c = Z} prf =
+        rewrite plusZeroRightNeutral a in
+        rewrite plusZeroRightNeutral b in
+        prf
+addLteRight {a} {b} {c = S k} prf =
+        rewrite sym $ plusSuccRightSucc a k in
+        rewrite sym $ plusSuccRightSucc b k in
+        LTESucc $ addLteRight prf
+
+addLteLeft : LTE b c -> LTE (a + b) (a + c)
+addLteLeft {a = Z} {b} {c} prf = prf
+addLteLeft {a = S k} {b} {c} prf =
+        LTESucc $ addLteLeft prf
+
+subtractLteLeft : LTE (a + b) (a + c) -> LTE b c
+subtractLteLeft {a = Z} {b} {c} prf = prf
+subtractLteLeft {a = S j} {b} {c} prf = subtractLteLeft $ fromLteSucc prf
+
+subtractEqLeft : {a, b, c : Nat} -> a + b = a + c -> b = c
+subtractEqLeft {a = Z} prf = prf
+subtractEqLeft {a = S k} {b} {c} prf = subtractEqLeft $ succInjective (k + b) (k + c) prf
+
+subtractEqRight : {a, b, c : Nat} -> a + c = b + c -> a = b
+subtractEqRight {a} {b} {c = Z} prf =
+        rewrite sym $ plusZeroRightNeutral a in
+        rewrite sym $ plusZeroRightNeutral b in
+        prf
+subtractEqRight {a} {b} {c = S k} prf =
+        subtractEqRight $
+        succInjective (a + k) (b + k) .
+        replace {P = \x => S (a + k) = x} (sym $ plusSuccRightSucc b k) $
+        replace {P = \x => x = b + S k} (sym $ plusSuccRightSucc a k) prf
+
+leftFactorLteProduct : a * S b = c -> LTE a c
+leftFactorLteProduct {a} {b} {c} prf =
+    let cGteA =
+        (c) ={ sym prf }=
+        (a * (S b)) ={ multRightSuccPlus a b }=
+        (a + a * b) QED
+    in
+    rewrite cGteA in
+    lteAddRight a
+
+rightFactorLteProduct : S a * b = c -> LTE b c
+rightFactorLteProduct {a} {b} {c} prf =
+    leftFactorLteProduct {a = b} {b = a} $
+        rewrite multRightSuccPlus b a in
+        rewrite multCommutative b a in
+        prf
--- a/libs/contrib/Data/Nat.idr
+++ b/libs/contrib/Data/Nat.idr
@@ -31,6 +31,11 @@ ltePlus (LTESucc lte1) lte2 = LTESucc $ ltePlus lte1 lte2
 lteCongPlus : (k : Nat) -> LTE m n -> LTE (m + k) (n + k)
 lteCongPlus k lte = ltePlus lte lteRefl
 
+lteMultRight : LTE a b -> (c : Nat) -> LTE a (b * S c)
+lteMultRight {b} prf c =
+        rewrite multRightSuccPlus b c in
+        lteTransitive prf $ lteAddRight b
+
 lteMult : LTE m1 n1 -> LTE m2 n2 -> LTE (m1 * m2) (n1 * n2)
 lteMult LTEZero _ = LTEZero
 lteMult {m1=S k} (LTESucc _) LTEZero = rewrite multZeroRightZero k in LTEZero
@@ -101,3 +106,8 @@ nonZeroRightFactor : {a, b : Nat} -> a * b = S n -> LTE 1 b
 nonZeroRightFactor {a} {b = Z} eq {n} =
         void . SIsNotZ $ replace (multZeroRightZero a) (sym eq)
 nonZeroRightFactor {b = S k} prf = LTESucc LTEZero
+
+notLteAndGt : (a, b : Nat) -> LTE a b -> GT a b -> Void
+notLteAndGt Z b aLteB aGtB = succNotLTEzero aGtB
+notLteAndGt (S k) Z aLteB aGtB = succNotLTEzero aLteB
+notLteAndGt (S k) (S j) aLteB aGtB = notLteAndGt k j (fromLteSucc aLteB) (fromLteSucc aGtB)
--- a/libs/contrib/Data/Nat/Factor.idr
+++ b/libs/contrib/Data/Nat/Factor.idr
@@ -1,5 +1,7 @@
 module Data.Nat.Factor
 
+import Data.Fin
+import Data.Fin.Extra
 import Data.Nat
 
 %default total
@@ -16,10 +18,10 @@ data Factor : Nat -> Nat -> Type where
     CoFactorExists : {n, f : Nat} -> (q : Nat) -> FactorsOf n (f, q) -> Factor n f
 
 data NotFactor : Nat -> Nat -> Type where
-    ProperRemExists : (n, p, q, r : Nat) ->
+    ProperRemExists : (n, p, q : Nat) ->
+        (r : Fin (pred p)) ->
         LTE 1 n ->
-        {auto rltp : LT (S r) p} ->
-        p * q + S r = n ->
+        p * q + S (finToNat r) = n ->
         NotFactor n p
 
 data DecFactor : Nat -> Nat -> Type where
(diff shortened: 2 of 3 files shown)


libs/contrib/Data/Nat/Factor.idr
```diff
@@ -0,0 +1,332 @@
...
+                        in
+                        greatestSbSr q' (CommonFactorExists q' qfb rightPrf)
+                in
+                (f ** MkGCD (CommonFactorExists f prfSa prfSb) greatest)
+
+||| An implementation of Euclidean Algorithm for computing greatest common
+||| divisors. It is proven correct and total; returns a proof that computed
+||| number actually IS the GCD. Unfortunately it's very slow, so improvements
+||| in terms of efficiency would be welcome.
+gcd : (a, b : Nat) -> {auto aok : LTE 1 a} -> {auto bok : LTE 1 b} -> (f : Nat ** GCD a b f)
```
nickdrozd, line 313: There is a `gcd` defined in [Prelude.Nat](https://github.com/idris-lang/Idris-dev/blob/0417c53fb6b91fc0692245f269b98a3bd4820136/libs/prelude/Prelude/Nat.idr#L419)

libs/contrib/Data/Nat.idr
```diff
@@ -16,10 +31,83 @@ ltePlus (LTESucc lte1) lte2 = LTESucc $ ltePlus lte1 lte2
...
+
+subtractLteLeft : LTE (a + b) (a + c) -> LTE b c
+subtractLteLeft {a = Z} {b} {c} prf = prf
+subtractLteLeft {a = S j} {b} {c} prf = subtractLteLeft $ fromLteSucc prf
+
+subtractEqLeft : {a, b, c : Nat} -> a + b = a + c -> b = c
+subtractEqLeft {a = Z} prf = prf
+subtractEqLeft {a = S k} {b} {c} prf = subtractEqLeft $ succInjective (k + b) (k + c) prf
+
+subtractEqRight : {a, b, c : Nat} -> a + c = b + c -> a = b
```
nickdrozd, line 70: `subtractEqRight` is can be proved as an instance of `subtractEqLeft`:

```idris
subtractEqRight : {a, b, c : Nat} -> a + c = b + c -> a = b
subtractEqRight {a} {b} {c} prf =
  subtractEqLeft {a = c} {b = a} {c = b} $
    trans (plusCommutative c a) $
      trans prf $
        plusCommutative b c
```

libs/contrib/Data/Nat.idr
```diff
@@ -16,10 +31,83 @@ ltePlus (LTESucc lte1) lte2 = LTESucc $ ltePlus lte1 lte2
...
+addLteRight {a} {b} {c = Z} prf =
+        rewrite plusZeroRightNeutral a in
+        rewrite plusZeroRightNeutral b in
+        prf
+addLteRight {a} {b} {c = S k} prf =
+        rewrite sym $ plusSuccRightSucc a k in
+        rewrite sym $ plusSuccRightSucc b k in
+        LTESucc $ addLteRight prf
+
+addLteLeft : LTE b c -> LTE (a + b) (a + c)
```
nickdrozd, line 57: Same here, `addLteRight` would be simpler to prove as an instance of `addLteLeft`. It's probably true for any of these left / right proofs that one side is easier to prove, and the other side can be derived from that one.

libs/contrib/Data/Nat/Factor.idr
Sventimir: True. But that definition does not provide any proof that the returned number is indeed what it claims to be. That is the whole point of this module: to define what a factor is and start proving things about factors and factorisation.

libs/contrib/Data/Nat.idr
Sventimir: Good point!

libs/contrib/Data/Nat/Factor.idr
nickdrozd: Yes, my point is just that there is a name collision.

Sventimir: I thought about naming it `gcf` for greatest common factor. That would be along with the naming convention in this module, but on the other hand it would *not* be along with what people expect that function to be named. Type of this function is different, so in some cases (when you supply implicit arguments explicitly) calls will be unambiguous. Otherwise you'll have to qualify it. I don't think it's too bad…

libs/contrib/Data/Nat.idr
Sventimir: Then I'll try rewriting them.

nickdrozd:
When I compile and load `Factor` into the repl (`:module Data.Nat.Factor`), it can't find `gcd` (or `gcdProof`, as I've renamed it locally). In the Emacs buffer, all the `data` types (`Factor`, `NotFactor`, etc) get underlined in red. I don't know what it is, but it seems like there is some kind of path / visibility problem.

Sventimir:
> When I compile and load `Factor` into the repl (`:module Data.Nat.Factor`), it can't find `gcd` (or `gcdProof`, as I've renamed it locally)

It's because of the `%access private` in the `GCD` namespace. I thought it would be restricted to that block, but apparently it is not. I will add `export`qualifier to the `gcd` function and it should be fine. As to the underlining, I don't know. In my Atom nothing is underlined. Does Emacs suggest anything about what it thinks is wrong?

libs/contrib/Data/Nat/Factor.idr
```diff
@@ -0,0 +1,333 @@
...
+multNAlsoFactor : Factor n p -> (a : Nat) -> {auto aok : LTE 1 a} -> Factor (n * a) p
+multNAlsoFactor _ Z {aok} = absurd $ succNotLTEzero aok
+multNAlsoFactor (CofactorExists {n} {p} q positN prf) (S a) =
+        CofactorExists (q * S a) (lteMultRight positN a) $
+            rewrite sym prf in
+            multAssociative p q (S a)
+
+||| If p is a factor of both n and m, then it is also a factor of their sum.
+plusFactor : Factor n p -> Factor m p -> Factor (n + m) p
+plusFactor {n} {p} nFactor@(CofactorExists qn positN prfN) (CofactorExists qm positM prfM) =
```
nickdrozd, line 164: What does the `@` mean?

nickdrozd:
> It's because of the `%access private` in the `GCD` namespace. I thought it would be restricted to that block, but apparently it is not. I will add `export`qualifier to the `gcd` function and it should be fine. As to the underlining, I don't know. In my Atom nothing is underlined. Does Emacs suggest anything about what it thinks is wrong? Maybe it's some style issue, indentation size perhaps?

That fixed it. 

I don't know what the lines are about. It could be a bug in the mode. Actually, it is certainly a bug in the mode. For: the underlining is either accurate or not. If it isn't accurate, then that's obviously a bug. On the other hand, if it is accurate, then the lines are an indication that something is wrong, but that something isn't being reported, which is a bug.

Is that an intuitionistically valid proof?

libs/contrib/Data/Nat/Factor.idr
```diff
@@ -0,0 +1,333 @@
...
+||| This actually follows directly from the definition of GCD.
+commonFactorAlsoFactorOfGCD : Factor a p -> Factor b p -> GCD a b q -> Factor q p
+commonFactorAlsoFactorOfGCD {p} pfa pfb (MkGCD _ greatest) =
+        greatest p (CommonFactorExists p pfa pfb)
+
+
+||| A decision procedure for whether of not p is a factor of n.
+decFactor : (n, d : Nat) -> {auto nok : LTE 1 n} -> {auto dok : LTE 1 d} -> DecFactor n d
+decFactor n (S d) {nok} {dok} with (Data.Fin.Extra.divMod n (S d))
+        | (Fraction n (S d) q r prf) = case r of
```
nickdrozd, line 207: This case statement can be pushed back into the arg list:

```idris
decFactor n (S d) {nok} {dok} with (Data.Fin.Extra.divMod n (S d))
  | (Fraction n (S d) q FZ prf) =
    let prf =
            replace {P = \x => x = n} (plusZeroRightNeutral (q + (d * q))) $
            replace {P = \x => x + 0 = n} (multCommutative q (S d)) prf
    in
    ItIsFactor $ CofactorExists q nok prf
  | (Fraction n (S d) q (FS pr) prf) =
    ItIsNotFactor $ ProperRemExists q pr nok $
      rewrite multCommutative d q in
      rewrite sym $ multRightSuccPlus q d in
      prf
```

(Indent to taste)

libs/contrib/Data/Nat/Factor.idr
Sventimir: `@` binds the whole pattern on the right to the name on the left. In this case it binds `CofactorExists qn positN prfN` to the name `nFactor`. This way you can simultaneously refer to the whole expression and to its subexpressions.

libs/contrib/Data/Nat/Factor.idr
Sventimir: Yes… I usually make a double indentation after the function pattern, because if I'm going to add a `where` clause later, I like it to be less indented than the function body. I feel it's more apparent where the body ends that way. And if I'm not going to add a `where`, at least all functions are indented according to the same rule. However, I'm not sure if it's the right approach. I also like 4-space-wide indents, while you apparently prefer 2-spaces only. I should probably have a closer look at older modules and adapt my style to what there already is…

Sventimir:
> Is that an intuitionistically valid proof?

Seems valid to me, although rather informal.

libs/contrib/Data/Nat/Factor.idr
nickdrozd: Mostly I stick to the indentation levels provided by the Emacs mode, and I don't think it ever provides 4-spacing. (It doesn't matter, of course.)

libs/contrib/Data/Nat/Factor.idr
```diff
@@ -0,0 +1,333 @@
...
+commonFactorSym : CommonFactor a b p -> CommonFactor b a p
+commonFactorSym (CommonFactorExists p pfa pfb) = CommonFactorExists p pfb pfa
+
+||| 1 is a factor of any natural number.
+oneIsFactor : (n : Nat) -> {auto ok : LTE 1 n} -> Factor n 1
+oneIsFactor (S k) {ok} =
+        CofactorExists (S k) ok (rewrite plusZeroRightNeutral k in Refl)
+
+||| For all natural numbers n, n is (the greatest) a factor of n.
+selfFactor : (n : Nat) -> {auto ok : LTE 1 n} -> Factor n n
```
nickdrozd, line 104: I would call this `factorReflexive`. The factor relationship is a [partial order](https://github.com/idris-lang/Idris-dev/blob/0417c53fb6b91fc0692245f269b98a3bd4820136/libs/contrib/Decidable/Order.idr#L23), right? That might be worth exploring, although the `ok` thing complicates it a little.

Sventimir: Yeah, these `LTE`s are annoying. However, considering zeros destroys most of the properties of factors, since anything is a factor of zero. On the other hand, maybe I should exclude zero only where it really hurts and allow considering factors of zero in general case.

Hmm, indeed! The factor relation is a partial order. That's a thing worth proving. I'll add it.

libs/contrib/Data/Nat/Factor.idr
```diff
@@ -0,0 +1,333 @@
...
+        )
+
+||| 1 is a common factor of any pair of natural numbers.
+oneCommonFactor : (a, b : Nat) -> {auto aok : LTE 1 a} -> {auto bok : LTE 1 b} -> CommonFactor a b 1
+oneCommonFactor a b {aok} {bok} = CommonFactorExists 1
+        (CofactorExists a aok (rewrite plusZeroRightNeutral a in Refl))
+        (CofactorExists b bok (rewrite plusZeroRightNeutral b in Refl))
+
+||| Any natural number is a common factor of itself and itself.
+selfIsCommonFactor : (a : Nat) -> {auto ok : LTE 1 a} -> CommonFactor a a a
```
nickdrozd, line 244: Same here, I would call this `commonFactorReflexive` and move it next to `commonFactorSym`. Is there a sense in which `CommonFactor` is transitive? If it is, you could implement it as an [equivalence relation](https://github.com/idris-lang/Idris-dev/blob/0417c53fb6b91fc0692245f269b98a3bd4820136/libs/contrib/Decidable/Order.idr#L29).

libs/contrib/Data/Nat/Factor.idr
nickdrozd: The other `gcd` function only stipulates that the inputs can't both be zero. Could that work here? I've looked at a few expositions of the algorithm, and nobody ever spends much time discussing the nature of the inputs. It would certainly be cleaner if there weren't `LTE 1 n` proofs floating around all over, but I don't know how necessary they are.

libs/contrib/Data/Nat/Factor.idr
Sventimir: Well, in order to be transitive, it would have to be a two-argument predicate. This can be done like so:

```
CommonFactor (a, b) p
```
However now sides of the relation have different types, which is no good. However, if you stretch the definition a little bit, you can come up with something like this:

```
cfTransitive : CommonFactor (a, b) p -> CommonFactor (b, c) p -> CommonFactor (a, c) p
```
So it is transitive when you consider it a binary relation (either there is or there isn't a common factor of two numbers in question), without considering what that factor actually is. When looked upon like this, it is indeed transitive, and this is the sense in which it is symmetric in `commonFactorSym` proof.

libs/contrib/Data/Nat/Factor.idr
Sventimir: Actually I think it does matter in the sense that it should be consistent within a project.

libs/contrib/Data/Nat/Factor.idr
```diff
@@ -0,0 +1,333 @@
...
+||| An implementation of Euclidean Algorithm for computing greatest common
+||| divisors. It is proven correct and total; returns a proof that computed
+||| number actually IS the GCD. Unfortunately it's very slow, so improvements
+||| in terms of efficiency would be welcome.
+export
+gcd : (a, b : Nat) -> {auto aok : LTE 1 a} -> {auto bok : LTE 1 b} -> (f : Nat ** GCD a b f)
+gcd Z _ {aok} = absurd aok
+gcd _ Z {bok} = absurd bok
+gcd (S a) (S b) with (cmp (S a) (S b))
+    gcd (S (b + S d)) (S b) | CmpGT d =
```
nickdrozd, line 318: One way to handle Euclid's algorithm is to arrange it so that the first argument is the smaller one. If the first argument is larger, just flip the arguments around and proceed as usual, as in `gcd 27 18 = gcd 18 27 = 9`. That would simplify the logic (just keep the simpler of the two branches), although I think it would require an `assert_total`.

This is discussed at the very beginning of Knuth's _The Art of Computer Programming_.

Sventimir: I believe it's exactly what I have done. If `a > b`, then the recursion is run with `(a, b)` pair, otherwise – with `(b, a)`. The `GCD.Search` type even contains a proof that `a >= b`. Totality of the function is implied by the use of `sizeInd` function from `WellFounded` module, which provides looping capability already proved to be total.

nickdrozd: Ah, okay, I'm still working through the details.

libs/contrib/Data/Nat/Factor.idr
Sventimir: Zeros are problematic, because according to this definition at least, everything is a factor of zero. This destroys some nice properties, like that a factor is always less than or equal to the number it divides. Also there is consideration that according to this setup, zero is a factor of zero, which implicitly assumes that you can actually divide by zero. Also when you want to implement rational numbers, that non-zero property comes in handy, because all ratios are either positive or negative and you just add a special constructor for zero. This way you avoid the problem `ZZ` has, that `0` and `-0` are distinct values, which messes up the `(=)` proofs. Also the requirement that divisor must be non-zero is graciously provided by this arrangement.

On the other hand, as I review the code, I see that the non-zero property is not used that often (mostly is just passed forward as is or with some alteration). So probably we could require non-zero-ness for those proofs only, and we could require it explicitly fro rational numbers as well. I'm currently looking into the possibilities of removing that constraint and it seems possible.

libs/contrib/Data/Nat/Factor.idr
```diff
@@ -0,0 +1,333 @@
...
+        (rec : (y : Search) -> Smaller y x ->  (f : Nat ** GCD (left y) (right y) f)) ->
+        (f : Nat ** GCD (left x) (right x) f)
+    step (SearchArgs Z _ bLteA {bNonZero}) _ = absurd . succNotLTEzero $ lteTransitive bNonZero bLteA
+    step (SearchArgs _ Z _ {bNonZero}) _ = absurd $ succNotLTEzero bNonZero
+    step (SearchArgs (S a) (S b) bLteA {bNonZero}) rec with (divMod (S a) (S b))
+        | Fraction (S a) (S b) q FZ prf =
+            let sbIsFactor = the (plus q (mult b q) = S a) $
+                    rewrite multCommutative b q in
+                    rewrite sym $ multRightSuccPlus q b in
+                    replace {P = \x => x = S a} (plusZeroRightNeutral (q * S b)) prf
```
nickdrozd, line 274: `replace` can be replaced with

```idris
                    rewrite sym prf in
                    sym $ plusZeroRightNeutral (q * S b)
```

IME explicit use of `replace` is a "proof smell", and there is almost always a nicer way to do it.

libs/contrib/Data/Nat/Factor.idr
```diff
@@ -0,0 +1,333 @@
...
+import Syntax.PreorderReasoning
+
+%default total
+%access public export
+
+
+||| Factor n p is a witness that p is indeed a factor of n,
+||| i.e. there exists a q such that p * q = n.
+data Factor : Nat -> Nat -> Type where
+    CofactorExists : {n, p : Nat} -> (q : Nat) -> LTE 1 n -> p * q = n -> Factor n p
```
nickdrozd, line 15: I notice that there are a lot of places with `rewrite sym prf` or
something like that. I think the `sym`s are requried because this
proof has the form `p * q = n`. If it were `n = p * q`, I think the
`sym`s could be done away with, which would be a little less verbose.

Generally proof equations are written as complexity-reducing devices,
with the more complex part on the left and the less complex on the
right. For example, the monoid neutral law looks like `n + 0 = n`,
because `n` is simpler than `n + 0`. But with factors, the reduction
in complexity is to say that a larger number can be written as the
product of smaller numbers, so I think `n = p * q` is appropriate.

(It might be too late to implement this minor suggestion.)

libs/contrib/Data/Nat/Factor.idr
Sventimir: I try to use `rewrite` rather than `replace` whenever I can, but it's not always possible. There is the huge difference between them, however. `replace` makes a substitution in a proof you already have to create another proof, whereas `rewrite` makes a substitution in the goal (which is a proof you do *not* have yet) instead. In practice it means that with `rewrite` you work kind of backwards, rewriting your goal until it matches something you already have, while `replace` has a more direct approach. Also with `rewrite` you can only use equality `(=)` proofs (the same with `replace` of course), which means that you you need use a custom proof-altering function like `cong` or something in the middle of replacements, you actually can no longer use `rewrite` from that point on. Unless of course I am missing something here.

libs/contrib/Data/Nat/Factor.idr
nickdrozd: https://en.wikipedia.org/wiki/Order_theory#/media/File:Lattice_of_the_divisibility_of_60.svg

libs/contrib/Data/Nat/Factor.idr
Sventimir: That's a fair point. And I don't think it's too late for the change. It's only 300 lines of code. I have done much larger refactorings with much less type-checking to support me, so I believe I will do it.

Sventimir:
I implemented `Preorder` interface for `Nat` and `Factor`. I also tried to implement `Poset`, but proving the antisymmetric property is messy, because there are 4 variables and only two equations, so squeezing contradictions from them takes a lot of pattern matching on those `Nat`s. I also inverted the equations in `Factor` and `NorFactor` proof constructors and eliminated these pesky `LTE 1`s.

libs/contrib/Data/Nat/Factor.idr
```diff
@@ -0,0 +1,332 @@
...
+    transitive a b c (CofactorExists qb prfAB) (CofactorExists qc prfBC) =
+        CofactorExists (qb * qc) (
+            rewrite prfAB in
+            rewrite prfBC in
+            rewrite sym $ multAssociative c qc qb in
+            rewrite multCommutative qc qb in
+            Refl
+        )
+
+    reflexive Z = CofactorExists 1 Refl
```
nickdrozd, line 65: This one doesn't require case splitting: `reflexive n = CofactorExists 1 $ sym $ multOneRightNeutral n`

libs/contrib/Data/Nat/Factor.idr
```diff
@@ -0,0 +1,332 @@
...
+import Decidable.Order
+import Syntax.PreorderReasoning
+
+%default total
+%access public export
+
+
+||| Factor n p is a witness that p is indeed a factor of n,
+||| i.e. there exists a q such that p * q = n.
+data Factor : Nat -> Nat -> Type where
```
nickdrozd, line 15: This is maybe getting into bikeshedding territory, but what about
changing this from `Factor n p` to `Factor p n`? `Factor n p` reads as
"n has p as a factor", but I feel like the more natural way to say it
would be "p is a factor of n", or alternatively, "p evenly divides n".
With infix syntax, it could even be `p | n` or something like that.
Similar with `CommonFactor`. Instead of saying "a and b both have p as
a factor", the reading would be "p is a factor of both a and b" or "p
divides both a and b".

Besides being more natural (IMO), this would clarify the way in which
`CommonFactor` is symmetric. Given some "p", the relation is "has
factor p in common with", as in "a has factor p in common with b".
Something like this:

```idris
using (p : Nat)
  Preorder Nat (CommonFactor p) where

  Equivalence Nat (CommonFactor p) where
```

libs/contrib/Data/Nat/Factor.idr
```diff
@@ -0,0 +1,332 @@
...
+||| If p is a common factor of a and b, then it is also a factor of their GCD.
+||| This actually follows directly from the definition of GCD.
+commonFactorAlsoFactorOfGCD : Factor a p -> Factor b p -> GCD a b q -> Factor q p
+commonFactorAlsoFactorOfGCD {p} pfa pfb (MkGCD _ greatest) =
+        greatest p (CommonFactorExists p pfa pfb)
+
+
+||| A decision procedure for whether of not p is a factor of n.
+decFactor : (n, d : Nat) -> DecFactor n d
+decFactor n Z = ?whaatIfZ
```
nickdrozd, line 197: hole

Sventimir: Dmnit. sorry.

libs/contrib/Data/Nat/Factor.idr
Sventimir: Good point

libs/contrib/Data/Nat/Factor.idr
Sventimir: Unfortunately, we can't have `(|)` as a type constructor, since it is already reserved as a part of Idris syntax. Nontheless, this is probably a better choice, so I'll change it.

Sventimir:
I have switched places of arguments to `Factor` and `NotFactor`.

As to the leftover hole, I'm not really proud of the way I filled in, but I see no other way. The problem stems, I think, from the fact that 0 is a singularity of the division function. Standard procedure of proving not-a-factor statement requires to specify a non-zero remainder of the division, which is also strictly less than the divisor. Without the last restriction, you could for instance prove that 2 is not a factor of 4, since `2 * 1 + 2 = 4`. However, when divisor is 0, there is no such remainder you could possibly construct for any number. Yet, you can't prove that 0 *is* a divisor of a non-zero number either. So the only way to fill that vacuum I can think of is to provide a sepcial constructor that asserts that 0 is not a factor of any successor. This way `decFactor` function is total once more and `factorNotFactoAbsurd` property still holds.

But I don't like it really, so if anyone could come up with a simpler/nicer solution, that would be welcome.

PR #4860, opened 2020-05-15
Update docs link in FAQ

alexhumphreys:
Not sure is this new link the correct one, but the current link 404s so at least it's a bit of an improvement.


melted:
Thanks!

PR #4874, opened 2020-06-13
Fix missing import

conjunctive:
This PR adds a missing import of `throwIO` for the conditional usage:
https://github.com/idris-lang/Idris-dev/blob/cf481823edef2130b550b87287f1a2085037e4dc/src/Util/DynamicLinker.hs#L118-L120


melted:
Thanks!

PR #4891, opened 2020-12-11
Update link to Idris 2 repo in README

lynn:
It looks like idris-lang/Idris2 is the “main” repo now.


melted:
Thanks for the PR!

PR #4892, opened 2020-12-16
Fix compatibility with megaparsec 9

felixonmars:


melted:
Thanks!

Jake-Gillberg:
Please ignore the linking, this pull request has nothing to do with pregnancy-induced hypertension, I was just using this as an example of a patch that got temporarily applied to a project dependency.

PR #4902, opened 2021-07-27
README: clarify Idris 1 vs 2 situation, and the compilation guide

attila-lendvai:
this is the fruit of my experience as a newcomer to Idris, and trying to compile Idris 1 from source.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Idris
+# Idris 1
 
 [![Build Status](https://travis-ci.org/idris-lang/Idris-dev.svg?branch=master)](https://travis-ci.org/idris-lang/Idris-dev)
 [![Appveyor build](https://ci.appveyor.com/api/projects/status/xi8yu81oy1134g7o/branch/master?svg=true)](https://ci.appveyor.com/project/idrislang/idris-dev)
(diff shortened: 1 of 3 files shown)


juhp:
I haven't verified all the details but this LGTM anyway.

README.md
```diff
@@ -26,9 +26,18 @@ More information about building Idris from source has been detailed in the [Inst
...
+More information about code generators can be found
+[here](https://idris.readthedocs.io/en/latest/reference/codegen.html).
+
+## Status
+
+This is Idris 1, implemented in Haskell. Idris 1 is not actively
+worked on anymore.
+
+[Idris 2](https://github.com/idris-lang/Idris2) is an early preview of
+the next generation of Idris, implemented in Idris 1.
```
melted, line 40: Idris 2 is implemented in Idris 2, it was bootstrapped from Idris2-boot, which is the one implemented in Idris 1

melted: Also, it's probably less unstable than an "early preview" now. But I don't know how Edwin would characterize it.

attila-lendvai:
ok, i've addressed the concerns by not saying anything about the bootstrap of idris 2. i guess it's more complicated than being discussed in the readme of idris 1... :)

melted (approved):

PR #4903, opened 2021-08-21
allow ansi-terminal-0.11 and megaparsec-9

juhp:
With this it should be possible to build with Stackage LTS 18


melted:
Thanks!

juhp:
By the way is there an plan to make a new minor release soon?

PR #4907, opened 2021-09-22
update test/pkg010/ to follow changes in idris --help output

attila-lendvai:
i'm not sure why this has changed. maybe because i'm compiling with a newer version of a dependency?


attila-lendvai:
hrm, this makes it fail in the CI. maybe it depends on the exact version of a dependency?

feel free to close this if it's not a step forward.

PR #4910, opened 2021-12-30
fix ffi008 test: add -fcommon to accommodate for gcc v9+

attila-lendvai:
Otherwise it fails with: "multiple definition of `mystruct'".

For further details see:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85678
https://stackoverflow.com/questions/52294245/how-to-declare-global-variable-for-an-anonymous-structure-in-c


melted:
Looks like this one can be merged. Thanks!

PR #4916, opened 2022-04-14
Add a guix.scm for easy building on Guix.

attila-lendvai:
This enables the following on Guix:

To build and install a local, possibly modified checkout:
  guix package -f guix.scm

To build it, but not install it:
  guix build -f guix.scm

To use as the basis for a shell with the development environment:
  guix shell


melted:
Thanks!

PR #4925, opened 2023-12-17
Support mtl 2.3

Vekhir:
`mtl` 2.3 removes re-exports of `Control.Monad`. This leads to various `Variable not in scope` errors.

This PR solves them by adding the explicit import of `Control.Monad` to the affected files. This also works with earlier versions of `mtl`, as it just replaces the implicit import with an explicit one.
This also enables support for GHC 9.6.


melted:
Thanks!

Vekhir:
Great! I've just made another PR at #4926 bringing more dependencies up-to-date

